PR #20, opened 2020-01-26
adding a openocd option for flashing with support for stlink v2 clones

luben93:


CMakeLists.txt
```diff
@@ -11,9 +11,9 @@ if (NOT NRF5_SDK_PATH)
   message(FATAL_ERROR "The path to the NRF52 SDK must be specified on the command line (add -DNRF5_SDK_PATH=<path>")
 endif ()
 
-if(NOT USE_JLINK AND NOT USE_GDB_CLIENT)
-  set(USE_JLINK true)
-endif()
+#if(NOT USE_JLINK AND NOT USE_GDB_CLIENT)
+#  set(USE_JLINK true)
+#endif()
```
JF002: The goal of this condition was to ensure that a default behavior (jlink) was enabled if no USE_* variable was specified.

We could either keep it or remove it:
 - if we want to keep it, we should add USE_OPENOCD in the condition
 - if we do not keep it, we must ensure that the project builds correctly and update the documentation (no USE_* means that the project builds but does not provide support for flashing, or one of the USE_* variable is defined, and in this case, the project builds and provide targets to flash the firmware in the device.

What do you think?

luben93: I think the first alternative works, I just forgot to review the code twice

JF002: For now, I think it's enough. We'll refactor this part if we need to add support for more debugger. Thanks for your pull request!

JF002:
It seems that the files interfaces/stlink.cfg and interfaces/nrf52.cfg are needed.
Do the users need to copy them in a specific folder? Or are these files custom to the project and need to be placed into the source code of the project?

luben93:
They are part of the openocd package compiled from AUR openocd-git

JF002:
Do they have to be copied in the working directory of the project?

Avamander:
In this case, yes? I suspect it would be nicer if a variable for OpenOCD's root could be specified, the location varies across OSs and distros.

luben93:
No, openocd has them in a its include path, mine on arch is `/usr/share/openocd/scripts` and that will differ between distributions

luben93:
But the cmake script is not defining the default root path so it should find the correct path as it does on my machine

Avamander:
I haven't installed my build and I have to define either a full or a correct relative path to the script folder for OpenOCD to find them.

luben93:
Please feel free to submit any PR to make it easier to build with locally built openocd, or was this enough

PR #24, opened 2020-04-09
sdk as submodule

geoffrey-vl:


JF002:
Thanks for this pull request, @geoffrey-vl . I appreciate!

However, I'm wondering : how do we ensure that the repo you use to fetch the SDK (https://github.com/justaddobvious/nRF5_SDK_15.3.0_59ac345) is genuine, as it does not seem to be an official repo?

geoffrey-vl:
You're correct.
I would have preferred to use the official [Nordic](https://github.com/NordicSemiconductor) repo, but I couldn't find it on github. This was the next best thing I could find. For hosting concerns (and the lack of it) it would be better for you to fork it so that we can point to your fork instead.
But that's your choice however :)

As far is integrity goes, its off course better to set up your own SDK repo from scratch and trust only but yourselves, or Nordic if you want.

I'll try to get someone from the Nordic team to elaborate the chance for the SDK to ever emerge on GH.

JF002:
Sorry for the delay !

Indeed, an official repo from Nordic would be a solution. However, according to [the link you mentionned](https://devzone.nordicsemi.com/f/nordic-q-a/27851/why-isn-t-the-nrf51-sdk-on-git), they are working on it for more than 5 years.

Do you know if the licence allows me to host a copy of the SDK in a Github repo?

endian-albin:
> Do you know if the licence allows me to host a copy of the SDK in a Github repo?

I've now skimmed through all of the license notices in the nRF5_SDK_16.0.0 SDK including the external libraries. Everything is redistributable under various permissive (e.g. BSD 3-clause) or silly terms, e.g. that it may only be used with Nordic hardware or that reverse-engineering / decompilation of the SDK is forbidden. The only license notice I found to be borderline is the one for "Tile" (under external/) which says:

> Copyright 2017 Tile Inc.  All Rights Reserved.
> All code or other information included in the accompanying files ("Tile Source Material")
> is PROPRIETARY information of Tile Inc. ("Tile") and access and use of the Tile Source Material
> is subject to these terms. The Tile Source Material may only be used for demonstration purposes,
> and may not be otherwise distributed or made available to others, including for commercial purposes.
> Without limiting the foregoing , you understand and agree that no production use
> of the Tile Source Material is allowed without a Tile ID properly obtained under a separate
> agreement with Tile.
> You also understand and agree that Tile may terminate the limited rights granted under these terms
> at any time in its discretion.
> All Tile Source Material is provided AS-IS without warranty of any kind.
> Tile does not warrant that the Tile Source Material will be error-free or fit for your purposes.
> Tile will not be liable for any damages resulting from your use of or inability to use
> the Tile Source Material.

Perhaps one could write a script that removes all of the components that are not used...

borgmanJeremy:
Could we consider just grabbing the depenencies with Conan? Then we can keep the original hosted location and tie into cmake nicely.

redengin:
CMake FetchContent https://cmake.org/cmake/help/latest/module/FetchContent.html can already provide dependency pulls.

While Conan is an option, it's yet another syntax and build-host installation.

JF002:
I'm closing this PR as it's really old. I don't think we need this right now, as the docker build downloads all the dependencies for us. 
Thanks anyway for your work!

PR #27, opened 2020-04-30
Basic time setting and notification service

piggz:
For comment/discussion as this is p-o-c at the moment.
My initial thought regarding the "Pinetime Service" was as a vendor-style dumping ground where any number of things could be implemented, starting with setting the time.  This is modelled on my knowledge of the Huami service on the various huami bands/watches.

If you would like to start simpler, we could implement this as a CTS service.  my reading of the spec here https://www.bluetooth.com/xml-viewer/?src=https://www.bluetooth.com/wp-content/uploads/Sitecore-Media-Library/Gatt/Xml/Services/org.bluetooth.service.current_time.xml says that writing the time characteristic is optional, so, it would just be a case of changing the uuid's and allowing reading, and we would have a basic CTS that allowed reading/writing time.

### DIFF
--- a/src/Components/Ble/CurrentTimeService.cpp
+++ b/src/Components/Ble/CurrentTimeService.cpp
@@ -50,7 +50,6 @@ int CurrentTimeService::OnTimeAccessed(uint16_t conn_handle, uint16_t attr_handl
 
   }
 
-  //!TODO need to support reading the time.
   return 0;
 }
 
--- a/src/Components/Ble/CurrentTimeService.cpp
+++ b/src/Components/Ble/CurrentTimeService.cpp
@@ -34,7 +34,22 @@ int CurrentTimeService::OnTimeAccessed(uint16_t conn_handle, uint16_t attr_handl
     m_dateTimeController.SetTime(result.year, result.month, result.dayofmonth,
                         0, result.hour, result.minute, result.second, nrf_rtc_counter_get(portNRF_RTC_REG));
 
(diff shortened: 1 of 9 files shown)


JF002:
Thanks for this PR!
I think we should try to implement standard services as much as possible (as long as it does not prevent easy integration on the companion app side).

I'm Ok with renaming the current PinetimeService to CTS and to use the standard UUIDs.

In the future, we'll add a custom service if we need to add functionalities that are not covered by the standards.

piggz:
Ive now renamed the time service and change the UUIDs, and also implemented reading, which is mandatory according to the spec, writing is optional, but used here to set the device time.

src/Components/Ble/CurrentTimeService.cpp
```diff
@@ -0,0 +1,84 @@
...
+    currentDateTime.second = m_dateTimeController.Seconds();
+    currentDateTime.millis = 0;
+
+
+    int res = os_mbuf_append(ctxt->om, &currentDateTime, sizeof(CtsData));
+    return (res == 0) ? 0 : BLE_ATT_ERR_INSUFFICIENT_RES;
+
+  }
+
+  //!TODO need to support reading the time.
```
JF002, line 53: Is this TODO still relevant?

piggz:
Hmm, something has gone horribly wrong!

piggz:
Ah, it looks lots better with the correct base branch!

PR #39, opened 2020-07-19
Docker image

Ripsnapper:
The Docker image contains the arm gcc toolchain 2019q3
and the NRF SDk 15.3.0_59ac345.

The build script currently builds the docker Image if not present.
Then it starts the container with the current directory mounted
which then runs cmake and make.

Flashing with this image does not work yet but that should be possible.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,28 @@
+FROM debian:10.4 
+
+ARG UNAME=user
+ARG UID=1000
+ARG GID=1000
+
+COPY checksums.md5 .
+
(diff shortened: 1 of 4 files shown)


JF002:
Thank you for this pull request! I was thinking about writing this kind of script!

Could you please add documentation about how to use it? You could maybe add an item in the list *Build, flash and debug* (Readme.md) and a new file buildWithDocker.md in the *doc* folder?

I tried to use it, it built the docker image and started the container, but CMake complained that -DNRFJPROG was not specified.

JF002:
Closing because I received no answer since end of July, and because another docker build has been integrated.

PR #40, opened 2020-07-20
Music

piggz:


JF002:
Thanks, great app and works out of the box with the application Media in SailfishOS on PinePhone!

PR #41, opened 2020-07-21
Correct including path

memiks:
Hi JF002,

When I tried to compile the new code, I was not able since you include "sdk/" in your path, and mine is in another folder.
So I corrected headers inclusion

BR,
Memiks.


JF002:
Thanks for this pull request!
I really have to find a wait to prevent this kind of error (this is not the first time I commit wrong include path...).

PR #47, opened 2020-07-25
Fixed typo, fixed caps and clarified one line more + iOS caps fix (not IoS)

DRG-developer:
I don't know what you mean with line 86, because it's the same as, or different than. I just assumed the same as

Basically i did nothing that would actually help the project, but since i don't own the skills or a pinewatch to contribute code (i can't even write a wayland compositor yet, and my only watch related projects are 2 garmin watchfaces with 8k combined downloads) ai did this, to at least help you guys with something.

And i do not expect/want to appear on this list of peolle, as i did nothing useful.


JF002:
Thanks for this PR, which I personally consider useful. As you have already noticed, I'm not that good at writing docs in English, and I'm happy that people take some time to improve them!

PR #61, opened 2020-09-07
Hardcoded sdk

jacobrosenthal:
I was going through the guide and was forced to move and rename my sdk directory


JF002:
Hi, thanks for this PR and sorry for the delay! 
This #include error should be already fixed in the branch ```develop```. Could you have a look and confirm it works for you?

PR #64, opened 2020-09-23
Update buildAndProgram : Use latest ARM-GCC toolcahain to build.

Rajssss:
Using the older (8.x.x) version of toolchain causes unexpected "template" errors, So we should always use version 9.x and above.


JF002:
Thanks!

PR #68, opened 2020-10-02
Rename folders to follow a consistent style

Avamander:
This PR renames most of the inconsistent folders (`FreeRTOS` was left) and fixes the imports that were broken.


JF002:
Thx for this huge PR!

`FreeRTOS` is the folder name used in the NRF52-SDK, we better not rename it :-)

PR #90, opened 2020-10-09
Addition of commented regions to Watchface for inclusion of new widgets

ZephyrLabs:
Added commented regions for adding new widgets in Clock.cpp (/develop/src/displayapp/screens/Clock.cpp)

### DIFF
--- a/src/displayapp/screens/Clock.cpp
+++ b/src/displayapp/screens/Clock.cpp
@@ -1,5 +1,4 @@
 #include <cstdio>
-
 #include <libs/date/includes/date/date.h>
 #include "components/datetime/DateTimeController.h"
 #include <libs/lvgl/lvgl.h>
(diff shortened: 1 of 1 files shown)


JF002:
I'm not sure this is the right place for this kind of comments.

I understand these comments makes sense when this code is included into the web based simulator for people that are designing and implementing their own watch face based on the screen `Clock`.

But outside of this context, when you just want to read the code to see how it works, I think they are confusing. The comments should explain the code when necessary, not instruct how to create a new app or how to extend the current one.

I'm not saying that these comments are useless, on the contrary! I think we should find a better place for them. Maybe in a tutorial, article or documentation explaining how to create a custom watch face based on InfiniTime? Or in the documentation of the base class `Screen`, where we can explain how to extend the class to implement a new app or a new watch face? Or in a template file developers could use as the base of a new app?

Also, note that it's rarely a good idea to declare variables/widgets in the global scope. Most of the time, these variables can be declared as member variable of the class, in the `private` section of the class definition (header file). I know you had to use global variables because of a limitation of the web based simulator but I would like to avoid global variables as much as possible.

ZephyrLabs:
In that case, I shall add this copy of clock.cpp to the Infinitime watchface tutorial, many newcomers would find it useful while testing out the the online simulator and trying out watchface development by themselves :-)

Thank you!

PR #92, opened 2020-10-09
Fix TwiMaster lockup

TT-392:
This should fix the TWIM lockup problem. 

I left the timeout at 5000, in my testing this seems to be a number that isn't really noticeable, but also seems to not false trigger.

I did leave in a static global variable for the twimaddress and 2 hardcoded pins (line 75), good chance that this isn't the best way of doing this here. But that is probably best fixed by someone who knows more about C++ than me.

### DIFF
--- a/src/drivers/TwiMaster.cpp
+++ b/src/drivers/TwiMaster.cpp
@@ -5,6 +5,9 @@
 
 using namespace Pinetime::Drivers;
 
+
+static NRF_TWIM_Type* twimaddress;
+
 // TODO use shortcut to automatically send STOP when receive LastTX, for example
 // TODO use DMA/IRQ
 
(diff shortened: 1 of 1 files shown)


JF002:
Thanks for this fix! I think the code could be slightly modified to look a bit  more like C++ and avoid the static global variable. I'll try to have a look soon!

Also, do you think [this comment](https://github.com/JF002/Pinetime/issues/79#issuecomment-706568547) correspond to this bug, and that we can try to fix the issue using the workaround from the NRF anomaly documentation?

TT-392:
> Thanks for this fix! I think the code could be slightly modified to look a bit more like C++ and avoid the static global variable. I'll try to have a look soon!
> 
> Also, do you think [this comment](https://github.com/JF002/Pinetime/issues/79#issuecomment-706568547) correspond to this bug, and that we can try to fix the issue using the workaround from the NRF anomaly documentation?

It might be related. But I wouldn't be surprised if it is specifically a twim issue or maybe even related to the slave device getting stuck though. 

I think I also, found the nordic forum thread where I originally got the idea for this fix:
https://devzone.nordicsemi.com/f/nordic-q-a/23601/twi-stops-working---sda-bus-always-low

JF002:
@TT-392 Thank you very much for your PR and for all the information you provided in this PR and in https://github.com/JF002/Pinetime/issues/79, they were invaluable.
I've eventually implemented your fix in a C++ way in https://github.com/JF002/Pinetime/pull/111. I hope you don't mind if I close this one?

TT-392:
Sounds good

JF002:
Merged and release in 0.9.0. Thanks!

PR #108, opened 2020-10-20
Improved notification UI

JF002:
The goal of this PR is to provide a better notification UI. The current UI consists in a modal dialog box that can only display the first 18 characters of the last notification received.

This new UI provide a new app that can display more content (full screen) and the last 5 notification received.

When a new notification is received 
 - while the Clock app is shown, the notification is displayed for 5s.
 - while the display is OFF, the display is first switched on and the notification is displayed for 5s.
 - while another app is shown, nothing is displayed (don't want to disturb the user). When the user goes back to the Clock app, an icon is displayed on the top left to notify him that a notification is available. He can then see all the notifications using the new Notification app.

Here is a demo of the current implementation : https://video.codingfield.com/videos/watch/5395d13a-a287-4689-9ec6-750f291e23b8

My goal here is to provide a simple notification UI that is usable (the previous one was just a proof of concept, but barely usable). Other ideas and improvements have been discussed on the Pinetime chat room, but I think we can implement it in the future (maybe someone with more UI experience will be able to do something a bit more shiny ?).

What do you think of it? Should I changed/improve something?

TODO:

- [x] Improve BLE API to support notifications with more than 18 characters
- [x] Fix https://github.com/JF002/Pinetime/issues/109
- [ ] [OPTIONAL] Vibrate when a notification is received
- [ ] [OPTIONAL] Display more than one notification on 1 screen of there's enough room


JF002:
Here is a new video of the Notification app : https://video.codingfield.com/videos/watch/fb8c766c-7cb6-4ad5-af62-8f9a3dc85233.

PR #110, opened 2020-10-23
Update main.cpp

maidenone:
changed "include Logging" to "include logging", now compiles with RTT ;)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -29,7 +29,7 @@
 
 
 #if NRF_LOG_ENABLED
-#include "Logging/NrfLogger.h"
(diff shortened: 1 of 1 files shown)


Avamander:
Nice, thanks! Related to https://github.com/JF002/Pinetime/issues/107 again.

JF002:
This is why I don't like `#ifdefs` : I always forgot to fix some parts of the code! Thanks, @maidenone !

jones139:
Thanks for this - I will try it because I had trouble with (I think) RTT crashing the watch after a few seconds when I enabled it - I might have done it wrong!

PR #112, opened 2020-10-24
Update arm-gcc toolchain in example

SteveAmor:
The difference in the example caused some confusion for at least one person.


JF002:
Thanks for this PR and sorry for the confusion ;)

PR #120, opened 2020-11-01
Fixed typo (Transfert -> Transfer)

ryanplusplus:
I assume this was intended to be "Transfer"


JF002:
Transfer is one of these word that are similar in French and English, but not exactly the same... It's written "Transfert" in French.

Thanks for this PR!

PR #124, opened 2020-11-03
Reduced interval where the display shows the previous time on wake

IoTPanic:
Now when you wake up your watch you'll notice the correct time is waiting for you! The display would show the last time displayed for the amount of time it took to start advertising over BLE and wake peripherals. Note that I think the final change should be some restructuring but now the time updating on wakeup isn't annoying.

### DIFF
--- a/src/systemtask/SystemTask.cpp
+++ b/src/systemtask/SystemTask.cpp
@@ -193,7 +193,7 @@ void SystemTask::Work() {
 
     monitor.Process();
     uint32_t systick_counter = nrf_rtc_counter_get(portNRF_RTC_REG);
-    dateTimeController.UpdateTime(nrf_rtc_counter_get(systick_counter));
+    dateTimeController.UpdateTime(systick_counter);
     if(!nrf_gpio_pin_read(pinButton))
       watchdog.Kick();
   }
--- a/src/systemtask/SystemTask.cpp
+++ b/src/systemtask/SystemTask.cpp
@@ -193,7 +193,7 @@ void SystemTask::Work() {
 
     monitor.Process();
     uint32_t systick_counter = nrf_rtc_counter_get(portNRF_RTC_REG);
-    dateTimeController.UpdateTime(nrf_rtc_counter_get(systick_counter);
+    dateTimeController.UpdateTime(nrf_rtc_counter_get(systick_counter));
     if(!nrf_gpio_pin_read(pinButton))
       watchdog.Kick();
   }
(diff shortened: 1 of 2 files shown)


src/systemtask/SystemTask.cpp
```diff
@@ -190,12 +191,8 @@ void SystemTask::Work() {
       }
     }
 
-    uint32_t systick_counter = nrf_rtc_counter_get(portNRF_RTC_REG);
-    dateTimeController.UpdateTime(systick_counter);
-    batteryController.Update();
-
     monitor.Process();
-
+    dateTimeController.UpdateTime(nrf_rtc_counter_get(portNRF_RTC_REG););
```
pfeerick, line 195: Semi-colon after `_REG)` shouldn't be there... only the end one ;)

pfeerick (changes requested):
Other than this, it appears to work great... I no longer see the remnants of the previous time as the screen turns on, but the current time - where it be valid or post-boot/pre-sync.

src/systemtask/SystemTask.cpp
IoTPanic: Weird, I just checked the branch out and on that line I only see one semicolon, do you know what happened?

JF002: I confirm, there is an extra ";" after _REG. Forgot to commit/push one last modification ? ;)
Other than that, I've just tested your branch and it works perfectly, the result is very pleasant, Thanks!

Oh and for info, in my code, I explicitely declared the variable 'systick_counter' to tell the reader of the code what's the meaning of `nrf_rtc_counter_get(portNRF_RTC_REG)`. It looks more readable to me.

pfeerick: It is more readable - and it would nice if the original two line version stayed.

IoTPanic: Done! Latest commit keeps the two line version

pfeerick:
Latest version works fine. Now, if the 2.6MB MP4 didn't encode to 18.9MB (this bigger than GitHubs 10MB upload limit), you'd be able to see the difference below, but such is life... a link to the video will have to do!

[Before and after video](https://streamable.com/ekbznx)

pfeerick (approved):
Still working great... and the `.gitignore` change makes one less less for me to do... not bumping this PR at all! 😇

JF002:
Sorry for the delay to merge this PR, and thanks!

Issue #119, opened 2020-10-31
Battery level always shows 100% or 0%


thelovebug:
Firmware 0.8.3 running on a sealed PineTime

The battery level on the information screen (bottom left icon on the first menu) always shows 100% or 0%.  I can't tell you at what point the level changes, but it was showing 100% for at least a day before changing to 0%, and it's been at least 3 days since then.

The battery icon on the main time view seems to reflect the battery level correctly, however.

![400097000710_118170](https://user-images.githubusercontent.com/2915687/97778069-8d8a4880-1b6c-11eb-8696-ad008af750c5.jpg)

JF002:
Thanks for your feedback! I'll have a look when I have a bit of time ;)

IoTPanic:
+1 Same issue, battery state in the battery icon on home seems fine, however after updating from 0.7.1 to 0.8.3 broke the menu battery percentage as pictured in thelovebugs post. Hope this tad extra info helps.

PR #125, opened 2020-11-03
Fixed battery percentage readout in the menu display

IoTPanic:
Fix to #119! Simple fix.

### DIFF
--- a/src/displayapp/screens/SystemInfo.cpp
+++ b/src/displayapp/screens/SystemInfo.cpp
@@ -43,10 +43,9 @@ bool SystemInfo::OnTouchEvent(Pinetime::Applications::TouchEvents event) {
 }
 
 std::unique_ptr<Screen> SystemInfo::CreateScreen1() {
-  auto batteryPercentF = batteryController.PercentRemaining();
-  uint16_t batteryPercent = 0;
-  if(batteryPercentF > 100.0f) batteryPercent = 100;
-  else if(batteryPercentF < 0.0f) batteryPercent = 0;
+  auto batteryPercent = batteryController.PercentRemaining();
+  if(batteryPercent > 100.0f) batteryPercent = 100;
+  else if(batteryPercent < 0.0f) batteryPercent = 0;
 
   uint8_t brightness = 0;
   switch(brightnessController.Level()) {
@@ -98,7 +97,7 @@ std::unique_ptr<Screen> SystemInfo::CreateScreen1() {
           dateTimeController.Day(), static_cast<uint8_t>(dateTimeController.Month()), dateTimeController.Year(),
           dateTimeController.Hours(), dateTimeController.Minutes(), dateTimeController.Seconds(),
           uptimeDays, uptimeHours, uptimeMinutes, uptimeSeconds,
-          batteryPercent, brightness, resetReason);
+          (int) batteryPercent, brightness, resetReason);
 
   return std::unique_ptr<Screen>(new Screens::Label(app, t1));
 }
--- a/src/displayapp/screens/SystemInfo.cpp
+++ b/src/displayapp/screens/SystemInfo.cpp
@@ -43,7 +43,7 @@ bool SystemInfo::OnTouchEvent(Pinetime::Applications::TouchEvents event) {
 }
 
 std::unique_ptr<Screen> SystemInfo::CreateScreen1() {
-  int8_t batteryPercent = (int)batteryController.PercentRemaining();
+  auto batteryPercent = static_cast<uint8_t>(batteryController.PercentRemaining());
 
   uint8_t brightness = 0;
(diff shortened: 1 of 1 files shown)


pfeerick:
Looks like this should fix the battery state showing only 0% or 100% (also present on 0.9.0)

src/displayapp/screens/SystemInfo.cpp
```diff
@@ -43,10 +43,9 @@ bool SystemInfo::OnTouchEvent(Pinetime::Applications::TouchEvents event) {
 }
 
 std::unique_ptr<Screen> SystemInfo::CreateScreen1() {
-  auto batteryPercentF = batteryController.PercentRemaining();
-  uint16_t batteryPercent = 0;
-  if(batteryPercentF > 100.0f) batteryPercent = 100;
-  else if(batteryPercentF < 0.0f) batteryPercent = 0;
+  auto batteryPercent = batteryController.PercentRemaining();
+  if(batteryPercent > 100.0f) batteryPercent = 100;
+  else if(batteryPercent < 0.0f) batteryPercent = 0;
```
okaestne, line 48: We can remove the boundary check completely, as it is already checked by `Battery::Update()`

pfeerick: Good catch... looks like

https://github.com/JF002/Pinetime/blob/f0e1f98823e41bfc2d9743fa8de70c882f26f93b/src/components/battery/BatteryController.cpp#L39-L40

will prevent it going out of range, and can thus be used directly. This makes it so only the conversion from float to int is needed and reducing this down to just `int8_t batteryPercent = (int)batteryController.PercentRemaining();` ??

IoTPanic:
@okaestne @pfeerick that is a good catch, thats my bad I was just trying to fix the broken fragment and didn't dig into the controller. Please review my change, I copied the fragment peter provided, it compiles but I have yet to test it.

src/displayapp/screens/SystemInfo.cpp
```diff
@@ -98,7 +95,7 @@ std::unique_ptr<Screen> SystemInfo::CreateScreen1() {
           dateTimeController.Day(), static_cast<uint8_t>(dateTimeController.Month()), dateTimeController.Year(),
           dateTimeController.Hours(), dateTimeController.Minutes(), dateTimeController.Seconds(),
           uptimeDays, uptimeHours, uptimeMinutes, uptimeSeconds,
-          batteryPercent, brightness, resetReason);
+          (int) batteryPercent, brightness, resetReason);
```
pfeerick, line 98: The cast isn't needed here any more, as it is done as part of L46...

pfeerick (changes requested):
Looks good to me... just remove that last bit and that should be it. This PR, with that additional removal, works fine for me. 
pr-125-test

src/displayapp/screens/SystemInfo.cpp
JF002: Not a big deal, but I prefer the C++ syntax  :)
`auto batteryPercent = static_cast<uint8_t>(batteryController.PercentRemaining());`

pfeerick:
btw, @IoTPanic if you change `Fix to #119!` to `Fixes #119!`, GitHub _should_ auto-link this PR to that issue, and close it when the PR is accepted. ;) It seems it is *that* fussy about the syntax... you can [see more here](https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) if you are interested in finding out more about that.

IoTPanic:
Thank you very much @pfeerick, I am not as familiar with the github collaboration features as I should be. I made both changes as outlined in the review. Compiled not tested, if I go in to grab the charger and wake others up I might get smacked. I was trying to get my watch to die because I got the heart rate sensor kinda working, then I tried someone else's code and now the heart rate sensor doesn't work at all.

pfeerick:
No problem ;) I'm still learning the ins and outs of git and github also. If you have a look at how your last commit is shown, you'll see it has a dotted underline under `Fixes`, and one of the tooltips that comes up says it closes the linked issue. And if you go to the issue, it will say that it will be closed by this PR. 

I'm actually waiting for mine to go flat also... as it currently says it is fully charged (100%)... but since that last change was just fancifying the cast, there's no reason to expect it to behave any different. 

image

JF002:
Thanks everyone for this PR, fix and awesome collaborative review! 

My PT is also fully charged, I'll let it run for a while to see it displays less than 100% to confirm it works, but from what I can see in the code, it should be OK :)

pfeerick:
I got distracted and didn't end up posting the photo I took this morning of the latest PR state... the PineTime decided to show 100% briefly still, then started quivering in the ~90%s, and several hours later is now is showing78%... 

image

image

PR #128, opened 2020-11-07
Add Cyrillic symbols into jetbrains font

AirHamster:
In goal to see Russian text in notifications I converted fonts with Cyrillic symbols.

You can see that there are many changes in old Latin symbols.
But I used the newest font version from Jetbrains. It was updated ~3 month ago. And looks like they changed it a little.

<a href="url"></a>


JF002:
Thank you very much for this PR! It's a first step towards a internationalized version of InfiniTime!
What's the impact of this PR on the code size of the application?

AirHamster:
+ 1584 bytes.

My opinion about multi-language support. I think in the future will be better to make it configurable at the compile time. For example, you define your firmware with en_ru or en_fin or en_tur (russian, finland, turkish - examples of alphabets that have non-latin symbols) and offline version of font converter prepares .c file with two languages array.

Anyway, when Infinitime will grow, some configurator will be needed. Different peoples want different languages, apps and etc.

JF002:
I'm really sorry it took so long to review this PR (health issue, job,...).

The impact on the binary is significant impact but not too big. The total binary size goes from ~264kB to 266kB, we still have a lot of room.
In the long term, in order to add more characters and alphabets, we'll certainly need to implement something smarter (like what you described just above, or storing the fonts on the external spi flash, for example).

I hope this addition will give the motivation to other developpers (maybe you, @AirHamster) to implement the "internationalisation" (is that the correct word?) of InfiniTime

PR #131, opened 2020-11-09
Fix companion app documentation links broken during #68

pfeerick:
As it says on the tin... fixes some links broken during path changes in #68 ...


JF002:
Thanks!

PR #133, opened 2020-11-14
fix error install cmake

miguemoya:
When installing Cmake you get Error: The `add-path` command is deprecated and will be disabled on November 16th...
And the file pinetime-mcuboot-app-dfu.zip was being generated as pinetime-mcuboot-app-dfu.zip.zip


ryanplusplus:
Make sure that pull requests are against `develop` and not `master`

miguemoya:
Close by #134

PR #134, opened 2020-11-14
Fix error install Cmake  main.yml

miguemoya:
When installing Cmake you get Error: The add-path command is deprecated and will be disabled on November 16th...
And the file pinetime-mcuboot-app-dfu.zip was being generated as pinetime-mcuboot-app-dfu.zip.zip


JF002:
Thanks!

PR #135, opened 2020-11-15
Flash the correct file when running FLASH_pinetime-app with JLink

ryanplusplus:
Without this change it attempts to flash using `pinetime-app.hex` (which doesn't exist) instead of `pinetime-app-${version}.hex` (which does exist).


JF002:
Thanks!

PR #136, opened 2020-11-15
Includes cleanup

okaestne:
This PR cleans up all `#include`s.

* remove all unused includes
* remove duplicate includes (in header *and* source)
* add missing includes, that previously were included indirectly
* fix circular includes by using forward-declarations
* use brackets (`<...>`) for libraries (also the included lvgl, date, NimBLE) and other external code; quotes (`"..."`) for own code.
* sort includes:
  * source files include their header files first
  * library includes next
  * own code includes last
  * (most of the time) sorted alphabetically

This also fixes compilation complications caused by NimBLE's `min` and `max` macros by undefining the macros behind the includes. See also: https://github.com/apache/mynewt-nimble/pull/866 and https://github.com/espressif/esp-nimble/issues/17

Let me know if this major change causes any regressions, thanks :)


Avamander:
The auto-sorting seems weird to me, the very least I'd place `<>` higher than `""`.

Avamander:
NimBLE's `min` and `max` sounds to me like a bug that should be reported upstream as well.

src/components/ble/MusicService.h
```diff
@@ -31,40 +34,40 @@ namespace Pinetime {
     class SystemTask;
   }
   namespace Controllers {
-    
+
```
Avamander, line 37: Seems like a potential autoformatter fight, I wouldn't include this in this PR

src/displayapp/screens/Music.cpp
```diff
@@ -47,7 +51,7 @@ inline void lv_img_set_src_arr(lv_obj_t *img, const lv_img_dsc_t *src_img) {
  */
 Music::Music(Pinetime::Applications::DisplayApp *app, Pinetime::Controllers::MusicService &music) : Screen(app), musicService(music) {
   lv_obj_t *label;
-  
+
```
Avamander, line 54: Same here

src/displayapp/screens/Tab.h
```diff
@@ -1,22 +1,24 @@
...
           explicit Tab(DisplayApp* app, Components::Gfx& gfx);
           ~Tab() override;
-          void Refresh(bool fullRefresh) override;
+          void Refresh(bool fullRefresh);
           void OnObjectEvent(lv_obj_t* obj, lv_event_t event);
 
-        private:
-
+        protected:
+          Components::Gfx& gfx;
```
Avamander, line 21: This isn't very nice compared to the rest and when Components will contain more members it's going to have to be moved back

src/components/ble/MusicService.h
Avamander: At some point cmake-format should be configured and applied, and that'll act as an authoritative source.

src/displayapp/screens/Tab.cpp
```diff
@@ -19,7 +16,7 @@ extern lv_font_t jetbrains_mono_bold_20;
 //  screen->OnObjectEvent(obj, event);
 //}
 
-Tab::Tab(DisplayApp* app, Pinetime::Components::Gfx &gfx) : Screen(app, gfx) {
+Tab::Tab(DisplayApp* app, Pinetime::Components::Gfx &gfx) : Screen(app), gfx(gfx) {
```
Avamander, line 19: To be honest, this is a weird autoformatting. The duplicated `gfx` doesn't seem to be adding any clarity.

src/displayapp/screens/Tab.h
okaestne: Yeah there I had some question marks regarding the whole thing. The Tab class isn't used anywhere yet, I wanted to keep it compile-able though, so I added the Gfx object.  As it looks like a base class for me, I set it as protected, so that it can be used by child-classes.

src/displayapp/screens/Tab.cpp
okaestne: No it's correct. The screen constructor only takes one argument, which is the DisplayApp pointer `app`. `gfx` is the member reference-variable, that is initialized with the parameter which is also called `gfx`.

src/components/ble/MusicService.h
okaestne:
```suggestion
    
```

src/displayapp/screens/Music.cpp
okaestne:
```suggestion
  
```

src/displayapp/screens/Tab.h
JF002: I confirm : the Tab class is not used/built anymore and I can be safely removed. The class Gfx is not used in the main firmware either. I used this class to draw pixels on the display using the class before I integrated LVGL. 
But do not remove Gfx, it is still use as a lightweight UI class in pinetime-graphics

So, if you added this protected variable only for the tab class, I think you can safely remove the variable **and** the Tab Screen :)

src/displayapp/screens/Tab.cpp
JF002: As I wrote above, I think we can remove this Tab class and forget about it :)

JF002:
> The auto-sorting seems weird to me, the very least I'd place `<>` higher than `""`.

Is this the result of an automatic format tool?

joaquimorg (approved):

JF002:
Sorry for the delay, and thank you very much for this PR!

PR #138, opened 2020-11-18
Use GitHub Pages for Docs

pfeerick:
Was playing around with gh-pages support, and thought I'd see if this is worth pursuing further. 

Basically the documentation 'site' is generated from all the markdown docs already existing, and it is just a matter of (in your case) specifying the `develop` branch and `/ (root)` as the GitHub Pages source (under the repo settings) for this to be used if merged. 

*i.e.*
image

https://pfeerick.github.io/InfiniTime/

Obviously the last change will be change the url... although it would probably be a good idea to rename the repo before that happens (if that is going to happen), as I don't believe gh-page urls redirect if the source repo moves... when I renamed a repo in the past github redirected properly... gh-pages just wants to be different. 

If you like this approach, I'll continue further reading through the documentation as I spotted some other typos or formatting glitches... otherwise, I'll start again, and just fix up the docs as they are...

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,36 +43,14 @@ As of now, here is the list of achievements of this project:
  - **[Experimental]** Bootloader based on [MCUBoot](https://juullabs-oss.github.io/mcuboot/)
  
 ## Documentation
+[Documentation site](https://pfeerick.github.io/InfiniTime/): For information on how to use, build, and program InfiniTime.
 
-### Develop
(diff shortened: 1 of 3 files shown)


JF002:
It looks nice! I'll try to have a better look at this soon !

pfeerick:
Closing this in favour of having #138 implemented.

PR #140, opened 2020-11-21
Add answer call notification

roka:
Adds the ability to answer calls via a notification that works with the app https://github.com/t0tten/PinetimeApp


Avamander:
The commit history is not clean, will you clean it up or should it be squash merged?

src/displayapp/screens/Modal.cpp
```diff
@@ -41,13 +38,44 @@ void Modal::OnEvent(lv_obj_t *event_obj, lv_event_t evt) {
   if(evt == LV_EVENT_DELETE && event_obj == mbox) {
     Hide();
   } else if(evt == LV_EVENT_VALUE_CHANGED) {
-    /* A button was clicked */
-    lv_mbox_start_auto_close(mbox, 0);
-//    Hide();
+    if(event_obj == mbox) {
+      if(strcmp(lv_mbox_get_active_btn_text(event_obj), this->positiveButton.c_str()) == 0) {
+        if(alertNotificationService != nullptr)
```
Avamander, line 43: Does not follow the agreed brace style.

Avamander:
Thanks for the brace fix, can you please also resolve the conflicts?

JF002:
Hi @roka, and thank you very much for your PR!
As @Avamander said, you should try to fix conflicts from current develop branch. These conflicts come from another PR that fixes issues with min()/max() macro from NimBLE. This way, `mynewt_min()` and `mynewt_max()` won't be needed anymore.

Is your Android companion app working for now? If yes, I would like to try it :) Do you have a package I can install on my phone?

roka:
Hi @JF002,
The merge conflicts have been fixed.
An apk for the application can be downloaded from https://github.com/t0tten/PinetimeApp/releases/tag/0.01
Currently you will get a notification on the watch if you receive a message on SMS, Snapchat, facebook messenger or call.
You need to have the application in the foreground for Spotify to work properly. 

Let me know if the app is working OK:)

src/components/ble/AlertNotificationClient.h
```diff
@@ -1,4 +1,4 @@
-#pragma once
+#pragma once
```
okaestne, line 1: here are some (178) trailing spaces

src/components/ble/AlertNotificationClient.h
```diff
@@ -83,4 +83,4 @@ namespace Pinetime {
         bool isDescriptorFound = false;
     };
   }
-}
\ No newline at end of file
+}
```
okaestne, line 86: why removed newline?

src/components/ble/AlertNotificationService.h
```diff
@@ -1,4 +1,4 @@
-#pragma once
+#pragma once
```
okaestne, line 1: also many trailing spaces

JF002:
@roka Sorry again for taking so long to review your PR. I tried to focus on the bootloader these last few weeks.

Are you the developer of this Android application ? Anyway, I installed the apk, and it detected and connected to my pinetime flawlessly! :+1:  I couldn't try the new call notification because... I have only one phone at home for now (I'll ask my wife to call me later :) ). It might be useful to add the possibility to trig a call notification from the app screen.

So, I had  a look the your PR. Before diving into the code, I was wondering if the addition you made to the AlertNotificationService is compliant with the BLE specification? 
According to ble specification of the [alert notification service](https://github.com/oesmith/gatt-xml/blob/master/org.bluetooth.service.alert_notification.xml), characteristic `0x2a47` is [new alert category](https://github.com/oesmith/gatt-xml/blob/master/org.bluetooth.characteristic.supported_new_alert_category.xml), which allows the host (PineTime) to specify which types of notifications it supports.

In this case, if I understand correctly, this characteristic (event, 0x2a47) is used notified by InfiniTime to notify the app that the call has been accepted or not?

It's a totally valid use-case, but I wouldn't use a characteristic ID that is part of the BLE specification for a "custom" use.
The annoying thing is that this use-case doesn't seem to be part of the specification. I don't know if the BLE spec allows to chose another ID to extend the specification, and if yes, which one.

Here is the PDF file that describe the alert notification service (you can find it on the BLE website, but it's well hidden) : [ANS_SPEC_V10.pdf](https://github.com/JF002/Pinetime/files/5735895/ANS_SPEC_V10.pdf)

petterhs:
I would like to help getting this merged into Infinitime. I have tested the call functionality with the android application  https://github.com/t0tten/PinetimeApp/ and it is so far working as intended. I don't use facebook messenger or Snapchat and I use Signal for SMS, so haven't gotten to test the notifications from messages.

What is working:
Calling my android phone with the companion app from my Pinephone prompts a window with caller ID and one button for accepting the call and one for rejecting. Both accepting and rejecting the call works from the PineTime. If the call is accepted I get a new notification on the PineTime showing the caller ID or number if the ID is unknown. This notification is also prompted when doing an outgoing call.

Avamander:
@petterhs The issue currently is that:

> It's a totally valid use-case, but I wouldn't use a characteristic ID that is part of the BLE specification for a "custom" use.
The annoying thing is that this use-case doesn't seem to be part of the specification. I don't know if the BLE spec allows to chose another ID to extend the specification, and if yes, which one.

This shouldn't be done, it can cause weird issues later on and it'll be a massive pain to fix later.

JF002:
> I would like to help getting this merged into Infinitime.

I would like to merge this feature too, and I appreciate your PR!

I'm just worried about the implementation that does not seem to comply with the standard specification of the Alert Notification Service.

petterhs:
I have taken a look through the specification now and agree that there does not seem to be any 16-bit UUID for this usecase in there. The usecase is implemented in different bluetooth headsets though and working across brands, so there must be a standard out there, right? Maybe not BLE?

As I understand it there are 16-bit UUID's for Bluetooth SIG-Adopted Attributes, and 128-bit UUID's for custom attribute types or (vendor-specific UUID's) [[Bluetooth GATT tutorial]](https://www.novelbits.io/bluetooth-gatt-services-characteristics/)

 Is there an issue with the community choosing a 128-bit UUID for this usecase(And other future usecases outside of the specifications)? The problem, as I see it, will be to be consistent with the UUID's across PT implementations and companion applications.

Avamander:
@petterhs 
> Is there an issue with the community choosing a 128-bit UUID for this usecase(And other future usecases outside of the specifications)? 

That's how it's theoretically supposed to work. If you do proprietary, you don't mess with known standard services and standard characteristics. 

Alternatively, it's probably okay (nobody's going to beat us) to place a unspecified/unknown characteristic under a known service that is related to this function though. Depends on if you wish to signify it's potential purpose more than you wish to comply with BT SIG's unenforceable rules. 

>  The problem, as I see it, will be to be consistent with the UUID's across PT implementations and companion applications.

That's not really a concern because InfiniTime would be the first.

petterhs:
@Avamander Ok i see. It's not crucial to follow the specifications then

I have been trying to see what other open source smartwatches and companion apps has done and try to get some inspiration. 

Most of the ones i have seen is either:

1.  using a custom 16-bit UUID like this Pebble implementation: [pebble_protocol.h](https://github.com/pebble-dev/RebbleOS/blob/680db4ff69082c854bb1090ab11cb769d2bf88b2/rcore/protocol/pebble_protocol.h#L14) where they use `0x0021` for messages to the phone (translates to 0000**0021**-0000-1000-8000-00805f9b34fb)
2. using a custom 128-bit UUID but still using the base 128-bit UUID 0000**0000**-0000-1000-8000-00805f9b34fb and changing the bold numbers. Like some of the supported devices by Gadgetbridge: [WatchXPlus(0xa800)](https://github.com/Freeyourgadget/Gadgetbridge/blob/27f8d195c11f17bd8de4b28ec4ce7d286142e997/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/lenovo/watchxplus/WatchXPlusConstants.java#L24), [watch9(0xa800)](https://github.com/Freeyourgadget/Gadgetbridge/blob/27f8d195c11f17bd8de4b28ec4ce7d286142e997/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/watch9/Watch9Constants.java#L22) and [xwatch(0xfff0)](https://github.com/Freeyourgadget/Gadgetbridge/blob/27f8d195c11f17bd8de4b28ec4ce7d286142e997/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/xwatch/XWatchService.java#L25)

Is there a reason people is using the 16-bit UUID or its 128-bit equivalent instead of just creating a completely random 128-bit UUID, like i presume was done when implementing the music functionality in Infinitime. [Pinetime music service UUID:c7e50000-78fc-48fe-8e23-43b37a1942d0 ](https://github.com/JF002/Pinetime/blob/50ae0ae5e073ac48652e6c26549f9b19655e8da3/src/components/ble/MusicService.h#L30)

If not, I can create a random UUID and implement the changes to this branch and the companion app. Unless you want to do it @roka and @t0tten. What do you think @JF002

Avamander:
@petterhs

> It's not crucial to follow the specifications then

It's not strictly enforced, yes, stuff won't die and there's a lot of devices out there that advertise a characteristic but deviate from the standard massively. **But** as I've said before, the very least, we shouldn't reuse a standard 128-bit UUID if we aren't implementing the standard, reusing the base 96-bits is less grave. However, read the last paragraph please.

> Is there a reason people is using the 16-bit UUID or its 128-bit equivalent instead of just creating a completely random 128-bit UUID

They probably haven't spent much thought on it and picked something "close" (and probably pray that the UUIDs won't be used in the future).

> If not, I can create a random UUID and implement the changes to this branch and the companion app.

I think incrementing the music service's lower 96-bit base by one and resetting 16-bits would be better, then they get listed next to after one another in the order of addition. So `c7e5xxxx-78fc-48fe-8e23-43b37a1942d**1**` where xxxx starts from 0. Random is fine too if you think that's needed.

JF002:
@Avamander  @petterhs Thanks for this analysis about custom UUIDs !
My understanding is that we should use the standard 16bits UUIDs only to implement the BLE specifications. Meaning that if we want to derive from the spec or implement something new, we should choose a custom 128bits UUID.

We started a debat about this right here : https://github.com/JF002/Pinetime/pull/163#discussion_r556798887. Stay tuned :)

JF002:
I'm closing this PR as I've just merged #161 which includes this PR. Thanks for your work and patience!

PR #141, opened 2020-11-22
Paddle app

ZephyrLabs:
This pull request is for the Paddle app to be included in Infinitime


JF002:
Thanks for your PR ;) I'm sorry it took me so long to review it!

I'm also sorry, I merged some other PRs and now, there're some conflicts, could you try to fix them?

About the code, I noticed that some variables are declared in the global scope in paddle.cpp and paddle.h. I think all these variables should be declared as member variables of the class Paddle. Some of them (constants) should also be declared as `static contexpr` variables to ensure they are stored in flash and not in RAM. It might be a good opportunity to try to find a better name (more explicit) for some of these them (it's not easy to understand the meaning of pos_y, y_1 and dx without reading the whole code).

This last point is maybe not mandatory, but I think you could avoid allocating the buffers for the ball and paddle bitmap by using lvgl objects (at least, the paddle could be replaced by a simple lv_line).

ZephyrLabs:
sure will do, 
I will replay here once those conflicts have been resolved :) 

good day!

PR #143, opened 2020-11-26
Add doxyfile

tobrie:
This is to be able to easily generate html-docs from the source code documentation using the following commands.

```
$ cd src
$ doxygen
```

The files are then located in `src/html.`

It has helped me and may help others.


JF002:
Thanks! However, I'm wondering if we actually want to use doxygen to generate documentation? For now, most of the documentation is written into mardown (.md) files. Do we want to split the documentation (high level -> md, low level -> doxygen) or keep everything at the same place?
I know that some method have doxygen tag, they are also useful for some IDEs (like CLion) to display the documentation in tooltips.

This is an open question, I haven't thought yet how to manage the documentation in the long term.

redengin:
Documentation should be as close to the implementation as informative. While Doxygen (all the other alternatives) are sexy, keeping the focus on the code is more important. I only leverage doc-generators that leverage source to satisfy regulation.

My two cents:
Create Markdown documents to describe the large concepts
Use source-level comments to inform developers of the design(reference to a markdown) and intent (inline comments about what's going on)

As modern editors/IDEs already provide syntax completion, having a source-generated document simply makes yet-another-thing to maintain.

Avamander:
@redengin 

> While Doxygen (all the other alternatives) are sexy, keeping the focus on the code is more important. I only leverage doc-generators that leverage source to satisfy regulation.

Doxygen does leverage source for generation and the documentation in the code has to be good for it to work properly. 

The maintenance aspect is true, if it's not automated. But it can be.

Riksu9000:
We've decided not to add doxygen support for now. Doxygen is useful for generating documentation for libraries, however it's not clear how InfiniTime would benefit from it. Documenting how every function in InfiniTime works would be a lot of work, when the names of the functions should ideally be enough information.

PR #145, opened 2020-11-27
Fix build toolchain

tobrie:
with the help of @ZephyrLabs

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,13 +43,10 @@ jobs:
           restore-keys: ${{ runner.os }}-build-${{ env.cache-name }}
 
       - name: Install Embedded Arm Toolchain arm-none-eabi-gcc
-        if:   steps.cache-toolchain.outputs.cache-hit != 'true'  # Install toolchain if not found in cache
-        uses: fiam/arm-none-eabi-gcc@v1.0.2
-        with:
-          # GNU Embedded Toolchain for Arm release name, in the V-YYYY-qZ format (e.g. "9-2019-q4")
-          release: 9-2020-q2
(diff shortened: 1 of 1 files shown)


JF002:
What's the issue with the current github actions build?

I don't know/use Github Actions, so I'm not sure I know what I'm talking about, but I think the file was written to use the cache functionality of Github action (to avoid downloading the toolchain for each build), and I'm not sure your modification still uses it.

Could you give more info about the modification?

ZephyrLabs:
@JF002 the current github actions script used a inbuilt action to install arm-none-eabi-gcc compiler, that action is now deprecated due to an update in github actions itself, which does not permit the function from using the "add-path" (which is the actual deprecated feature)

this PR is for replacing it with standard linux commands that can be run instead as a replacement for that function :)

JF002:
Thanks for the info! 
Could you confirm that the toolchain is still cached and won't be downloaded/installed for each build?

JF002:
A response to my last question would be appreciated :)

pfeerick:
I haven't used caching with github actions, but judging by the changes and how the nrf52 SDK is installed, I suspect not - I think this PR will still *check* if it is cached, and download it regardless of if it is in the cache or not, rather than using a similar process to the nrf52 SDK install step.

JF002:
Wouldn't it be fair to not download the sdk if it is already in the cache, if possible ?

pfeerick:
Since there is the facility to do so, yes it would... it's a pity @tobrie didn't use the NRF52 installation stage as a reference, and just change the download commands... copy/paste/tweak/done.

pfeerick:
I believe this is no longer relevant due to #294.

PR #147, opened 2020-12-03
Paddle-app (updated)

ZephyrLabs:
addition of Paddle application into Infinitime.
changes:

**_I noticed that some variables are declared in the global scope in paddle.cpp and paddle.h. I think all these variables should_** **_be declared as member variables of the class Paddle._**

this has been changed, the variables have been moved to the Paddle class

**_Some of them (constants) should also be declared as static contexpr variables to ensure they are stored in flash and not_**
**_in RAM._**

there only 2 constants, and those have been declared under `const`, the rest need to be in RAM, as there are constantly changing while the app is running.

**_It might be a good opportunity to try to find a better name (more explicit) for some of these them (it's not easy to_** **_understand_ _the meaning of pos_y, y_1 and dx without reading the whole code)._**

I have tried my best to replace this with more understandable variable names, I won't be the judge of much more they are understandable to other users, and developers ;)

**_This last point is maybe not mandatory, but I think you could avoid allocating the buffers for the ball and paddle bitmap by_**
**_using lvgl objects (at least, the paddle could be replaced by a simple lv_line)._**

this hasn't been really addressed, they occupy a small amount of buffer space, its not too much noticeable, if there is a need to replace this with LVGL objects, it will be carried out through a future PR.

I hope this is better than before, I'm not an expert at coding, but this is the best I can offer in my current experience :) 
hoping for the best,
good day!


ZephyrLabs:
This is the compiled firmware, as seen it compiles successfully.
you can test this for app verification ;)

[pinetime-mcuboot-app-dfu(paddle_app).zip](https://github.com/JF002/Pinetime/files/5637085/pinetime-mcuboot-app-dfu.paddle_app.zip)

JF002:
Thanks for this remastered version of your PR :)

I would like to add a few comments:

- `x_paddle_bottom` and `x_paddle_top` are not used, they can be removed
- To be more consistent with the rest of the code, please use the PascalCase (`ballX` instead of `ball_x` for example).
- The naming `ball_x` and `y_paddle_top` are not consistent. `PaddleTopY` or `PaddleTopLine` seem more appropriate to me.
- `Val` could use a better naming. `scoreStr` or `scoreText`, for example?
- There are useless empty lines at the beginning of `Paddle::Refresh()` and `Paddle::OnTouchEvent()`
- Could you use `nullptr` instead of `NULL`?
- `lv_img_dsc_t paddle` and `lv_img_dsc_t ball` can also be member variables of the class Paddle (not static)
- As `paddle_map` and `ball_map` are used only into the implementation of Paddle, I think you could declare them in an anonymous namespace in Paddle.cpp. This will reduce the global namespace pollution:
At the beginning of Paddle.cpp:
```
namespace {
  const uint8_t paddle_map[] = {
    ...
  }

  const uint8_t ball_map[] = {
    ...
  }
}
```

> I hope this is better than before, I'm not an expert at coding, but this is the best I can offer in my current experience :)

Don't worry about that, I appreciate your contributions! And sorry if I look a bit picky, but I would like to keep the code as clean and consistent as possible ;)

ZephyrLabs:
@JF002 I have changed the things you have mentioned and corrected them all :)
in addition I have reduced the space taken by the paddle data from the old value to just 68 bytes by using a color palette of 2. reducing some unnecessary data :)

> Don't worry about that, I appreciate your contributions! And sorry if I look a bit picky, but I would like to keep the code as clean and consistent as possible ;)

I'm new to adding proper commits to the codebase, so I'll try to keep up with the level of tidyness ;)

JF002:
Thanks for the update, and sorry again it takes so long to review this PR!
2 small details:
 - there are still 3 unused blank line at lines 137, 139 and 188
 - Variables `PaddleBottomY`, `PaddleTopY`, `BallX` and `BallY` should not start with a capital letter

Otherwise, the code is fine. 

BUT... when I play the game, the ball never bounces on the paddle, it always goes through. Could you check it?

ZephyrLabs:
I've fixed the uppercase variables, they should be lowercase now, eg. `ballX` instead of `BallX`

also fixed a small bug that may have caused the paddle to not function properly, 

here is a zipped static binary image (not DFU package), with the fixes:

[pinetime-mcuboot-app-img(paddle).zip](https://github.com/JF002/Pinetime/files/5720492/pinetime-mcuboot-app-img.zip)

cheers!

JF002:
Thanks!

Issue #50, opened 2020-08-09
Add copying conditions to the firmware


endian-albin:
To help distributors follow the GPLv3 license, it would be helpful to add the [license document](https://www.gnu.org/licenses/gpl-3.0.txt) (or a link to it) along with a link to this repository in an About screen.

pfeerick:
This issue was closed by https://github.com/JF002/Pinetime/pull/151 ;)

ptman:
please close issue

MysteriousLog6:
Prithee closeth this issue

JF002:
Thanks for the reminder :)

PR #151, opened 2020-12-28
Add source code license and link to github to system app

nobodywasishere:
For #50


JF002:
Thanks!

Just wondering... maybe the first line "Hello from the developer" is not needed anymore? What do you think?

nobodywasishere:
When I first found it I was excited because it was like an easter egg. I left it in because of that (though maybe it doesn't have as much of an effect now) and because I didn't want to overwrite another developer's work. Maybe instead I could push it to a 4th page instead of squishing it on the same page?

JF002:
That's fine, thanks! At first, I put this label as a place holder to show that the widget can handle more than 2 pages, and I thought that would be funny is someone would see the message ;)

So... let's leave it there, thanks!

Issue #130, opened 2020-11-09
Music app, track text scrolls too fast with longer track titles.


dns53:
When using the music app on the watch the track artist and track name are displayed.
I have a track name with a longer than will display on the screen so this text is automatically scrolling forwards and backwards.

This seems to be moving the text forwards and backwards one character each frame with this appearing to update more than 20 characters each second making it hard to read.

Could we change the update speed of the text so the text updates at a more easy to read rate such as moving the text forward or backwards a character every second, every half second etc.

Avamander:
Duplicate of https://github.com/JF002/Pinetime/issues/97

pfeerick:
This can be closed as was fixed in #259.

MysteriousLog6:
prithee closeth this issue f'r t doest the not existeth anym're @JF002

Issue #97, opened 2020-10-11
Music: long titles move back and forth too fast


arteeh:
0.8.2

When playing a song with a title that's longer that the Pinetime screen can draw (about 18 or 19 characters?), the title moves back and forth extremely fast, making it nearly unreadable. 

Expected: The title moves back and forth slowly, so it's easy to read.

Ideal: The title moves forward, then when it's at the end, it immediately goes back to the front. (People don't read backwards so text moving backwards is unintuitive)

arteeh:
This may be fixed with #96

Avamander:
It is not fixed in the PR, it's something in LVGL. That's a topic for an another PR.

jedmijares:
I took a shot at adjusting this in #152, here's a video showing what it looks like: https://www.youtube.com/watch?v=PvBWN1-NBNU

PR #152, opened 2020-12-29
Slowly scrolls long titles/artists in music app

jedmijares:
Addresses #97 / #130.

### DIFF
--- a/src/displayapp/screens/Music.cpp
+++ b/src/displayapp/screens/Music.cpp
@@ -142,10 +142,10 @@ bool Music::OnButtonPushed() {
 }
 
 inline void scrollText(std::string text, unsigned int &textIndex, lv_obj_t *obj, unsigned int counter) {
-  if (text.length() > 20 && ((counter % 5) == 0)) {
-    lv_label_set_text(obj, (text.substr(textIndex, 19)).data());
+  if (text.length() > 20 && ((counter % 8) == 0)) {
+    lv_label_set_text(obj, ((text + " | " + text).substr(textIndex, 19)).data());
     textIndex++;
-    if (textIndex == text.length()) textIndex = 0;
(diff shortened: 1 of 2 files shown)


JF002:
Thanks for this PR and for [the video demonstrating that it works](https://github.com/JF002/Pinetime/issues/97#issuecomment-752286261) @jedmijares !

However, there is the function [lv_label_set_anim_speed](https://docs.lvgl.io/v6/en/html/object-types/label.html#_CPPv423lv_label_set_anim_speedP8lv_obj_t8uint16_t) in lv_label, couldn't it be used to achieve the same goal (without having to manually scroll the text) ?

jedmijares:
Interesting, I hadn't seen that before. I just tried implementing it and was indeed able to slow down the scrolling, but ran into another error where if the string is too long (not sure how long is too long), the string [will pause before displaying the whole thing and skip back to the beginning instead of scrolling back.](https://www.youtube.com/watch?v=3A7CCb1JHwc)

In any case, since that is another approach, perhaps I should open a separate pull request for comparison's sake?

JF002:
If the function provided by LVGL works the way we want, I would prefer LVGL implementation (so that we don't reinvent the wheel). 
If it's buffy or do not fit our needs, then, a custom implementation will be the way to go ;)

I agree, you can open a 2nd PR for the lvgl implementation and we'll see which one we'll merge into the project !

The video you linked correspond to the lvgl implementation?

JF002:
This was fixed in v1.0.

jedmijares:
Sounds good! I got kind of tied up with school so I'm glad someone else got around to it, ha.

JF002:
Yep, since then, we upgraded LVGL to version 7 and the whole UI have been greatly improved :)

PR #155, opened 2021-01-04
add motor controller and vibration on notifications

petterhs:
Adding the motor controller from [ZephyrLabs](https://github.com/ZephyrLabs/Watchfaces/tree/gh-pages/docs/Terminal) to enable vibrations on notifications.

More info from ZephyrLabs: https://zephyrlabs.github.io/Watchfaces/Terminal/

Not sure why this has not been put in a PR yet and tried to get in contact with @ZephyrLabs through this issue to ask why. https://github.com/ZephyrLabs/Watchfaces/issues/3

This works on my Pinetime atleast, though it also vibrates when opening the notification tab from the menu aswell. Could easily be modified if needed.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -374,6 +374,7 @@ list(APPEND SOURCE_FILES
         components/ble/ImmediateAlertService.cpp
         components/ble/ServiceDiscovery.cpp
         components/firmwarevalidator/FirmwareValidator.cpp
+        components/motor/MotorController.cpp
         drivers/Cst816s.cpp
         FreeRTOS/port.c
         FreeRTOS/port_cmsis_systick.c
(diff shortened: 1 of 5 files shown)


ZephyrLabs:
hello @petterhs,
I have just viewed your post, 
I have replied in that thread itself, please ask @jlukanc1 for more info

cheers!

jlukanc1:
How I set up vibration was kinda messy and could use some improvements like a global on/off to check before running the motor for those who would not prefer it. 

I'm not too sure how to cleanly PR it from an existing branch, so I guess I will make a new one when I have the time today.

JF002:
Thanks for this PR. 
However, I think the management of the vibrator could be improved.
 
- activating the motor will actually block the current task during the vibration. As this will mostly happen in the DisplayTask, this will make the UI sluggish. I think it would be better to drive the motor into a separate task or using a hardware timer.
- the vibration should only be activated when **receiving** the notification, not when opening the Notification app. I think this would be less confusing for the user.

petterhs:
Improvements in #167. Closing this PR

PR #157, opened 2021-01-05
Update main.yml

ComansServices:
allow the build to succeed 1.0.2 current version but currently broken


JF002:
Hi @ComansServices 

FYI, PR should be created on the branch **develop** instead of **master**. See [this page](https://github.com/JF002/Pinetime/blob/develop/doc/contribute.md) and [this one](https://github.com/JF002/Pinetime/blob/develop/doc/branches.md) about branches on the project.

If I understand correctly, you want to downgrade the toolchain version from 9-2020-q2 to 9-2019-q4 ?

PR #161, opened 2021-01-08
Call functionality with changed UUID for notification event characteristic

petterhs:
Rebased and squashed @t0tten and @roka's work in #140  into one commit to clean the commit history before continuing on their work. (Best if they do this themself, so my account doesn't end up in the commit. I don't want to take any credit from them)

Changed the custom notification event characteristic UUID under the AlertNotificationService. This is UUID can be used to send commands/notification-responses to a phone. Firstly, accept or decline a phonecall but can also be extended to other notification responses.

The new UUID: c7e50000-78fc-48fe-8e23-433b3a1942d1 is the UUID from the MusicService just iterated last character by one, as suggested by @Avamander. 

I wonder if it is better to seperate this usecase from AlertNotificationService and create a new "PhoneEventService" or similar, for all commands to a phone?


JF002:
We are discussing custom UUIDs in https://github.com/JF002/Pinetime/pull/163#discussion_r556798887. Changing the UUID might be needed when we decide how to choose a new UUID :)

src/components/ble/AlertNotificationService.h
```diff
@@ -24,8 +27,25 @@ namespace Pinetime {
         int OnAlert(uint16_t conn_handle, uint16_t attr_handle,
                                     struct ble_gatt_access_ctxt *ctxt);
 
+        void event(char event);
+
+        static const char EVENT_HANG_UP_CALL = 0x00;
+        static const char EVENT_ANSWER_CALL = 0x01;
```
piggz, line 33: From my experience with huami, perhaps the options should be hangup and ignore, not answer.  If you need to answer the call, you need the phone anyway.

petterhs: If you are wearing headphones it can be useful but not sure how often it is necessary. I can see myself out on a run/bicycle and using headphones for music and wanting to accept the call with the watch

piggz: Yeah, I cant disagree

roka:
> Rebased and squashed @t0tten and @roka's work in #140 into one commit to clean the commit history before continuing on their work. (Best if they do this themself, so my account doesn't end up in the commit. I don't want to take any credit from them)

Hi,
I've talked to @t0tten and you can merge this directly if you want. When you have decided what to do about the UUIDs.

src/components/ble/AlertNotificationService.h
```diff
@@ -24,8 +27,25 @@ namespace Pinetime {
...
+
 
       private:
+        static const char ALERT_UNKNOWN = 0x01;
```
piggz, line 37: these dont match the types defined in the spec, CALL should be 0x03 see https://developer.nordicsemi.com/nRF5_SDK/nRF51_SDK_v4.x.x/doc/html/group__ble__sdk__srv__ans__c.html

JF002:
I've just reviewed this PR and added a few comments. 
Basically, it looks good and will be really useful. There's just some formatting details I would like to improve. And also check the UUID of the characteristic.

I don't mind working on this PR in the next few days if you want me to ;)

src/components/ble/AlertNotificationService.h
petterhs: Thanks! Pushing fix to this now

petterhs:
Just go ahead and continue on this :) Might be best to create a CALL_NOTIFICATION_EVENT as a characteristic with UUID 0002-0001-xx....... and keeping the NOTIFICATION_EVENT_SERVICE as the base UUID. This way we can extend this further for other usecases where we want to respond to a notification with a command/message. I can't see the comments you have added btw?

JF002:
> Just go ahead and continue on this :) 

I think I'll release 0.11 first. If you've got some time to work on it in the meantime, feel free to commit :)

> Might be best to create a CALL_NOTIFICATION_EVENT as a characteristic with UUID 0002-0001-xx....... and keeping the NOTIFICATION_EVENT_SERVICE as the base UUID. This way we can extend this further for other usecases where we want to respond to a notification with a command/message. 

In this case, we are using the standard ANS (Alert Notification service) service and added a new characteristic with a custom UUID, right ?
So yes, I think the characteristic should have the UUID 00020001-... to conform with our new guidelines.

> I can't see the comments you have added btw?

I forgot to push the button "submit review" (which I didn't know it existed...).

JF002:
I created my own branch based on petterhs:calls to do some experiments.

In this branch, I tried to handle the call notification the same way than the other notifications in AlertNotificationService and SystemTask. 
So, call notifications end up in the Notification app. When displaying the notification, the app look at the `category` field and draws Call notification differently : it adds 2 buttons (accept/reject).

It turns out the code does not look as good as I would have wanted to (the reference to the notification service is passed to displayapp->Notification->NotificationItem, ...). The design of the notification handling might need some improvements... The question is : "is it good enough for now or should I refactor the whole thing?"

My branch : https://github.com/JF002/Pinetime/tree/calls
Video : https://video.codingfield.com/videos/watch/f46c0e8f-9ac3-4a84-b76b-6e484501e6a9
What do you think ? Better? Worse?

piggz:
Could we have 3 buttons, accept/reject/ignore ?

Avamander:
@piggz 
> Could we have 3 buttons, accept/reject/ignore ?

Ignore could be the physical button.

piggz:
It should also sent an event to the app, so the ringer can be muted

Avamander:
@piggz 

> It should also sent an event to the app, so the ringer can be muted

I haven't read the PR's code yet, but I assume there's a `notify` characteristic that could be used to pass user action info back to the companion?

JF002:
> Could we have 3 buttons, accept/reject/ignore ?

Yes, that would be possible. There is not enough width to display "Accept", "Ignore" and "Reject", but using icon, it'll be perfect

petterhs:
> I created my own branch based on petterhs:calls to do some experiments.
> 
> In this branch, I tried to handle the call notification the same way than the other notifications in AlertNotificationService and SystemTask.
> So, call notifications end up in the Notification app. When displaying the notification, the app look at the `category` field and draws Call notification differently : it adds 2 buttons (accept/reject).
> 
> It turns out the code does not look as good as I would have wanted to (the reference to the notification service is passed to displayapp->Notification->NotificationItem, ...). The design of the notification handling might need some improvements... The question is : "is it good enough for now or should I refactor the whole thing?"
> 
> My branch : https://github.com/JF002/Pinetime/tree/calls
> Video : https://video.codingfield.com/videos/watch/f46c0e8f-9ac3-4a84-b76b-6e484501e6a9
> What do you think ? Better? Worse?

Better! Pulled your commits here

petterhs:
I added a ignore/mute button now, with symbols. Also added a new response to the phone so it can mute the call.

JF002:
Looks good to me! Now, we need to add support for this feature in companion apps (gadgetbridge and amazfish) :)

PR #160, opened 2021-01-08
fix MusicService UUID typo in comment

petterhs:
Stumbled upon this small typo. Changes to whitespaces happened automatically..

PR #163, opened 2021-01-11
WIP Navigation service and application

piggz:
This adds a service with a few characteristics to the firmware, and an application to display the information.  Service accepts:
icon/flag - icon name to use
narrative - textual description of instruction
manDist - manoeuvre distance
progress - overall progress (0-100) 

All included icons are from pure-maps, which provides the actual routing from the client, via Amazfish.  The icon names ultimately come from the mapbox project "direction-icons", See https://github.com/rinigus/pure-maps/tree/master/qml/icons/navigation

The icons were batch converted from SVG to 64px PNG, then converted to lvgl 1bit arrays.

The client side is currently implemented in Amazfish, which gets instructions from the pure-maps dbus interface.  On other devices, amazfish either sends navigation instructions as messages or via the music app, so this is a unique feature for the Pinetime.

**Instructions for testing:**
Install Amazfish on device, youll want the infininav branch from https://github.com/piggz/harbour-amazfish/tree/infininav
Pair the watch with the app
In the app, settings, application, enable Navigation Instructions
It may be nescessary at this point to restart the daemon
Launch puremaps and set it off navigating
Open the watch app, and instructions should be displayed.


src/components/ble/NavigationService.h
```diff
@@ -0,0 +1,96 @@
...
+#include <cstdint>
+#include <string>
+#define min // workaround: nimble's min/max macros conflict with libstdc++
+#define max
+#include <host/ble_gap.h>
+#include <host/ble_uuid.h>
+#undef max
+#undef min
+
+//c7e60000-78fc-48fe-8e23-43b37a1942d0
```
petterhs, line 29: Small typo in this comment inherented from the MusicService.h, proposed fix in #160. Should be `c7e60000-78fc-48fe-8e23-433b3a1942d0` instead

piggz: thx

src/components/ble/NavigationService.h
```diff
@@ -26,7 +26,7 @@
 #undef max
 #undef min
 
-//c7e60000-78fc-48fe-8e23-43b37a1942d0
+//c7e60000-78fc-48fe-8e23-433b3a1942d0
 #define NAVIGATION_SERVICE_UUID_BASE {0xd0, 0x42, 0x19, 0x3a, 0x3b, 0x43, 0x23, 0x8e, 0xfe, 0x48, 0xfc, 0x78, 0x00, 0x00, 0xe6, 0xc7}
```
Avamander, line 30: If music uses c7e5xxxx-:scissors:-433b3a1942d0, call control uses c7e5xxxx-:scissors:-433b3a1942d1, could you please use c7e5xxxx-:scissors:-433b3a1942d3? It sounds more consistent to me personally.

petterhs: I was actually thinking about suggesting to change call control to c7**e7**xxxx-scissors-433b3a1942**d0**. As long as we iterate the same byte I guess its fine, but to me it seems easier to see the differences in UUID's with this variation. Might just be because of the 0's after.

Avamander: It kind-of makes more sense to increment the base than some other part. But in the end it does not matter much, if it's agreed upon now and assuming we won't receive thousands of services... there's more room to increment the base.

petterhs: I assume we should also the bundle all the UUID definition's in a dedicated file? Or at least documentation with all services listed in the same place

Avamander: Yes, having this documented would be very useful. A new documentation page?

piggz: So, what the conclusion, what is the preference for UUID?

JF002:
I've just tested the navigation app, it still need a bit of fine tuning (improve the readability of the instruction label), but it's working!
You need Amazfish and PureMaps. It's running fine on my PC, but I guess it should also work on... the pinephone! Someone wants to try ?

src/components/ble/NavigationService.h
JF002: I'm also wondering what is the state of the art to choose a custom UUID for a ble service.

piggz:
TODO good suggestion by TLLIM to add vibration notification.  Should this be on icon/narrative change only, maybe not needed on distance change?

src/components/ble/NavigationService.h
JF002: Ok, I read this comment again. If I understand correctly, the idea is to use the same "root" for all custom UUIDs defined in InfiniTime? 

So, for now, we have the following UUIDs :

```
- Music       c7e50000-78fc-48fe-8e23-433b3a1942d0
- Navigation  c7e60000-78fc-48fe-8e23-433b3a1942d0
- Call        c7e50000-78fc-48fe-8e23-433b3a1942d1
```

Note that for now, only the Music service has been released and implemented into companion apps.
@piggz and @petterhs chose to increment the 4th digit, while @Avamander and #161 chose the last one.

I honestly have no preference. @piggz choice is more similar than the standard ble 16bits uuid which also changes digits on "the left" for each service, and keep the same base on "the right" : 0000**XXXX**-0000-1000-8000-00805f9b34fb.

So if there's no good reason to change @piggz 's strategy, I think we can keep it.

piggz: For a bit more background, this is how huami have numbered their services https://github.com/piggz/harbour-amazfish/blob/master/daemon/src/devices/bipdevice.h

JF002: So they increment the digits [4..7], right?
It looks like most implementation read the uuid from right to left (when written in the human readable form), keep the first 96bits unchanged and number their services by changing 0000XXXX-0000-0000-0000-000000000000.
Do you think we should also follow this guideline?

piggz: We could, though we have already started using 0000XXXX for the characteristics following the service uuid, where huani have completely different characteristic uuids.  So, we could continue using XXXX0000 for the services, and 0000XXXX for the chars?

petterhs: I like this aswell. And start the services from 0001XXXX instead of c7e5XXX

JF002: Note that the Music service is already released and implemented in 2 companion apps (Amazfish and Gadgetbrigde). 
If we change this UUID, these apps will have to implement the new uuid and maybe keep the old one for retro-compatibility... Not good.

So, if we want to increment digits [4..7], we should use c7e5XXXX.

Avamander: I think it's fine if we break this compatibility once and force users to upgrade.

Avamander: So the final decision is:
```
- Music       00000000-78fc-48fe-8e23-433b3a1942d0
- Navigation  00010000-78fc-48fe-8e23-433b3a1942d0
- Call        00020000-78fc-48fe-8e23-433b3a1942d0
```
?

piggz: I cant speak for GB, but i dont mind changing Amazfish if it means getting this right.
Things to consider
-Using a common base, with a separation of services and characteristics
-Ensuring services can be expanded in the future
I think this can be achieved using the 000X000Y- method of naming, which would mean changing the music service.   The same could probably be achieved without changing the music service by using c7e[x]000[y] where y = 1 for services and 2+ for characteristics, and X is incremented for each new service.

But, as I say, im happy to change it to whatever @JF002 decides.

pfeerick:
>  add vibration notification
Might be good to keep PR https://github.com/JF002/Pinetime/pull/155 in mind or offer suggestions on how it can be improved, then a single vibration interface can be utilised. ;)

> Should this be on icon/narrative change only
I would have thought so... otherwise it would keep buzzing every time distance changes, or x distance units? Better it to be vibrate on navigation changes.

src/components/ble/NavigationService.h
JF002: I think we agree to use the following scheme : SSSSCCCC-78fc-48fe-8e23-433b3a1942d0, where SSSS is a service ID and CCCC is a characteristic ID (where 0001 is reserved for the service).
This is how the Music service is setup : 

```
- c7e50000-78fc-48fe-8e23-433b3a1942d0 is the base UUID
- c7e50001-78fc-48fe-8e23-433b3a1942d0 is the UUID of the service
- c7e50002-78fc-48fe-8e23-433b3a1942d0 and following are the characteristics UUIDs
```

I think we can apply this scheme to other custom services : 

```
- c7e60001-78fc-48fe-8e23-433b3a1942d0 : service
- c7e60002-78fc-48fe-8e23-433b3a1942d0 : characteristic 1
- c7e60003-78fc-48fe-8e23-433b3a1942d0 : characteristic 2
- ...
```

Nothing prevent us from using 00000001-78fc-48fe-8e23-433b3a1942d0 or abcd0001-78fc-48fe-8e23-433b3a1942d0 as UUID for a custom service, but we can also use **c7e** as a signature for PineTime services (and deadbeef-78fc-48fe-8e23-433b3a1942d0 as easter egg :D )

This scheme allows us to keep the music service as is, and add new services and characteristics easily.
It also allows to keep the UUID of the navigation service as it is now.

Ok for everyone?

Avamander: > This scheme allows us to keep the music service as is [...] Ok for everyone?

I don't think this is a problem to change and would look a tiny bit neater, but sure.

So with current specific services it'd be:

```
- Music       c7e50000-78fc-48fe-8e23-433b3a1942d0
- Navigation  c7e60000-78fc-48fe-8e23-433b3a1942d0
- Call        c7e70000-78fc-48fe-8e23-433b3a1942d0
```

Correct?

JF002: Changing means that companion app have to release a new version and maybe handle retrocompatibility with older fw. I'm not sure it worth it.

In fact, Music service uses c7e50000-... as a root UUID, but this one is never used. Instead the service is c7e50001 and chars are c7e50002, c7e50003,...

piggz: And that is what this service does too, c7e60000 as the base, then c7e60001 for the service, and the characteristics after that.

Avamander: > In fact, Music service uses c7e50000-... as a root UUID, but this one is never used. Instead the service is c7e50001 and chars are c7e50002, c7e50003,

Let's leave it weird, but not continue like that?

Avamander: > Changing means that companion app have to release a new version and maybe handle retrocompatibility with older fw. I'm not sure it worth it.

I think it's better rip that bandaid off while we can, especially if use in production is low so far. If we need to shift music service around a bit, that's fine as well.

JF002: I took some time to think about it, and, although there's no functional need to change the current scheme, it may be the only opportunity we have to start from scratch and avoid future debate and discussion about why c7exxxxx and why services starts at 1 and not 0... 
Here is my quick'n'dirty RFC : https://gist.github.com/JF002/c37d156c88f6dae7846e5252a7728163

This new scheme implies a modification of the MusicService. Hence, we'll need to update Amazfish and Gadgetbridge to take this changes into account.

piggz:
Updated with new UUID scheme

piggz:
Of note, it turns out we've been using the wrong UUIDs anyway, as the wrong indexes were used in the music service ctor, then copied here.  Old code used 11,12, should have been 12,13, and reversed.

piggz:
Scroll speed now nicer

JF002:
Mhm I had to merge this branch manually to fix conflict and... github didn't notice that it's actually merged...

pfeerick:
I'd close this PR pronto then, and act like nothing broke... before GitHung gets any ideas ;) Since it's merged, any further changes can be from the current develop branch.

Issue #146, opened 2020-11-28
Battery level is erratic


hrmckay:
In version 0.9.0 and earlier versions I noticed that the battery indicator at the top of the screen would sometimes change erratically.  It would show 25% full, then a few seconds later 50%, then after a few seconds back to 25%, etc.  I changed the clock face to show the actual battery percentage and it changes almost every second.  As I look at it now it shows, 32, 30, 34, 27, 30, 32 changing once a second or so.  This indicates the .IsUpdated() code in Clock.cpp is being executed every second which seems a lot.

Hugh

JF002:
The battery level is updated (via an ADC acquisition) periodically, maybe every 1s. No filtering is applied on the value acquired on the ADC, so it's subject to a lot of noise (depending on the actual battery level and cpu load, for example).

Knowing that, I would certainly expect a noisy battery level value. There are several ways to improve that:
 - Probe the ADC when we know that the CPU mostly idle, so that we always measure the voltage on low load
 - filter/average the value to reduce the noise
 - calibrate the %-value : store the min/max voltage value in NVM and use these min/max to provide a better value in %.

Interested to work on that ?

hrmckay:
Thanks for the info.  I will look into smoothing the data and let you know what I find out.

Hugh

Panky-codes:
I think we can close this issue for now.

hrmckay:
I was testing similar code, written in pure C since I have very limited C++ experience.  I got improvements but not as good as I would have liked.  Your code looks excellent and fits in better with the rest of the code.  I will close this issue.

Hugh

PR #168, opened 2021-01-16
Fix erratic battery

Panky-codes:
This addresses #146 issue by averaging out the battery percent value. 
As a part of this PR, I have done the following changes: 
- Added a very simple circular buffer. I was thinking initially to use the `nrf_ringbuf`, but the API is very C-like and probably I don't need all the features that come with it. 
- Changed the percent value to be an `int` instead of a `float`. Floating-point arithmetic is generally more expensive and the user is not really interested in decimal level accuracy for battery percentage. I also noticed in many places where they are used, they are being downcasted to a smaller data type anyway. (e.g., see SystemInfo.cpp:50)

Tested it on the devkit. The battery variation reduced from +- 8% to +- 1%.

### DIFF
--- a/src/components/battery/BatteryController.h
+++ b/src/components/battery/BatteryController.h
@@ -6,15 +6,21 @@
 
 namespace Pinetime {
   namespace Controllers {
-  // A simple circular buffer that can be used to average 
-  // out the sensor values
+  /** A simple circular buffer that can be used to average 
+   out the sensor values. The total capacity of the CircBuffer  
+   is given as the template parameter N.
+   */ 
   template <int N> 
   class CircBuffer {
   public:
-    CircBuffer() : arr{}, sz{}, cap{N}, loc{} {}
+    CircBuffer() : arr{}, sz{}, cap{N}, head{} {}
+    /**
+   insert member function overwrites the next data to the current 
+   HEAD and moves the HEAD to the newly inserted value.
+   */ 
     void insert(const int num) {
-      loc %= cap;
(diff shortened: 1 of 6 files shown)


src/components/battery/BatteryController.h
```diff
@@ -1,14 +1,42 @@
 #pragma once
 #include <cstdint>
 #include <drivers/include/nrfx_saadc.h>
+#include <array>
+#include <numeric>
 
 namespace Pinetime {
   namespace Controllers {
+  // A simple circular buffer that can be used to average 
+  // out the sensor values
```
Avamander, line 10: This should probably be in the Doxygen format so that it can be picked up by doc->HTML generators and IDEs
```
/* A simple circular buffer that can be used to average out the sensor values */
```

src/components/battery/BatteryController.h
```diff
@@ -1,14 +1,42 @@
...
+    }
+
+    int GetAverage() const {
+      int sum = std::accumulate(arr.begin(), arr.end(), 0);
+      return (sum / sz);
+    }
+
+  private:
+    std::array<int, N> arr;
+    uint8_t sz;
```
Avamander, line 30: Maybe an explanation about the abbreviation/usage would be useful.

Avamander:
Nice improvement, thanks!

JF002:
Nice improvement, indeed! Thanks for this PR, I'll merge it soon® :)

PR #169, opened 2021-01-17
Heart rate sensor

JF002:
This PR adds support for the HRS3300 heart rate sensor. It's based on an open source implementation of the low-level driver and algorithm from [wasp-os](https://github.com/daniel-thompson/wasp-os) (thanks to [Daniel](https://github.com/daniel-thompson) for this great work!).

A new app is available in the menu and allows to start/stop the HR acquisition and displays the HR value.
A new BLE service (Heart Rate service) has been added and implements the standard BLE heart rate service.

Video : https://video.codingfield.com/videos/watch/63a5d374-f911-4eb5-88a1-4ee184205675

image

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -375,6 +375,7 @@ list(APPEND SOURCE_FILES
         components/ble/BatteryInformationService.cpp
         components/ble/ImmediateAlertService.cpp
         components/ble/ServiceDiscovery.cpp
+        components/ble/HeartRateService.cpp
         components/firmwarevalidator/FirmwareValidator.cpp
         drivers/Cst816s.cpp
         FreeRTOS/port.c
@@ -464,6 +465,7 @@ set(INCLUDE_FILES
         components/ble/ImmediateAlertService.h
         components/ble/ServiceDiscovery.h
         components/ble/BleClient.h
+        components/ble/HeartRateService.h.h
         drivers/Cst816s.h
         FreeRTOS/portmacro.h
         FreeRTOS/portmacro_cmsis.h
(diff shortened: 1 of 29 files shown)


piggz:
Ill give this a test I think ... ill compare it against my GTS while doing some indoor cycling using golden cheetah....hoping GC just-works with it

piggz:
I tested this against GoldenCheetah.  It was good that the HRM was auto-discovered and values obtained, not so good that the measurement stopped when the watch went to sleep.  There should be some way to keep measurement on when performing an activity.

jones139:
> .....not so good that the measurement stopped when the watch went to sleep. There should be some way to keep measurement on when performing an activity.

I think this is because the I2C interface (twiMaster) goes to sleep when the watch display sleeps.   I have inhibited this in my (not fully working) accelerometer interface for the same reason.   I have just done it with a simple compiler directive because I would always want it running and offering data via BLE:   https://github.com/OpenSeizureDetector/Pinetime/blob/accel/src/systemtask/SystemTask.cpp#L217

Thanks for doing the heart rate monitor - I will look at how this gets the data to BLE, which is where I was struggling with the accelerometer (I have a memory leak somewhere which crashes the watch after a  couple of minutes....).

Graham.

piggz:
For now, it may be enough to keep the sensor alive when clicking the start button.  In the future, we will maybe need to think about interval measurement.

jones139:
Yes, it will be a lot more power hungry than the accelerometer unless it is
already using very short led pulses.   I've never worked out how garmin
does it- I think it is a combination of pulsing the leds and using
intervals that must change depending on the activity level.

JF002:
I agree, the fact that the HR acquisition stops when the watch goes to sleep is not very practical in real life.
I haven't measured the power consumption of the HR sensor, but I guess it draws a bit of power when running. I also wonder how Garmin and Fitbit do to keep it on most of the time while maintaining good battery live.

I have 2 solutions in mind:
 - do not go to sleep when the HR app is running or when the HR acquisition is started.
 - Go to sleep but do not stop HR acquisition and BLE notification when HR is running.

JF002:
I'll merge this PR, we'll improve the UX later on ;)

PR #174, opened 2021-01-19
Fix music service UUID

JF002:
Add custom BLE service UUID guidelines in the documentation.
Adapt the music service to these guidelines.


piggz:
LGTM

PR #175, opened 2021-01-21
changed command MemoryDataAccessControl from 036 to 0x36 in St7789 class

nscooling:
Fixed minor error in commands enum for class St7789. The MemoryDataAccessControl was mistakenly set to `036` and should be `0x36`. 

I have built and run the app on my test PineTime but unsure how to do further validation and not sure why it wan't an issue (as `0x1E` is not a valid command) in the first place without doing further investigation.


JF002:
Wow! That was a nasty typo! Thanks!

PR #178, opened 2021-01-23
check for cmake build-success before running post-build.sh

petterhs:
Goal: Don't run post-build.sh if the cmake build fail.


JF002:
Great! I think it happened to my once or twice that the build failed but it was trying to create the package anyway. 
Thanks!

PR #181, opened 2021-01-25
QR code application

petterhs:
With this application you can send text from a companion app (e.g. URL's, ticket's) and view it as a QR code on the PineTime. Works as is but will probably crash after a while so need some help. (see **Needs fixing before merge**)

### Testing: (WARNING: Don't test on sealed device. Can crash and cause bricked watch)
Not implemented in any of the companion apps yet but can be tested with nRFConnect.
- Go to QR app on page 3 of applications on the watch
- Press screen once and the initial QR-code should be visible(should be "https://github.com/JF002/Pinetime/" when scanned)
- Go to the list of services on nRFConnect when you are connected to the watch
- Press service with UUID: 00030000-78fc-.................
- Press the icon with arrow pointing upwards on the characteristic with UUID: 00030001-78fc-.........
- Change valuetype to TEXT
- Type your URL or text you want to encode to QR
- Send
- On watch:
- Exit and re-enter QR application(if not you will just overwrite the previously drawn code.)
- Press screen and try to scan the code with phone.

### Needs fixing before merge:
- Memory leaks. I need some help here. One place is atleast in Qr::drawQr() It support different sizes of QR codes so there is different sizes in the QR modules (QR pixels). In the solution now i create a buffer of the size of the module with white color and use this to draw modules. So a new buffer each time I draw the QR code... Instead my idea is to create a white buffer with MAX_SIZE in Qr.h and just use a part of this buffer when drawing modules to screen. How do I do this?
- Reset the screen when drawing a new QR code. Now you need to exit the app and go back in to not just overwrite the old one.

### Future:
- Support storing more than one QR code on the phone, and be able to alter between those. (Probably best to wait for littlefs)
- Update the QR code automatically when entering the app and when receiving new code


petterhs:
As suspected, there are still memory issues here as two more tried this PR (from the chat) experienced freezes that needed watch reboot

src/components/ble/QrService.cpp
```diff
@@ -0,0 +1,75 @@
...
+    char *s = (char *) &data[0];
+    NRF_LOG_INFO("DATA : %s", s);
+    
+    const char d[2] = "|";
+    char* qrId = strtok(s, d);
+    char* qrName = strtok(NULL, d);
+    char* qrText = strtok(NULL, d);
+    
+    if (ble_uuid_cmp(ctxt->chr->uuid, (ble_uuid_t *) &qrsTextCharUuid) == 0) {
+      if (std::stoi(qrId) < MAXLISTITEMS && qrId > 0 && qrName != NULL && qrText != NULL) {
```
MFAshby, line 64:
```suggestion
      if (std::stoi(qrId) < MAXLISTITEMS && std::stoi(qrId) > 0 && qrName != NULL && qrText != NULL) {
```

I think qrId should be parsed as an integer before comparing. 
I'm also not sure if qrId could ever be `NULL`, but it would cause a crash if it was as stoi doesn't handle it.

MFAshby (commented):
This looks like a fun little app :smile: 

I had a quick play with this after rebasing the change to latest `develop`. I noticed a couple of things:
- sending any malformatted request causes a crash (so sending something like `hello!` instead  of `0|hello|world`)
- sending item 4 doesn't extend the menu.
- the menu background gets brighter every time I open a QR then leave it again (took a video but GitHub complains it's too large, will find another way to share it if necessary)

I Am Not A C++ developer, but I had a read through the code for educational purposes and I think I spotted one bug, see inline comment about qrId variable comparison.

I am also not a UX designer, but I had a thought about this: I don't think it makes too much sense to have a dedicated QR option in the menu. I think instead maybe: 
1. Include the project link QR codes in the settings/about screen - currently the GitHub project URL is there as text but it would be way cooler have the QR code as well :sunglasses: 
2. When a QR is received over the bluetooth service, generate and display it immediately on the screen instead of assigning it to a slot in the menu. I can't think of a reason to save it as a menu entry (to be fair, I also can't think of a really good reason to send a QR code to the watch via bluetooth, but this doesn't mean we shouldn't do it :smile_cat: )

Izaic:
Bump?

src/components/ble/QrService.cpp
```diff
@@ -0,0 +1,75 @@
...
+int QRCallback(uint16_t conn_handle, uint16_t attr_handle, struct ble_gatt_access_ctxt *ctxt, void *arg) {
+  auto qrService = static_cast<QrService *>(arg);
+  return qrService->OnCommand(conn_handle, attr_handle, ctxt);
+}
+
+QrService::QrService(Pinetime::System::SystemTask &system) : m_system(system) {
+  qrsTextCharUuid.value[11] = qrsTextCharId[0];
+  qrsTextCharUuid.value[12] = qrsTextCharId[1];
+
+  characteristicDefinition[0] = {.uuid = (ble_uuid_t *) (&qrsTextCharUuid),
```
Avamander, line 15: I think these should be updated to the non-c-style cast style. Like other services today.

src/displayapp/screens/Qr.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+  lv_obj_set_pos(container1, 0, 0);
+  lv_obj_set_width(container1, LV_HOR_RES - 15);
+  lv_obj_set_height(container1, LV_VER_RES);
+  lv_cont_set_layout(container1, LV_LAYOUT_COLUMN_LEFT);
+
+  lv_obj_t * labelBt;
+
+  for(int i = 0; i < MAXLISTITEMS; i++) {
+    if ( qrList[i].text != "") {
```
Avamander, line 74: Needs formatting using `clang-format`

src/displayapp/screens/Qr.cpp
```diff
@@ -0,0 +1,159 @@
...
+  lv_obj_set_style_local_border_width(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, 0);
+
+  lv_obj_set_pos(container1, 0, 0);
+  lv_obj_set_width(container1, LV_HOR_RES - 15);
+  lv_obj_set_height(container1, LV_VER_RES);
+  lv_cont_set_layout(container1, LV_LAYOUT_COLUMN_LEFT);
+
+  lv_obj_t * labelBt;
+
+  for(int i = 0; i < MAXLISTITEMS; i++) {
```
Avamander, line 73: I suspect uint16_t would be less wasteful.

src/displayapp/screens/Qr.cpp
```diff
@@ -0,0 +1,159 @@
...
+  showingQrCode = true;
+  
+  bool ok = qrcodegen_encodeText(qrText.c_str(), tempBuffer, qrcode, qrcodegen_Ecc_LOW,
+    qrcodegen_VERSION_MIN, qrcodegen_VERSION_MAX, qrcodegen_Mask_AUTO, true);
+  if (ok) {
+    qrSize = qrcodegen_getSize(qrcode);
+    qrModuleSize = LV_HOR_RES_MAX / (qrSize + 2*border);
+    bufferSize = qrModuleSize * qrModuleSize;
+    
+    offset = (LV_HOR_RES_MAX - (qrSize + 2*border)*qrModuleSize)/2;
```
Avamander, line 112: Needs formatting using `clang-format`

src/libs/qrcodegenerator/qrcodegen.h
```diff
@@ -0,0 +1,311 @@
...
+ * Returns the color of the module (pixel) at the given coordinates, which is false
+ * for white or true for black. The top left corner has the coordinates (x=0, y=0).
+ * If the given coordinates are out of bounds, then false (white) is returned.
+ */
+bool qrcodegen_getModule(const uint8_t qrcode[], int x, int y);
+
+
+#ifdef __cplusplus
+}
+#endif
```
Avamander, line 311: I think C-compatibility is not necessary.

Avamander (changes requested):

src/components/ble/QrService.cpp
```diff
@@ -0,0 +1,75 @@
...
+                                                   struct ble_gatt_access_ctxt *ctxt) {
+  if (ctxt->op == BLE_GATT_ACCESS_OP_WRITE_CHR) {
+    size_t notifSize = OS_MBUF_PKTLEN(ctxt->om);
+    uint8_t data[notifSize + 1];
+    data[notifSize] = '\0';
+    os_mbuf_copydata(ctxt->om, 0, notifSize, data);
+    char *s = (char *) &data[0];
+    NRF_LOG_INFO("DATA : %s", s);
+    
+    const char d[2] = "|";
```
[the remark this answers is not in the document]
petterhs: Thank you :)

src/components/ble/QrService.cpp
```diff
@@ -0,0 +1,75 @@
...
+using namespace Pinetime::Controllers;
+
+int QRCallback(uint16_t conn_handle, uint16_t attr_handle, struct ble_gatt_access_ctxt *ctxt, void *arg) {
+  auto qrService = static_cast<QrService *>(arg);
+  return qrService->OnCommand(conn_handle, attr_handle, ctxt);
+}
+
+QrService::QrService(Pinetime::System::SystemTask &system) : m_system(system) {
+  qrsTextCharUuid.value[11] = qrsTextCharId[0];
+  qrsTextCharUuid.value[12] = qrsTextCharId[1];
```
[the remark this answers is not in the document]
petterhs: old way we used to set the UUID for different service characteristics, updated now so not like this anymore

src/components/ble/QrService.cpp
petterhs: Thank you :+1:

src/components/ble/QrService.cpp
petterhs: Thanks :+1: Done now

src/libs/qrcodegenerator/qrcodegen.h
petterhs: As it is from a third party library i don't think i will change it now. Might make more sense to have the library as a git submodule though?

petterhs:
Thanks for the review and comments!! I'm sorry I have not been active here, and not kept the PR up to date. I have worked a little on it the last days though:

- rebased onto develop branch
- changed UUID to not interfere with the Motion UUID
- fixed suggestions above
- clang-formated files
- Fix some edgecases when sending data not on the right format. When you send with the wrong format it overrides the nr.1 slot now.
- debugging of new bugs....

The PR is quite a mess now with the rebases and force-pushes, I will try to keep it cleaner from now so it is easier to review.

Status:
- It is still not stable, having some inconstant bugs still (Testing at your own risk)
- New display bug in bottom right corner when displaying codes
- Temporarily hijacked the Metronome position in the app menu instead of creating a new page. (If we want a new page at all)
- Should the qrencode library be a git submodule instead?
- Needs to add a doc file and add the service UUID to BLE docs

I think the app could be useful to me, but i realize that not everyone agrees, so we might just want to close the PR and I can maintain it on a separate branch myself. Up to the rest of the community :smiley: 

QR codes i want to have on my watch:
- Boarding Pass when traveling
- Matrix account info (Or other chat apps/social media)
- Spotify Group session??
- Links to personal website/portfolio?
- BTC Lightning node (Scan to connect to my node)

src/libs/qrcodegenerator/qrcodegen.h
Avamander: I think a submodule sounds great. Can you do that?

Avamander:
> QR codes i want to have on my watch:

GPG pubkey and WiFi password for me :)

Avamander:
> Needs to add a doc file and add the service UUID to BLE docs

I think a barebones one for now is fine. It's much easier to update and improve docs than it is to constantly keep a PR from bitrot when it hasn't been merged. I really want to make it less effort for contributors.

petterhs:
I think I have figured out the crashes. Its because the application uses to much memory when I try to support all QR sizes. (up to version 40, 177x177 modules, max 2953 bytes, which is probably not readable on the 240x240 screen anyway)

The Qr.cpp screen creates two buffers of size dependent on max QR version. with version 40 that is 3918 bytes per buffer which is alot.. So the app crashes if there is not enough memory for these buffers, so dependent on other memory usage and that is why it was unstable.

From nRFConnect i'm only able to send 162 bytes to the watch (not sure where this limit comes from, and if we can increase it). So it makes sense to reduce the max QR version down to just support the 162 bytes, at least for now.

The choice is then what error correction to chose:
image

| ECC Level | Max QR version | Module size | Pixels per module | max size(bytes) | buffer sizes(bytes)
| ------------- | ------------- | ------------- | ------------- | ------------- |  ------------- | 
| L  | 8  | 49x49  | 4x4 | 192  | 2*302
| M  | 9  | 53x53  | 4x4 | 180  | 2*353
| Q  | 11  |  61x61  | 3x3 | 177  | 2*467
| H  | 13  |  69x69  | 3x3 | 177  | 2*597


All of these are easily readable by my phone. I have not been able to test with more than 162 bytes, so not sure when it gets problematic, if we want to support bigger QR codes in the future.

So L and Q have the same pixel sizes as M and H so the choice is probably either M or H.

The buffers:
M : 2x353 bytes
H:  2x597 bytes

Buffers are deleted when exiting app so these should both be small enough to always work.

For now I choose QR version 9 with error correction Medium. As i think the extra pixel per module is worth more than having correction level HIGH.

Another potential problem is the QrService that actually store the text that gets encoded. If we send 4 max size codes to the watch we take up 4*162 = 648 bytes of RAM that can potentially be needed by other applications. This is the same for other services that we send data to; Music, navigation, notifications, etc. Hopefully we can move this to external memory  soon. (#321 #750)

src/libs/qrcodegenerator/qrcodegen.h
petterhs: I can do that, what is the best practice when i have made a change in external file? Lvgl is using joaqimorg's own fork but i don't like that approach

petterhs: Also the repo contains qr libraries for other languages aswell, so that is not really suitable as a submodule either

petterhs:
> From nRFConnect i'm only able to send 162 bytes to the watch (not sure where this limit comes from, and if we can increase it). So it makes sense to reduce the max QR version down to just support the 162 bytes, at least for now.

This was just a MTU thing, with nRFConnect i'm able to increase it to 517 bytes.

I also tested with the QR Version 40, 177*177 modules, with only one pixel per qr module.. Was not able to scan with my phone.. but version 25 was.

Version 25 is 117*177 so its the biggest version we can fit into our screen (240x240) with 2 pixels per module. Interestingly it also supports 535 bytes with error correction HIGH. (Max 1,273 bytes with error correction low)

Then if we want to support the biggest, actually readable QR code version it is probably 25. with buffer size 2*1713B. Seems to work fine now, RAM vice, but maybe not in the future.. Not sure what to choose.

IMG_20211113_121905_edit_9333989304825

Of course the app generates the smallest QR code needed, so all codes will not look like this

src/displayapp/screens/Qr.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+    for (int16_t y = 0; y < qrSize + border * 2; y++) {
+      for (int16_t x = 0; x < (qrSize + border * 2); x++) {
+        if (!qrcodegen_getModule(qrcode, x - border, y - border)) {
+          area.x1 = qrModuleSize * x + offset;
+          area.y1 = qrModuleSize * y + offset;
+          area.x2 = qrModuleSize * (x + 1) + offset - 1;
+          area.y2 = qrModuleSize * (y + 1) + offset - 1;
+          lvgl.SetFullRefresh(Components::LittleVgl::FullRefreshDirections::None);
+          lvgl.FlushDisplay(&area, colorBuffer);
```
petterhs, line 120: Is this OK even if colorBuffer is larger than area?? It works but wonder if I overwrite something on the SPI

Avamander (approved):

Avamander:
Recent changes have fixed the qualms I had with the code. Now it would be great to test this out further.

JF002:
@petterhs Thanks for this PR! I'm really sorry I didn't review it sooner. It looks quite nice and useful, especially in those time of "covid pass".
However, from my measurement, it increases the size of the binary of 15KB, which, while not critical, is quite big knowing that the FLASH memory is already quite packed.

I'll delay this PR for some time, until we have a better view on how we can reduce the memory usage and keep it under control.

petterhs:
@JF002 Completely understandable :) If there is a usecase for smaller qr-codes it is possible to reduce the memory usage by removing the ability to create large codes. I will try to keep this up to date for people that want to use it in the meantime

Danimations:
Nice concept. I'd like to share another couple of use cases. I would use mine to display QR codes from e-tickets to various events, bought from vendors like Eventbrite. It would definitely be preferable than digging through emails on a phone to enter an event with. Thanks for this initiative!

devnoname120:
I'd love to use this to quickly share my contact card and my Wi-Fi profile.

Riksu9000:
Displaying QR codes seems like a neat idea, but perhaps there's a better way to implement it. As this feature is already dependent on the companion app to enter the data, the conversion to a qr code could also be done on the companion, where memory is not so limited. I was reminded about this after seeing #1237, which could also be used for this, so simply send the codes as images to the external flash.

Eloitor:
Normal barcodes could be also be useful, to store multiple customer cards

PR #187, opened 2021-01-28
Upgrade to the LVGL version 7.10.0

joaquimorg:
Add external module to LVGL version 7.10.0

Update main theme, and screens to use the new lvgl functions.
Notifications screen, Application List and the game "2" had new colors and some design change.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "src/libs/lvgl"]
+	path = src/libs/lvgl
+	url = https://github.com/joaquimorg/lvgl.git
```
petterhs, line 3: What is the reason to use your fork instead of main repo?

src/displayapp/DisplayApp.cpp
```diff
@@ -208,7 +208,9 @@ void DisplayApp::RunningState() {
       case Apps::SysInfo: currentScreen.reset(new Screens::SystemInfo(this, dateTimeController, batteryController, brightnessController, bleController, watchdog)); break;
       case Apps::Meter: currentScreen.reset(new Screens::Meter(this)); break;
       case Apps::Twos: currentScreen.reset(new Screens::Twos(this)); break;
-      case Apps::Gauge: currentScreen.reset(new Screens::Gauge(this)); break;
+
+      //case Apps::Gauge: currentScreen.reset(new Screens::Gauge(this)); break;
```
petterhs, line 212: Why is Gauge not working properly?

src/displayapp/screens/ApplicationList.cpp
```diff
@@ -8,77 +8,75 @@
...
-          }
-        } {}
-
+ApplicationList::ApplicationList(Pinetime::Applications::DisplayApp *app) : Screen(app),
+                                                                            screens{app, {
+                                                                                             [this]() -> std::unique_ptr<Screen> { return CreateScreen1(); }, [this]() -> std::unique_ptr<Screen> { return CreateScreen2(); },
+                                                                                             //[this]() -> std::unique_ptr<Screen> { return CreateScreen3(); }
+                                                                                         }}
+{
+}
```
petterhs, line 17: The changes in this file is just code formating right? I'm pretty sure the previous formatting is preferred in this project.

src/displayapp/screens/Tile.cpp
```diff
@@ -22,10 +20,12 @@ Tile::Tile(DisplayApp* app, std::array<Applications, 6>& applications) : Screen(
       appIndex++;
     }
   }
-  modal.reset(new Modal(app));
+  
+  // ????
```
petterhs, line 24: ????

petterhs (commented):
Nice! Added some comments and I'm going to try and test it out now. Interesting to see the firmware size comparison

petterhs:
This is a really nice improvement! Everything works fine :)

Two things i notice:
- App buttons in application list feel slower. Takes longer time from clicking a button until the app launches. Don't know what causes this
- 2048 game: Tiles are could be a bit "taller" so they feel more square.

Increases 0.11.0 firmware size from 332kB to 358kB but i suspect that is a small price to pay.

.gitmodules
joaquimorg: The main repo don't have the function "lv_disp_set_direction"

src/displayapp/DisplayApp.cpp
joaquimorg: The Gauge is not used, I did not make any corrections to the new version.

src/displayapp/screens/ApplicationList.cpp
joaquimorg: in so many files something escaped ...

src/displayapp/screens/Tile.cpp
joaquimorg: this call to the modal must have been lost code, the ????? was to check later ....

src/displayapp/DisplayApp.cpp
petterhs: Okok just commenting so it will be removed before this get merged. Same with the places you have added "joaquimorg to FIX". Shouldn't Gauge and Modal get removed completely?

JF002:
Thank you very much for this PR, I appreciate! I'll look at it as soon as possible, but it may take some time as I'm a bit overwhelmed with big PRs and some other things in real life :)

JF002:
Ok, I had a quick look at the code, and it looks quite good!

While testing, I noticed that there is a small delay when we tap an app in the menu before it opens. I wonder what's causing this.

joaquimorg:
I'm going to do some tests, but it can be from the animation of the buttons.

JF002:
I've just done some test : the button is redrawn when it's pushed. I gives a nice feedback to the user, but takes ~280ms. It's useful when the UI does not change when the user taps the button (for example : the start button in the HR app).

But in this case, the feedback is provided by the display that is redrawn with the app. In this case, showing that the button was pushed is not needed. And it takes a bit too much time, IMO.

I managed to remove this delay by doing the following quick'n'dirty changes : 
image

Then I tried to create a new style (btn_no_anim) but ... I find the styling functionality very confusing... How to apply a non-default style to a widget?

joaquimorg:
I’ve already fixed the animation, I don’t think it’s necessary, on such a small screen you don’t have the notion of animation and don’t add anything to the button style

joaquimorg:
> I've just done some test : the button is redrawn when it's pushed. I gives a nice feedback to the user, but takes ~280ms. It's useful when the UI does not change when the user taps the button (for example : the start button in the HR app).
> 
> But in this case, the feedback is provided by the display that is redrawn with the app. In this case, showing that the button was pushed is not needed. And it takes a bit too much time, IMO.
> 
> I managed to remove this delay by doing the following quick'n'dirty changes :
> image
> 
> Then I tried to create a new style (btn_no_anim) but ... I find the styling functionality very confusing... How to apply a non-default style to a widget?

The theme is just to define the generic styles of the widgets, to create something specific the style must be created and applied to each case.

JF002:
Nice, I'll try your update this evening!

PR #196, opened 2021-01-30
Added clang-tidy configuration

Avamander:
This PR adds a clang-tidy configuration file that should help turn people's linting more consistent. It also enables more linting that wasn't previously enabled. There are a lot of warnings displayed with the current codebase that now would have to be phased out and will need a bit of attention. 

A few rules have been disabled because they're project-specific, unsuitable for embedded, massively too noisy or conflict with just the current project's style. It is unlikely that this list does not have to be updated when new linters are added to clang-tidy. I would not turn this into a PR gate just yet, ruleset is not stable enough.


JF002:
Thanks ! 

What's the difference between clang-format and clang-tidy? And how do you use this one?

JF002:
From what I understand, clang-format checks the... formatting of the source code (space, tabs, ...) while clang-tidy does static analysis of the code to prevent some bugs.

Avamander:
@JF002

> What's the difference between clang-format and clang-tidy? And how do you use this one?

clang-format is a formatting tool, clang-tidy is a linter basically. 

You have CLion so it again takes just enabling the option in settings: 
Screenshot that shows the location in CLion settings

JF002:
Perfect, thanks!

PR #199, opened 2021-02-03
Added a helper script for docker

shuhaowu:
Just a script to make recompiling in docker a little less typing. Simply invoke `docker/b` from the root of the repository.

### DIFF
--- a/docker/b
+++ b/docker/b
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+docker image build -t infinitime-build --build-arg PUID=$(id -u) --build-arg PGID=$(id -g) ./docker
+docker run --rm -it -v $(pwd):/sources infinitime-build /opt/build.sh "$@"
+


petterhs:
I also want this merged :) I have a similar script in docker/ that I have to make sure not gets added to my commits. The script could probably have a more describing name though. Something like docker-build maybe?

shuhaowu:
The problem is that there is a script called build.sh already in the docker directory, which is copied into the container. Can name this script docker/build if we rename that other script ...

pfeerick:
Why not just call it `docker-build.sh`..., i.e. `docker/docker-build.sh` ... after all, it's not that often you need to rebuild it, right?

JF002:
I agree that the file could be renamed as something more expressive like `docker-build.sh`.

JF002:
I close this PR since there was no activity since end of February. Feel free to reopen it if you want to continue your work.

PR #182, opened 2021-01-25
Fix build warnings coming from HR implementation

petterhs:
My OCD told me to fix this lol

PR #206, opened 2021-02-12
Eliminated warnings in Hrs3300

nscooling:
The .cpp file was emitting indentation warnings for `WriteRegister` as the while loop wasn't wrapped in a block.

Also removed the unnecessary static associated with the constexpr for
```
  constexpr uint8_t maxGain = 64U;
```
and added the `U` just for completeness.


Avamander:
I'm not mandating anything, but did you use the clang-tidy configuration in the develop branch to detect this? Also, is it formatted using clang-format configuration? It's just that I want to know if they have been useful.

nscooling:
I'm building using docker and I don't think that's setup to use clang-tidy at the moment. I'm pretty sure these are just standard GCC warnings. 

Clang-tidy is great, especially most of the modernise warnings. I haven't tried out the confit file you've setup yet. I'm just getting back into the swing of things after a couple of weeks not being able to contribute. Cppcheck is also useful.

petterhs:
There is already #182 suggesting some of these changes🙂

nscooling:
> There is already #182 suggesting some of these changes🙂

Sorry I'd missed these. I'd suggest it's always better to do a PR per file change. I know it's a minor thing but it makes merging simpler. But great minds think alike 😉 or maybe we're both showing our OCD 🤔

PR #211, opened 2021-02-19
Flash correct file name

roblabla:
Currently, trying to flash the firmware with `make FLASH_pinetime-app` fails because it's trying to flash `pinetime-app.hex`, which doesn't exist. The correct filename is `pinetime-app-0.13.0.hex`, which can be found in the EXECUTABLE_FILE_NAME variable.


JF002:
Right! I do not use these targets anymore, as CLion takes care of that for me... Thanks for the fix!

PR #217, opened 2021-02-24
Multi face support, analog clock, 12/24 config

joaquimorg:
Changes made to this pull request:

- Possibility of having several watch faces, to change the watch face just make a long press on the watch.
- New analog watch face.
- Configuration to choose between 12h or 24h
- Added some color to the default digital clock
- New component to manage configurations and save generic states (Work in progress)
- Go back to the same menu page when we leave an app


JF002:
Thanks for this high quality PR, @joaquimorg !
I reviewed most of it and it looks good to me!

Apart from a few formatting details, I have only 1 point of attention : `bg_clock`. If I understand correctly, it's the backgound of the analog clock (and it's quite beautiful!). But, as this is a 240*240px bitmap, it uses a bit of space (14KB). That's ok because we still have a bit of space available, but this new asset increases the total size of the FW to 397KB. Not an issue right now  as we still have some room (the max size if ~460KB) buuuut... we are getting closer and closer to the limit!

I can see 2 ways to improve this : 
 - compress/RLE encode the bitmap so that it needs less space
 - store it in the external spi flash memory.

I'm not saying it should be done in this PR as it works fine right now, but that's definitely something we'll have to think about in the near future to be able to add other watchfaces that also needs graphical assets.

I'll add some comments to fix the formatting issue and then we'll be able to fix the conflicts and merge :)

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    "files.associations": {
+        "*.sql": "oracle",
```
JF002, line 3: SQL / Oracle in InfiniTime ?

src/components/settings/Settings.cpp
```diff
@@ -0,0 +1,18 @@
...
+
+using namespace Pinetime::Controllers;
+
+
+// TODO (team): 
+// Read and write the settings to Flash
+//
+
+void Settings::Init() {
+
```
JF002, line 11: These empty lines at the beginning  and end of the function are not needed.

src/displayapp/screens/Clock.cpp
```diff
@@ -10,243 +10,73 @@
...
+          {
+                [this]() -> std::unique_ptr<Screen> { return WatchFaceDigitalScreen(); },
+                [this]() -> std::unique_ptr<Screen> { return WatchFaceAnalogScreen(); },
+                //[this]() -> std::unique_ptr<Screen> { return WatchFaceMinimalScreen(); },
+                //[this]() -> std::unique_ptr<Screen> { return WatchFaceCustomScreen(); }
+          },
+          Screens::ScreenListModes::LongPress          
+        } {
+
+          settingsController.SetAppMenu(0);
```
JF002, line 42: Is this necessary? I think the app menu index is handled by Tile ?

src/displayapp/screens/ScreenList.h
```diff
@@ -9,16 +9,33 @@
...
     namespace Screens {
+
+      enum class ScreenListModes {UpDown, RightLeft, LongPress};
       template <size_t N>
       class ScreenList : public Screen {
         public:
-          ScreenList(DisplayApp* app, std::array<std::function<std::unique_ptr<Screen>()>, N>&& screens)
-          : Screen(app), screens{std::move(screens)}, current{this->screens[0]()} {
 
+          ScreenList(
```
JF002, line 18: I'm not a huge fan of this indentation. Maybe apply the changes using the original indentation?

src/displayapp/screens/ScreenList.h
```diff
@@ -9,16 +9,33 @@
...
+            screens{std::move(screens)}, 
+            mode{mode},
+            current{this->screens[initScreen]()}
+          {
+            screenIndex = initScreen;
           }
 
           ~ScreenList() override {
-
+            current.reset(nullptr);
```
JF002, line 34: I think this is not necessary : the using pointer will be deleted by the default destructor, and this class does not allocate any LVGL objects that must be de-allocated.
Or did I miss something?

src/displayapp/screens/WatchFaceAnalog.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+  lv_style_init(&hour_line_style_trace);
+  lv_style_set_line_width(&hour_line_style_trace, LV_STATE_DEFAULT, 3);
+  lv_style_set_line_color(&hour_line_style_trace, LV_STATE_DEFAULT, LV_COLOR_WHITE);
+  lv_style_set_line_rounded(&hour_line_style_trace, LV_STATE_DEFAULT, false);
+  lv_obj_add_style(hour_body_trace, LV_LINE_PART_MAIN, &hour_line_style_trace);
+
+  UpdateClock();
+
+  /*lv_obj_t* backgroundLabel = lv_label_create(lv_scr_act(), nullptr);
```
JF002, line 106: Is there something that needs to be done or can these comments be removed?

src/displayapp/screens/WatchFaceDigital.cpp
```diff
@@ -0,0 +1,254 @@
...
+        Controllers::NotificationManager& notificatioManager,
+        Controllers::Settings &settingsController,
+        Controllers::HeartRateController& heartRateController): Screen(app), currentDateTime{{}},
+                                           dateTimeController{dateTimeController}, batteryController{batteryController},
+                                           bleController{bleController}, notificatioManager{notificatioManager},
+                                           settingsController{settingsController},
+                                           heartRateController{heartRateController} {
+  settingsController.SetClockFace(0);
+
+  // init
```
JF002, line 33: Is there something to be done or can we remove these comments?

.vscode/settings.json
joaquimorg: strange... I installed same extensions for another project, VSC making stange things.

src/displayapp/screens/Clock.cpp
joaquimorg: One of the changes made in this PR is to save the position of the application menu for when leaving an app we return to the menu where it is and not to the beginning.
Here we reset this value to go back to the first page when we call the laucher from the clock.

joaquimorg:
Size really starts to be a problem, but we can only solve this when we are able to load this type of images from Flash.
I think that RLE does not work with LVGL, since it had to be written directly to the lcd or create a buffer which caused memory problems.

JF002:
> Size really starts to be a problem, but we can only solve this when we are able to load this type of images from Flash.
> I think that RLE does not work with LVGL, since it had to be written directly to the lcd or create a buffer which caused memory problems.

I *think* it should be possible to use RLE with LVGL using the [image encoder API](https://docs.lvgl.io/v7/en/html/overview/image.html#image-decoder). The doc say that we can implement a decoder than provides images line-by-line to lvgl, which is exactly how the RLE decoder I use works.

joaquimorg:
> > Size really starts to be a problem, but we can only solve this when we are able to load this type of images from Flash.
> > I think that RLE does not work with LVGL, since it had to be written directly to the lcd or create a buffer which caused memory problems.
> 
> I _think_ it should be possible to use RLE with LVGL using the [image encoder API](https://docs.lvgl.io/v7/en/html/overview/image.html#image-decoder). The doc say that we can implement a decoder than provides images line-by-line to lvgl, which is exactly how the RLE decoder I use works.

I hadn't thought about that option.
What I have been trying to develop is a way to use the flash, and have a basic way of managing files, without using litefs, to save memory, the upload would be done via bluetooth of a zip with resources, these resources would be written in the flash where all the images, fonts, etc ... would be easier to customize the fonts and also the background images, without taking up space in the firmware.

JF002:
> What I have been trying to develop is a way to use the flash, and have a basic way of managing files, without using litefs, to save memory, the upload would be done via bluetooth of a zip with resources, these resources would be written in the flash where all the images, fonts, etc ... would be easier to customize the fonts and also the background images, without taking up space in the firmware.

Yes, that kind of features would be a better solution in the long term. Even with optimizations, we'll fill the whole internal memory one day or another :)

So you're implementing your own file system instead of using a lib like littlefs? Is there a reason? Or is Littlefs using a lot of (RAM?) memory?

Anyway, I'm interested in this feature, let us know when you have something to show us!

In the mean time, I'll try to finish the review of this PR, sorry for the delay !

PR #223, opened 2021-03-04
Add link to README.md to WebBLEWatch to update time from web browser

hubmartin:


JF002:
Thanks @hubmartin ! I like your idea to build a web app that  connects to InfiniTime! I hope you'll continue your work and add support for OTA in the near future :-)

PR #229, opened 2021-03-12
Fix #228 (Tap to Wake)

### LINKED ISSUE #228: Tap to wake?
Hello, i would greatly enjoy the ability to tap to wake up my watch. here is the code and what you need to change it to.

code:
```
void SystemTask::OnTouchEvent() {
  if(isGoingToSleep) return ;
  NRF_LOG_INFO("[systemtask] Touch event");
  if(!isSleeping) {
    PushMessage(Messages::OnTouchEvent);
    displayApp->PushMessage(Pinetime::Applications::DisplayApp::Messages::TouchEvent);
  }
}
```
change to:
```
void SystemTask::OnTouchEvent() {
  if(isGoingToSleep) return ;
  NRF_LOG_INFO("[systemtask] Touch event");
  if(!isSleeping) {
    PushMessage(Messages::OnTouchEvent);
    displayApp->PushMessage(Pinetime::Applications::DisplayApp::Messages::TouchEvent);
  } 
  else {
      if(!isWakingUp) {
      NRF_LOG_INFO("[systemtask] Screen tapped, waking up");
      GoToRunning();
      }
    }
}
```

FlameKat53:
Simple as that :)

-Sincerely, Evan Rodgers.

### DIFF
--- a/src/systemtask/SystemTask.cpp
+++ b/src/systemtask/SystemTask.cpp
@@ -251,7 +251,13 @@ void SystemTask::OnTouchEvent() {
   if(!isSleeping) {
     PushMessage(Messages::OnTouchEvent);
     displayApp->PushMessage(Pinetime::Applications::DisplayApp::Messages::TouchEvent);
-  }
+  } 
+  else {
+      if(!isWakingUp) {
+      NRF_LOG_INFO("[systemtask] Screen tapped, waking up");
+      GoToRunning();
+      }
+    }
 }
 
 void SystemTask::PushMessage(SystemTask::Messages msg) {
--- a/src/systemtask/SystemTask.cpp
+++ b/src/systemtask/SystemTask.cpp
@@ -246,18 +246,18 @@ void SystemTask::GoToRunning() {
 }
 
 void SystemTask::OnTouchEvent() {
-  if(isGoingToSleep) return ;
+  if (isGoingToSleep)
+    return;
   NRF_LOG_INFO("[systemtask] Touch event");
-  if(!isSleeping) {
+  if (!isSleeping) {
     PushMessage(Messages::OnTouchEvent);
     displayApp->PushMessage(Pinetime::Applications::DisplayApp::Messages::TouchEvent);
-  } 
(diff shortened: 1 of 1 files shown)


Avamander:
Indentation is not according to the rules specified in clang-format configuration file, please run clang-format before making a PR, thanks.

FlameKat53:
> Indentation is not according to the rules specified in clang-format configuration file, please run clang-format before making a PR, thanks.

my bad!

joaquimorg:
Would be great if we could choose between single tap or double tap, I prefer double tap to avoid turning the screen on unintentionally.

FlameKat53:
> Would be great if we could choose between single tap or double tap, I prefer double tap to avoid turning the screen on unintentionally.

a challenge :)

JF002:
I also think a double tap would be more effective to avoid false positive when you are wearing the watch.
The touch driver is able to detect double taps, so it should be easy to implement :)

JF002:
I've just had a look at the gesture detection from the touch controller : you can check the gesture using the driver:

```
auto touch = touchPanel.GetTouchInfo();
if(touch.isTouch && touch.gesture == Drivers::Cst816S::Gestures::DoubleTap) {
  // A double tap is detected!
}
```

However, this will only work if the TWI master and touch controller are woken up:

```
twiMaster.Wakeup();
touchPanel.Wakeup();
```

But you cannot call these methods in `SystemTask::OnTouchEvent()` because this method is called by an IRQ, and calls FreeRTOS functions than cannot be called from an IRQ.

There are probably multiple solutions to fix it. Here is one on the top of my head : when a touch event is detected, push a message in the event queue of SystemTask. In the handler of this event, you can check if the system is sleeping. If it's sleeping, you can enable the touch controller, read the gesture, and go back to sleep or continue the wake up according to the gesture returned by the driver.

muesli:
This looks like something @joaquimorg already solved in his fork. I'm not sure (yet) how much these projects diverged, but maybe this is something we could backport?

PR #232, opened 2021-03-15
Refactor Clock to use `lv_label_set_text_fmt` and other minor updates

nscooling:
Minor refactoring of Clock.h/.cpp

1. Replaced unnecessary `sprintf` statements with the build-in `lvgl` `lv_label_set_text_fmt` function (designed for that exact purpose)
2. Cleaned up Clock construction and member-initialisation list, where the NSDMI can be used in preference (e.g. defaults zero values and `nullptr`)
3. Cleaned up some logic code
4. Change some instances of `||` to `or` (this is a C++ keyword, as is `and` and `not` and has been since the original C++98 standard)
5. Ran Clock.* through clang-tidy based on `.clang-format` file at project root
6. Replaced any remaining NULL with nullptr
7. Replaced C-style casts with C++ casts

### DIFF
--- a/src/displayapp/screens/Clock.cpp
+++ b/src/displayapp/screens/Clock.cpp
@@ -17,62 +17,30 @@ using namespace Pinetime::Applications::Screens;
 
 namespace {
 
-char const *DaysString[] = {
-        "",
-        "MONDAY",
-        "TUESDAY",
-        "WEDNESDAY",
-        "THURSDAY",
-        "FRIDAY",
-        "SATURDAY",
-        "SUNDAY"
-};
+  char const* const DaysString[] = {"", "MONDAY", "TUESDAY", "WEDNESDAY", "THURSDAY", "FRIDAY", "SATURDAY", "SUNDAY"};
 
-char const *MonthsString[] = {
-        "",
-        "JAN",
-        "FEB",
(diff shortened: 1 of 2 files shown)


JF002:
Thanks for this PR! And good idea to get rid of these `sprintf`!
Also, today, I learned that you can use `and` and `or` in C++! I honestly didn't know these were keywords from the language!

Could you try to fix those conflicts that appears since I merged #217 ? Note that the content of Clock.cpp is now split across Clock and WatchFaceDigital.

Thanks!

nscooling:
I'll leave this until the #238 C++14 PR is accepted as I'll end up doing lots of repetitive merging. Once we have a stable CMake setup I can realign all the PRs with the post #217 merge.

JF002:
I've just merged #238 :)

Avamander:
Autoformat PR has been merged as well, so it should be possible to rebase without a lot of future conflicts.

JF002:
As this PR a too old, it cannot be easily merged (Clock code was split in Clock / Analog / Digital and the new PineTimeStyle). I decided to re-apply those changes on those classes manually in PR #520.
Thanks @nscooling for your work, and sorry I cannot merge your branch directly!

PR #233, opened 2021-03-15
Clean up contribute.md

muesli:
Fixes a bunch of typos, improves grammar, sticks to a consistent line
width.

### DIFF
--- a/doc/contribute.md
+++ b/doc/contribute.md
@@ -1,40 +1,70 @@
 # How to contribute?
+
 ## Report bugs
-You use your Pinetime and find a bug in the firmware? [Create an issue on Github](https://github.com/JF002/Pinetime/issues) explaining the bug, how to reproduce it, the version of the firmware you use...
+You use your Pinetime and found a bug in the firmware?
+[Create an issue on Github](https://github.com/JF002/Pinetime/issues) explaining
+the bug, how to reproduce it, which version of the firmware you use, etc.
+
 ## Write and improve documentation
-Documentation might be incomplete, or not clear enough, and it is always possible to improve it with better wording, pictures, photo, video,... 
+Documentation might be incomplete or not clear enough, and it is always possible
+to improve it with better wording, pictures, photo, video, ...
 
-As the documentation is part of the source code, you can submit your improvements to the documentation by submitting a pull request (see below).
-## Fix bugs, add functionalities and improve the code
-You want to fix a bug, add a cool new functionality or improve the code? See *How to submit a pull request below*.
+As the documentation is part of the source code, you can submit your
+improvements to the documentation by submitting a pull request (see below).
+
+## Fix bugs, add functionality, and improve the code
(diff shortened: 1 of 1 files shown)


JF002:
Thanks for your PR!

However, it's not easy to see what was actually changed since the whole file has changed when adding new lines.
Was this really necessary? I mean most of IDE and text editor can do "soft wrap" to adapt the length of the line to you window and screen. For example, in CLion : 
image

Avamander:
Line width should be left to the client, hard wrapping ruins display on narrow screens such as phones.

muesli:
Similar to source code, this is a general recommendation for markdown from the CommonMark style guide. The intention is to improve readability on the terminal. Unlike source code, markdown renderers typically use soft-wraps, so the content automatically reflows and adapts to the output width (see @JF002's screenshot above for an example).

Avamander:
> The intention is to improve readability on the terminal.

This content is not viewed only on terminals, it's also visible on web, both mobile and desktop. The age of VT-100 terminal-oriented web or text is seriously over.

muesli:
> This content is not viewed only on terminals, it's also visible on web, both mobile and desktop. The age of VT-100 terminal-oriented web or text is seriously over.

Then you really don't need to worry, because on the web these newlines don't get interpreted. For everyone who's still on the terminal, this is generally considered an improvement.

JF002:
I agree with  @Avamander, not everyone use terminal to view and edit markdown (or source code) files. I personally work on 2 1920x1080 display and find it irritating when a file renders on less than a third of the width available.
Doesn't terminal apps allow to softwrap long lines?

Anyway, it looks like this PR (together with #234 contains 2 modifications : typos and grammar, and reformat (fixed width). Maybe you could split these changes in 2 PR so that we can at least fix typos and grammar until even if we don't agree with the fixed with?

muesli:
@JF002 I'll try to get to that, I just need to figure out how I can actually prevent my markdown editor from re-wrapping these lines :laughing:

pfeerick:
Whilst there are  [pros and cons of wrapping and not wrapping](https://about.gitlab.com/blog/2016/10/11/wrapping-text/), I also prefer to let wrapping for text documentation at least to be handled by the viewer / editor... as that allows for the various screen sizes to be taken into account (code is different, especially Python because of those damn PEP8 style guide requirements). 

Out of curiosity though, I made a [simple markdown doc with some limited width text and running text](https://github.com/pfeerick/junkTester/blob/master/wordWrap.md) to double check, and both actually wrap to fit the screen width, because both the GitHub and GitLab Markdown renderer are looking for the double line break before a line break.

muesli:
> Out of curiosity though, I made a [simple markdown doc with some limited width text and running text](https://github.com/pfeerick/junkTester/blob/master/wordWrap.md) to double check, and both actually wrap to fit the screen width, because both the GitHub and GitLab Markdown renderer are looking for the double line break before a line break.

That's what I've been pointing out twice in this thread already.

PR #235, opened 2021-03-15
Small correction to git URL in documentation files

MFAshby:
Looks like some references to the old repository name remain. GitHub redirects if you click them, but I think it's better to keep them up to date.


JF002:
Yes, I renamed the project just after the release of InfiniTime 0.14, and haven't changed all the links yet. Thanks!

PR #238, opened 2021-03-16
Resolved C++14 Cmake build issues so correctly building to C99/C++14

nscooling:
The CMake configuration is, to put it mildly, a bit of a mess. Multiple files are configuring and reconfiguring compile and link flags. As a result the previous update to set C++14 was being overridden by multiple other directives.

The CMake setup is now configured to use one, project wide, language directive for C99 and C++14 in the base CMakeFiles.txt file.

This is confirmed by building all versions (Pinetime-app, pinetime-graphics, etc.) to C++14 standard.

As a test, `src/systemtask/SystemTask.cpp` now uses C++14 `std::make_unique` rather than `new`

Assuming this PR is accepted I will, in another PR, replace all other occurrences of `new` with `make_unique`.


Avamander:
Oh, absolutely awesome.

I can also confirm it builds:
```
post build steps for pinetime-mcuboot-app-0.14.0
 368764	    672	  44544	 413980	  6511c	pinetime-mcuboot-app-0.14.0.out
[100%] Built target pinetime-mcuboot-app
```

Panky-codes:
Is it possible to jump to C++ 17? I think features like std::optional would come in handy. Also, the compiler we use is GCC 9, so it should support 17.

nscooling:
It is possible as I've been using C++17 for some time. However I didn't want to push the project too far too quickly. 
I also deferred 17 due to the issues you raised, it's frustrating to have annoying warnings. Register is the biggest issue as it's been removed in 17 (deprecated in 11) 
Std::optional is very useful, but much more so is guaranteed copy elision, structured bindings and if-unit. These really change the style of programming. I'd suggest getting C++14 accepted and then look at the best way to introduce 17 (or even 20). 
Gcc10 is available and I'm using it locally. So would be good to upgrade this first as well as the Nordic library.

JF002:
Awesome, thank you! Yes, I agree, CMake files in the project are a bit messy... They come from another project using CMake on the NRF52, and since then, I did many modifications to support new libraries, build DFU, use Docker, remove parts of the NRF SDK,... and they really need some refactoring!

> The above is error is mainly popping up from date.h
Note that the date library I integrated comes from https://github.com/HowardHinnant/date (if I recall correctly). I think newer version of C++/GCC now integrate this library natively. If we switched to the native date.h, the error might disappear :)

JF002:
Looks good to me! And I totally agree with using `make_unique` instead of `new` wherever possible!
Could you try to fix those conflicts that appeared since I merged other PRs?

nscooling:
All updated and merged with v0.15.0. Built and tested on local unit.

Issue #213, opened 2021-02-21
screen corruption


ashkitten:
sometimes parts of the screen will get corrupted a bit, making text hard to read. i've noticed this in both the notification view and the main watchface, usually a horizontal line of incorrect pixels that might spread vertically when the screen updates (such as notification progress bar updates at the bottom of the screen). it seems to work itself out when swiping between screens.

JF002:
Yes, I also noticed there glitches on the display. They are easily reproducible in the paddle app.
They have already been reported in #77 and #91.

JF002:
Fixed in 0.15

PR #240, opened 2021-03-18
FIX screen corruption (Issue #213)

joaquimorg:
After some investigation I managed to solve the problem that caused the image on the screen to be left with some "garbage".

I also took the opportunity to optimize the function  **LittleVgl::FlushDisplay** a little.

Before :

IMG_20210318_181330

After :

IMG_20210318_192919


JF002:
Awesome! I spent so much time trying to integrate the hardware scrolling effect with LVGL that I hadn't had the time/energy to refactor and optimize it! But I didn't know it was buggy!

Hopefully it'll also fix glitches in the game 2 and the brightness app!

PR #244, opened 2021-03-20
Update to LVGL 7.11.0 Final

joaquimorg:
Update to LVGL 7.11.0 Final


JF002:
Thanks, git does not see any conflict any more. Do you know what was the issue? I'm still learning submodules, and I'm still afraid to break things if I'm not using it correctly.

Also, in this PR (https://github.com/JF002/InfiniTime/pull/231/files), submodule got updated to, but with different file... Any idea what is happening?

PR #249, opened 2021-03-23
add siglo as a companion app

alexr4535:


JF002:
Thanks, and welcome aboard!

Issue #247, opened 2021-03-22
Digital clock  - placeholder text/icons visble before final render


pfeerick:
Noticed with 0.15 (although it exists as at least as far back as 0.12) that the digital clock watchface is first populated with placeholder text / icons, and then the final display set, when returning to it from the menu (not from sleep). It became more obvious with the 0.15 version as there is a flash of colour up the top as the charge and bluetooth icons appear and disappear (they disappear to reflect the true state of the device). 

![20210322_125517](https://user-images.githubusercontent.com/5500713/111935437-397a0380-8b0f-11eb-85de-fb1a1fa78538.gif)

petterhs:
This was fixed in #258 and can be closed now

pfeerick:
Thanks for the reminder... yes, this has been gone for a while now!

PR #251, opened 2021-03-28
Fix issue #247

joaquimorg:
Fix the issue #247


joaquimorg:
This code is in new PR, this is now obsolete.

PR #252, opened 2021-03-28
Battery non blocking read, fix calibration and System Info redesign

joaquimorg:
- Battery voltage non blocking read.
- Battery calibration.
- System Info redesign.

IMG_20210328_163657
IMG_20210328_163704


joaquimorg:
Not ready, needs more tests...

Issue #62, opened 2020-09-14
P8 support


nextfullstorm:
Hi,
can anyone tell me how to build Infinetime for Colmi P8? It's wide-known clone of Pinetime (or Pinetime clone of P8? I'm not sure). [This](https://github.com/atc1441/ATCwatch) awesome project proves that there is little to no difference in hardware and pinout differs. I don't have enough skill to resolve it myself. After blind dfu upgrade from stock fw to latest Infinitime release my device was bricked and I have to restore it via SWD programmer.

JF002:
Hi!
I'm not working on a P8 port for now, because I do not have one and I doubt I would have the time to maintain both of them.
You're right, there are very few differences between the PineTime and the P8, so the port should not be too complicated. I'm open to pull requests if someone wants to implement the support for the P8 :)

Please note that the DFU upgrade from ATCWatch and InfiniTime are **not** compatible. ATCWatch and InfiniTime do not use the same bootloader and memory map, which means that you can DFU from ATCWatch to InfiniTime or from InfiniTime to ATCWatch.
There is a project to build a tool that would allow that but I don't think it's ready to be used for now.

maidenone:
i have worked with DFU and NRF52, it should be possible to create a new DFU package that contains the correct bootloader.

hrmckay:
If you look at issue # 69 for the wasp-os support for the Pinetime at https://github.com/daniel-thompson/wasp-os/issues you will see what had to be changed to get things working for that operating system.

Hugh

nextfullstorm:
Tnx guys. As I understand, main problem here is memory map difference. And THAN maybe pinout/accelerometer difference. I will look into this and write back if I have meaningful result.

maidenone:
where can i buy a p8?

nextfullstorm:
> 
> where can i buy a p8?

On aliexpress for sure. I got mine there. IDK if I can post direct links.

maidenone:
well i ordered 2, will se when i get them. 
do not know when i will get the time, but i intend to look into it.

JF002:
> i have worked with DFU and NRF52, it should be possible to create a new DFU package that contains the correct bootloader.

@maidenone The tool I was talking about is [wasp-reloader](https://github.com/daniel-thompson/wasp-reloader). You might want to have a look at it before you start something from scratch :)
And thanks in advance for your futur PR for the support of the P8 in InfiniTime :)

[...]

PR #255, opened 2021-03-31
[WIP] Motion sensor integration : step counting and wake on wrist rotation

JF002:
This PR adds support for the BMA421 motion sensor and add the step counting and "wake up on wrist rotation" functionalities. There is also a small demo app that displays the value of the 3 axis in real time.

wrist_wake_up_short

motion_graph

TODO:

 - [ ] Check OTA (during development, I noticed that the watch would sometimes crash/freeze when booting the new FW. It didn't brick the watch, but the bootloader would revert to the previous version)
 - [ ] Check that further OTA are still possible (install a new version of the FW with motion sensor enabled onto a version that already use the motion sensor.
 - [x] Check power usage (the system task now wakes up every 100ms to check for wrist rotation) - see comment from @vbelloir bellow :  https://github.com/JF002/InfiniTime/pull/255#issuecomment-812920335
 - [x] Usage axis remapping functionality from the BMA instead of manually swapping X and Y values -- **Won't do** : The axis remapping has no effect on the raw x/y/z values returned by the sensor. Instead, the remapping is only apply on 'features' like step counting. I found no way to check that the remapping was actually enabled, and the documentation contains too few info on how to use it. I've decided to not use this functionality and to swap X and Y axis at the driver level.
 - [x] Add a configuration app to enable/disable wake up on wrist rotation (especially if it uses a lot of power) - Done in https://github.com/JF002/InfiniTime/pull/258
 - [x] Reset step counter everyday at midnight
 - [x] If the init of the BMA421 fails, disable it instead of assert and crash.
 - [ ] Check that the step count is correct
 - [x] Fix I²C (TWIMaster) driver / peripheral that become unresponsive under heavy load


JF002:
DFU for testing (**_might not be safe to install on a sealed device_**)
[pinetime-mcuboot-app-dfu-0.16.0.zip](https://github.com/JF002/InfiniTime/files/6238656/pinetime-mcuboot-app-dfu-0.16.0.zip)

MFAshby:
Installed, will charge to full now and report how long it lasts with this firmware!

JF002:
Here are 2 DFU files with different versions (0.16.10 & 0.16.11) so that you can test multiple OTAs (these files are not "official" releases, just test DFU).

[pinetime-mcuboot-app-dfu-0.16.10.zip](https://github.com/JF002/InfiniTime/files/6250047/pinetime-mcuboot-app-dfu-0.16.10.zip)
[pinetime-mcuboot-app-dfu-0.16.11.zip](https://github.com/JF002/InfiniTime/files/6250048/pinetime-mcuboot-app-dfu-0.16.11.zip)

JF002:
My Pinetime is running for more than 2 days on this FW, and there ~40% of battery available.
It looks like this feature doesn't have a dramatic impact on the battery life!

motion_batterylife

ildar:
пт, 2 апр. 2021 г., 1:19 JF002:

@JF002 <https://github.com/JF002> pushed 1 commit.

   - f8fde06
   <https://github.com/JF002/InfiniTime/commit/f8fde069abe7a1c73518bc210bac42602c5b4c13>
Code
   cleaning in BMA421 driver. Do the axis inversion in the driver and not in
   the application.

One more hardware-specific thing hardcoded. Ought to change for P8 port.

JF002:
> One more hardware-specific thing hardcoded. Ought to change for P8 port.

I can make this configurable by adding a parameter to the constructor of the driver. You'll then be able to enable it or not according to the hardware.

vbelloir:
I made some quick current measurements this evening, with this version.
I measured in "standby" 1.65mA, with bootloader, whereas I was at about 1.43mA qhith 0.14.0.

JF002:
> I made some quick current measurements this evening, with this version.
> I measured in "standby" 1.65mA, with bootloader, whereas I was at about 1.43mA with 0.14.0.

Thanks for these measurements! Do you think that increase in power consumption is reasonable for the motion sensor (polling every 100ms) ?

vbelloir:
Yes, from my point of view, it's totally reasonable.

Also, I have a big difference in step counter in comparison with my Suunto Ambit Peak 3.
In 2h, my suunto counted 2700 step, while my pinetime saw 5700 steps.

JF002:
> Yes, from my point of view, it's totally reasonable.
Thanks!

> Also, I have a big difference in step counter in comparison with my Suunto Ambit Peak 3.
> In 2h, my suunto counted 2700 step, while my pinetime saw 5700 steps.

I also find that the step count is very optimistic, but I haven't compared yet with another device. Maybe people who own a P8 device can compare the step count from the P8 with the one from InfiniTime (Linking #62 to ping potential users of P8 :) )

ildar:
P8 doesn't yet see the gyro with this branch. I need to  check pinout
settings.

JF002:
> P8 doesn't yet see the gyro with this branch. I need to check pinout settings.

It would be interesting to compare the step count from InfiniTime with the original firmware of the P8 (sorry for the confusion). 

Apart from the axis remapping, I'm not sure there's anything we can configure that would have an effect on the step counting feature.

pfeerick:
Hey JF... does that 0.16.11 have the same version number as the 0.16.0 one... as if they don't... either Siglo's OTA did something... wrong... or the pinetime crashed and auto-reverted (which is what I think it just did)...

MFAshby:
Firmware 0.16.0 lasted just under 2 days from 90% charge.

I was fairly active during this time (read: not working at my desk) and I noticed the watch would light up fairly frequently just while I was moving around.

I wonder if some longer term experiment could help to refine the activation trigger; e.g. confirming a desired activation by tapping the screen and logging the accelerometer output during desired vs undesired activations.

pfeerick:
> I was fairly active during this time (read: not working at my desk) and I noticed the watch would light up fairly frequently just while I was moving around.

Yes, that reminds me... I was going to ask since there was mention of configurable on time / different on time when motion triggered... perhaps that could/should be paired with a lockout interval? i.e. can only re-trigger 10-20 seconds after last trigger? Something like that?

JF002:
> Hey JF... does that 0.16.11 have the same version number as the 0.16.0 one... as if they don't... either Siglo's OTA did something... wrong... or the pinetime crashed and auto-reverted (which is what I think it just did)...

0.16.11 displays that version in the SystemInfo app. If you see 0.16.0, then it's 0.16.0 that's running. I've also noticed that "sometimes", the new FW crashes (probably because it coulnd't init the motion sensor) and then the bootloader reverts to the previous version. It does not brick the device, but it's a bit annoying. Since then, I committed this (https://github.com/JF002/InfiniTime/pull/255/commits/52a90288fd2b744b68d584db316fcf3fede84262) to disable the motion sensor if it's frozen. In my observations, if you wait a bit and reset, it would work again. Other developers told me they also noticed that the BM421 would not initialize from time to time, with no known workaround :/

@MFAshby @pfeerick  Yes, of course, there are probably many ways to improve the wake on wrist functionality. Feel free to contribute to this :)

JF002:
I noticed that *sometimes*, the firmware crashes under heavy load (swipe the display to switch to/from the menu and the clock face, espcially with the analog one, when ble is connected). 
At first sight, the issue comes from the I²C driver (TwiMaster) and/or peripheral. This freeze is probably linked to other crashes noticed in games.
This issue becomes more visible when this branch and #258 are merged together. I think #258 is a bit more taxing on the CPU time.

It's a bit worrisome as we have already implemented a workaround for a HW bug similar to this one on the TWI side. I'll try to investigate a bit more, but any help on this subject is welcome!

joaquimorg:
I was cleaning up the twi driver and used the sdk functions to see if it improves.
At first glance I got the feeling that the touch scree was more responsive.
However, the problem of bma421 remains.

JF002:
> I was cleaning up the twi driver and used the sdk functions to see if it improves.

It's funny because I'm doing exactly the same thing with the same conclusion! No more frozen TwiMaster, but BMA still fails to init from time to time.

joaquimorg:
There is something in the initialization of bma421 that is not being done correctly, which leaves it in a strange state.

JF002:
When it fails, it looks like it fails sending the configuration file needed for the step counter.

JF002:
I spent most of the day on this issue : I cannot find a way to make sure that the BMA will init correctly each time.
As we said on the chat, the simplest solution to this issue would be to disable the motion sensor until the firmware is verified. This will ensure that the OTA won't be automatically reverted because the bma failed its init just after the OTA.

I tried another solution:
 - Before the init of the bma, start a timer which elapses in 3s and fires an interrupt.
 - If the init is succesful in less than 3s, stop the timer and continue normally
 - If the init failed (driver stuck in an infinite loop)
   - The timer will trig the IRQ
   - In the handler, set a variable that'll unlock the infinite loop in the driver (and make the driver return an error code)
   - The error code will propagate up to the Bma421 class, and the motion sensor won't be enabled
   - But the rest of the application (and the touch sensor) works!

I don't like this solution as it requires to use global variable and to modify the NRFX driver and... it's just a hack.

I'm going to sleep on this, but feel free to tell me what you think about it and help to find a better solution :)

joaquimorg:
It is a solution, it does not mean that it is definitive, but it is a solution.

joaquimorg:
I found this :
https://community.bosch-sensortec.com/t5/MEMS-sensors-forum/BMA423-lockup/td-p/8927
the problem is identical, and has a solution !

JF002:
Indeed, the issue looks similar to ours:

> The first time I ran the code, calling bma423_init() was successfully able to communicate with the chip and read the chip ID. But when calling bma423_write_config(), the chip accepted a few dozens of 8-byte chunks, but at some point, it stopped responding (doesn't ACK the I2C address or pulls the bus down anymore). Now bma423_init() fails every time.

And the solution is:

> After some trial and error the problem was that I was splitting the I2C write access into 2 I2C transactions (1st: write register address without stop condition, 2nd: write data (single byte or burst) with stop condition). This doesn't always work.
Now I buffer the reg address + data to write and do the write access in one single transaction. This works every time.

But... The driver from InfiniTime already packs the write operation (address + data) into 1 transaction:
```

TwiMaster::ErrorCodes TwiMaster::Write(uint8_t deviceAddress, uint8_t registerAddress, const uint8_t *data, size_t size) {
  ASSERT(size <= maxDataSize);
  xSemaphoreTake(mutex, portMAX_DELAY);
  TwiMaster::ErrorCodes ret;

  internalBuffer[0] = registerAddress;
  std::memcpy(internalBuffer+1, data, size);
  auto err = nrfx_twim_tx(&m_twi_master, deviceAddress, internalBuffer , size+1, false);
  if(err != 0){
    return TwiMaster::ErrorCodes::TransactionFailed;
  }

  xSemaphoreGive(mutex);
  return TwiMaster::ErrorCodes::NoError;
}
```

Read happens in 2 transactions however (write address + read data) :

```
TwiMaster::ErrorCodes TwiMaster::Read(uint8_t deviceAddress, uint8_t registerAddress, uint8_t *data, size_t size) {
  xSemaphoreTake(mutex, portMAX_DELAY);
  TwiMaster::ErrorCodes ret;

  auto err = nrfx_twim_tx(&m_twi_master, deviceAddress, &registerAddress, 1, true);
  if(err != 0){
    return TwiMaster::ErrorCodes::TransactionFailed;
  }

  err = nrfx_twim_rx(&m_twi_master, deviceAddress, data, size);
  if(err != 0){
    return TwiMaster::ErrorCodes::TransactionFailed;
  }
  xSemaphoreGive(mutex);

  return TwiMaster::ErrorCodes::NoError;
}
```

Is there any way to write the address and then read data into 1 single operation to ensure there's no additional delay between write and read ?

JF002:
Ok, here's the datasheet of the bma423 : https://www.bosch-sensortec.com/media/boschsensortec/downloads/datasheets/bst-bma423-ds004-00.pdf

They say that the read happens in 2 phases : 1 write and 1 read, with a start condition for both. So the implement should be OK.
But the data sheet also say this : 
image

and this:
image

Next thing I'm going to try : add delays after read and write operations!
And enable that watchdog if it doesn't work!

JF002:
Great news : we found a solution!

Post-write access timings were already taken care of by the Bosch SDK, adding delays would not help.

We noticed that when we removed the call to the soft reset functions in the init method, the BMA would initialize correctly (no crash), but the step counter would not work... 

Finally, the solution was to reset (de-init and re-init) the TWI driver (TwiMaster) just after the reset! After that, we (@joaquimorg  and I) couldn't reproduce any error during the init of the BMA!

I'll do the modification in this branch (and also switch to the NRF TWI driver as it fixes the random freezes of the touch panel) in this branch before merging into develop along with other PRs!

Thanks everyone for your help and support!

danielh186:
> 
> 
> > P8 doesn't yet see the gyro with this branch. I need to check pinout settings.
> 
> It would be interesting to compare the step count from InfiniTime with the original firmware of the P8 (sorry for the confusion).
> 
> Apart from the axis remapping, I'm not sure there's anything we can configure that would have an effect on the step counting feature.

So I still have a P8 with the original Firmware because I didn't have the time to reflash it. But I don't have a second watch running InfiniTime so I can't test them side to side...

PR #256, opened 2021-04-01
src/CMakeLists.txt: use iHEX format for mcuboot image creation

ildar:
for better control of addresses used by img/fw
+ fix typo (RECOVERYY)


ildar:
please note two fixes this PR brings

JF002:
I thought that `imgtool.py` and `adafruit-nrfutil`only worked with .bin file. Could you confirm they both work as expected with .hex files?
If they do, do the DFU file contain the .hex file or does adafruit-nrfutil conver the hex in bin for the DFU zip file?

I think some companion app (gadgetbridge, amazfish, siglo) expect a .bin file for the OTA functionality, and could be confused if they found a .hex file instead.

ildar:
Ah! that's unfortunate I missed your comment.
This PR surely doesn't bring any regressions. Everything works absolutely
as it should.
DFU zip files contain BIN only. And NRFutil flavors accept HEX and convert
them to BIN as required by DFU uploaders.
Check my latest P8 DFUs, which were produced with this patch applied.

ildar:
NRFutil is the tool that produces DFU. The result must be the same whether
you give it a bin or hex. the resulting DFU zip contains BIN.

ildar:
added one more fix commit

ildar:
up-to-date with the current `develop`.

geekbozu:
Does switching to iHEX actually give is any benefits here? The other changes look good thanks for updating this.

ildar:
On Sun, Oct 3, 2021 at 9:09 AM Tim Keller wrote:

> Does switching to iHEX actually give is any benefits here? The other
> changes look good thanks for updating this.
>
missed this comment because of all the renames...
1. iHEX oftentimes is better because it contains address information. With
IT we have bootloader-targeted images which have tricky offsets that are
good to control tightly. One can do it like this:
> $ hexinfo.py *.hex
> - file: 'hwmonitor.hex'
>   entry: 0x000000DD
>   data:
>   - { first: 0x00008020, last: 0x0000F4D3, length: 0x0000F4D4 }
>
> - file: 'hwmonitor-mcuboot.hex'
>   data:
>   - { first: 0x00008000, last: 0x0000F51B, length: 0x0000F51C }

2. iHEX is accepted by all the software around: openocd, imgtool and many
others. Easily converted by objcopy.

geekbozu:
Thanks for the info, I am further un-opinionated, So I await my fellow devs @JF002 and @Avamander opinions :)

JF002:
We talked about this PR in #1051 and my concerns have been answered so... we can merge this PR.
Thanks @ildar for this PR, and my apologies for the time it took to finally accept it !

ildar:
incredible!
👍

PR #258, opened 2021-04-04
New settings menu and more...

joaquimorg:
First of all I want to apologize for the size of this PR, but I think that this way I can bring to Infinitime the improvements in navigation and aspect that the project deserves.

### In the image we can see the new menus, as well as the comparison between the old application menu and the new one.
infinitime_menus

### I will try to describe as much as possible the changes made:
- Major UI changes and navigation rewriting.
I made visual changes to some applications as well as the application menu, added a new smaller font to be used for example in hart reate.
- Navigation is now done as follows:
Being on the clock and slide down we have access to notifications. 
In the clock we slide it up and access the application menu. 
If we do slide right we access the menu with quick actions.
The logic associated with DisplayApp.cpp that manages navigation and loads applications has been changed,
there is a "LoadApp" function that contains all the applications that can be called, this is where we also indicate what must be done to exit the application and to which we return.
- Alteration of the Sysinfo application.
- New application to see the battery status.
- New flashlight application.
- New way to wake up smarwatch, single tap, double tap, the configuration is prepared for Raise Wrist.
- It is possible to turn off the vibration in the quick settings menu.
- Reading the battery status is done in non blocking mode.
- When we are reading the heart rate the watch does not turn off.
- Great cleanliness in the navigation application, the images that were being used took up a lot of space, in addition to having many images that were the same but with different names, I reduced these images by almost half and turned them into a compressed font which made a big difference in the size of the final binary.
- The settings are saved to the external flash.
At the end of the flash there are 10 blocks of 4k to store these settings, the 10 blocks serve to not always be recording in the same place and in this way minimize the cycles of writing in the same place, so every time we are going to save the settings we advance a block and we mark the old one as invalid, when we reach the 10th block it returns to the 1st block.
- I think that's all, there must be a few more points that I touched but they were just cosmetics.

### File for testing this PR :
*(might not be safe to install on a sealed device)*
[pinetime-mcuboot-app-dfu-0.15.0.zip](https://github.com/JF002/InfiniTime/files/6254766/pinetime-mcuboot-app-dfu-0.15.0.zip)


JF002:
I think this is the biggest PR InfiniTime has ever received :1st_place_medal:  !

And it looks... awesome! It looks like you are porting many of the improvements you made in your PineTimeLite fork into this repo, and I thank you for that!

I'll try to review it asap! With the step counter, the last bugfix and this PR, I think the next release will be awesome!

Avamander:
The rudimentary filesystem is unsuitable for long-term use. It needs to be changed at some point, if not now.

petterhs:
Very nice improvements! Will try to help with review as best I can

src/displayapp/screens/ApplicationList.cpp
```diff
@@ -33,47 +37,41 @@ bool ApplicationList::Refresh() {
...
-          {Symbols::music, Apps::Music},
-          {Symbols::sun, Apps::Brightness},
-          {Symbols::list, Apps::SysInfo},
-          {Symbols::check, Apps::FirmwareValidation},
-          {Symbols::heartBeat, Apps::HeartRate}
+          {
+            {Symbols::stopWatch,    Apps::StopWatch},
+            {Symbols::music,        Apps::Music},
+            {Symbols::map,          Apps::Navigation},
+            {Symbols::shoe,         Apps::Clock},
```
petterhs, line 50: Is it intentional that the "shoe application" returns to Clock?

JF002: Maybe that's a place holder for the new app I added in the motion PR that displays the value of the axis from the accelerometer ?

JF002:
> The rudimentary filesystem is unsuitable for long-term use. It needs to be changed at some point, if not now.

It maybe not as advanced as LittleFS, but it has the merit of existing and of being integrated in InfiniTime :) There's also a wear leveling mechanism (it uses a pool of 10 pages to avoid writing in the same page every time), and a field to store the version (so that we can handle new version of the settings struct).
@Avamander  Do you see any blocking point that would make it unusable to now?

@joaquimorg I'm curious : Is there any reason why you didn't use an existing filesystem like littlefs?

Avamander:
@JF002 

> It maybe not as advanced as LittleFS, but it has the merit of existing and of being integrated in InfiniTime :) There's also a wear leveling mechanism (it uses a pool of 10 pages to avoid writing in the same page every time), and a field to store the version (so that we can handle new version of the settings struct).
> Do you see any blocking point that would make it unusable to now?
 
No, it's usable, but the current solution should be refactored into something more long-term, known, tested and cross-firmware, sooner rather than later..

src/displayapp/fonts/Readme.md
```diff
@@ -22,3 +22,11 @@ Add new symbols:
...
 ```
 
+#### Navigation font
+
+navigation.ttf
+0xe900-0xe929
+
+0xe900,0xe901,0xe902,0xe903,0xe904,0xe905,0xe906,0xe907,0xe908,0xe909,0xe90a,0xe90b,0xe90c,0xe90d,0xe90e,0xe90f,0xe910,0xe911,0xe912,0xe913,0xe914,0xe915,0xe916,0xe917,0xe918,0xe919,0xe91a,0xe91b,0xe91c,0xe91d,0xe91e,0xe91f,0xe920,0xe921,0xe922,0xe923,0xe924,0xe925,0xe926,0xe927,0xe928,0xe929
+
+lv_font_conv --font navigation.ttf -r '0xe900-0xe929' --size 70 --format lvgl --bpp 2 --no-prefilter -o lv_font_navi_70.c
```
JF002, line 32: If I understand correctly, you generated navigation.ttf from the SVG files, and then convert this TTF into a LVGL font?

src/drivers/TwiMaster.cpp
```diff
@@ -61,9 +62,11 @@ void TwiMaster::Init() {
 }
 
 TwiMaster::ErrorCodes TwiMaster::Read(uint8_t deviceAddress, uint8_t registerAddress, uint8_t *data, size_t size) {
-  xSemaphoreTake(mutex, portMAX_DELAY);
+  // this is causing an error when came from sleep
```
JF002, line 65: I think we should analyze this is more details : the bus can be accessed by multiple tasks (Display for the touch and HeartRate for the heart rate sensor). Without this mutex, 2 tasks can stsart Read operations on the bus at the same time, which will cause undefined behaviors.

There's probably an issue is the Sleep/Wakeup order: 
 - the bus was still in use when going to sleep (a transaction couldn't finish because it went to sleep)
 - a task started a transaction before the bus was woken up

src/drivers/Cst816s.cpp
```diff
@@ -32,67 +32,47 @@ void Cst816S::Init() {
   vTaskDelay(5);
   twiMaster.Read(twiAddress, 0xa7, &dummy, 1);
 
+  /*
+  [2] EnConLR - Continuous operation can slide around
+  [1] EnConUD - Slide up and down to enable continuous operation
+  [0] EnDClick - Enable Double-click action
+  */
+  static constexpr uint8_t motionMask = 0b00000101;
```
JF002, line 40: Did you find more detailed info about the CST816S? 
Does this register enables functionalities that weren't working before?

joaquimorg: Yes, this is for double tap to work and for continuous reading.

src/displayapp/screens/ApplicationList.cpp
joaquimorg: yes, is for the app to show the axis, or in the future to show the steps...

src/displayapp/fonts/Readme.md
joaquimorg: Yes, a compressed lvgl font.

src/drivers/TwiMaster.cpp
joaquimorg: yes that point was difficult, in my fork is working fine, but here I couldn’t get him to read the touch ... I didn’t investigate much, if anyone can find out why,

src/drivers/Cst816s.cpp
JF002: Maybe we can enable multitouch using this register? I'm not sure we need it, but I remember I've never manager to get more than one touch point from the touch panel.

joaquimorg:
> > The rudimentary filesystem is unsuitable for long-term use. It needs to be changed at some point, if not now.
> 
> It maybe not as advanced as LittleFS, but it has the merit of existing and of being integrated in InfiniTime :) There's also a wear leveling mechanism (it uses a pool of 10 pages to avoid writing in the same page every time), and a field to store the version (so that we can handle new version of the settings struct).
> @Avamander Do you see any blocking point that would make it unusable to now?
> 
> @joaquimorg I'm curious : Is there any reason why you didn't use an existing filesystem like littlefs?

I agree that this is a point that must be discussed.
It may not be the best method but it was what I chose, we are talking about internal information that resides in an area of ​​the flash that can be reserved for this and there is no need to have the complexity of a FS just to store this information.
My starting point is always the same limit of memory and space in the firmware, in the case of these configurations we actually have more written than readings, so I implemented the model of having the 10 blocks, but they could be more, in total losing the 40k to saving this information is reasonable.
In the tests that I have done with the resources like images saved in the flash, what I have noticed is the access speed, we are at the same limit as the 8mhz lcd, which makes the loading of the images to the lcd perceptible, and I am reading sequentially, if we add the over head of an FS it can be even bigger, in the near future I will do tests with litefs to see the result.
Or if someone wants to do these tests it was great so we had a way to compare and make a decision.

src/drivers/TwiMaster.cpp
JF002: I'll definitely investigate this :)

src/displayapp/fonts/Readme.md
JF002: If you still have it, would you like to document how to generate the ttf from the source svg files?

Avamander:
@joaquimorg JF002 said that reads from flash do not use DMA, that might provide a big speedup for UI.

src/displayapp/fonts/Readme.md
joaquimorg: Yes, the process is simple, I use https://icomoon.io/app in this site you can create ttf from svg files.
I will add this information to the documentation.

src/drivers/Cst816s.cpp
joaquimorg: I didn't see anything in the documentation about it, what I realized was that the double touch didn't work and I went to look at the documentation and I found this information, which also indicated the continuous reading mode and activated to try to solve the problem when it was always reading the touch.

joaquimorg:
One point I forgot to mention, is that the InfiniPaint application now has one more functionality, long press changes the color !

IMG_20210404_225323

src/drivers/Cst816s.cpp
JF002: > and activated to try to solve the problem when it was always reading the touch

Does it actually fix these freezes, by any chance?

JF002:
> One point I forgot to mention, is that the InfiniPaint application now has one more functionality, long press changes the color !

Let's create the hashtash #DrawnWithPineTime on Mastodon/Twitter :-D

src/drivers/Cst816s.cpp
joaquimorg: From what I tested, I was being able to play pong for several minutes without blocking, what I realized was that the touch is sometimes unresponsive, but just lift your finger to get it working again, but it doesn't block the application.

src/components/battery/BatteryController.cpp
```diff
@@ -1,49 +1,89 @@
 #include "BatteryController.h"
 #include <hal/nrf_gpio.h>
+#include <nrfx_saadc.h>
 #include <libraries/log/nrf_log.h>
 #include <algorithm>
+#include <math.h>
 
 using namespace Pinetime::Controllers;
 
+#define SAMPLES_IN_BUFFER 1
```
JF002, line 10: These constants and static variables can be move into the Battery class, as member variables.

src/drivers/Cst816s.cpp
JF002: I think the touch panel automatically goes to sleep mode after a while. That's maybe why it would appear unresponsive from time to time.

src/systemtask/SystemTask.cpp
```diff
@@ -199,10 +212,14 @@ void SystemTask::Work() {
...
           }
           lcd.Sleep();
-          touchPanel.Sleep();
-
           spi.Sleep();
-          twiMaster.Sleep();
+
+          // Double Tap needs the touch screen to be in normal mode
+          if ( settingsController.getWakeUpMode() != Pinetime::Controllers::Settings::WakeUpMode::DoubleTap ) {
+            touchPanel.Sleep();
```
JF002, line 219: I thought that the touch panel would automatically wake up (and send an IRQ) when it was touched, even in sleep mode. Do you confirm sleep mode disabled double tap detection?

joaquimorg: In sleep mode he continues to send the irq but afterwards we are unable to know what the type of event was, if he is sleeping when doing the reset to wake we lost that information.

JF002:
I did a first pass across all the changes in this PR, and it look mostly good to me! Impressive work, @joaquimorg !

Here's a small todo list:
- [ ] Check the mutex issue in TwiMaster.cpp (https://github.com/JF002/InfiniTime/pull/258#discussion_r606838925)
- [ ] Discussion around filesystem if needed (see my opinion below)
- [ ] If possible, try to map the memory usage, especially regarding the RAM usage : is it full or is there still a bit of room left?
- [ ] Pay attention to the recovery firmware : it might be needed to disable the motion sensor (as it tends to fail its initialization) and the setting manager when the recovery firmware is built

Regarding the file system : I know we talked a lot about using littlefs as filesystem to store persistent data like settings, applications settings, activity data, graphical assets,... but... it's not done yet.
The FS implemented in this PR is really simple (113 lines of C++ code!) and does exactly what we need right now : store a simple binary struct containing a few user settings. It even implement a wear leveling mechanism on 10 flash pages. Maybe it won't scale as well as littlefs when we'll extend the usage of this fs to other things than user settings, but it does the work for now.
In my opinion, this is good enough for now, and I think it would be a shame to delay the integration of this PR because we think we can do better on the FS side.
That does not prevent us from exploring other implementations like the one introduced in #162 (it integrates littlefs but does not implement any use-case with it) and changing the storage backend in future PR/release.

src/drivers/TwiMaster.cpp
JF002: I think the issue comes from `SystemTask::OnTouchEvent()` (SystemTask.cpp:271). This method is called in the context of the IRQ from the touch panel. It's not called in the context of the scheduler of FreeRTOS. 

When the wake up on double-tap is enabled, it receives the interrupt from the touchpanel and checks that the touch event is a double tap by reading the status from the touch panel. The method TwiMaster::Read() tries to get the semaphore, but freertos throws an assert because the scheduler is suspended (sleep mode). It probably does so to prevent deadlocks in case the semaphore were not available at that time.

I think about 2 solutions to work around this issue:
- Add *_FromIRQ() methods in CST816S and TwiMaster and remove the semaphore from these implementations
- Instead of checking the touch event in `SystemTask::OnTouchEvent()`, send a message to the SystemTask so that SystemTask checks the event and decides to wake up or stay asleep. In this case, SystemTask would run in the context of the scheduler and everything would be fine.

I think 2nd solution is the best, even if it's a bit more complex to implement.

joaquimorg: Nice, the second solution worked.

src/drivers/Cst816s.cpp
JF002: Mhm I noticed a crash while playing 2048, but I could play paddle for a few minute without any crash!

petterhs:
Why is the transition from clock to the quick settings menu on the left not as smooth as moving down to application list? Is is because the display writes pixels from top to bottom? And is that something that we can solve?

petterhs:
>     * [ ]  If possible, try to map the memory usage, especially regarding the RAM usage : is it full or is there still a bit of room left?

Do you mean something other than the memory info added to the About section in settings?

src/components/battery/BatteryController.cpp
JF002: Ok, now I understand why you declared them as static variables instead of member variables... Too bad that function pointer does not provide a 'user data' like lvgl to avoid these statics or this ::Instance() method!

src/drivers/Cst816s.cpp
joaquimorg: I think it must not be related to touch. Debug time...

JF002: Right, 2048 does not enable the continuous mode, so that's probably another bug !

src/components/battery/BatteryController.cpp
joaquimorg: yes the sdk functions have this problem.
but I still rewrote it and now it's better.

JF002:
> Why is the transition from clock to the quick settings menu on the left not as smooth as moving down to application list? Is is because the display writes pixels from top to bottom? And is that something that we can solve?

The vertical transition is assisted by the hardware : you write 4 lines, set an offset and the display shifts the whole frame buffer up or down. 
The lcd controller only supports vertical scrolling, not horizontal scrolling. Joaq had to implement in 100% software. I also think LVGL is design to refresh the display by lines and not by columns. It's true that the animation looks slower when going from the quick actions menu to the analog watchface.

JF002:
> > Why is the transition from clock to the quick settings menu on the left not as smooth as moving down to application list? Is is because the display writes pixels from top to bottom? And is that something that we can solve?
> 
> The vertical transition is assisted by the hardware : you write 4 lines, set an offset and the display shifts the whole frame buffer up or down.
> The lcd controller only supports vertical scrolling, not horizontal scrolling. Joaq had to implement in 100% software. I also think LVGL is design to refresh the display by lines and not by columns. It's true that the animation looks slower when going from the quick actions menu to the analog watchface.

> Do you mean something other than the memory info added to the About section in settings?

I'll have a look at it, but I think it only take the memory buffer reserved to LVGL into account, not the one from freertos and the global one. As long as there buffer do not overlap or overflow, we're good ;)

JF002:
While testing the TWI driver, I open/close a lot of apps and menu. I noticed this in the log:

```
<info> app: ---------------------------------------
Free heap : 576
<info> app: Task [IDL] - 43
<info> app: Task [MAI] - 113
<info> app: Task [dis] - 1
<info> app: WARNING!!! Task dis task is nearly full, only 4B available
<info> app: Task [Hea] - 447
<info> app: Task [Tmr] - 141
<info> app: Task [LOG] - 94
<info> app: Task [ll] - 470
<info> app: Task [ble] - 411
<info> app: MOTION : 0 - -331
```

From my observations, the available space in the display task is shrinking every time an app is opened, but it does not grow when it's closed. Are we leaking memory from the display task?

JF002:
Regarding the memory issue I raised just above : The monitor displays the "water mark" from the stack of each task : the number is the minimum number of bytes available since the beginning of the execution of the task. In this case, there was only 4B available in display task. 

It means that it's not a leak, but that we should increase the size of the stack of the display task.

Issue #262, opened 2021-04-09
Version 0.16.0 testing


joaquimorg:
I have notice some problems with the touch when is in Polling mode.
Testing InifiPaint and Pong they freeze making the watchdog go into action.

JF002:
I've also noticed some glitches around the handling of the TWI driver (probably sleep/wakeup) since the merge of your BigRewrite PR and my motion sensor branch.  I'll try to debug them this week-end :)

JF002:
For anyone wondering where the 0.16.0 version is : it's not released yet ! It's the version of the *develop* branch and, as you can see, it's not stable enough to be released yet ;)

joaquimorg:
I think we have to go back to the original TWI, I reverted the code and there are no more blocks in the apps.
bma421 starts well and does not block on restart.

joaquimorg:
I created a PR #263 with some corrections, for those I already tested I do not have blocks of the apps, the BMA is working.

PR #263, opened 2021-04-10
Fix Issue #262 for 0.16.0 version

joaquimorg:
0.16.0 Fixes...
TWI problems fix, by reverting to old code, it works better, I think it's because of the fix to not let the bus block.
More memory for freertos heap and timer stack, I have random reboots, with the increase they stopped happening.
Fix warning in watchface.
Fix number of bytes read by cst816, 63 bytes were being read when in reality the first 10 are needed.
Debug app to show freertos tasks, this can then be removed, just to see the free memory in each of the tasks.
Increased the number of bytes of the twi write buffer,to write bma421 configuration faster.


JF002:
Wow you were faster than me to analyze those issues!

Originally, I switched to the NRFX driver because the driver from InfiniTime would sometimes freeze, especially when the display and the touch panel were under heavy use (in games). Using the NRFX driver seemed to improve that situation.

Have you already tested this branch in games?

JF002:
I played a longer 2048 for a longer time with no issue:


Regarding the Tasks app, maybe we can add it in the SystemInfo (About) app?

joaquimorg:
Yes good idea. In the afternoon I'll change the task to about.

joaquimorg:
Tasks information added to About app.

JF002:
Thanks @joaquimorg ! I'll try to test and merge it ASAP, but I have some issues with my computer and I couldn't find the time to work on pinetime these last few days. Sorry for the delay !

petterhs:
Sent this PR to your branch to fix the style of the 2048 game: https://github.com/joaquimorg/PinetimeLite/pull/1

Picture in the PR

joaquimorg:
> Sent this PR to your branch to fix the style of the 2048 game: [joaquimorg#1](https://github.com/joaquimorg/PinetimeLite/pull/1)
> 
> Picture in the PR

Nice.

JF002:
Awesome, thanks @joaquimorg  @petterhs !

Regarding these comments:
```
      // call the battery controller or use the MSG in DisplayApp to get the battery status ???
      // it is necessary to validate which is the most efficient
      batteryController.Update();
      //displayApp->PushMessage(Pinetime::Applications::Display::Messages::UpdateBatteryLevel);
      // analyze a more efficient way to do this refreshment
      // this and the UpdateMotion(); can be called on a timer to be independent of the main process ???
```

When I wrote this code, the idea was to do the analog acquisition in SystemTask, and only the display in DisplayTask. This would allow the UI to not have to worry about timing, choosing when it's the good time to read the voltage,... 
On the other hand, SystemTask has the responsability of doing the acquisition when needed (for example : only when it's awake, but not when the cpu is under heavy usage,...). Of course, all of this intelligence is not implemented yet :)

It's the same for the motion acquisiton, it could be improved by refreshing the motion only when the sensor actually detects motion (it can be configure to trig an IRQ when it detects 'any motion' and 'no motion'. This is an improvement we'll do in the future ;)

PR #264, opened 2021-04-11
Add new watchface PineTimeStyle

kieranc:


kieranc:
This watchface is a copy of [TimeStyle](https://github.com/freakified/TimeStylePebble) for the Pebble smartwatch, created with permission from the original author. It looks like this:
image

PR #266, opened 2021-04-14
Add Python modules to build instructions

Caton101:
I had issues building InfiniTime for the first time because the instructions never mentioned needing these Python modules. Including them in the build documentation will help people not be confused like I was.

I recommend adding all needed modules to this list. I only added the ones I knew I was missing.


pfeerick (commented):
These are actually a requirement of mcuboot, not InfiniTime itself. I'm not saying not to add them, but it would probably be a good idea to list all the packages from it's [requirements.txt](https://github.com/mcu-tools/mcuboot/blob/master/scripts/requirements.txt)...

JF002:
Python scripts are used to build the recovery firmware (pinetime-recovery, pinetime-recovery-loader, pinetime-mcuboot-recovery and pinetime-mcuboot-recovery-loader) and to build the DFU files (enabled with -DBUILD_DFU).

I think `cbor` and `intelhex` are needed by `bin2c.py` to convert the binary file of the reocvery firmware into a C array for the recovery-loader. As not everyone will build the recovery firmware, it would be a nice improvement to add a CMake switch to disable the generation of the recovery firmware by default, and enable it only on request using a command line variable (-D...).

There are probably other modules that are necessary when you enable -DBUILD_DFU (`adafruit-nrfutil` for example).

PR #275, opened 2021-04-17
Changed text in Navigation

panos:


JF002:
Could you explain why  this change is needed?

panos:
It's not _needed_, it was just a hunch that I made because I felt that "Navigation" was clear enough and it seemed more consistent and laconic (in the spirit of "KISS"), since the icon seemed slightly unclear to me.

JF002:
I discussed about this PR with @panos  and I think he's right : there's no need to welcome the user in applications (and we don't do it in other apps).

PR #279, opened 2021-04-17
Overwrite labels if their corresponding metadata is empty

panos:
Fixing some edge cases, see: https://github.com/JF002/InfiniTime/pull/278#issuecomment-821789432

### DIFF
--- a/src/displayapp/screens/Music.cpp
+++ b/src/displayapp/screens/Music.cpp
@@ -138,25 +138,29 @@ Music::~Music() {
   lv_obj_clean(lv_scr_act());
 }
 
+// If some of the metadata of a specific track happens to be empty, it should be
+// overwritten, instead of retaining and displaying the older metadata. The artist
+// or the track may not change and get reported again, while some other metadata
+// may not get reported to InfiniTime.
 bool Music::Refresh() {
-  if (artist != musicService.getArtist()) {
+  if (artist != musicService.getArtist() or musicService.getArtist() = "") {
     artist = musicService.getArtist();
     currentLength = 0;
     lv_label_set_text(txtArtist, artist.data());
   }
   
-  if (track != musicService.getTrack()) {
+  if (track != musicService.getTrack() or musicService.getTrack() = "") {
     track = musicService.getTrack();
     currentLength = 0;
     lv_label_set_text(txtTrack, track.data());
   }
   
-  if (album != musicService.getAlbum()) {
+  if (album != musicService.getAlbum() or musicService.getAlbum() == "") {
     album = musicService.getAlbum();
     currentLength = 0;
   }
   
-  if (playing != musicService.isPlaying()) {
+  if (playing != musicService.isPlaying() or musicService.isPlaying() == False) {
     playing = musicService.isPlaying();
   }
   


Avamander:
This is related to https://github.com/JF002/InfiniTime/pull/278, and as mentioned there, if the track changes, the companion app can push a full update and this PR and resulting unnecessary screen updates aren't necessary.

src/displayapp/screens/Music.cpp
```diff
@@ -138,25 +138,29 @@ Music::~Music() {
   lv_obj_clean(lv_scr_act());
 }
 
+// If some of the metadata of a specific track happens to be empty, it should be
+// overwritten, instead of retaining and displaying the older metadata. The artist
+// or the track may not change and get reported again, while some other metadata
+// may not get reported to InfiniTime.
 bool Music::Refresh() {
-  if (artist != musicService.getArtist()) {
+  if (artist != musicService.getArtist() or musicService.getArtist() = "") {
```
JF002, line 146: I guess you meant `or musicService.getArtist() == ""` ?
Or better : `or musicService.getArtist().empty()`

src/displayapp/screens/Music.cpp
```diff
@@ -138,25 +138,29 @@ Music::~Music() {
...
   }
   
-  if (album != musicService.getAlbum()) {
+  if (album != musicService.getAlbum() or musicService.getAlbum() = "") {
     album = musicService.getAlbum();
     currentLength = 0;
   }
   
-  if (playing != musicService.isPlaying()) {
+  if (playing != musicService.isPlaying() or musicService.isPlaying() = "") {
```
JF002, line 163: IsPlaying is a `bool`, it cannot be empty.

JF002:
I've seen that there was a debate in #278. Did @panos and @Avamander find a consensus on how to handle missing info from companion apps? Or should we write recommendations on how the companion app should send data to the InfiniTime to allow a consistent display?

src/displayapp/screens/Music.cpp
panos: oh my god sorry

panos:
> I've seen that there was a debate in #278. Did @panos and @Avamander find a consensus on how to handle missing info from companion apps? Or should we write recommendations on how the companion app should send data to the InfiniTime to allow a consistent display?

I'm still concerned about the fact that even with @Avamander's approach, which seems pretty good, the initiated text will cause problems.

Also, is it a good idea to keep doing what I'm trying to do (resource-wasting) by default, and then have the companion apps send sets of data with a flag that will tell the watch not to wipe and retain (or to wipe) the metadata that's already there and allowing the companion app to decide? I could think of a few case scenarios where this would be useful.

panos:
A few scenarios could be:
- Everytime the app in use changes, the companion app asks the watch to wipe all metadata. That would also resolve the problem regarding the default text.
- The flag ('wipe everything') could be sent if the notification disappears, and the UI could work with that and gray out too instead of the watch assuming that a song is still being played indefinitely, until it gets overwritten, and then bring the "Waiting for track information..." text back. That's something that really annoys me-- I am always trying to play a song from the watch, but then realize that Android has OOM'd the music player.

I could think of a few more, but these are the primary issues that have bothered me, as a user, personally and others.

src/displayapp/screens/Music.cpp
panos: Makes sense.

JF002:
I agree, there are a lot of edge cases we can think of, depending on how the companion app sends the data.
So, I'm wondering if this PR should be merged (one syntax errors will be fixed), or it if fixes a few edge cases but creates new ones?

panos:
> I agree, there are a lot of edge cases we can think of, depending on how the companion app sends the data.
> So, I'm wondering if this PR should be merged (one syntax errors will be fixed), or it if fixes a few edge cases but creates new ones?

I'm more concerned about whether I overlooked any edge cases, and some external user feedback may be required. Considering that, by the look of how things currently are, companion apps only send whatever they have available and do not conditionally remove or add any more data, I don't see how this could go wrong any further, unless if my excessively Python-adjusted self managed to forget another `=` again,

Avamander:
> I'm wondering if this PR should be merged (one syntax errors will be fixed)

I'm not sure what the PR will really achieve to be honest, the checks are for determining discrepancies between display state and service state. The question being, in what case is the `display state not different and blank and needs to be updated`?

Avamander:
> how to handle missing info from companion apps? Or should we write recommendations on how the companion app should send data to the InfiniTime to allow a consistent display?

Well, due to the fact that we try to do partial updates to reduce lag and battery usage, the fix would be to determine legitimate cases where some metadata is empty and then explicitly still send that. The recommendation is really "update what needs to be updated" :þ 

Trying to detect this on the watch's side (as the other PR did) breaks current companion apps' partial updates in the cases where the same artist's songs are seeked, artist remains the same but title doesn't - but artist gets blanked.

Currently I think this can be fixed on GB's side, it filters out empty updates (because some Intents sent do not contain all the information each and every time) but I think it's safe to assume on GB's side that `if title is not empty and artist is empty` then `artist is actually empty and it can be sent`. It'll fix this inconsistency. 

Other bugs should probably discussed separately, a music app dying is something that would have to be detected on the companion side as well. The watch doesn't have the visibility required to provide good UX there.

panos:
I'm assuming that the returned value of `musicService.getTrack()` will be empty, if e.g. the companion app doesn't send anything. That should prevent most visual glitches like the ones I (and others) have described.I'm trying to take advantage of that and remove the name of the previous label of the track by setting the corresponding labels with the empty value. Have I gotten anything wrong?

Avamander:
> I'm assuming that the returned value of musicService.getTrack() will be empty, if e.g. the companion app doesn't send anything.

Yeah but then the first part in the if statemet `display state is different from service content` evaluates to true though. Does it not?

panos:
> Trying to detect this on the watch's side (as the other PR did) breaks current companion apps' partial updates in the cases where the same artist's songs are seeked, artist remains the same but title doesn't - but artist gets blanked.

> Well, due to the fact that we try to do partial updates to reduce lag and battery usage, the fix would be to determine legitimate cases where some metadata is empty and then explicitly still send that. The recommendation is really "update what needs to be updated" :þ

I wasn't aware that partial updates were already a thing, oops. Either way, I brought up an idea that could probably fix both the problems that I brought up and not break more stuff.
 
> Currently I think this can be fixed on GB's side, it filters out empty updates (because some Intents sent do not contain all the information each and every time) but I think it's safe to assume on GB's side that `if title is not empty and artist is empty` then `artist is actually empty and it can be sent`. It'll fix this inconsistency.

It won't. A track may miss metadata, so the watch can falsely attribute a different artist as the creator of a song.

> Other bugs should probably discussed separately, a music app dying is something that would have to be detected on the companion side as well. The watch doesn't have the visibility required to provide good UX there.

Agreed, I tried to fix the problem for now while I was under the impression that partial updates were something to be added later on.

panos:
> Yeah but then the first part in the if statemet `display state is different from service content` evaluates to true though. Does it not?

Not necessarily, I believe. See the bug that I shared regarding the initialized text staying in place if a track doesn't have a specific piece of metadata.

Avamander:
> It won't. A track may miss metadata, so the watch can falsely attribute a different artist as the creator of a song.

In which case after that fix? False attribution only happens when the companion phone switches from a song with known artist into a one without an artist. After the proposed fix: In that case the title updates, the artist is blank. If the title updates and the artist is blank, force update both. No more false attribution.

Avamander:
> Not necessarily, I believe. See the bug that I shared regarding the initialized text staying in place if a track doesn't have a specific piece of metadata.

Yeah, but that's only when the string in the service itself doesn't get updated, if that gets updated to a blank one, the display updates.

panos:
> > It won't. A track may miss metadata, so the watch can falsely attribute a different artist as the creator of a song.
> 
> In which case after that fix? False attribution only happens when the companion phone switches from a song with known artist into a one without an artist. After the proposed fix: In that case the title updates, the artist is blank. If the title updates and the artist is blank, force update both. No more false attribution.

Force update both... that could work.

panos:
Sorry, I've been programming a lot this past day and I'm a bit exhausted, is this a matter of adding one more additional condition?

Avamander:
> Sorry, I've been programming a lot this past day and I'm a bit exhausted, is this a matter of adding one more additional condition?

Fortunately or unfortunately, it's only a matter of keeping the state in the music service accurate and that can't be done on the watch.

panos:
I see. Huge thanks for your patience and the useful feedback!

PR #280, opened 2021-04-17
bin2c: Don't assume that the default Python version is higher than 2

panos:


panos:
Something went clearly wrong, closing and pushing again.

PR #281, opened 2021-04-17
bin2c: Don't assume that the default Python version is higher than 2

panos:
Continuation of the failed #280


Avamander:
Dealing with EOL Python2 is relatively worthless though. If you still have py2 as the default, well, update.

JF002:
Thanks :)

PR #290, opened 2021-04-18
Multiple wakeup sources

### LINKED ISSUE #277: Feature Request: Allow waking with a double tap, even when Wake Up is set to "Raise Wrist"
I get that it doesn't make perfect sense, but the accelerometer is not infallible. That can make using InfiniTime much harder.

I think that it'd make sense to enable waking by double tapping, even in this mode, instead of only allowing people to choose either "Raise Wrist" or "Double Tap". After all, it's considerably harder to accidentally wake the watch in comparison to "Single Tap" and many people familiar with smartwatches are already conditioned to that feature by default, so I'm not sure why the user shouldn't get to enjoy the best of both worlds.

Kozova1:
This PR should add the ability to set multiple wakeup sources.

Warning: I do not have a devkit, only a sealed one. I am not even sure that this compiles, so beware.
Also, does anyone know if there's a better way to know whether a checkbox is checked?

This would (try to) fix #277

### DIFF
--- a/src/components/settings/Settings.h
+++ b/src/components/settings/Settings.h
@@ -1,5 +1,6 @@
 #pragma once
 #include <cstdint>
+#include <bitset>
 #include "components/datetime/DateTimeController.h"
 #include "components/brightness/BrightnessController.h"
 #include "components/fs/FS.h"
@@ -8,69 +9,111 @@
 namespace Pinetime {
   namespace Controllers {
     class Settings {
-      public:
-        enum class ClockType {H24, H12};
-        enum class Vibration {ON, OFF};
-        enum class WakeUpMode : uint8_t {
-            None = 1,
-            SingleTap = 2,
-            DoubleTap = 4,
-            RaiseWrist = 8
-        };
+    public:
+      enum class ClockType : uint8_t { H24, H12 };
+      enum class Vibration : uint8_t { ON, OFF };
+      enum class WakeUpMode : uint8_t {
+        SingleTap = 0,
+        DoubleTap = 1,
+        RaiseWrist = 2,
+      };
 
-        Settings( Pinetime::Drivers::SpiNorFlash &spiNorFlash );
+      Settings(Pinetime::Controllers::FS& fs);
 
-        void Init();
-        void SaveSettings();
+      void Init();
+      void SaveSettings();
 
-        void SetClockFace( uint8_t face ) {
-          if ( face != settings.clockFace ) settingsChanged = true;
-          settings.clockFace = face; 
-        };
-        uint8_t GetClockFace() const { return settings.clockFace; };
+      void SetClockFace(uint8_t face) {
+        if (face != settings.clockFace) {
+          settingsChanged = true;
+        }
+        settings.clockFace = face;
+      };
+      uint8_t GetClockFace() const {
+        return settings.clockFace;
+      };
 
-        void SetAppMenu( uint8_t menu ) { appMenu = menu; };
-        uint8_t GetAppMenu() { return appMenu; };
+      void SetAppMenu(uint8_t menu) {
+        appMenu = menu;
+      };
+      uint8_t GetAppMenu() {
+        return appMenu;
+      };
 
-        void SetSettingsMenu( uint8_t menu ) { settingsMenu = menu; };
-        uint8_t GetSettingsMenu() const { return settingsMenu; };
+      void SetSettingsMenu(uint8_t menu) {
+        settingsMenu = menu;
+      };
+      uint8_t GetSettingsMenu() const {
+        return settingsMenu;
+      };
 
-        void SetClockType( ClockType clocktype ) { 
-          if ( clocktype != settings.clockType ) settingsChanged = true;
-          settings.clockType = clocktype; 
-        };
-        ClockType GetClockType() const { return settings.clockType; };
+      void SetClockType(ClockType clocktype) {
+        if (clocktype != settings.clockType) {
+          settingsChanged = true;
+        }
+        settings.clockType = clocktype;
+      };
+      ClockType GetClockType() const {
+        return settings.clockType;
+      };
 
-        void SetVibrationStatus( Vibration status ) { 
-          if ( status != settings.vibrationStatus ) settingsChanged = true;
-          settings.vibrationStatus = status; 
-        };
-        Vibration GetVibrationStatus() const { return settings.vibrationStatus; };
+      void SetVibrationStatus(Vibration status) {
+        if (status != settings.vibrationStatus) {
+          settingsChanged = true;
+        }
+        settings.vibrationStatus = status;
+      };
+      Vibration GetVibrationStatus() const {
+        return settings.vibrationStatus;
+      };
 
-        void SetScreenTimeOut( uint32_t timeout ) { 
-          if ( timeout != settings.screenTimeOut ) settingsChanged = true;
-          settings.screenTimeOut = timeout; 
-        };
-        uint32_t GetScreenTimeOut() const { return settings.screenTimeOut; };        
+      void SetScreenTimeOut(uint32_t timeout) {
+        if (timeout != settings.screenTimeOut) {
+          settingsChanged = true;
+        }
+        settings.screenTimeOut = timeout;
+      };
+      uint32_t GetScreenTimeOut() const {
+        return settings.screenTimeOut;
+      };
 
-        void setWakeUpMode( WakeUpMode wakeUp ) { 
-          if ( !isWakeUpModeOn(wakeUp) ) settingsChanged = true;
-          settings.wakeUpMode |= static_cast<uint8_t>(wakeUp); 
-        };
-        void unsetWakeUpMode( WakeUpMode wakeUp ) {
-            if ( isWakeUpModeOn(wakeUp) ) settingsChanged = true;
-            settings.wakeUpMode &= ~(static_cast<uint8_t>(wakeUp));
-        };
-        uint8_t getWakeUpModes() const { return settings.wakeUpMode; };
-        bool isWakeUpModeOn( const WakeUpMode mode ) const {
-            return (getWakeUpModes() & static_cast<uint8_t>(mode)) != 0;
-        };
+      void setWakeUpMode(WakeUpMode wakeUp, bool enabled) {
+        if (!isWakeUpModeOn(wakeUp)) {
+          settingsChanged = true;
+        }
+        settings.wakeUpMode.set(static_cast<size_t>(wakeUp), enabled);
+        // Handle special behavior
+        if (enabled) {
+          switch (wakeUp) {
+            case WakeUpMode::SingleTap:
+              settings.wakeUpMode.set(static_cast<size_t>(WakeUpMode::DoubleTap), false);
+              break;
+            case WakeUpMode::DoubleTap:
+              settings.wakeUpMode.set(static_cast<size_t>(WakeUpMode::SingleTap), false);
+              break;
+            case WakeUpMode::RaiseWrist:
+              break;
+          }
+        }
+      };
 
-        void SetBrightness( Controllers::BrightnessController::Levels level ) { 
-          if ( level != settings.brightLevel ) settingsChanged = true;
-          settings.brightLevel = level; 
-        };
-        Controllers::BrightnessController::Levels GetBrightness() const { return settings.brightLevel; };
+      std::bitset<3> getWakeUpModes() const {
+        return settings.wakeUpMode;
+      }
+
+      bool isWakeUpModeOn(const WakeUpMode mode) const {
+        return getWakeUpModes()[static_cast<size_t>(mode)];
+      }
+
+      void SetBrightness(Controllers::BrightnessController::Levels level) {
+        if (level != settings.brightLevel) {
+          settingsChanged = true;
+        }
+        settings.brightLevel = level;
+      };
+      Controllers::BrightnessController::Levels GetBrightness() const {
+        return settings.brightLevel;
+      };
 
       void SetStepsGoal( uint32_t goal ) { 
         if ( goal != settings.stepsGoal ) {
@@ -94,42 +137,21 @@ namespace Pinetime {
         ClockType clockType = ClockType::H24;
         Vibration vibrationStatus = Vibration::ON;
 
-          ClockType clockType = ClockType::H24;
-          Vibration vibrationStatus = Vibration::ON;
+        uint8_t clockFace = 0;
 
-          uint8_t clockFace = 0;
+        std::bitset<3> wakeUpMode {0};
 
-          uint32_t stepsGoal = 1000;
-          uint32_t screenTimeOut = 15000;
+        Controllers::BrightnessController::Levels brightLevel = Controllers::BrightnessController::Levels::Medium;
+      };
 
-          uint8_t wakeUpMode = static_cast<uint8_t>(WakeUpMode::None);
+      SettingsData settings;
+      bool settingsChanged = false;
 
-          Controllers::BrightnessController::Levels brightLevel = Controllers::BrightnessController::Levels::Medium;
-
-        };
-
-        SettingsData settings;
-        bool settingsChanged = false;
-
-        uint8_t appMenu = 0;
-        uint8_t settingsMenu = 0;
+      uint8_t appMenu = 0;
+      uint8_t settingsMenu = 0;
 
       void LoadSettingsFromFile();
       void SaveSettingsToFile();
-        // There are 10 blocks of reserved flash to save settings
-        // to minimize wear, the recording is done in a rotating way by the 10 blocks
-        uint8_t settingsFlashBlock = 99; // default to indicate it needs to find the active block
-
-        static constexpr uint32_t settingsBaseAddr =  0x3F6000; // Flash Settings Location
-        static constexpr uint16_t settingsVersion =  0x0100; // Flash Settings Version
-
-        bool FindHeader();
-        void ReadSettingsData();
-        void EraseBlock();
-        void SetHeader( bool state );
-        void SaveSettingsData();
-        void LoadSettingsFromFlash();
-        void SaveSettingsToFlash();
     };
   }
 }
--- a/src/components/settings/Settings.h
+++ b/src/components/settings/Settings.h
@@ -8,89 +8,69 @@
 namespace Pinetime {
   namespace Controllers {
     class Settings {
-    public:
-      enum class ClockType : uint8_t { H24, H12 };
-      enum class Vibration : uint8_t { ON, OFF };
-      enum class WakeUpMode : uint8_t { None, SingleTap, DoubleTap, RaiseWrist };
+      public:
+        enum class ClockType {H24, H12};
+        enum class Vibration {ON, OFF};
+        enum class WakeUpMode : uint8_t {
+            None = 1,
+            SingleTap = 2,
+            DoubleTap = 4,
+            RaiseWrist = 8
+        };
 
-      Settings(Pinetime::Controllers::FS& fs);
+        Settings( Pinetime::Drivers::SpiNorFlash &spiNorFlash );
 
-      void Init();
-      void SaveSettings();
+        void Init();
+        void SaveSettings();
 
-      void SetClockFace(uint8_t face) {
-        if (face != settings.clockFace) {
-          settingsChanged = true;
-        }
-        settings.clockFace = face;
-      };
-      uint8_t GetClockFace() const {
-        return settings.clockFace;
-      };
+        void SetClockFace( uint8_t face ) {
+          if ( face != settings.clockFace ) settingsChanged = true;
+          settings.clockFace = face; 
+        };
+        uint8_t GetClockFace() const { return settings.clockFace; };
 
-      void SetAppMenu(uint8_t menu) {
-        appMenu = menu;
-      };
-      uint8_t GetAppMenu() {
-        return appMenu;
-      };
+        void SetAppMenu( uint8_t menu ) { appMenu = menu; };
+        uint8_t GetAppMenu() { return appMenu; };
 
-      void SetSettingsMenu(uint8_t menu) {
-        settingsMenu = menu;
-      };
-      uint8_t GetSettingsMenu() const {
-        return settingsMenu;
-      };
+        void SetSettingsMenu( uint8_t menu ) { settingsMenu = menu; };
+        uint8_t GetSettingsMenu() const { return settingsMenu; };
 
-      void SetClockType(ClockType clocktype) {
-        if (clocktype != settings.clockType) {
-          settingsChanged = true;
-        }
-        settings.clockType = clocktype;
-      };
-      ClockType GetClockType() const {
-        return settings.clockType;
-      };
+        void SetClockType( ClockType clocktype ) { 
+          if ( clocktype != settings.clockType ) settingsChanged = true;
+          settings.clockType = clocktype; 
+        };
+        ClockType GetClockType() const { return settings.clockType; };
 
-      void SetVibrationStatus(Vibration status) {
-        if (status != settings.vibrationStatus) {
-          settingsChanged = true;
-        }
-        settings.vibrationStatus = status;
-      };
-      Vibration GetVibrationStatus() const {
-        return settings.vibrationStatus;
-      };
+        void SetVibrationStatus( Vibration status ) { 
+          if ( status != settings.vibrationStatus ) settingsChanged = true;
+          settings.vibrationStatus = status; 
+        };
+        Vibration GetVibrationStatus() const { return settings.vibrationStatus; };
 
-      void SetScreenTimeOut(uint32_t timeout) {
-        if (timeout != settings.screenTimeOut) {
-          settingsChanged = true;
-        }
-        settings.screenTimeOut = timeout;
-      };
-      uint32_t GetScreenTimeOut() const {
-        return settings.screenTimeOut;
-      };
+        void SetScreenTimeOut( uint32_t timeout ) { 
+          if ( timeout != settings.screenTimeOut ) settingsChanged = true;
+          settings.screenTimeOut = timeout; 
+        };
+        uint32_t GetScreenTimeOut() const { return settings.screenTimeOut; };        
 
-      void setWakeUpMode(WakeUpMode wakeUp) {
-        if (wakeUp != settings.wakeUpMode) {
-          settingsChanged = true;
-        }
-        settings.wakeUpMode = wakeUp;
-      };
-      WakeUpMode getWakeUpMode() const {
-        return settings.wakeUpMode;
-      };
+        void setWakeUpMode( WakeUpMode wakeUp ) { 
+          if ( !isWakeUpModeOn(wakeUp) ) settingsChanged = true;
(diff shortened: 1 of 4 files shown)


Caton101:
I compiled your branch and flashed it without any issues. I can offer some feedback after some testing:

- I can check multiple wake sources, but I can't uncheck them
- `none` being checked when a different source is checked
- `Single Tap` and `Double Tap` can both be checked at the same time

I also noticed some strange issues when trying to wake it with every source checked. Sometimes it wakes on one tap and other times on two taps. There have also been times where it would only wake by turning my wrist or pressing the button.

I think most of these are related so there are only two things left to fix:

- allow users to uncheck wake sources
- fix the logic that determines if the watch should wake up

_This is a very useful PR and all my feedback is meant to be constructive and offer improvements._

Kozova1:
Glad to hear it somewhat worked.

I don't really have experience with lvgl so ideally someone else can help me fix issues 1 & 2.  
As for the 3rd - I can see how that would cause a problem, but I don't think it's possible using the current (proposed) implementation to separate them without a lot of ugly code.

Caton101:
I see it as another restraint like having None selected when something else is selected. So there could be two statements: one unchecks None if anything else is selected and the other will ensure only Single Tap or Double Tap can be selected at a time. I assume this should also fix the random behavior where taps stop working correctly.

Kozova1:
@Caton101, any chance you can test the new commit? Hopefully, it should fix the issues you showed above.

Caton101:
I will test it later today and reply with any issues I find.

Caton101:
The default config after flashing had both `Single Tap` and `Double Tap` checked. It was working really well until I checked `None` and it checked all of them. When I try to uncheck `None`, it just checks all of them. I might have more time today to look at the logic and suggest some changes.

Caton101:
I'll try to help you with the logic at some point this week. There still appears to be issues.

- I noticed that when `None` checks all the wake up options, this is purely visual and refreshing the setting page will fix it. The solution might be to force a refresh on the settings page.
- Unchecking all the wake options will revert to the old settings instead of `None`.
- Unless you use `None` to reset them, you can not uncheck `Raise Wrist` or `-Tap`.
- `-Tap` does not work when `Raise Wrist` is checked

Kozova1:
Great, is there any way to talk with you using discord / irc / etc? That would be a bit faster than having the entire conversation over github pr comments.

Caton101:
My Discord is Caton101#3699. I'm also `@Caton101` in the Pine64 Discord server.

Caton101:
If you are in the Pine64 discord server, just message me using `@Caton101`

src/displayapp/screens/settings/SettingWakeUp.cpp
```diff
@@ -90,18 +90,48 @@ bool SettingWakeUp::Refresh() {
 
 void SettingWakeUp::UpdateSelected(lv_obj_t *object, lv_event_t event) {
   if(event == LV_EVENT_VALUE_CHANGED) {
```
petterhs, line 92: Tested out your branch, and had the same problems as @Caton101. Tried to find the cause and I think at least one problem is using `LV_EVENT_VALUE_CHANGED`. Since this causes the `UpdateSelected` function to run when a value is changed, not just when it is clicked but also when changed with `lv_checkbox_set_checked`. So you get a loop of `UpdateSelected` functions when `lv_checkbox_set_checked` inside `UpdateSelected`. I'm not 100% sure but thats how i understand lvgl.

You could could use the `LV_EVENT_PRESSED` event instead but then you need to handle the conflict with lvgl's "hidden" checking/unchecking of the box. If not, then lvgl will try to check/uncheck and at the same time you check/uncheck inside `Updateselected`

Kozova1: > handle the conflict with lvgl's "hidden" checking/unchecking of the box.  
Could you elaborate a bit about this? I'm not sure that I got what you mean.

petterhs: Ok I will try to explain it better.

We have our `event_handler`:
```
  static void event_handler(lv_obj_t * obj, lv_event_t event) {
    SettingWakeUp* screen = static_cast<SettingWakeUp *>(obj->user_data);
    screen->UpdateSelected(obj, event);
  }
```
which runs when a lvgl event occurs. Like when a button is clicked, dragged etc. More about events: [lvgl events](https://docs.lvgl.io/v7/en/html/overview/event.html#special-events). In our event_handler we call the function `screen->UpdateSelected(obj, event);` and then handles what to do with the event in this memberfunction of the screen.

`UpdateSelected` starts with this if statement: 
```  
void SettingWakeUp::UpdateSelected(lv_obj_t *object, lv_event_t event) {
    if(event == LV_EVENT_VALUE_CHANGED) {
```
```LV_EVENT_VALUE_CHANGED``` is a special event that is sent when the checkbox is checked/unchecked. [special event](https://docs.lvgl.io/v7/en/html/overview/event.html#special-events).

So when you click a button:

1. LVGL internally checks/unchecks the box and sends a LV_EVENT_VALUE_CHANGED event.
2. This event gets picked up by event_handler and passed to UpdateSelected
3. In UpdateSelected you change the values of this and other checkboxes multiple times with `lv_checkbox_set_checked`
4. If I'm correct each of your calls to lv_checkbox_set_checked sends out new LV_EVENT_VALUE_CHANGED events which in turn gets picked up by the event_handler and you get a loop.

Comment (Kozova1) [reply src/displayapp/screens/settings/SettingWakeUp.cpp:92]: I see. Would introducing a "lock" variable, that would make UpdateSelected ignore events while it's set, and setting it when starting to handle a checkbox work?

Comment (petterhs) [reply src/displayapp/screens/settings/SettingWakeUp.cpp:92]: Yh that could work. Another solution could be to set the state of the checkboxes to disabled so lvgl dont do the checking/unchecking itself. And then instead use `LV_EVENT_CLICKED` and handle all the checking/unchecking in UpdateSelected

**Comment by petterhs** [comment]
I think your isLocked should work, but I found some other problems now :) 

Since lvgl changes the checked status internally before it sends the LV_EVENT_VALUE_CHANGED, the if statements should be the other way around.

`auto currentCheckboxState = lv_checkbox_is_checked(currentCheckbox);` in this case return the currentCheckboxState after it is changed not before it is clicked.

Changed it to ``auto currentCheckboxState = !lv_checkbox_is_checked(currentCheckbox);` and it worked but is not the cleanest solution. I suggest going through your logic once more and only handle the cases where you have to intervene with the automatic lvgl checking/unchecking.

Example:
- If WakeOnWrist is pressed don't set the value of this button in UpdateSelected let lvgl take care of it only set the None button in if everything else is unchecked

File: src/displayapp/screens/settings/SettingWakeUp.cpp [hunk 2]
```diff
@@ -89,19 +89,51 @@ bool SettingWakeUp::Refresh() {
 
 
 void SettingWakeUp::UpdateSelected(lv_obj_t *object, lv_event_t event) {
-  if(event == LV_EVENT_VALUE_CHANGED) {
+  if(event == LV_EVENT_VALUE_CHANGED && !isLocked) {
+    isLocked = true;
+    bool anyCheckboxChecked = false;
```
petterhs, line 94: Setting anyCheckboxChecked to false here does not make much sense since you will have to check the status of other buttons.

Result now:
- Click single tap
- Click wake on wrist
- click single tap again
- This removes wake on wrist also and sets it to None

It makes more sense to have it as:
`bool anyCheckboxChecked = !lv_checkbox_is_checked(cbOption[0]);`

petterhs (changes requested):

src/displayapp/screens/settings/SettingWakeUp.cpp
Kozova1: Can you try the new commit?

petterhs: Works :)

Kozova1: Great!

petterhs:
My bad that last change did add some additional edge cases. And even made it possible to toggle the None button. I suggest dropping the anyCheckboxChecked statements inside each case and just do a check if the boxes are checked at the end of the UpdateSelected instead.

petterhs:
Now the checkboxes work visually as intented :+1: It is still problems with either the setting of Wakeupmodes or when checking which modes are active though. `None` and `RaiseWrist` works as intended but the others do not. I'm not familiar with the fancy operators you use in setWakeupMode() so you will need to describe your  thought process there. 

Aswell as `return (getWakeUpModes() & static_cast<uint8_t>(mode)) != 0;` in isWakeupModeOn();

src/displayapp/screens/settings/SettingWakeUp.cpp
```diff
@@ -89,19 +89,50 @@ bool SettingWakeUp::Refresh() {
...
-        
-        if ( i == 0 ) { settingsController.setWakeUpMode(Pinetime::Controllers::Settings::WakeUpMode::None); };
-        if ( i == 1 ) { settingsController.setWakeUpMode(Pinetime::Controllers::Settings::WakeUpMode::SingleTap); };
-        if ( i == 2 ) { settingsController.setWakeUpMode(Pinetime::Controllers::Settings::WakeUpMode::DoubleTap); };
-        if ( i == 3 ) { settingsController.setWakeUpMode(Pinetime::Controllers::Settings::WakeUpMode::RaiseWrist); };
-        
-      } else {
-        lv_checkbox_set_checked(cbOption[i], false);
+        auto& currentCheckbox = cbOption[i];
+        auto currentCheckboxState = !lv_checkbox_is_checked(currentCheckbox);
```
petterhs, line 98: I think it is cleaner now to take this back to.
```
auto currentCheckboxState = lv_checkbox_is_checked(currentCheckbox);
```
And instead change to
```
if (currentCheckboxState) {
```
the 3 times it is used. Easier to read and makes more sense that currentCheckboxState is the actual checkbox state at that time.

petterhs (commented):
Everything is working as intended now :+1: So here is just some cleanup proposals.

I still think the `|= ~()` operators etc. are a bit hard to read but maybe because I have not used them before. Lets see what @JF002 thinks.

src/displayapp/screens/settings/SettingWakeUp.cpp
Kozova1: good idea, done!

petterhs (approved):

src/components/settings/Settings.h
```diff
@@ -98,7 +128,7 @@ namespace Pinetime {
         uint32_t stepsGoal = 1000;
         uint32_t screenTimeOut = 15000;
 
-        WakeUpMode wakeUpMode = WakeUpMode::None;
+        uint8_t wakeUpMode = static_cast<uint8_t>(WakeUpMode::None);
```
JF002, line 131: Maybe we can investigate the use of std::bitset for these flags? Bitsets provides more readable API (set(), clear(),...) and might allow to write the code without all these static_cast.

I haven't used bitset that much, so I might be wrong...

src/displayapp/screens/settings/SettingWakeUp.cpp
```diff
@@ -85,27 +85,46 @@ bool SettingWakeUp::Refresh() {
 }
 
 void SettingWakeUp::UpdateSelected(lv_obj_t* object, lv_event_t event) {
-  if (event == LV_EVENT_VALUE_CHANGED) {
+  if (event == LV_EVENT_VALUE_CHANGED && !isLocked) {
+    isLocked = true;
```
JF002, line 89: Why do you need this variable `isLocked`?

`SettingWakeUp::UpdateSelected()` will never be called from 2 different tasks (this callback is called by LVGL which only runs in the context of DisplayApp) so it should not be needed to protect against race conditions here.

Kozova1: later on we use `lv_checkebox_set_checked()`, so isLocked is there to prevent those extra events from causing issues

src/components/settings/Settings.h
Kozova1: I've played around a bit with bitsets, it would still require the static casts.

Kozova1: This new push should convert all of those to bitsets

src/displayapp/screens/settings/SettingWakeUp.cpp
JF002: I still doesn't look good to me... I mean... couldn't we find another way to implement the logic so that we don't have to protect ourselves again event we generate ?

JF002: Another idea I've just had : would it be possible to explicitly disable the event on the checkboxes while we are changing their states, and re-enabling it once we are done?

JF002: My concern about this is that the code is confusing. When I read the code the first time, when I saw the variable `IsLocked`, I thought this was a simple implementation of a mutex to protect against race conditions against multiple tasks. And then, as I said before, it makes little sense to protect ourselves against event we generate ourselves.

I'm just trying to see if there is any better solution that would make the code more expressive and readable..

If we cannot find a better solution right now, I think we should at least find a better name for IsLocked and document why it's needed so that someone can try to improve that in the future.

JF002:
@Kozova1 I don't  what happened to your branch, but it doesn't build any more:

```
/home/jf/git/PineTime/src/displayapp/screens/Steps.cpp: In constructor 'Pinetime::Applications::Screens::Steps::Steps(Pinetime::Applications::DisplayApp*, Pinetime::Controllers::MotionController&, Pinetime::Controllers::Settings&)':
/home/jf/git/PineTime/src/displayapp/screens/Steps.cpp:29:76: error: 'class Pinetime::Controllers::Settings' has no member named 'GetStepsGoal'
   29 |   lv_arc_set_value(stepsArc, int16_t(500 * stepsCount / settingsController.GetStepsGoal()));
```

Step goal come from another PR I merged this morning... You'll probably need to rebase your branch on current develop.

Also, the issue with `IsLocked` is a minor one, really. May I suggest to rename it `ignoreEvent` and add a comment when you set it to true, something like this:
```
// Ignore all intermediate events that will be generated by checkboxes while we modify their state in this event handle
```

Kozova1:
Seems like I messed up my rebase

Kozova1:
@JF002 I changed `isLocked` to `ignoreEvent` as suggested and added the comment in the event handler.
Build issues were caused by my incorrect rebasing of the Steps app, should be fixed now.

JF002:
> @JF002 I changed `isLocked` to `ignoreEvent` as suggested and added the comment in the event handler.
> Build issues were caused by my incorrect rebasing of the Steps app, should be fixed now.

Thanks!

However... Did you test it? It doesn't work on my setup... I select 1 or 2 wake up sources, quite the settings app, go back to the wake up configuration and the settings changed (select none, and all checkboxes will be enabled when leaving and re-entering into the app).

Could you have a look ? If you cannot test the next changes, could you please specify it so that we are aware that we'll potentially need to debug it?

Kozova1:
Hey, I do not have access to a non-sealed device so I can  not test the build - I've pushed a new version that should hopefully work

JF002:
@Kozova1 Thanks for your work! Knowing that you develop the whole feature without testing on an actual device is impressive!

In the future, could you please specify that you didn't test the code on an actual device ? Reviewing and testing PR takes a lot of time, and reviewing un-tested PR requires extra attention from me and other contributors. 

I'll probably delay the merge of this feature until I have enough time to do more testing. Unless someone else can help on this ?

Kozova1:
No problem.  
For the record, the first two commits (up until the change to `std::bitset`) were tested with the help of @petterhs and @Caton101

kieranc:
I've tested this, and while the logic in the settings seems to work well, tap to wake doesn't work when wrist raise is enabled. It works when only tap is selected.

Kozova1:
Good to know, will take a look at it

src/components/settings/Settings.h
```diff
@@ -11,7 +12,12 @@ namespace Pinetime {
     public:
       enum class ClockType { H24, H12 };
       enum class Vibration { ON, OFF };
-      enum class WakeUpMode { None, SingleTap, DoubleTap, RaiseWrist };
+      enum class WakeUpMode : uint8_t {
+        None = 0,
```
jonvmey, line 16: I don't think `None` needs to be its own enumerated value here. `None` should effectively just be equivalent to all three other options disabled. This would allow getting rid of the extra logic to clear the None checkbox when other are selected or set it when no other modes are enabled.

src/components/settings/Settings.h
```diff
@@ -106,7 +139,7 @@ namespace Pinetime {
         uint32_t stepsGoal = 10000;
         uint32_t screenTimeOut = 15000;
 
-        WakeUpMode wakeUpMode = WakeUpMode::None;
+        std::bitset<4> wakeUpMode {static_cast<size_t>(WakeUpMode::None)};
```
jonvmey, line 142: This doesn't initialize the `bitset` with bit 0 set and all other bits cleared which I think was the intended behaviour. It initializes the `bitset` with the 4 least significant bit of the value 0, so all bits start cleared.

src/displayapp/screens/settings/SettingWakeUp.cpp
```diff
@@ -85,27 +85,48 @@ bool SettingWakeUp::Refresh() {
...
+            settingsController.unsetWakeUpMode(Pinetime::Controllers::Settings::WakeUpMode::SingleTap);
+          }
+        } else if (i == 2) {
+          lv_checkbox_set_checked(cbOption[1], false);
+          if (currentCheckboxState) {
+            settingsController.setWakeUpMode(Pinetime::Controllers::Settings::WakeUpMode::DoubleTap);
+          } else {
+            settingsController.unsetWakeUpMode(Pinetime::Controllers::Settings::WakeUpMode::DoubleTap);
+          }
+        } else if (i == 3) {
```
jonvmey, line 116: I know that it was like this before but replacing the magic integer literals with the relevant enum value would make this code easier to follow without needing to reference what each index refers to.

Eg. Changing
```
} else if (i == 1) {
  lv_checkbox_set_checked(cbOption[2], false);
```
to
```
} else if (i == static_cast<int>(Pinetime::Controllers::Settings::WakeUpMode::SingleTap)) {
  lv_checkbox_set_checked(cbOption[static_cast<size_t>(Pinetime::Controllers::Settings::WakeUpMode::DoubleTap)], false);
```
is easier to follow without memorizing the values of `WakeUpMode` or switching to another file to look them up, IMO at least (except maybe for all of the namespace identifiers but that's solvable too).

src/components/settings/Settings.h
```diff
@@ -69,13 +75,38 @@ namespace Pinetime {
       };
 
       void setWakeUpMode(WakeUpMode wakeUp) {
```
jonvmey, line 77: Changing `setWakeUpMode(WakeUpMode wakeUp)` to `setWakeUpMode(WakeUpMode wakeUp, bool enable)` and setting the relevant bit of the `bitset` to `enable` would eliminate the need for a separate `unsetWakeUpMode` method.

jonvmey (commented):
Just a couple possible changes that I thought might be improvements.

src/components/settings/Settings.h
Kozova1: Good idea!

src/displayapp/screens/settings/SettingWakeUp.cpp
Kozova1: Good point - but you have to add 1 to the static_casts (because the first checkbox is None).
Nevertheless, taking this advice

src/components/settings/Settings.h
Kozova1: Shouldn't be a problem anymore.

src/displayapp/screens/settings/SettingWakeUp.cpp
```diff
@@ -85,27 +85,39 @@ bool SettingWakeUp::Refresh() {
...
+            lv_checkbox_set_checked(cbOption[1 + static_cast<size_t>(WakeUpMode::DoubleTap)], false);
+          settingsController.setWakeUpMode(WakeUpMode::SingleTap, currentCheckboxState);
+        } else if (i == 1 + static_cast<size_t>(WakeUpMode::DoubleTap)) {
+          // SingleTap and DoubleTap checkboxes cannot be both checked at the same time.
+          if (!currentCheckboxState)
+            lv_checkbox_set_checked(cbOption[1 + static_cast<size_t>(WakeUpMode::SingleTap)], false);
+          settingsController.setWakeUpMode(WakeUpMode::DoubleTap, currentCheckboxState);
+        } else if (i == 1 + static_cast<size_t>(WakeUpMode::RaiseWrist)) {
+          settingsController.setWakeUpMode(WakeUpMode::RaiseWrist, currentCheckboxState);
+        }
```
jonvmey, line 115: Up to you, but in the same way that the `None` enumeration wasn't really needed, I don't think that the None checkbox is really needed either since selecting it is effectively the same as just unchecking all of the other options.

src/components/settings/Settings.h
```diff
@@ -68,14 +73,34 @@ namespace Pinetime {
...
       };
 
-      void setWakeUpMode(WakeUpMode wakeUp) {
-        if (wakeUp != settings.wakeUpMode)
+      void resetWakeUpModes() {
+        settings.wakeUpMode.reset();
+      }
+
+      void setWakeUpMode(WakeUpMode wakeUp, bool enable) {
+        if (!isWakeUpModeOn(wakeUp) != enable) // Boolean XOR
```
jonvmey, line 81: I think the ! before `isWakeUpModeOn` here needs to be removed as XOR is just `A != B`.

src/displayapp/screens/settings/SettingWakeUp.cpp
Kozova1: I'd prefer not to change the UI too much, and a "None" checkbox can be useful to turn all other modes at once.

src/components/settings/Settings.h
Kozova1: good point. fixed in a sec.

src/systemtask/SystemTask.cpp
```diff
@@ -388,10 +388,8 @@ void SystemTask::OnTouchEvent() {
   if (!isSleeping) {
     PushMessage(Messages::OnTouchEvent);
     displayApp.PushMessage(Pinetime::Applications::Display::Messages::TouchEvent);
-  } else if (!isWakingUp) {
-    if (settingsController.getWakeUpMode() == Pinetime::Controllers::Settings::WakeUpMode::None or
-        settingsController.getWakeUpMode() == Pinetime::Controllers::Settings::WakeUpMode::RaiseWrist)
-      return;
+  } else if(!isWakingUp) {
+    if(settingsController.isWakeUpModeOn(Pinetime::Controllers::Settings::WakeUpMode::RaiseWrist) ) return;
```
Riksu9000, line 392: Fix formatting

src/displayapp/screens/settings/SettingWakeUp.cpp
```diff
@@ -45,31 +45,31 @@ SettingWakeUp::SettingWakeUp(Pinetime::Applications::DisplayApp* app, Pinetime::
   lv_checkbox_set_text_static(cbOption[optionsTotal], " None");
   cbOption[optionsTotal]->user_data = this;
   lv_obj_set_event_cb(cbOption[optionsTotal], event_handler);
-  if (settingsController.getWakeUpMode() == Pinetime::Controllers::Settings::WakeUpMode::None) {
+  if (settingsController.getWakeUpModes().none()) {
     lv_checkbox_set_checked(cbOption[optionsTotal], true);
   }
```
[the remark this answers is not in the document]
Kozova1: None checkbox can still be there to easily deselect all of them, or to provide indication whether a checkbox is checked or not (having checked checkboxes be just green rectangles is a bit unintuitive imo).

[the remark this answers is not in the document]
Riksu9000: The user probably won't want to disable all wakeup options, and it adds unnecessary complexity.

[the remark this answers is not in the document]
Kozova1: Good point. Removed None checkbox.

src/systemtask/SystemTask.cpp
Kozova1: Whoops! think I fixed it now.

Kozova1:
@JF002 After testing this for a bit on my watch, it seems to work almost perfectly - When `Raise Wrist` and `Double Tap` are both selected at the same time, Double Tap detection seems to be a bit dodgy, but still works. Would like some help with fixing this before this is accepted.

Riksu9000:
> When `Raise Wrist` and `Double Tap` are both selected at the same time, Double Tap detection seems to be a bit dodgy, but still works. Would like some help with fixing this before this is accepted.

This is probably due to:
```
if (settingsController.isWakeUpModeOn(Pinetime::Controllers::Settings::WakeUpMode::RaiseWrist) or
    settingsController.getWakeUpModes().none()) {
    return;
}
PushMessage(Messages::TouchWakeUp);
```
On line 397 returning if RaiseWrist is enabled, so TouchWakeUp message doesn't get pushed. This should probably be changed to
```
if (wakeupmode singletap or doubletap) {
  pushmessage
}
```

Kozova1:
Thanks! After that change, everything seems to work perfectly.

src/displayapp/screens/settings/SettingWakeUp.cpp
JF002: If I understood correctly, `SettingWakeUp::UpdateSelected()` is called whenever the state of a checkbox changes. This change happens when the user taps a checkbox or... when the code explicitly changes the state of the checkbox.
That's what happens here : when the user taps a checkbox, the code needs to uncheck  mutually exclusives checkboxes. This call to `lv_checkbox_set_state()` will trig another (not needed) call to `SettingWakeUp::UpdateSelected()`.

If we decide to keep this workflow, I would just like to rename `isLocked` with something that shows what it's actually doing : ignore event. Maybe `ignoreSelectEvent` ?

Riksu9000: I don't think that's correct. I even looked at the lvgl sources and couldn't find any evidence of an event being generated when `lv_checkbox_set_state` is called. It wouldn't really make sense either since there is `lv_event_send(obj, LV_EVENT_..., &custom_data)` if we needed to simulate an actual tap.

I would just try this, but I'm worried that messing with settings could cause problems on my sealed pinetime.

Kozova1:
[pinetime-mcuboot-app-dfu-1.2.0.zip](https://github.com/JF002/InfiniTime/files/6803697/pinetime-mcuboot-app-dfu-1.2.0.zip)
This is a DFU that has this feature in it

Itai-Nelken:
Just tested and it works great!

JF002:
Thank you very much to @Kozova1 for this PR, and to everyone else for contributing, testing and reviewing those changes :1st_place_medal: !
And I'm really sorry it took so long for these changes to be merged! 

And thank you for this new feature, I was actually waiting for it :+1:  !

JF002 (approved):

PR #292, opened 2021-04-18
Added a pre-commit hook that should simplify commiting pre-formatted code

Avamander:
In order to reduce forgetfulness when commiting code it's a good idea to have a pre-commit hook. This also creates the opportunity of creating a PR gate that checks that it is enabled.

This PR adds a simple pre-commit hook that autoformats files that are in a known safe location, that's why the relatively ugly regex. Just in case it also prints out that it is doing so. I also included a README in the folder with installation instructions.

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -1,13 +1,19 @@
 #!/bin/bash
+if clang-format --version | grep -q 'version 11\.'; then
+   CLANG_FORMAT_EXECUTABLE="clang-format"
+else
+   CLANG_FORMAT_EXECUTABLE="clang-format-11"
+fi
+
 for FILE in $(git diff --cached --name-only)
 do
 	if [[ "$FILE" =~ src/[A-Za-z0-9\ \-]+*\.(c|h|cpp|cc)$ ]]; then
-		echo Autoformatting $FILE with clang-format
-		clang-format-11 -style=file -i -- $FILE
+		echo Autoformatting $FILE with $CLANG_FORMAT_EXECUTABLE
+		$CLANG_FORMAT_EXECUTABLE -style=file -i -- $FILE
 		git add -- $FILE
 	elif [[ "$FILE" =~ src/(components|displayapp|drivers|heartratetask|logging|systemtask)/.*\.(c|h|cpp|cc)$ ]]; then
-		echo Autoformatting $FILE with clang-format
-		clang-format-11 -style=file -i -- $FILE
+		echo Autoformatting $FILE with $CLANG_FORMAT_EXECUTABLE
+		$CLANG_FORMAT_EXECUTABLE -style=file -i -- $FILE
 		git add -- $FILE
 	fi
 done
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -1,0 +1,13 @@
+#!/bin/bash
+for FILE in $(git diff --cached --name-only)
+do
+	if [[ "$FILE" =~ src/[A-Za-z0-9\ \-]+*\.(c|h|cpp|cc)$ ]]; then
+		echo Autoformatting $FILE with clang-format
+		clang-format-11 -style=file -i -- $FILE
+		git add -- $FILE
+	elif [[ "$FILE" =~ src/(components|displayapp|drivers|heartratetask|logging|systemtask)/.*\.(c|h|cpp|cc)$ ]]; then
+		echo Autoformatting $FILE with clang-format
+		clang-format-11 -style=file -i -- $FILE
+		git add -- $FILE
+	fi
+done
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -5,6 +5,12 @@ else
    CLANG_FORMAT_EXECUTABLE="clang-format-11"
 fi
 
+if ! command -v $CLANG_FORMAT_EXECUTABLE &> /dev/null
+then
+    echo $CLANG_FORMAT_EXECUTABLE does not exist, make sure to install it
+    exit 1
+fi
+
 for FILE in $(git diff --cached --name-only)
 do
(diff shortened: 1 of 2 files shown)


hooks/pre-commit
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+for FILE in $(git diff --cached --name-only)
+do
+	if [[ "$FILE" =~ src/[A-Za-z0-9\ \-]+*\.(c|h|cpp|cc)$ ]]; then
+		echo Autoformatting $FILE with clang-format
+		clang-format-11 -style=file -i -- $FILE
```
JF002, line 6: On my system (Manjaro with clang 11 installed), there's no `clang-format-11` command available. `clang-format` is available, though.
Is there anyway to make this script portable across Linux distros?

Avamander: Parsing the output of `clang-format --version` could work to detect clang-format 11. Though maybe an environment variable would work better?

JF002: Mhm I don't know... does that `clang-format-11` command exists on your system?

Avamander: @JF002 Yes, `clang-format` on my system is 10 not 11, that's why I used that binary. I'll try to figure out how to autodetect this. If you want to use it meanwhile you can install a modified version in your `.git` folder.

Avamander:
@JF002

I added a simple check to use `clang-format` if it's the correct version.

pfeerick:
Be nice to see some basic checks like `if ! command -v clang-format &> /dev/null` to make sure the executable is there before the shell script breaks spectacularly.

Avamander:
It'll error, as it should when it is installed when the executable doesn't exist and you've manually installed it. Additional checking doesn't change pretty much anything here.

pfeerick:
Huh? 

The additional checking would be to check for the presence of clang-format, and instead allowing the commit to go ahead with merely a 'clang-format not found' which could go unnoticed instead do a  `echo "clang-format not installed - install it" && exit 1` so that git knows there is a problem with the pre-commit job and to not continue with the commit. If the user wants to force the commit still, that's what the `--no-verify` flag is there for. 

I would also expect all the requirements of any hook would be documented - in the `hooks\README.md` it should state that the pre-commit hook requires the clang-format package/executable.

Avamander:
That's a better point, that no `exit` allows the commit to complete. I'll add that nagging.

jwillikers:
Why not use the existing ClangFormat integration? https://github.com/doublify/pre-commit-clang-format

Avamander:
@jwillikers extra dependencies people have to install. Relatively unnecessary.

hooks/pre-commit
```diff
@@ -0,0 +1,25 @@
...
+for FILE in $(git diff --cached --name-only)
+do
+	if [[ "$FILE" =~ src/[A-Za-z0-9\ \-]+*\.(c|h|cpp|cc)$ ]]; then
+		echo Autoformatting $FILE with $CLANG_FORMAT_EXECUTABLE
+		$CLANG_FORMAT_EXECUTABLE -style=file -i -- $FILE
+		git add -- $FILE
+	elif [[ "$FILE" =~ src/(components|displayapp|drivers|heartratetask|logging|systemtask)/.*\.(c|h|cpp|cc)$ ]]; then
+		echo Autoformatting $FILE with $CLANG_FORMAT_EXECUTABLE
+		$CLANG_FORMAT_EXECUTABLE -style=file -i -- $FILE
+		git add -- $FILE
```
Riksu9000, line 23: These are quite hard to read. It seems that the top one matches all files in src and the bottom one ignores libs and FreeRTOS folders. Would it make more sense to specifically ignore these
```
if [[ "$FILE" =~ src/libs/.* ]] || [[  "$FILE" =~ src/FreeRTOS/.* ]]
then
  continue
fi
```
and match everything else based on just the extension `if [[ "$FILE" =~ .*\.(c|h|cpp|cc)$ ]]`

hooks/pre-commit
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+if clang-format --version | grep -q 'version 11\.'; then
+   CLANG_FORMAT_EXECUTABLE="clang-format"
+else
+   CLANG_FORMAT_EXECUTABLE="clang-format-11"
+fi
```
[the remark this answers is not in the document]
Avamander: Support is currently limited to Clang 11.

Installing it would be a reasonable approach. If you confirmed it working, a future PR would be welcome.

hooks/pre-commit
Avamander: It's intentionally an explicit and a predefined choice to avoid any unnecessary formats.

Have you tested your suggestion?

Riksu9000: I tested it separately and now in the repo and it works fine.

hooks/pre-commit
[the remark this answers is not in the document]
Riksu9000: To me it seems more likely for this to stop working after a system update, than a newer clang-format to break.

Riksu9000:
This also gets run when merging branches, potentially creating new unrelated changes if some code in the repo isn't formatted correctly. This might not matter, but I thought I should mention.

jwillikers:
> This also gets run when merging branches, potentially creating new unrelated changes if some code in the repo isn't formatted correctly. This might not matter, but I thought I should mention.

LLVM provides [clang-format-diff.py](https://github.com/llvm/llvm-project/blob/main/clang/tools/clang-format/clang-format-diff.py), which makes it pretty straight-forward to apply formatting only on the specific lines changed by a commit / diff.
Formatting changed lines instead of files should prevent unrelated changes from cropping up.

Avamander:
`clang-format-diff.py` is interesting, not too difficult to locate on Ubuntu systems but Arch *et al* support should be added by someone else.

PR #293, opened 2021-04-18
Reformatted all the files according to clang-format style

Avamander:
This PR autoformats every InfiniTime file according to the clang-format (also updated in this PR). 

The PR will conflict with each and every other code PR!

Following the code style should be enforced in some way after the PR.


JF002:
The alignment of `public` and `private` keyword in class declaration is a bit weird:
```
namespace Pinetime {
  namespace Applications {
    namespace Screens {

      class Motion : public Screen {
          public:
        Motion(DisplayApp* app, Controllers::MotionController& motionController);
        //...
```

Is this intentional? I'm more used to this:
```
namespace Pinetime {
  namespace Applications {
    namespace Screens {

      class Motion : public Screen {
      public:
        Motion(DisplayApp* app, Controllers::MotionController& motionController);
        //...
```

Or this: 
```
namespace Pinetime {
  namespace Applications {
    namespace Screens {

      class Motion : public Screen {
        public:
        Motion(DisplayApp* app, Controllers::MotionController& motionController);
        //...
```

Also, template class declarations are on 1 line:
```
    template <class T> class SystemMonitor {
        public:
      SystemMonitor() = delete;
    };
 ```
when I'm used to declare them on 2 lines:
```
    template <class T> 
    class SystemMonitor {
        public:
      SystemMonitor() = delete;
    };
```

What is your opinion about theses 2 points?

Avamander:
@JF002 I don't think there's a good way to achieve either currently. `AccessModifierOffset` can be changed into `0` which leaves it on the same level as the items after the access modifier, indentation after access modifier is not yet in (https://reviews.llvm.org/D60225).

JF002:
`-2` looks more natural to me.
In the current state, the access modifier (public/private) is more indented than the fields of the class, and that looks weird to me...

Avamander:
@JF002 I've changed it to -2 and updated the files.

JF002:
@Avamander Thanks !

Issue #268, opened 2021-04-14
Build is currently failing


nanch:
The GitHub "Build PineTime Firmware" workflow is currently failing.

nanch:
I'm working on this.

PR #294, opened 2021-04-18
fixes GitHub automated build

nanch:
#268 

This resolves the build issues.

I added workflow_dispatch so that builds can be triggered manually on the workflow page. This is useful for debugging and monitoring the build process.

the cbor python module is not in the mcuboot/scripts/requirements.txt file, so I added that into the workflow. Interestingly, cbor is in the source, but not in the version that gets checked-out (perhaps a caching issue or checking out the wrong version?) Either way, installing it as a build step is trivially adequate.

The input to the imgtool.py was not correct because multiple files were being passed into the input. I used a grep to select the correct file to use as the input to imgtool.py

submodules are required in the checkout because lvgl is a submodule (I think). Without submodules, the lvgl.h wasn't being included properly.

ACTIONS_ALLOW_UNSECURE_COMMANDS is needed since github updated their workflow configuration.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,13 +9,17 @@ name: Build PineTime Firmware
 # When to run this Workflow...
 on:
 
+  # Run this Workflow if a user presses the "Run workflow" button on GitHub
+  workflow_dispatch:
+    branches: [ master, nan_fix_build, develop ]
+    
   # Run this Workflow when files are updated (Pushed) in the "master" Branch
   push:
-    branches: [ master ]
+    branches: [ develop, master ]
 
   # Also run this Workflow when a Pull Request is created or updated in the "master" Branch
   pull_request:
-    branches: [ master ]
+    branches: [ develop, master ]
 
 # Steps to run for the Workflow
 jobs:
@@ -28,7 +32,7 @@ jobs:
 
       #########################################################################################
       # Download and Cache Dependencies
-
+      
       - name: Install cmake
         uses: lukka/get-cmake@v3.18.3
 
@@ -45,11 +49,14 @@ jobs:
       - name: Install Embedded Arm Toolchain arm-none-eabi-gcc
         if:   steps.cache-toolchain.outputs.cache-hit != 'true'  # Install toolchain if not found in cache
         uses: fiam/arm-none-eabi-gcc@v1.0.2
-        with:
+        with:          
           # GNU Embedded Toolchain for Arm release name, in the V-YYYY-qZ format (e.g. "9-2019-q4")
           release: 9-2020-q2
           # Directory to unpack GCC to. Defaults to a temporary directory.
           directory: ${{ runner.temp }}/arm-none-eabi
+        env:
+          # needed for set-env (https://github.com/actions/toolkit/issues/641)
+          ACTIONS_ALLOW_UNSECURE_COMMANDS: 'true'
 
       - name: Check cache for nRF5 SDK
         id:   cache-nrf5sdk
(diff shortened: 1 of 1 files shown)


petterhs:
Great work! Nice to see this building again.

I don't have any experience with github actions but it seems like there is still something wrong with the uploaded dfu artifact. The size is too small and it fails if I try to DFU with it. Optimally it should also upload the pinetime-mcuboot-app-image-x.y.z.bin file aswell.

pfeerick:
I can't reproduce that failure... I was able to flash the DFU zip present at https://github.com/JF002/InfiniTime/actions/runs/761649804 ... *BUT* it seems it was rebuild about 5 minutes ago, even though there was no change to this PR... so I can't discount variation in the file.

nanch:
@petterhs 

> Great work! Nice to see this building again.
> 
> I don't have any experience with github actions but it seems like there is still something wrong with the uploaded dfu artifact. The size is too small and it fails if I try to DFU with it. Optimally it should also upload the pinetime-mcuboot-app-image-x.y.z.bin file aswell.

I agree that there are improvements to be made. I didn't want to make too many changes at once without a successfully merged pull request first.

My goal here was to "get the build working" with minimal changes.

Once this pull request is merged I'm happy to define the exported artifacts optimally.

There are many artifacts that can/should be exported. Which artifact is useful at any given time depends on what you're trying to do. For most users who are simply flashing to update, **artifact D** (below) is all they need.

All possible artifacts (as far as I understand):

A: infinitime image without MCUBoot bootloader support (flash-write at offset 0x0000**0**000) [No bootloader support]

B: MCUBoot bootloader image (flash-write with offset 0x0000**0**000) [Bootloader support]
C: infinitime image with MCUBoot bootloader support [Bootloader support]

**D: Device Firmware Upgrade (DFU) .zip file ([C] infinitime image with MCUBoot bootloader support)** [Bootloader support]

E: raw pinetime-recovery-1.0.0.bin (flash-write with offset 0x0000**0**000) [No bootloader support]
F: pinetime-mcuboot-recovery-loader-1.0.0.bin [Bootloader support]


For ease of the reader, the MCUboot bootloader images are most recently referenced in the releases at https://github.com/JF002/InfiniTime/releases/tag/0.14.1

nanch:
@petterhs @pfeerick

> something wrong with the uploaded dfu artifact

> I can't reproduce that failure

I'll fork, define the artifacts, test each of the artifacts, and create a follow-up pull request.

.github/workflows/main.yml
```diff
@@ -124,12 +139,17 @@ jobs:
         run:  |
           cd build
           make pinetime-mcuboot-app
+          echo ""
 
       - name: Create firmware image
```
DavidVentura, line 144: this step should be deleted -- `make pinetime-mcuboot-app` already does this (except the file name is `image` instead of `img`)

.github/workflows/main.yml
```diff
@@ -124,12 +139,17 @@ jobs:
...
           ls -l build/src/pinetime-mcuboot-app*.bin
-          ${{ runner.temp }}/mcuboot/scripts/imgtool.py create --align 4 --version 1.0.0 --header-size 32 --slot-size 475136 --pad-header build/src/pinetime-mcuboot-app*.bin build/src/pinetime-mcuboot-app-img.bin
+          IMGTOOL_INPUT_ARGUMENT=$( ls build/src | grep "pinetime-mcuboot-app-[0-9][0-9]*\.[0-9][0-9]*\.[0-9][0-9]*\.bin" )
+          echo ""
+          echo -e "\u001b[32m ${IMGTOOL_INPUT_ARGUMENT} \u001b[0m"
+          echo ""
+          ${{ runner.temp }}/mcuboot/scripts/imgtool.py create --align 4 --version 1.0.0 --header-size 32 --slot-size 475136 --pad-header build/src/${IMGTOOL_INPUT_ARGUMENT} build/src/pinetime-mcuboot-app-img.bin
           ${{ runner.temp }}/mcuboot/scripts/imgtool.py verify build/src/pinetime-mcuboot-app-img.bin
 
       - name: Create DFU package
```
DavidVentura, line 155: this one also could be removed by enabling DFU in the Cmake step

JF002: Right! It was not the case before, but now, images and DFU files are generated directly by CMake so that you don't need to add custom step after the build. You can have a look at the docker file to see how it's done : 
 - build : https://github.com/JF002/InfiniTime/blob/develop/docker/build.sh
 - post build : https://github.com/JF002/InfiniTime/blob/develop/docker/post_build.sh.in

Previously, the post build script generated images and DFU, but now it just copies the binaries into the output folder.

Basically, all you have to do is add  `-DBUILD_DFU=1` to your cmake command line.

JF002:
@nanch Could you have a look at the review @DavidVentura and myself did about this PR?

JF002:
CI builds were (finally) fixed in #579. Feel free to contribute and improve Github Actions configuration in a new PR ;)

PR #298, opened 2021-04-19
New Steps app

joaquimorg:
- New Steps app
- Settings to set the steps goal
- More detail in Motion app
- New 42px Font

IMG_20210419_152300

IMG_20210419_153435

IMG_20210419_152316


petterhs (approved):
Very nice! Everything works and looks good. 
Minor improvements:
- Make the +/- buttons in step settings a bit bigger
- Make 10000 steps default goal instead of 1000

Future possibilities:
- Have a notification when goal is hit
- Possible to turn step counting off in settings
- (Move Motion app somewhere in the settings as it is more of a testing/debug application now at least)

pfeerick:
Looks great! I agree on all the points @petterhs made. Particularly the 10,000 steps part, as I was part of a active living promotion/campaign 10 or so years ago... funnily enough, called [10,000 steps](https://www.cqu.edu.au/cquninews/stories/research-category/2019/10,000-steps-a-staggering-success) 😆

joaquimorg:
> Very nice! Everything works and looks good.
> Minor improvements:
> 
> * Make the +/- buttons in step settings a bit bigger
> * Make 10000 steps default goal instead of 1000
> 
> Future possibilities:
> 
> * Have a notification when goal is hit
> * Possible to turn step counting off in settings
> * (Move Motion app somewhere in the settings as it is more of a testing/debug application now at least)

Yes, I will make the changes.

pfeerick:
I've just tried this, and it looks and works great. Rather strangely, it seems like it still defaulted to 1000, but I'm definitely on  5fc07a8df7d18d1fa9c4971f2c7b69ea68db559b, so it should be using the new default of 10000. I DFU updated from the 1.0.0 RC4, so it's not like I'd already set and saved a value. Just in case someone else gets the same behaviour, or it was just something in the Matrix ;)

petterhs:
> I've just tried this, and it looks and works great. Rather strangely, it seems like it still defaulted to 1000, but I'm definitely on [5fc07a8](https://github.com/JF002/InfiniTime/commit/5fc07a8df7d18d1fa9c4971f2c7b69ea68db559b), so it should be using the new default of 10000. I DFU updated from the 1.0.0 RC4, so it's not like I'd already set and saved a value. Just in case someone else gets the same behaviour, or it was just something in the Matrix ;)

The stepsGoal value in settings is already present in the RC's so then value 1000 was already loaded to you SPI flash

pfeerick:
Ah, fair enough... thank you @petterhs! :) In that case, it works perfectly. 

As far as the motion app... I don't see why it can't stay where it is... it's as much a toy as the paint app is... I see it more as a demo of the sensors that can be read, which can get the boot when space becomes an issue or there are enough apps to warrant the slot wanted.

MysteriousLog6:
Really excited about this! The default app should be moved somewhere else , maybe a hidden settings menu? Similar to what @petterhs was saying.

JF002:
Nice PR, as always, @joaquimorg. I'll try to review it soon!
I checked the memory usage : this PR adds ~12KB in flash memory. That's ok, there's still some room, but we are getting closer and closer to the limit of our internal flash space...

joaquimorg:
What I was able to validate is that the problem is not the size of the button but the interval between each touch, if it is too fast it does not detect the touch on the button, but if we take a short break between each touch it works well, I will try to understand whatever might be behind it.

As for the Motion application, I think that for now it can stay, I use it to validate if the accelerometer is working, later on when there is a need for space we can remove it.

JF002:
@joaquimorg you might be right about the interval between 2 taps. If it's the case, that would be an issue that could be handled in another issue/PR.
Let me know if you find something interesting about this!

And I agree with you about the motion app ;)

PR #308, opened 2021-04-21
Update LVGL

joaquimorg:
Merge from upstrem :
 - fix(table) fix memory leak


petterhs:
Is it updated to v7.11.0?

joaquimorg:
We are already at 7.11.0 this is a fix
https://github.com/lvgl/lvgl/commits/release/v7

MysteriousLog6:
I think the LVGL is a Submodule?

joaquimorg:
> I think the LVGL is a Submodule?

Yes it is, but it is linked to a certain version of this module when there are updates we have to update that link to the new one.

MysteriousLog6:
Can't we link it to the main branch

pfeerick:
Submodule versions have to be pinned, otherwise you will have the "what version of the submodule is the code written for" issue. So when you want to use a newer version, you update the submodule version, which then brings in whatever changes were made in that repo.

PR #311, opened 2021-04-23
Add Hebrew symbols from SimpleCLM font

yehoshuapw:
First thing first: thanks a lot for the great work!

This pull request adds 28 Hebrew symbols (and, I assume since something updated in the generator or fonts, makes some changes to `jetbrains_mono_bold_20.c`)
It has been tested and used in the past few versions successfully.
(for those interested, Gadgetbridge has a RTL feature which can be enabled) 

It adds another 632 bytes to the code size.

Nonetheless, as has been commented in touching issues - this impact might be too much.


Man-with-Arrow:
I'm having trouble setting up the build environment and getting the DFU binaries to compile. @yehoshuapw  - would you mind sharing them? I (and at least two others) would greatly appreciate it.

yehoshuapw:
This is the 1.0.0 version: [pinetime-mcuboot-app-dfu-1.0.0.zip](https://github.com/JF002/InfiniTime/files/6413563/pinetime-mcuboot-app-dfu-1.0.0.zip)

Man-with-Arrow:
It works perfectly, with no additional observed issues or performance drops 

Thank you very, very much.

yehoshuapw:
Released 1.1.0: [Pinetime-hebrew 1.1.0](https://github.com/yehoshuapw/InfiniTime/releases/tag/1.1.0)
(tested)

Man-with-Arrow:
> Released 1.1.0: [Pinetime-hebrew 1.1.0](https://github.com/yehoshuapw/InfiniTime/releases/tag/1.1.0)
> (tested)

Works here. I've encountered an issue with [long notifications not properly displaying](https://github.com/JF002/InfiniTime/issues/388), can you reproduce it?

Kozova1:
I tried this but for some reason the letters are in reverse order when receiving messages, so the message:
`הודעת בדיקה`  
appears like this:  
`תעדוה הקידב`

Man-with-Arrow:
> I tried this but for some reason the letters are in reverse order when receiving messages, so the message:
> `הודעת בדיקה`
> appears like this:
> `תעדוה הקידב`

If you're using Gadgetbridge, you need to enable RTL under Settings/Language and region.

Kozova1:
Ah, that fixed it.

yehoshuapw:
Now that littlefs is in, this can (and should) be solved in a better way. (I might try myself)

Until that time, I shall continue to make releases with the relevant modifications - but there is no point having this merged.

PR #319, opened 2021-05-01
Use int8_t for PPG data array

### LINKED ISSUE #209: PPG array could be int8_t to reduce RAM footprint
Currently the array in Pinetime::Controllers::Ppg is std::array<int, 200> . 

The PPG algorithm is tuned for microcontrollers and part of that included designing the automatic gain control step so that the history buffer fits into a signed char. Specifically the output should be from the final biquad should be +-100.

Changing the type to int8_t or signed char will reduce the memory footprint of the PPG algorithm by 600 bytes (a.k.a. almost 1% of the total RAM ;-) ).

DavidVentura:
This fixes #209

### DIFF
--- a/src/components/heartrate/Ppg.cpp
+++ b/src/components/heartrate/Ppg.cpp
@@ -11,7 +11,7 @@ using namespace Pinetime::Controllers;
 
 /** Original implementation from wasp-os : https://github.com/daniel-thompson/wasp-os/blob/master/wasp/ppg.py */
 namespace {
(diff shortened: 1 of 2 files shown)


JF002:
Thanks @DavidVentura. Did you check that the processing of the heart rate data still work as expected?

JF002:
Oh and it might be interesting to check if we can reduce the size of the stack of the heart rate task, now that the memory requirement are reduced.

PR #322, opened 2021-05-02
SPI flash sleep if bootloader >= 1.0.0

nlfx:
This PR provide the following improvements:
- Extracts version exported by bootloader
- Displays bootloader version on System Info screen (Settings -> About menu)
- Enable SPI flash sleep mode if bootloader version >= 1.0.0

Unfortunately, I recently sealed my dev kit after flashing InfiniTime 1.0.0, so I would appreciate if someone with an open dev kit could help with the following tests:
- Is the reported bootloader version == 0.0.0 on watches with a bootloader older than 1.0.0 (before it exported its version in `NRF_TIMER2->CC[0]`)? Or is it a random value?
- Does this patch improve the power consumption of the watch when the screen is off? @vbelloir @nanch if any of you still have a setup to measure current in the μA range, I would love to have your feedback.

I'm also open to code improvement advice. I'm not a C++ developer so I'm not sure if storing the bootloader version in a class static variable is an acceptable way to do it or not.

Thanks a lot!

### DIFF
--- a/src/BootloaderVersion.cpp
+++ b/src/BootloaderVersion.cpp
@@ -1,26 +1,36 @@
 #include <cstdint>
+#include <cstdio>
 #include "BootloaderVersion.h"
 
 using namespace Pinetime;
 
-// NOTE : current bootloader does not export its version to the application firmware.
+// NOTE : version < 1.0.0 of bootloader does not export its version to the application firmware.
 
-uint32_t BootloaderVersion::Major() {
-  return 0;
+uint32_t BootloaderVersion::version = 0;
+char BootloaderVersion::versionString[BootloaderVersion::VERSION_STR_LEN] = "0.0.0";
+
+const uint32_t BootloaderVersion::Major() {
+  return (BootloaderVersion::version >> 16u) & 0xff;
 }
 
-uint32_t BootloaderVersion::Minor() {
-  return 0;
+const uint32_t BootloaderVersion::Minor() {
+  return (BootloaderVersion::version >> 8u) & 0xff;
 }
 
-uint32_t BootloaderVersion::Patch() {
-  return 0;
+const uint32_t BootloaderVersion::Patch() {
+  return BootloaderVersion::version & 0xff;
 }
 
 const char* BootloaderVersion::VersionString() {
-  return "0.0.0";
+  return BootloaderVersion::versionString;
 }
 
-bool BootloaderVersion::IsValid() {
-  return false;
+const bool BootloaderVersion::IsValid() {
+  return BootloaderVersion::version >= 0x00010000;
+}
+
+void BootloaderVersion::SetVersion(uint32_t v) {
+  BootloaderVersion::version = v;
+  snprintf(BootloaderVersion::versionString, BootloaderVersion::VERSION_STR_LEN, "%ld.%ld.%ld",
+           BootloaderVersion::Major(), BootloaderVersion::Minor(), BootloaderVersion::Patch());
 }
--- a/src/components/ble/DfuService.cpp
+++ b/src/components/ble/DfuService.cpp
@@ -121,6 +121,11 @@ int DfuService::WritePacketHandler(uint16_t connectionHandle, os_mbuf* om) {
       NRF_LOG_INFO(
         "[DFU] -> Start data received : SD size : %d, BT size : %d, app size : %d", softdeviceSize, bootloaderSize, applicationSize);
 
+      // wait until SystemTask has finished waking up all devices
+      while (systemTask.IsSleeping()) {
+        vTaskDelay(50); // 50ms
+      }
+
       dfuImage.Erase();
 
       uint8_t data[] {16, 1, 1};
--- a/src/BootloaderVersion.h
+++ b/src/BootloaderVersion.h
@@ -3,10 +3,15 @@
 namespace Pinetime {
   class BootloaderVersion {
   public:
-    static uint32_t Major();
-    static uint32_t Minor();
-    static uint32_t Patch();
+    static const uint32_t Major();
+    static const uint32_t Minor();
+    static const uint32_t Patch();
     static const char* VersionString();
-    static bool IsValid();
+    static const bool IsValid();
+    static void SetVersion(uint32_t v);
+  private:
+    static uint32_t version;
+    static constexpr size_t VERSION_STR_LEN = 12;
+    static char versionString[VERSION_STR_LEN];
   };
 }
--- a/src/displayapp/screens/SystemInfo.cpp
+++ b/src/displayapp/screens/SystemInfo.cpp
@@ -3,6 +3,7 @@
 #include "../DisplayApp.h"
 #include "Label.h"
 #include "Version.h"
+#include "BootloaderVersion.h"
 #include "components/battery/BatteryController.h"
 #include "components/ble/BleController.h"
 #include "components/brightness/BrightnessController.h"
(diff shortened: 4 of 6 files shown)


JF002:
Nice PR! I forgot to add the version from the bootloader after it was updated, so your modification is welcome!
I'll try to test it soon(tm), but someone else might be able to test it sooner than me :)

vbelloir:
Hi @nlfx 
What I can say:

**Without bootloader**
- when display is on, consumption drops from 27mA to 15mA
- when display is off, and adv off  (after 3min, consumption drops from 1.2mA to to 1.02mA.

**With Bootloader 1.0.0**
- when when both display and adv are off : 0.720mA.

It's really nice!

vbelloir:
I made these measurements, compiling source from your fork.  
I don't see in your commit when you are enabling this sleep mode, can you explain?

nlfx:
Hi @vbelloir

Thanks a lot for these measurements and your feedback!

The SPI flash sleep mode was [already implemented](https://github.com/JF002/InfiniTime/blob/e3ead332b9db3ad7b6acde933b6a7943b3ac81e2/src/systemtask/SystemTask.cpp#L254) by @JF002.
The only missing part was to retrieve the bootloader version and make `IsValid()` return true if its version is >= 1.0.0. And also wake up the flash when starting an OTA DFU with the screen turned off.

Thanks again for your help!

JF002:
This PR looks lile a quick-win to gain a bit of battery life!

But I think we'll need to do extensive testing before we actually release it to avoid unwanted side-effects. For info, we disabled the sleep mode of the flash memory just a few days before sending version 0.7.0 (which became 0.7.1) because we detected it was causing a bug in the bootloader (bootloader frozen, device, temporarily bricked). Now, the bootloader should not freeze, and, if we handled it correctly, the sleep mode should not be enabled with older  bootloader but I think we really should test that (OTA on older bootloader, newer bootloder, from infinitime 0.7.1, 1.0.0,..).

If tests are succesful, we'll add this in release 1.2 :)

nlfx:
Devkits are finally back in stock. I'll get a new one so I can test this with older bootloaders. I just don't know when I'll receive it, so we might need to shift the milestone to a later revision.
If no one else has time to test this PR until then, I'll report back once I have my new dev kit.

Raupinger:
I flashed my devkit with this branch and the bootloader from 6.0.0 and it shows up in system Info as 0.0.0. Tried both OTA updates from this branch to 0.7.1 and from that to this branch. Didn't notice anything out of the ordinary there. Flashing from 0.7.1 took a few tries but that's to be expected.

nlfx:
@Raupinger Thank you so much for testing with old bootloaders :-)

If it shows `0.0.0` for bootloaders < `1.0.0`, then I think this patch should not cause any problem as it doesn't enable SPI flash sleep in this case.

For bootloader >= `1.0.0`, I've been using this patch on my watch for more than a month, rebooted 30+ times (with screen on and off), made ~ 10 OTA updates, and didn't experience any problem.

src/BootloaderVersion.cpp
```diff
@@ -1,26 +1,36 @@
...
-  return 0;
+  return (BootloaderVersion::version >> 8u) & 0xff;
 }
 
 uint32_t BootloaderVersion::Patch() {
-  return 0;
+  return BootloaderVersion::version & 0xff;
 }
 
-const char* BootloaderVersion::VersionString() {
```
JF002, line 20: I think this method can still return `const char*` as the caller is not supposed to modify the string. In fact, other getters should also probably be `const`.

src/components/ble/DfuService.cpp
```diff
@@ -388,6 +388,7 @@ void DfuService::DfuImage::WriteMagicNumber() {
 }
 
 void DfuService::DfuImage::Erase() {
+  spiNorFlash.Wakeup();
```
JF002, line 391: This should not be needed, as `SystemTask` already takes care to wake up all devices, the SPI NOR flash included.
Did you spot any issue without waking the memory here?

JF002:
Thanks @nlfx, @Raupinger  and @vbelloir for this PR and these tests!

I am very sorry for the delay for this review! I've just looked at the code and did some tests, it looks good to me! 
The only think I'm not sure about is the call to `spiNorFlash::Wakeup()` in DfuService : it should not be needed to call it as SystemTask should have already waken up all device.

src/BootloaderVersion.cpp
nlfx: Thanks a lot for the feedback. I wrongly thought that a `const` method in C++ meant that the returned value was constant. I'll push a new version fixing that.

src/components/ble/DfuService.cpp
nlfx: Without this, future OTA DFU would fail if started with the screen turned off.

I see that the [`BleFirmwareUpdateStart` message is pushed](https://github.com/JF002/InfiniTime/blob/883700fca13a45b6cb0af4d74e3ef276a2d69283/src/components/ble/DfuService.cpp#L208) before going to the [`Start` state](https://github.com/JF002/InfiniTime/blob/883700fca13a45b6cb0af4d74e3ef276a2d69283/src/components/ble/DfuService.cpp#L116) where the SPI flash is erased. So my guess is that this pushed message, or [the `GotoRunning` message](https://github.com/JF002/InfiniTime/blob/883700fca13a45b6cb0af4d74e3ef276a2d69283/src/systemtask/SystemTask.cpp#L382), is not processed before arriving at the flash erase step.

nlfx:
@JF002 Thanks a lot for your review!

I've tested again without the `spiNorFlash::Wakeup()` call and the next OTA DFU fails if started with the screen turned off. It displays "Firmware update \n 0%" for a few seconds and then "Firmware update \n Error!".

Here is the buggy version without the call to `spiNorFlash::Wakeup()` if someone wants to test it: [pinetime-mcuboot-app-dfu-1.2.0-no-spi-wakeup.zip](https://github.com/JF002/InfiniTime/files/6698524/pinetime-mcuboot-app-dfu-1.2.0-no-spi-wakeup.zip)

src/components/ble/DfuService.cpp
JF002: Thanks for the explanation! You are right : the message `GotoRunning` is handled by SystemTask, while the DFU service is handled by the BLE task. 
On my setup, SystemTask wakes the memory up before the BLE task tries to erase the flash memory but... nothing prevent the BLE task to run before SystemTask (which is probably what you experienced).

So yeah, calling Wakeup() in DFUService might seem a good idea but there might still be a risk of race condition (both systemtask and ble task call that method) and wakeup will be called twice...

I wondering if we can find a nice and explicit solution to ensure that DFU does not use the memory until it's woken up by SystemTask.
Or ensure that the priorities of the tasks are correctly set to ensure that this race condition won't happen ?

nlfx:
I've pushed a new version where the BLE tasks waits for the System task to wakeup all devices before erasing the SPI flash.

I had to make the atomic `isSleeping` public, please tell me if you would prefer a different approach.

jonvmey:
Generally you don't want to expose internal data members with invariants as part of a classes public interface. If some other piece of code needs read-only access to that data then you would typically add a "getter" method, something like
```
bool SystemTask::IsSleeping() const {
  return isSleeping;
}
```
that way users of that information can't accidentally (or intentionally) modify the value of `isSleeping` when they shouldn't be in a way that could break things.

Raupinger:
TL;DR: use a getter. 

Here are my two cents: I generally prefer public data members over the "pure" object-oriented style of having getters and setters assuming you would have both and they don't have any side effects. This is mainly because it adds extra steps to things like checking usages (instead of seeing all places where some member is access some are hidden behind the getter/setter which you need to look up usages for) and general clutter. Also, while the actual impact is negligible, every function call pushes a bunch of stuff on the stack and that just seems wasteful to me. 

With that bit of philosophy aside, since the value is only read in this case, isSleeping should be private and using a getter. This way when you think about what values isSleeping might have, you only need to consider SystemTask and not (potentially) everything that holds a reference to it. In this case doing things the objet-orientated way has clear benefits.

Ultimately its about using OO style where it helps, but not dogmatically sticking to it when it does more harm than good. In this case it helps. Also, this is just the personal opinion of a single, fairly inexperienced person, so make of that what you wish.

nlfx:
@jonvmey Thanks a lot for the feedback. I've pushed a new version with a getter.

I assumed that using a getter defeated the purpose of the atomic, because the value could changed between when it's read and when it's returned by the function. But I guess the compiler will probably optimize that away and inline it? I'll let someone more knowledgeable than me in C++ decide what is the right approach, and change this back if necessary.

jonvmey:
`isSleeping` being atomic provides no guarantees that it's value in memory won't have changed between when it's read and when you use that value. It only makes guarantees that 1) accessing the variable simultaneously from multiple threads does not constitute a data race (which would be UB), and 2) depending on the `memory_order` specified for the operation, whether the effects of certain previous operations will be seen by the current thread or not.

Also trivial setters and getters are very likely to be inlined by the compiler.

TLDR: using a getter on an atomic here should be fine.

JF002:
Thanks everyone! This PR looks good to me, merging :)

PR #334, opened 2021-05-07
Add PineTimeStyle watchface

kieranc:
I created this watchface based on TimeStyle for the Pebble with the permission of the original creator.

It looks like this:

image


JF002:
Thanks for this PR, your watchface looks really nice!

As we are trying to rationalize the usage of the flash memory ([see here](https://github.com/JF002/InfiniTime/projects/3)), would it be possible to measure the impact on memory usage (mostly flash, I guess) of this PR  (just compare the size of the binary file with `master` or the 1.0 tag, for example) ?

Avamander:
@kieranc whoops, this branch is a bit off, a lot of old commits. Please rebase on latest JF002:develop.

src/displayapp/screens/PineTimeStyle.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+  notificationState = notificatioManager.AreNewNotificationsAvailable();
+  if (notificationState.IsUpdated()) {
+    if (notificationState.Get() == true) {
+      lv_label_set_text(notificationIcon, NotificationIcon::GetIcon(true));
+      lv_obj_realign(notificationIcon);
+      //lv_obj_align(bleIcon, sidebar, LV_ALIGN_IN_TOP_MID, -8, 25);
+    } else {
+      lv_label_set_text(notificationIcon, NotificationIcon::GetIcon(false));
+      //lv_obj_align(bleIcon, sidebar, LV_ALIGN_IN_TOP_MID, 0, 25);
```
Avamander, line 214: If these are not required, they should probably be removed.

src/libs/lv_conf.h
```diff
@@ -758,4 +759,4 @@ typedef void* lv_obj_user_data_t;
 
 /*--END OF LV_CONF_H--*/
 
-#endif /*LV_CONF_H*/
\ No newline at end of file
+#endif /*LV_CONF_H*/
```
Avamander, line 762: This change is probably not needed.

src/displayapp/screens/PineTimeStyle.cpp
```diff
@@ -0,0 +1,294 @@
...
+        minutesChar[1] != displayedChar[3]) {
+      displayedChar[0] = hoursChar[0];
+      displayedChar[1] = hoursChar[1];
+      displayedChar[2] = minutesChar[0];
+      displayedChar[3] = minutesChar[1];
+
+      char hourStr[3];
+      char minStr[3];
+
+      /*Display the time as 2 pairs of digits*/
```
Avamander, line 251: A space is usually used after `/*`

src/displayapp/screens/PineTimeStyle.cpp
```diff
@@ -0,0 +1,294 @@
...
+  lv_obj_align(batteryPlug, sidebar, LV_ALIGN_IN_TOP_MID, 0, 2);
+
+  bleIcon = lv_label_create(lv_scr_act(), nullptr);
+  lv_obj_set_style_local_text_color(bleIcon, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(0x000000));
+  //lv_label_set_text(bleIcon, BleIcon::GetIcon(false));
+  lv_obj_align(bleIcon, sidebar, LV_ALIGN_IN_TOP_MID, 0, 25);
+
+  notificationIcon = lv_label_create(lv_scr_act(), nullptr);
+  lv_obj_set_style_local_text_color(notificationIcon, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(0x000000));
+  //lv_label_set_text(notificationIcon, NotificationIcon::GetIcon(false));
```
Avamander, line 99: Unnecessary dead code?

src/displayapp/screens/PineTimeStyle.cpp
```diff
@@ -0,0 +1,294 @@
...
+#include "NotificationIcon.h"
+#include "Symbols.h"
+#include "components/battery/BatteryController.h"
+#include "components/ble/BleController.h"
+#include "components/ble/NotificationManager.h"
+#include "components/heartrate/HeartRateController.h"
+#include "components/settings/Settings.h"
+#include "../DisplayApp.h"
+
+/*
```
Avamander, line 17: You might just want to use GPLv3 license header and add this comment to its comment section. Would look a bit nice.

src/displayapp/screens/Clock.cpp
```diff
@@ -91,4 +105,4 @@ std::unique_ptr<Screen> Clock::WatchFaceCustomScreen() {
   return std::make_unique<Screens::WatchFaceCustom>(app, dateTimeController, batteryController, bleController, notificatioManager,
 settingsController);
 }
-*/
\ No newline at end of file
+*/
```
Avamander, line 108: Unnecessary file ending change.

src/displayapp/screens/settings/SettingWatchFace.cpp
```diff
@@ -81,4 +90,4 @@ void SettingWatchFace::UpdateSelected(lv_obj_t* object, lv_event_t event) {
       }
     }
   }
-}
\ No newline at end of file
+}
```
Avamander, line 93: Unnecessary file ending change

Avamander (changes requested):

src/displayapp/screens/settings/SettingWatchFace.cpp
kieranc: It's adding a newline at the end of the file - some files have them, some don't, it seemed easier to add them when they're missing than to strip them out everywhere.

src/displayapp/screens/PineTimeStyle.cpp
kieranc: Done

src/displayapp/screens/PineTimeStyle.cpp
kieranc: Removed

src/displayapp/screens/PineTimeStyle.cpp
Avamander: Usually placed before any includes

Avamander (changes requested):

src/displayapp/screens/settings/SettingWatchFace.cpp
Avamander: I think this is not currently enforced as a standard, the next person will revert again and so on and on before tooling allows a easy way to enforce. So it's better not to change.

Avamander (approved):
Needs testing, the rest is great!

JF002:
@kieranc I've just tried PineTimeStyle on my devkit : it's really nice! I might be able to merge it since we've freed some space in flash memory :)

By any chance, do you still have the original font file? It would be nice to add it to the project and update the [font documentation](https://github.com/kieranc/InfiniTime/blob/pinetimestyle/src/displayapp/fonts/README.md) with instruction on how to re-generate it (mostly the characters ranges).

kieranc:
I downloaded the font from [here](https://www.1001fonts.com/open-sans-font.html)
I used the existing documentation to generate the font using the lvgl online converter, just listing the symbols and not using a range.
It looks like I have a conflict before it can be merged - I'll try deal with it in the next couple of days.

JF002:
Don't forget to check the license of the font, to ensure we can use it in an opensource project :)

I can help with the conflict, they are caused by merges I did sooner today.

JF002:
I've just fixed the conflicts :)

kieranc:
Thanks!
As someone else commented, [Open Sans](https://fonts.google.com/specimen/Open+Sans#license) is licensed with Apache 2.0 so as far as I'm aware we can use it without issues.

JF002:
> Thanks!
> As someone else commented, [Open Sans](https://fonts.google.com/specimen/Open+Sans#license) is licensed with Apache 2.0 so as far as I'm aware we can use it without issues.

Indeed, Apache 2.0 should be OK :)

Could you add the source file of the font to the project so we can easily re-generate it if needed ?

src/displayapp/screens/PineTimeStyle.cpp
```diff
@@ -0,0 +1,341 @@
...
+  lv_label_set_text(dateDay, "25");
+  lv_obj_align(dateDay, sidebar, LV_ALIGN_CENTER, 0, 3);
+
+  dateMonth = lv_label_create(lv_scr_act(), nullptr);
+  lv_obj_set_style_local_text_color(dateMonth, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(0x000000));
+  lv_label_set_text(dateMonth, "MAR");
+  lv_obj_align(dateMonth, sidebar, LV_ALIGN_CENTER, 0, 32);
+
+  // Step count gauge
+  static lv_color_t needle_colors[1];
```
JF002, line 175: Is `static` needed here?

kieranc: I don't feel qualified to confidently answer that, but it didn't work correctly until it was static, and the LVGL docs state:
"Use the lv_gauge_set_needle_count(gauge, needle_num, color_array) function to set the number of needles and an array with colors for each needle. **The array must be static or global variable because only its pointer is stored.**"

kieranc:
> > Thanks!
> > As someone else commented, [Open Sans](https://fonts.google.com/specimen/Open+Sans#license) is licensed with Apache 2.0 so as far as I'm aware we can use it without issues.
> 
> Indeed, Apache 2.0 should be OK :)
> 
> Could you add the source file of the font to the project so we can easily re-generate it if needed ? This way, we ensure that everyone use the same version of the font and avoid [this](https://github.com/JF002/InfiniTime/pull/420).

Font source file added!

src/displayapp/screens/PineTimeStyle.cpp
JF002: Ok, I think I understand. When you declare the array on the stack like this : 

```
 PineTimeStyle::PineTimeStyle(...) {
...
lv_color_t needle_colors[1];
lv_gauge_set_needle_count(stepGauge, 1, needle_colors);
...
}
```
The variable will only exist during the execution of the method. As the doc says that it only keep the address of the array and does not copy its content, we muse ensure that the variable will still be valid in memory when LVGL will need to access it.

Declaring the variable as `static` work, but with the drawback that the memory is reserved... statically, and only once, even if multiple instances of PineTimeStyle are instanciated. There's very few chances we would need multiple instances of this class at the same time, but this is not a good pattern.

Instead, I would recommend to declare the array as **member variable** of the class `PineTimeStyle`. This way, you ensure that variable and its pointer will be valid as long as the instance of PineTimeStyle exists and is not destroyed.

kieranc: Resolved I think, thanks for the input

JF002:
This is looking really good!

You know what might be missing ? A bit of doc on how to generate the font for this watchface (parameters for the generator, range of symbols,...).

JF002:
Thanks for this PR @kieranc  !
I've been using this watchface for a few days and it's been awesome ! Really good looking and easily readable! Merging for InfiniTime 1.3!

PR #342, opened 2021-05-13
Call improvements

Raupinger:
Incoming calls now cause a periodic vibration. The vibration stops when the call is accepted, rejected, muted or the notification is dismissed. The timeout for the call notification preview doesn't start until one of the three buttons is pressed. 

Overview of the changes:
- introduced new methods for stopping and starting the vibration. `MotorController::startRunning(uint8_t motorDuration)` starts a vibration pattern with a period of motorDuration and 50% duty cycle. stopRunning() stops that pattern. In the meantime no other vibrations can be triggered.
- renamed `APP_TIMER_DEF(shortVibTimer)` to `MotorController::runForDuration(uint8_t motorDuration)` for clarity.
- `Screens::Notifications` and `Notifications::NotificationItem` now require a reference to the motor Controller. This is needed so they can turn off the vibrations based on user interaction.
- `SystemTask::Work()` will check the notification category and either starts a vibration pattern or a one time vibration based on it.
- `Notifications::NotificationItem` now gets a pointer to `timeoutTickCountEnd` and `timeoutTickCountStart`. This is needed because it needs to set those times based on when one of the buttons is pressed.
- the constructor of `NotificationItem` sets the `timeoutOnHold` flag when the notification is an incoming call
- the event handlers for the buttons on the call notification now reset the timeout, clear the `timeoutOnHold` flag and stop the vibration pattern. A helper function is used to prevent dublicating that code. 
- `Notifications::Refresh()` checks if the `timeoutOnHold` flag is set before moving the progress bar. It also includes a check to make sure the vibration pattern gets stopped before the app quits.

Caveats: 
- the motor takes a lot of energy. It might be worth thinking about reducing the duty cycle of the pattern. I choose 50% because it allowed me to use a single timer. It might also be a good idea to add a timeout after which the vibration stops.
- while working on this, the SWD interface on my devkit stopped working. Because of that, **I could not test the finished code.** In particular the sections relevant for resting the timeout when the `timeoutOnHold` flag is cleared. It looks fine to me and I didn't want to wait for a new kit to open this PR. I would appreciate if somebody could confirm that its working.


Raupinger:
https://user-images.githubusercontent.com/29229062/118059773-9c5e7d00-b391-11eb-9d95-61c65f6ead69.mp4

Video of Gadgetbridges "find me" feature. As you can see the progress bar makes a jump when the mute button is pressed. The reason is that this build doesn't reset the timeout.

Raupinger:
I found a different adapter. tested the remaining changes and everything works.

JF002:
Thanks for this PR, and sorry for the delay to review it!

It looks mostly good to me, and the vibration is a lot more perceptible when receiving a call. No one will ever miss a call with this PR!

1 small formatting detail : Could you use the **C**amelCase for the methods you created (first letter is a capital one) so they are consistent with (most of) the other ones?

Also, could you fix the conflicts with develop so I can merge it?

Thanks!

Raupinger:
ok, i changed the method names and merged current develop into the branch. didint rebase this time because i cant test it right now and the conflicts were nothing 100% trivial so i would like to keep the working history.

JF002:
Thanks for the merge. However, I think something went wrong during conflict fixing, and new changes that are not related to this PR got added (in DfuService.cpp (IsSleeping() is called 2x, once as a field, once as a method), SystemTask.h (IsSleeping was a method, it's now an atomic field), SystemTask.cpp (I think the code that actually enables the new vibrations for the call have been removed).
Could you check that out?

Raupinger:
Sure. Ill look it once im back at my desk

JF002:
Of course, take your time, we're not in a hurry ;)

src/components/motor/MotorController.cpp
```diff
@@ -13,19 +14,43 @@ MotorController::MotorController(Controllers::Settings& settingsController) : se
...
     return;
 
   nrf_gpio_pin_clear(pinMotor);
   /* Start timer for motorDuration miliseconds and timer triggers vibrate() when it finishes*/
-  app_timer_start(vibTimer, APP_TIMER_TICKS(motorDuration), NULL);
+  app_timer_start(shortVibTimer, APP_TIMER_TICKS(motorDuration), NULL);
 }
 
-void MotorController::vibrate(void* p_context) {
+void MotorController::startRunning(uint8_t motorDuration) {
```
Riksu9000, line 34: uint8_t isn't big enough

src/components/motor/MotorController.cpp
```diff
@@ -13,19 +14,43 @@ MotorController::MotorController(Controllers::Settings& settingsController) : se
...
   nrf_gpio_pin_set(pinMotor);
+  isBusy = false;
+}
+
+void MotorController::vibrate(void* p_context) {
+  if (nrf_gpio_pin_out_read(pinMotor) == 0) {
+    nrf_gpio_pin_set(pinMotor);
+  } else {
+    nrf_gpio_pin_clear(pinMotor);
+  }
```
Riksu9000, line 55: Maybe this function should play a pattern on each call. This would work for both a short and continuous notification and make them more noticable, and nicer. Just two short vibrations could work.

src/components/motor/MotorController.cpp
Riksu9000: Or maybe we don't need this parameter at all.

Raupinger: It defines the length of a the individual vibration and wait times the continuous pattern is made up of. You're right that using uint8_t limits that lenght to 256ms. Ill switch to 16 bit, but its really not that big of a deal (the only place this is currently used, the call notification sets it to 10ms). Removing this would mean hard-coding the "period" of the pattern which would mean applications cant vary their patterns.

Riksu9000: Exactly, however I think it might be better for consistency to not let applications change it. And "hard coding" isn't always a bad thing if the alternative is to re define the same thing in many places.

In one of the earlier commits, there was this line, which overflowed the parameter. I don't know if you changed this later, as currently this line doesn't exist in this PR.
`motorController.startRunning(500);`

JF002: Hardcoding this parameter is a possibility, indeed, mostly if we only use 1 single value.
If you want to future proof the class, the parameter is nice too. IMHO, the best solution for the type is not uint8_t or uint16_t or uint32_t but a stronger type that clearly represents a duration : `std::chrono::millisecond` or a custom type `Duration`, for example.

src/components/motor/MotorController.cpp
JF002: Vibration patterns would be a nice addition to this class `MotorController`. However, I think we can create a new PR to add them :)

src/components/motor/MotorController.cpp
jonvmey: Agree with using `std::chrono::milliseconds`, it's large enough and eliminate any ambiguity as to what the expected units of `motorDuration` are without needing to dig into the implementation.

src/components/motor/MotorController.cpp
jonvmey: I was looking at extending the `MotorController` interface with something like `void vibrate(std::chrono::milliseconds pulseDuration, int pulseCount = 1, std::chrono::milliseconds = some_reasonable_default)` (or maybe a 0-100 duty cycle as the last parameter, defaulted to 50). If you think something like that would be useful here I can keep working on it.

src/components/motor/MotorController.cpp
```diff
@@ -13,19 +14,43 @@ MotorController::MotorController(Controllers::Settings& settingsController) : se
...
+
+void MotorController::stopRunning() {
+  
+  app_timer_stop(longVibTimer);
   nrf_gpio_pin_set(pinMotor);
+  isBusy = false;
+}
+
+void MotorController::vibrate(void* p_context) {
+  if (nrf_gpio_pin_out_read(pinMotor) == 0) {
```
jonvmey, line 51: `nrf_gpio_pin_toggle` exists as an option here.

src/components/motor/MotorController.cpp
Riksu9000: I think that the current 50% duty cycle is too aggressive and uncomfortable for the user. We should also limit the maximum duration the motor is running, because even in Metronome we can hear it get weaker.

JF002:
@Raupinger do you plan on continuing working on this PR? Or should we focus on #544 instead? (no pressure, I'm just trying to plan my reviews of some PRs on the project :))

Raupinger:
Better to work on #544. I'll focus on other things.

PR #346, opened 2021-05-13
BugFix: Freeze `click` package (dependency of another package) version to 7.1.2 in *Dockerfile

jedi2light:
This fixes a build error caused by a newer version of the `click` package (for instance, 8.0.0)

### DIFF
--- a/docker/.gitpod.Dockerfile
+++ b/docker/.gitpod.Dockerfile
@@ -23,7 +23,7 @@ RUN apt-get update -qq \
 
 # Needs to be installed as root
 RUN pip3 install adafruit-nrfutil
-RUN pip3 install -Iv cryptography==3.3
+RUN pip3 install -Iv cryptography==3.3 click==7.1.2
 
 COPY docker/build.sh /opt/
 # Lets get each in a separate docker layer for better downloads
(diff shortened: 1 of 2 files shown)


Avamander:
A `requirements.txt` should be used for specifying the packages instead. This would also allow automated tools to notify of outdated modules in addition to being just easier to manage. 

But this PR fixes the problem for now.

JF002:
I've just rebuilt the docker image on my dev machine and encountered no error during the build of the image of of InfiniTime with this image.
Is there a specific use case affected by this issue?

jedi2light:
Can you then run the following inside of container based on that image?

```
pip3 list | grep click
```
And tell me the version of that package?
Afair, version 8.0.0 was problematic

jedi2light:
Also, here is a screenshot that shows that build error. Sorry for the low quality of the image.

image

JF002:
I've just removed all images from my system and re-build `infinitime-build` and then build InfiniTime with this new image with no error. 

Here is the version of `click':

```
[jf@jf-manjaro PineTime]$ docker run --rm -it -v $(pwd):/sources infinitime-build bash
infinitime@120a9870b6da:/$ pip3 list | grep click
DEPRECATION: The default format will switch to columns in the future. You can use --format=(legacy|columns) (or define a format=(legacy|columns) in your pip.conf under the [list] section) to disable this warning.
click (7.1.2)
```

I don't mind merging your fix, but I would like to understand what's different on your setup than on mine... Docker is supposed to give repeatable result on multiple host computer, right?

JF002:
I got no news from OP and haven't heard of any issues with the docker file since then. Feel free to re-open if necessary.

PR #347, opened 2021-05-13
Feature: Store and show git short ref on PineTime alongside other firmware information

jedi2light:
I think this feature would be useful for developers.
These three commits quite independent
 - first adds git package to docker image (for the people who use this build method)
 - second adds `git rev-parse --short HEAD => PROJECT_GIT_COMMIT_HASH` in CMakeLists.txt
 - and, finally, third adds the ability to show short ref on the SystemInfo app first screen and FirmwareValidation app screen

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,10 +55,22 @@ if(BUILD_DFU)
   set(BUILD_DFU true)
 endif()
 
+set(PROJECT_GIT_COMMIT_HASH "")
+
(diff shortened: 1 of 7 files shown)


JF002:
Nice PR, thanks! I needed that kind of features when releasing RC version of 1.0.0. I had to post the build date/time so that users had a way to check they were using the right version ;) Now, I'll be able to specify the commit hash, which will look a bit more professional :+1:

PR #354, opened 2021-05-14
Notification title fix

Raupinger:
Long notification titles now scroll. Newlines get replaced with spaces to prevent the title being draw over the notification text

### DIFF
--- a/src/displayapp/screens/Notifications.cpp
+++ b/src/displayapp/screens/Notifications.cpp
@@ -176,7 +176,17 @@ Notifications::NotificationItem::NotificationItem(const char* title,
   lv_obj_set_style_local_text_color(alert_type, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(0x888888));
   if (title == nullptr)
     title = "Notification";
+  char* pchar;
+  pchar = strchr(title, '\n');
+  while (pchar != nullptr) {
+    *pchar = ' ';
+    pchar =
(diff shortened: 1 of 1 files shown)


src/displayapp/screens/Notifications.cpp
```diff
@@ -176,7 +176,17 @@ Notifications::NotificationItem::NotificationItem(const char* title,
   lv_obj_set_style_local_text_color(alert_type, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(0x888888));
   if (title == nullptr)
     title = "Notification";
+  char* pchar;
+  pchar = strchr(title, '\n');
+  while (pchar != nullptr) {
+    *pchar = ' ';
+    pchar =
```
Avamander, line 183: Can you explain this row?

Avamander:
Also please if you can, please rebase the PR on latest `develop` keeps the commit history cleaner.

src/displayapp/screens/Notifications.cpp
Raupinger: It searches for new lines and replaces them with spaces. 
`strchr()` gives a pointer to the first occurrence of the newline or nullptr if none is found. the while loop keeps searching for newlines until none are left.

Raupinger: img

it prevents this from happening

Avamander: I meant more the assignment that seems weird and throws a warning.

Issue #344, opened 2021-05-13
Timer functionality. 


Raupinger:
Add a timer, that can be set, triggering a vibration when it runs out. A good way to put this in the UI might be with a left/right swipe on the stopwatch app.

pfeerick:
I was sure this had been mentioned in another issue, but perhaps it was only on the chat platform. 

Anyway, I would prefer this gets its own icon, rather than embedded in another app... It makes a little more sense on a phone as you have the screen realestate to have nice big buttons and link it all under a clock app, but I think single task apps suit a watch better.

JF002:
That's actually a feature I would like to see in InfiniTime in the near future :)
I agree with @pfeerick , I think this functionality deserves its own app.

Raupinger:
PR is merged, I'll close this

PR #355, opened 2021-05-16
Timer App

Raupinger:
this addresses #344

[video](https://cdn.discordapp.com/attachments/622348681538043924/843389057770061874/VID_20210516_091356-50mB.mp4)


Raupinger:
Would probably be a good idea to change the icon to something distinct, but I'll leave that up tho those who wanted the Timer to be its own App. :p

Raupinger:
Ok, I just noticed a bug while playing around. If the time gets changed while a timer is running, the timeout handler doesn't get triggered.

JF002:
Nice Timer app, thanks! Many people talked about that one, but you're the first to actually make a PR to implement it :+1:

src/components/motor/MotorController.cpp
```diff
@@ -13,7 +13,6 @@ MotorController::MotorController(Controllers::Settings& settingsController) : se
 void MotorController::Init() {
   nrf_gpio_cfg_output(pinMotor);
   nrf_gpio_pin_set(pinMotor);
-  app_timer_init();
```
JF002, line 16: 2 notes about app from the NRF SDK:
 - I tend to avoid them because they provide much more functionalities and options than what we need. Also, they will make the port to another MCU (I'm working on the BL602 on the side :) ) more difficult. It's not a big deal on the short term, and this won't prevent me from merging this branch (as we already use it for the motor controller), but on the long term, I think that would be better if we implemented timer from "scratch", probably using timers from FreeRTOS.
 -  As app_timer is already used by the motor driver, did you check these modifications have no side-effects on the vibrations and notifications?

JF002 (changes requested):
This PR contains modifications on the submodule for LVGL, is this intentional?
If not, I guess you forgot to update the submodule (`git submodule update`) on your working copy before pushing the branch.

src/components/timer/TimerController.cpp
```diff
@@ -0,0 +1,60 @@
...
+using namespace Pinetime::Controllers;
+
+
+APP_TIMER_DEF(timerAppTimer);
+
+
+TimerController::TimerController(System::SystemTask& systemTask) : systemTask {systemTask} {
+}
+
+
```
Avamander, line 19: Can you please make sure that you autoformat the code with clang-format?

(If you use CLion you can enable integration in the settings)

src/displayapp/DisplayApp.cpp
```diff
@@ -264,6 +275,10 @@ void DisplayApp::LoadApp(Apps app, DisplayApp::FullRefreshDirections direction)
         this, notificationManager, systemTask.nimble().alertService(), Screens::Notifications::Modes::Preview);
       ReturnApp(Apps::Clock, FullRefreshDirections::Up, TouchEvents::SwipeUp);
       break;
+    case Apps::Timer:
```
Avamander, line 278: Previous switch cases have braces around them, please add them here too for consistency.

src/displayapp/DisplayApp.cpp
```diff
@@ -264,6 +275,10 @@ void DisplayApp::LoadApp(Apps app, DisplayApp::FullRefreshDirections direction)
...
+      currentScreen = std::make_unique<Screens::Timer>(
```
Avamander, line 279: I think this here is misconfigured max line width.

src/displayapp/screens/Timer.cpp
```diff
@@ -0,0 +1,164 @@
...
+      }
+      if (obj == btnMinutesUp) {
+        if (minutesToSet == 59) {
+          minutesToSet = 0;
+        } else {
+          minutesToSet++;
+        }
+        lv_label_set_text_fmt(time, "%02d:%02d" , minutesToSet, secondsToSet);
+    
+      } else if (obj == btnMinutesDown) {if (minutesToSet == 0) {
```
Avamander, line 125: Missing newline I think

src/systemtask/SystemTask.cpp
```diff
@@ -228,6 +233,12 @@ void SystemTask::Work() {
           motorController.SetDuration(35);
           displayApp->PushMessage(Pinetime::Applications::Display::Messages::NewNotification);
           break;
+        case Messages::OnTimerDone:
+          if (isSleeping && !isWakingUp)
```
Avamander, line 237: Missing braces. This is a clang-tidy check you can enable. 

(If you use CLion, you can also enable this integration in the settings)

Avamander:
A few minor mistakes that would've been caught by tooling if it'd be enabled, rest seems relatively good.

src/components/motor/MotorController.cpp
Raupinger: i just moved the initialization out of the motor driver and into SystemTask

src/components/timer/TimerController.cpp
Raupinger: this part looks exactly the same after reformating

Raupinger:
I've now reformatted the code in the new files automatically and manually changed the sections @Avamander noted in other files. I've also updated the lvgl submodule.

src/displayapp/screens/Timer.cpp
Raupinger: fixed by new changes

src/displayapp/DisplayApp.cpp
Raupinger: its just the result of copying a line that was warped and editing it. fixed it now

src/displayapp/DisplayApp.cpp
Raupinger: I don't see any of the surrounding cases having braces.

src/components/motor/MotorController.cpp
Raupinger: the timer is kind of sketchy and i really don't like how I have to keep track of the time additionally for display purposes, but it's what works for now.

Raupinger:
Also note that you can't change the time while the timer is running anymore. I didn't have the patience to make that work, so i just sidestepped the problem :).  I removed the buttons when they would be inactive to clearly communicate that the time cant be changed.

Raupinger:
This PR still kind of a mess with 13 changed files and touching basically everything, but it seems kind of necessary to me if the timer is supposed to run in the background.

src/displayapp/DisplayApp.cpp
Avamander: `case Messages::TouchEvent: {`

src/components/motor/MotorController.cpp
Raupinger: moving app_timer_init() out of the motor driver isn't strictly needed, but I found it to be more logical to initialize the timer outside of it now that its no longer exclusively using that functionality.

src/components/timer/TimerController.cpp
```diff
@@ -0,0 +1,59 @@
...
+}
+
+void TimerController::StartTimer(uint32_t duration) {
+  app_timer_stop(timerAppTimer);
+  app_timer_start(timerAppTimer, APP_TIMER_TICKS(duration), this);
+  //dont ask why this random scaling factor is here. i got a difference between the time set and the time displayed and it works out to
+  //be exactly this linear factor
+  //you might be wondering why im not simply using app_timer_cnt_get() here. I am too. It is in app_timer.h, but the compiler says it
+  // doesnt exist
+  endTime = ((static_cast<float >(xTaskGetTickCount()) / static_cast<float >(configTICK_RATE_HZ)) * 1000 + APP_TIMER_TICKS(duration)) -
```
Avamander, line 32: Because one second is 1024 not 1000 ticks AFAIK

Raupinger: I mostly just stole that line from your StopWatch code. Should i change the 1000 to 1024?

Raupinger:
Another note: I find the vibration somewhat faint. It might make sense to use the continuous vibrations from #342 in the future.

src/components/timer/TimerController.cpp
Avamander: Oh whoops, sorry, I was confused there for a sec, `configTICK_RATE_HZ` makes it correct.

src/displayapp/DisplayApp.cpp
Raupinger: I see what you mean. GitHub collapses all the code between the changes. if you expand the path above this line, you see that in this particular section braces aren't used. Still kind of inconsistent. Probably a good idea to fix that in a dedicated PR.

src/components/motor/MotorController.cpp
JF002: Move app_timer_init() out of the motor driver makes sense, indeed. What I meant is that we should re-test the notification vibrations to ensure these changes did not break it : check that notifications and vibrations still works, still works after the timer was used, still works while the timer is running, ...

src/components/timer/TimerController.cpp
JF002: That's another reason why I would rather like we use custom implementation instead of blindly use these "apps" from the SDK : we use it as a black box without understanding what it does under the hood.

To me, it is needed we understand what this code does, and answer all the questions from this comment. We need to ensure we do the right thing and understand why we do it.

At first glance, app_timer uses RTC1. RTC1 is configured by FreeRTOS as its systick timer (I just hope app_timer does not alter this configuration): 1024 ticks of RTC1 correspond to 1s. That's probably why you need to apply these 1000/1024/0.024 factors in the code.

src/components/motor/MotorController.cpp
Raupinger: I don't think there's  any risk of thing breaking here since in principle there isn't any change to the motor controller, but I've tested some things anyways.

- [x] Notifications before the timer is launched once
- [x] Notifications while the timer is running
- [x] Notifications after the timer has stopped

src/components/timer/TimerController.cpp
Raupinger: the funny business here is that i have to use both freeRTOS and "apps" since I cant get the ticks from the apps timer for reasons that are beyond me. In theory  `APP_TIMER_TICKS` should give me milliseconds and I trust `(static_cast<float >(xTaskGetTickCount()) / static_cast<float >(configTICK_RATE_HZ)) * 1000)` to do the same since its already in StopWatch. Obviously one of those assumptions has to be wrong. If it helps you sleep, I can rip out the app_timer and replace it with xTimer from FreeRTOS. Maybe also open up a PR to change it in the Motor driver while I'm at it. I'm not sure I can still do that today, but there's really no need to rush anything here.

JF002: We can keep app_timer in a first iteration, I just want us to ensure that the code does what we think it's doing :) Let me have a closer look at that part...

JF002: There are 2 implementations of app_timer in the NRF SDK : app_timer.c and app_timer_freertos.c. As Infinitime uses FreeRTOS, you need to use the _freertos implementation (you'll get link error if you try to add app_timer.c in the project because both freertos and app_timer defines RTC1_Handler()). Sadly, this implementation does not provide `app_timer_cnt_get` and `app_timer_cnt_diff_compute` which would have been useful. 

`app_timer_cnt_get()` can be replaced by `xTaskGetTickCount()` as app_timer and FreeRTOS both use RTC1 as base timer.

I couldn't find an easy replacement for `app_timer_cnt_diff_compute()` but we'll need something similar the handle the counter overflow : RTC1 is a counter that is incremented 1024x per second. Once it reaches its max value (I  *think* this is a 24bit timer), it'll rollback to 0 and start counting again. We probably want to take this edge case into account.

So, RTC1 gives 1024 ticks per 1000ms, that's why we have to convert to/from ticks/ms. In order to simplify the code, I would recommend to use only 'ticks' in the internal of `TimerController` and convert them to milliseconds only in  `GetTimeRemaining()`.

Maybe you can have a look at these changes? 

```
void TimerController::StartTimer(uint32_t duration) {
  app_timer_stop(timerAppTimer);
  auto currentTicks = xTaskGetTickCount();
  app_timer_start(timerAppTimer, APP_TIMER_TICKS(duration), this);
  endTicks = currentTicks + APP_TIMER_TICKS(duration);
  timerRunning = true;
}

uint32_t TimerController::GetTimeRemaining() {
  if (!timerRunning) {
    return 0;
  }
  auto currentTicks = xTaskGetTickCount();
  auto deltaTicks = endTicks - currentTicks; // TODO Handle the counter overflow
  return (static_cast<float>(deltaTicks) / static_cast<float>(configTICK_RATE_HZ)) * 1000.0f;
}
```
I tried with a 1 minute timer and it looks like to counts and displays the time correctly.

What do you think?

Raupinger: This looks a lot more like the function in [StopWatch.ccp](https://github.com/JF002/InfiniTime/blob/develop/src/displayapp/screens/StopWatch.cpp) I originally lifted the conversion from now. Here it is 
```cpp
TickType_t calculateDelta(const TickType_t startTime, const TickType_t currentTime) {
    TickType_t delta = 0;
    // Take care of overflow
    if (startTime > currentTime) {
      delta = 0xffffffff - startTime;
      delta += (currentTime + 1);
    } else {
      delta = currentTime - startTime;
    }
    return delta;
  }
  ```
Let me quickly explain why I made the changes I made and why I was wrong.
- since I app_timer and xTaskGetTickCount() come from different libraries, I didn’t expect the underlying ticks to be the same. So I converted everything to milliseconds. Since they both use the same hardware timer configured in the same way, that assumption was wrong and things can be simplified by just using ticks.
- I thought I wouldn’t have to handle overflows since I’m only counting down. This obviously doesn’t take the potential for overflows from calculating the end time into account. 

So overall this seems like a decent way to do it. Just two more things:
Do we have to use floats here? I’d much rather not. The error we get from integer division is on the scale of microseconds which shouldn’t matter at all here. 
Secondly, this thing about the FreeRTOS implementation only implementing a subset of the API seems like an easy thing to trip over. Are there other instances of this kind of thing and is there anything we can do to make the inspector aware of it?

Comment (JF002) [reply src/components/timer/TimerController.cpp:32]: The fact that both timer_app and freertos use the same underlying hardware is a happy coincidence (well... I guess NRF designed it that way in purpose, but you cannot know it without looking at the implementation. Your first idea to convert to a common unit between both libs doesn make sense!
Using strong types here (Ticks, Milliseconds or std::chrono::millisecond) would make things a lo more clear and explicit, by the way.

Regarding floats, you are right, they are probably not needed here! and I also think it's better to avoid them when possible!

> Secondly, this thing about the FreeRTOS implementation only implementing a subset of the API seems like an easy thing to trip over. Are there other instances of this kind of thing and is there anything we can do to make the inspector aware of it?

NRF SDK provide both implementation, one that use the NRF scheduler and the other that is based on freertos. I don't know why they designed it that way. It looks like functions like `app_timer_cnt_get` and `app_timer_cnt_diff_compute` were supposed to be 'private' and should have been declared in the .c file instead of in the .h file...

Comment (Raupinger) [reply src/components/timer/TimerController.cpp:32]: @Avamander can you explain why you add 1 in this line from StopWatch?
```cpp
delta += (currentTime + 1);
```

Raupinger: Ok, changes are pushed. I'm using `TickType_t` internally for the timer controller but kept the inputs and outputs as ints so i don't have to include portmacro_cmsis.h in the app. I also did a quick sanity check to make sure that we cant have two overflows while the timer is running, which would make the remaining time ambiguous. The highest time that can be set are 59 minutes and 59 seconds, which is 3599 seconds in total. 1024*3599 is 0x383C00 which is less than (2^24-1) = 0xFFFFFF. So we're in the clear there. The time between overflows works out to be just about 4 hours and 33 minutes, so testing the overflow handling isn't really feasible. Best we can do is make sure it looks good ~~and wait for bug reports~~.

JF002: Awesome, thanks!

JF002:
The PR looks mostly good to me now. Thanks for your work, @Raupinger !

The only thing that's missing is a dedicated icon for the app (now, I understand why you suggested to extend the Stopwatch app instead of creating a new one :D). It would be nice if someone could help to find an icon and add it to the project ;)

JF002 (approved):

Raupinger:
what about this icon, maybe for the stopwatch.
Screenshot_20210519_135108

christianfl:
Nice PR! From a UI perspective I think it would make sense to keep the current stopwatch icon as it is and use this for the timer:

grafik
- hourglass-half from fontawesome, there are also other hourglass variations

The current stopwatch app literally uses the fontawesome "stopwatch" icon. The gadget everyone knows at least from some kind of school sports... Using that one for the timer and changing the stopwatch icon to some tachometer one could be confusing.

The hourglass on the other hand is THE symbol for a predefined time running up. :)

Raupinger:
Yeah, that's much better. Good find! I'll change the icon when I get back at my desk.

Raupinger:
timer Icon is a hourglass now. I asked some people and they correctly identified it as a timer.

PR #357, opened 2021-05-16
Improvements to /doc/filesInReleaseNotes.md

Itai-Nelken:


Itai-Nelken:
in line 41, is it supposed to be 'flashsed at **0x8000**' instead of 'flashed **@ 0x8000**'? I'm not sure and don't want to put a incorrect memory address.

JF002:
> in line 41, is it supposed to be 'flashsed at 0x8000' instead of 'flashed @ 0x8000'? 

For me, '@' mean 'at', but we can change it if you find it confusing.

Itai-Nelken:
I changed it

PR #358, opened 2021-05-16
Add note about getting GadgetBridge from F-Droid

cincodenada:
Hopefully this is useful and will save folks some frustration

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ As of now, here is the list of achievements of this project:
     * Firmware validation
     * System information
  - Supported by 3 companion apps (development is in progress):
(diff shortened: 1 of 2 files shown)


Avamander:
Should just mention the minimum compatible version, would be much clearer.

cincodenada:
My main problem is that the fork on the Play Store hasn't been updated since August 2020, so it seems unlikely that it will be updated. Might be helpful to mention a minimum version number too, but I think it's still useful to point people to the official version, since although PineTime users are more likely to use F-Droid, many folks will still default to the Play Store.

JF002:
Thanks! :+1:

PR #359, opened 2021-05-16
Used a macro for UUID generation, switched from C-style casts to reinterpret_cast, renamed callback

Avamander:
I tried to separate these changes a bit more, but couldn't.

The PR makes some minor improvements to the music watchapp:
1) Switches to a macro for UUID generation. I think this might save some RAM+ROM, but primarily gets rid of all these hard-to-read assignments in the constructor.
2) Switches from C-style casts to reinterpret_cast
3) To avoid a name conflict, renames MSCallback to MusicCallback, also did a minor variable inlining (that the compiler will also do above `-O0` very likely)
4) `const`-ified the value getters for more safety
5) Moved a few functions around a tiny bit to group getters

(It should remain functionally identical, at least my tests with GB didn't show any differences)

Let me know if I should strip out some of these changes. 

I would also really like to know if the macro is okay.

### DIFF
--- a/src/components/ble/MusicService.cpp
+++ b/src/components/ble/MusicService.cpp
@@ -18,117 +18,68 @@
 #include "MusicService.h"
 #include "systemtask/SystemTask.h"
 
-int MSCallback(uint16_t conn_handle, uint16_t attr_handle, struct ble_gatt_access_ctxt* ctxt, void* arg) {
-  auto musicService = static_cast<Pinetime::Controllers::MusicService*>(arg);
-  return musicService->OnCommand(conn_handle, attr_handle, ctxt);
+int MusicCallback(uint16_t conn_handle, uint16_t attr_handle, struct ble_gatt_access_ctxt* ctxt, void* arg) {
+  return static_cast<Pinetime::Controllers::MusicService*>(arg)->OnCommand(conn_handle, attr_handle, ctxt);
 }
 
 Pinetime::Controllers::MusicService::MusicService(Pinetime::System::SystemTask& system) : m_system(system) {
-  msUuid.value[14] = msId[0];
-  msUuid.value[15] = msId[1];
-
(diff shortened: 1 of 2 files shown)


JF002:
Looks good to me!

2 small details:
I *think* MusicService::characteristicDefinition could be initialized in the initialization list instead of in the core of the ctor() like it's done in other services (ex : https://github.com/JF002/InfiniTime/blob/develop/src/components/ble/DfuService.cpp#L35).

The macro is fine. But it would be even better if we find a way to write a constexpr method to replace the macro :)

Avamander:
> I think MusicService::characteristicDefinition could be initialized in the initialization list instead of in the core of the ctor() like it's done in other services (ex : https://github.com/JF002/InfiniTime/blob/develop/src/components/ble/DfuService.cpp#L35).

I'd see if it can be done in a future PR, if you wouldn't mind?

JF002:
> I'd see if it can be done in a future PR, if you wouldn't mind?

Yes, of course :)

jonvmey:
The `ble_uuid128_t` intializers can be replaced with a `constexpr` function that returns a `ble_uuid128_t` which eliminates the need for the macros. This has the additional benefit of allowing all of those members to be declared `constepxr`. I tried making it a member function function at one point but the compiler then complains about trying to call a method of a class before its definition is complete. It worked fine as a free function though.

Also the `reinterpret_cast<ble_uuid_t*>`s wouldn't be necessary if you changed `(&msSomeUuid)` to `(&msSomeUuid.u)` since `ble_uuid128_t`'s first member `u` is a `ble_uuid_t`.

I can open a new PR with these changes if you're interested. I was going to suggest them as possibilities before the PR got merged.

Avamander:
@jonvmey I would certainly like to see those changes. This PR was intended from the start to be a small incremental cleanup and I would certainly like to get rid of those `reinterpret_casts`.

PR #360, opened 2021-05-16
Turned sortById to a class method to avoid a name conflict

Avamander:
This PR turns sortById into a class method to avoid a name conflict.

There's also a dangling comment I removed, felt weird to make a separate PR for that.


JF002:
Thanks! :+1:

PR #362, opened 2021-05-17
Navigation app: reduce memory usage

JF002:
This PR slightly changes how `iconMap` is declared to reduce significantly the FLASH and RAM memory consumed by this app.

Before this change, `iconMap` was declare as a non-const member variable of Navigation. It means that this array was stored in FLASH. When Navigation was instantiated, the content of the flash was copied into RAM.
Also, `iconMap` was declared as `std::array<std::pair<std::string, const char*>, 89>`. Each item of the array was 28B long (24B for the string, 4B for the ptr to char). Using `const char*` instead of `std::string` frees 20B per element in the array.

Now, the array is declared as `static constexpr std::array<std::pair<const char*, const char*>, 86>` : 8B per element and it only lives in `.rodata` which is in FLASH. No need to copy it into RAM.

In the end, 5300B of flash were freed, and probably that much in the heap (RAM):
 - Before :  `424644	    960	  47676	 473280	  738c0	pinetime-app-1.0.0.out`
 - After    : `419344	    960	  47676	 467980	  7240c	pinetime-app-1.0.0.out`

### DIFF
--- a/src/displayapp/screens/Navigation.cpp
+++ b/src/displayapp/screens/Navigation.cpp
@@ -25,7 +25,7 @@ using namespace Pinetime::Applications::Screens;
 LV_FONT_DECLARE(lv_font_navi_80)
 
 namespace {
-  static constexpr std::array<std::pair<const char*, const char*>, 86> m_iconMap = {{
+  constexpr std::array<std::pair<const char*, const char*>, 86> m_iconMap = {{
     {"arrive-left", "\xEE\xA4\x81"},
     {"arrive-right", "\xEE\xA4\x82"},
     {"arrive-straight", "\xEE\xA4\x80"},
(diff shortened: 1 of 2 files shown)


src/displayapp/screens/Navigation.cpp
```diff
@@ -24,6 +24,106 @@ using namespace Pinetime::Applications::Screens;
 
 LV_FONT_DECLARE(lv_font_navi_80)
 
+namespace {
+  static constexpr std::array<std::pair<const char*, const char*>, 86> m_iconMap = {{
```
jonvmey, line 28: Somewhat a matter of style preferences but anything declared in an anonymous namespace is automatically `static` so the extra static here is redundant.

PR #369, opened 2021-05-19
Change Twos to make a tile with a value of 64 display a pinecone.

Raupinger:
I'm so sorry...


Raupinger:
image

Raupinger:
For those wondering if this is worth it: The resulting hex file is 270 byte larger than the current develop. Make of that what you wish.

Avamander:
Why the jetbrains font change though?

Raupinger:
I added the icon to the main font. because i figured this would be the simplest way to do it. The font generator changed a bunch of the comments to have an additional leading zero, which makes the diff look very big. the only relevant change in that file is the added character, but i didn't want to mess with auto-generated files. I can trim that down if there's actual interest in merging this

Avamander:
> I can trim that down if there's actual interest in merging this

Depends on if you want actual interest.

Raupinger:
I think its a neat idea. But I could totally understand if people think its silly. Probably a good idea to see what JF002 has to say about this.

Raupinger:
Ok, i've now reduced the diff in jetbrains_mono_bold_20. interestingly enough there were a few changed pixels in other characters. looks like the font conversion isn't perfectly repeatable

Roxxor91:
IMHO it looks nice, but it's silly. Not everyone will make the connection with pine64, even though it's seems obvious, it will certainly irritate people.
So I am for no, not that my vote would count :D Just leaving my opinion as user.

Riksu9000:
It would be more work to generate the font each time just to add this one icon that is used in one game. Even if that wasn't the case, I'm not sure this is a feature worth having.

Riksu9000:
I'll close this as I don't think we'll be adding this.

PR #372, opened 2021-05-19
fixing my boneheaded mistake

Raupinger:
thanks to @Avamander and @MysteriousLog6 for noticing this. I'm honestly surprised that code even compiled.

### DIFF
--- a/src/displayapp/screens/Notifications.cpp
+++ b/src/displayapp/screens/Notifications.cpp
@@ -180,7 +180,6 @@ Notifications::NotificationItem::NotificationItem(const char* title,
   pchar = strchr(title, '\n');
   while (pchar != nullptr) {
     *pchar = ' ';
-    pchar =
     pchar = strchr(pchar + 1, '\n');
   }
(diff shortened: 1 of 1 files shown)


MysteriousLog6:
Seems to have fixed it the compiler warning is no more.

JF002:
For my own comfort when reviewing PRs, could you be a bit more specific in the title/description?

Ok, I've had a look at the code. How come no one (me included) noticed that?

So, yes, it compiles because, for the compiler
```
    pchar =
    pchar = strchr(pchar + 1, '\n');
```
is the same than 
```
    pchar = pchar = strchr(pchar + 1, '\n');
```
which is probably optimized with
```
    pchar = strchr(pchar + 1, '\n');
```

Raupinger:
> Ok, I've had a look at the code. How come no one (me included) noticed that?

Avamander did. I just completely missed his point.

JF002:
@Avamander My apologies I missed that while reviewing the PR!

PR #382, opened 2021-05-21
Fix scrollbar on about page

Riksu9000:
This fixes the dimensions of the scrollbar on page 2 of about.


src/displayapp/screens/SystemInfo.cpp
```diff
@@ -161,7 +161,7 @@ std::unique_ptr<Screen> SystemInfo::CreateScreen2() {
                         brightnessController.ToString(),
                         resetReason);
   lv_obj_align(label, lv_scr_act(), LV_ALIGN_CENTER, 0, 0);
-  return std::unique_ptr<Screen>(new Screens::Label(1, 4, app, label));
+  return std::unique_ptr<Screen>(new Screens::Label(1, 5, app, label));
```
jonvmey, line 164: It's probably worth refactoring this constant out into a separate constant since it's repeated in several places. That would make an issue like this less likely to occur in the future.

JF002:
Thanks ! :+1:

PR #386, opened 2021-05-21
Update Amazfish.md

Roxxor91:
Add Navigation feature.

### DIFF
--- a/doc/companionapps/Amazfish.md
+++ b/doc/companionapps/Amazfish.md
@@ -8,6 +8,7 @@ The following features are implemented:
  - Time synchronization
  - Notifications
  - Music control
+ - Navigation with Puremaps
  
 ## Demo
 [This video](https://seafile.codingfield.com/f/21c5d023452740279e36/) shows how to connect to the Pinetime and control the playback of the music on the phone.


doc/companionapps/Amazfish.md
```diff
@@ -8,8 +8,9 @@ The following features are implemented:
...
  - Notifications
  - Music control
+ - Navigation with Puremaps
  
 ## Demo
 [This video](https://seafile.codingfield.com/f/21c5d023452740279e36/) shows how to connect to the Pinetime and control the playback of the music on the phone.
 Amazfish and Sailfish OS are running on the [Pinephone](https://www.pine64.org/pinephone/), another awesome device from Pine64.
- 
\ No newline at end of file
+
```
Avamander, line 16: This change is probably not necessary

Avamander (changes requested):

doc/companionapps/Amazfish.md
Roxxor91: Why? Navigation is obviously depended on the companion app. So it should be listed as a feature (or not be present in the feature list of the other companion apps).

Avamander: Make note of the line number in question, this is not about the line above.

Roxxor91: Ah ok. I created the commit via github GUI. No clue how to change that. But my commit adds one, and removes one line. Does it matter?

Avamander: > Does it matter?

It's just neater not to change lines that don't have to be changed. There's currently no specific consensus on line endings, so leaving them as-is would probably be the best course of actions.

This can't be undone using the web GUI most likely. Not sure if you're willing to learn CLI/Desktop Git just for the sake of it.

jonvmey: The missing newlines at the end of various files throughout the repo is rather annoying when I open a file, make no changes and save it (which then adds the newline) resulting in a git diff that I can't undo (without resorting to resetting the file via git).

JF002: I agree that the change about the new line at the end of the file is not needed, but that's not a big deal here :)

I also agree with

> Obviously this PR is not the correct place to fix this, but it would certainly be nicer if every file ended with a newline.

JF002:
Thanks :+1:

Issue #389, opened 2021-05-22
Add compiler flags : -fno-rtti, -fno-exceptions, -fno-non-call-exceptions,...


JF002:
I really thought these flags were already enabled but... it looks like they are not!

We are in an embedded system, we don't need some high level functionalities like exceptions and RTTI. Disabling these functionalities at build time should reduce the size of the final binary.

I think we should be able to add `-fno-exceptions -fno-non-call-exceptions -fno-rtti` to `COMMON_FLAGS` in src/CMakeLists.txt.
It might be interesting to experiment with other flags mentioned here : https://bitbashing.io/embedded-cpp.html

Avamander:
:S RTTI might exactly be the thing that I'm currently building upon. I'll investigate. The rest, possibly. 

Though lack of them does make debugging harder in any case I think.

Avamander:
This issue maybe should be about compiler flags in general?

I'd pull a few other topics from other issues to here:

`Os` has been mentioned multiple times, but I have to warn that it can turn an usable UI into an unusable one. Performance regressions are a real risk and it can't be blanket-applied. I've shot myself in the foot with this with a few other projects. So if at all possible, let's please leave stripping out code and space-optimizing compilation as the last straws.

`LTO` has also been mentioned in a few other issues. Someone even tried it, but didn't manage to achieve a fully working configuration. It really isn't trivial but the result can be great.

JF002:
> :S RTTI might exactly be the thing that I'm currently building upon

dynamic_cast?

> This issue maybe should be about compiler flags in general?

Yes, I think this is a good opportunity to investigate all compiler flags that would be useful for the project :)
And I agree with you, we won't apply flags and just be happy by the code size reduction! There's certainly good reasons to not enable these flags by default and we must be aware of potential side-effects.

`Os` (already mentionned [here](https://github.com/JF002/InfiniTime/issues/335)) and `LTO` are indeed good candidates!

I've done some quick tests: Add those flags in `src/CMakeLists.txt` at line 747 (`set(COMMON_FLAGS ...`):

 - original size : 426.512 B
 - `-fno-exceptions -fno-non-call-exceptions`: 414,264 B (-12,248 B)
 - `-fno-rtti`: 422,312 B (-4,200 B)
 - `-fno-exceptions -fno-non-call-exceptions -fno-rtti`: 409,320 B (-17,192 B)
 - `Os` for **the entire code** : 358,148B (-68,364 B)
[...]

Issue #335, opened 2021-05-07
Optimizing the build size


joaquimorg:
Using the option  "-Os" in the build reduces the size of the build.

The only problem is that it removes all the debug information, but also in builds on the master there is no need to have that information.

![build](https://user-images.githubusercontent.com/1682318/117511352-772ed080-af85-11eb-82fe-7bdd0c785615.png)

It may be an idea to change the build, since we cannot use the debug, so we could switch to this option and the debug removes the "-Os" and in the release makes the build with this option.

The idea cames from :

![discord](https://user-images.githubusercontent.com/1682318/117511688-118f1400-af86-11eb-8d95-41726109c9b0.jpg)

Avamander:
I mentioned this under some other issue, but `-Os` is not necessarily fast, hot paths should be built with `-O3`. Ideally other optimizations would be taken before losing out in speed.

JF002:
I suggest we test this in a PR in milestone 1.2 and decide if we keep it, or improve it before the release of 1.2

PR #401, opened 2021-05-25
Enable various compilation flags to reduce the binary size

JF002:
Enable the following compilation flags:

 * -fno-exceptions and -fno-non-call-exceptions : disable exception handling
 * -fno-rtti : disable run time type information (needed by dynamic_cast, for example)
 * -Os : optimize for size

These flags reduce the binary size by about 100KB!

Also, -fstack-usage generate debug info (not in final binary) to allow tools like Puncover to do a stack analysis.

#335 
#389

TODO:

- [x] Remove duplicated flags
- [x] Replace -O0 by -Og (for DEBUG builds)

### DIFF
--- a/cmake-nRF5x/CMake_nRF5x.cmake
+++ b/cmake-nRF5x/CMake_nRF5x.cmake
@@ -74,21 +74,6 @@ macro(nRF5x_setup)
                 )
     endif ()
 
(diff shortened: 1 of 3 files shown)


jonvmey:
I'd personally recommend `-Og` instead of `-O0` for debug builds. It just enables optimizations which won't make debugging harder which is a lot nicer (IMO) than `-O0`'s very unoptimized output.

JF002:
> I'd personally recommend -Og instead of -O0 for debug builds. It just enables optimizations which won't make debugging harder which is a lot nicer (IMO) than -O0's very unoptimized output.

You're probably right :+1: 
Wait... I've just tested on my working copy (cannot commit right now) and -Og generates a binary that is smaller than -O0 ! i'll probably add it to this PR :)

PR #403, opened 2021-05-25
Fix fallthrough on case statement

### LINKED ISSUE #402: Implicit case fallthrough
In [SpiMaster.cpp](../blob/develop/src/drivers/SpiMaster.cpp#L54) there is an fallthrough from a case statement to the default case which does not appear to be intentional as it won't allow the `SpiMaster` to be initialized with `BitOrder::Lsb_Msb`.

jonvmey:
Fixes (presumably) unintentional case fallthrough in `SpiMaster.cpp`.

Closes #402


JF002:
Thanks :+1:

PR #404, opened 2021-05-25
Warning cleanup

jonvmey:
Fix compiler warnings present in `pinetime-recovery` and `pinetime-recovery-loader` build targets.


JF002:
Thanks :+1:

PR #405, opened 2021-05-26
Improve stopwatch

Riksu9000:
Improve alignment
Larger buttons
Buttons on bottom
Keep buttons always visible, but disabled
initial
times


JF002:
Nice, thanks, @Riksu9000 :+1:

PR #407, opened 2021-05-27
Timer v2

### LINKED ISSUE #390: Improve setting the timer value
1. Three cheers to the timer! A great new feature!
2. Setting the time in the timer is somewhat awkward on my watch at least. I have to wait at least 1 second between presses, so it takes about 30 seconds to set the timer to 30 seconds. One proposal would be to keep incrementing/decrementing the value when one of the + or - buttons is held. Another would be to set the time faster by swiping up and down.

### LINKED ISSUE #391: Timer ending is not obvious enough
From what I understand, when a timer runs out, it vibrates once and then resets to 00:00.

- It should vibrate repeatedly until dismissed so you don't ignore it and have your water boil over or similar (you may be timing something very urgent)
- There should be some sort of message that your timer expired so if you look at it a minute late you remember what happened

I also think it should not clear the timer value, but maybe that would work best after implementing my idea (https://github.com/JF002/InfiniTime/issues/390#issuecomment-846558561) to save timer presets. It would then show something like 05:00 after your 5 minute timer ended, and you would press the back button to get to a screen that lists all your timers, similar to the Pebble timer app.

Raupinger:
making some requested changes to the timer. resolves #390.
- [x] Presets
- [x] +/- 10 seconds by swiping vertically
- [ ] longer vibrations (depends on #342). this would resolve #391.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -404,6 +404,7 @@ list(APPEND SOURCE_FILES
         displayapp/screens/HeartRate.cpp
         displayapp/screens/Motion.cpp
         displayapp/screens/FlashLight.cpp
+        displayapp/screens/settings/SettingsList.cpp
         displayapp/screens/List.cpp
         displayapp/screens/BatteryInfo.cpp
         displayapp/screens/Steps.cpp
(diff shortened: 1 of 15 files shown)


Riksu9000:
The situation has changed with #492 getting merged. Now it can detect holding down a button and tapping is much faster as well. I think at least the swipe part should be taken out, as it isn't very intuitive and overrides the exit gesture.

Riksu9000:
@Raupinger Will you be updating this PR, or should this be closed in favor of a newer PR, if there still is an issue?

Riksu9000:
I'm closing this due to being outdated and inactive.

PR #409, opened 2021-06-02
add basic metronome app

### LINKED ISSUE #361: Metronome app?
I know the pinetime my not be designed for this but can it act as a metronome?
For example an app that detects how fast you tap on the screen and starts vibrating with short pulses in the same rate. Or set specific bpm to pulse in.
Kind of like the [soundbrenner](https://www.soundbrenner.com/).

hall:
A simple app that triggers a vibration based on the user-selected beats per minute (closes #361).

screenshot

What's done:

  - [x] drag/click arc to select bpm
  - [x] tap to select bpm
  - [x] set bpb (beats per bar, for creating an accent beat)

### DIFF
--- a/src/displayapp/screens/Metronome.cpp
+++ b/src/displayapp/screens/Metronome.cpp
@@ -123,7 +123,7 @@ void Metronome::OnEvent(lv_obj_t* obj, lv_event_t event) {
       }
       break;
     }
-    case LV_EVENT_CLICKED: {
+    case LV_EVENT_PRESSED: {
       if (obj == bpmTap) {
         float timeDelta = calculateDelta(tappedTime, xTaskGetTickCount());
         if (tappedTime == 0 || timeDelta > 3) {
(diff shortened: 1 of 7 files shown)


hall:
Despite setting
```cpp
lv_arc_set_adjustable(bpmArc, true);
```
the arc doesn't show the knob (tried adding styles as seen in the diff as well) and cannot be clicked or dragged to a new value.

hall:
Fixed arc selection by enabling the polling touch mode.
```cpp
app->SetTouchMode(DisplayApp::TouchModes::Polling);
```

src/displayapp/screens/Metronome.cpp
```diff
@@ -0,0 +1,168 @@
...
+  lv_obj_set_event_cb(playPause, eventHandler);
+  lv_obj_align(playPause, lv_scr_act(), LV_ALIGN_IN_BOTTOM_RIGHT, -15, -10);
+  lv_obj_set_height(playPause, 39);
+  playPauseLabel = lv_label_create(playPause, nullptr);
+  lv_label_set_text(playPauseLabel, Symbols::play);
+
+  app->SetTouchMode(DisplayApp::TouchModes::Polling);
+}
+
+Metronome::~Metronome() {
```
GatewayBit, line 90: I would recommend setting
`systemTask.PushMessage(Pinetime::System::SystemTask::Messages::EnableSleeping);` here since it's modified on lines 153 and 158.

GatewayBit (changes requested):
Might be good to re-enable sleep state in destructor.

src/displayapp/screens/Metronome.cpp
hall: Good catch! I've added it now.

JF002:
@hall Nice app! As a (very amateur) drummer, I appreciate it!

Next step: find a nice icon for the app. 
I'll merge this PR, but feel free to update it with an icon if you find one ;)

PR #356, opened 2021-05-16
minor changes regarding building

afeuerstein:
I removed a check in `./CMakeLists.txt` that forced setting the `USE_JLINK`-variable when no variables regarding the use of debug and flashing hardware were set, and I fixed a typo in `doc/buildAndProgram.md` in the `BUILD_DFU`-example.

PR #412, opened 2021-06-04
Allow DFU only builds

MFAshby:
Adds a check so that USE_JLINK, USE_GDB_CLIENT and USE_OPENOCD can all be disabled if BUILD_DFU=1. This allows DFU only builds.

Fixes https://github.com/JF002/InfiniTime/issues/384

Tested by running: 
```bash
mkdir build
cd build
cmake .. -DBUILD_DFU=1 -DNRF5_SDK_PATH=/home/martin/dev/nRF5_SDK_15.3.0_59ac345 -DARM_NONE_EABI_TOOLCHAIN_PATH=/home/martin/dev/gcc-arm-none-eabi-9-2020-q2-update
make -j4
```
And checking that a DFU file is produced at build/src/pinetime-mcuboot-app-dfu-1.1.0.zip.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,7 @@ if (NOT NRF5_SDK_PATH)
   message(FATAL_ERROR "The path to the NRF52 SDK must be specified on the command line (add -DNRF5_SDK_PATH=<path>")
 endif ()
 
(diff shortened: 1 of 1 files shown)


JF002:
Similar changes have already been implemented by #356, I think we can close this one. Thanks for your contribution, though!

Issue #327, opened 2021-05-03
Freeze/crash in About menu since 9f15fd1


nlfx:
I've had frequent crashes since yesterday merges. Here is how to reliably reproduce the issue:
- Open and close About menu twice
- Open About menu a third time
- Swipe up to try to access screen 2
- Watch freezes and is reset by watchdog a few seconds later

https://user-images.githubusercontent.com/69880003/116891847-b9bb7a80-ac1e-11eb-9d82-254a69eaa988.mp4

I've bisected the problem to commit 9f15fd1820e620e42fb717fb5d692d44a3f0f70d. However given its content, I don't understand how it could create this crash. Maybe it's a problem with some completely unrelated part of the code which only becomes visible after this commit due to some timing or memory position constraint?

Could someone with a dev kit and SWD access please see if they manage to get more info about what is creating the above crash?

In case this bug is memory location dependent and strings with file path in the firmware change the behavior, here is [my version of the above commit](https://github.com/JF002/InfiniTime/files/6415662/infinitime-about-menu-crash.zip).

Thanks a lot for your help!

MysteriousLog6:
What exactly have you changed in the commit?
My first response too was "memory allocation maybe?" but there seems to be enough memory for doing a simple vibration.
It might be something else triggering this maybe the driver?

nlfx:
> What exactly have you changed in the commit?

No change, I just did a
```
git checkout 9f15fd1820e620e42fb717fb5d692d44a3f0f70d
git submodule update
```
compiled it and flashed it.

> My first response too was "memory allocation maybe?" but there seems to be enough memory for doing a simple vibration.

What is weird, is that the the crash happens on the About menu ([SystemInfo.cpp](https://github.com/JF002/InfiniTime/blob/9f15fd1820e620e42fb717fb5d692d44a3f0f70d/src/displayapp/screens/SystemInfo.cpp)) which has nothing to do with the motor controller change in this commit.

So indeed, I also initially though that the problem might be that the memory for the stack was at its limit and the additional `motorController` argument passed down the call stack was enough to start overwriting other memory region. But the fact that the crash happen only when accessing the About menu for the 3rd time would make no sense if that was the real problem. Unless I'm missing something?

[...]

PR #415, opened 2021-06-06
Move dynamically allocated variables to static variables.

JF002:
Initialize SystemTask, DisplayApp and HeartRateTask as global static variable instead of variables on the heap. We don't need them on the heap as we know their size at build time, it'll reduce memory fragmentation and it'll make memory analysis easier.

Heap usage is reduced from >6KB to ~2KB. Most of this memory is allocated by NimBLE.

Global heap and stack have also been reduced (4KB heap, 1KB stack)

- [x] Test, test, test and check that heap and stack do not overflow.


JF002:
Heap usage is reduced from >6KB to ~2KB. Most of this memory is allocated by NimBLE.

JF002:
I've just pushed new changes that should fix https://github.com/JF002/InfiniTime/issues/327. I would be grateful if some of you could test the branch and confirm it's working fine :)

kieranc:
> I've just pushed new changes that should fix #327. I would be grateful if some of you could test the branch and confirm it's working fine :)

I have compiled this branch and not seen any issues

JF002:
> I think these changes might be vulnerable to [Static Initialization Order Fiasco](https://en.cppreference.com/w/cpp/language/siof) issues.
> 
> `DisplayApp`'s constructor loads the `Clock` app, which itself accesses the static `SettingsController` which may or may not be constructed before `DisplayApp`. There might be more instances like this but that was just the first potential one I found.

I thought that the compiler would construct the static variables in the order of declaration. In this case, `settingsController` is declared before `displayApp` so it should be OK. But am I a bit naive here?

Anyway, I've just checked with the debugger and SettingsController is init'd before DisplayApp. One way to ensure we are not vulnerable to this SIOF is to call the first LoadApp in DisplayApp::Start() instead of in the ctor(). This way, we'll be sure everything is constructed in time.

PR #420, opened 2021-06-07
update main font to keep diffs for future changes small

Raupinger:
the c file generated from following the instructions in  [README.md](../tree/develop/src/displayapp/fonts/README.md) currently has some large but inconsequential differences to the one in the source. Update it.


JF002:
Which font files did you use for this PR? Using the files I have locally, the generated file is also different from the on in the project, but also different from yours.

Could you add the font files (ttf/woff/...) to the project so we ensure everyone is using the same font file?

Raupinger:
Its 2.2.5 downloaded a few weeks ago. I'll add that file. Should we maybe use the local generator so we don't end up having to do something like this again?

JF002:
> Should we maybe use the local generator so we don't end up having to do something like this again?

Probably :+1:  Is it a simple script we can add in the project ?

Raupinger:
> > Should we maybe use the local generator so we don't end up having to do something like this again?
> 
> Probably +1 Is it a simple script we can add in the project ?

its an npm package. I'd like to avoid making people install a bunch of node stuff.

JF002:
> its an npm package. I'd like to avoid making people install a bunch of node stuff.

Maybe https://github.com/JF002/InfiniTime/pull/439/files is a nice alternative?

Thanks for the file, I'll have a look!

JF002:
Looks good to me :+1: ! Feel free to do the same for the other fonts :)

PR #423, opened 2021-06-09
Improve when to display which battery-icon for the current battery percentage

hassless:
We have 5 different icons to represent the current battery percentage. At any time, we want to display the icon that best represents the actual current battery percentage.

For instance, we would ideally want to display the Symbols::batteryThreeQuarter icon when the battery percentage is between 62.5% and 87.5%. Any lower, then the 50% (Symbols::batteryHalf) icon becomes more appropriate. Any higher, then the 100% (Symbols::batteryFull) icon is most appropriate.

This PR changes the selection of the battery icon from the current battery percentage accordingly. The table below illustrates the new ranges.
```
Icon                          Perfectly represents   Used for battery % in range

Symbols::batteryEmpty                 0%                     0.0% -  12.5%
Symbols::batteryOneQuarter           25%                    12.5% -  37.5%
Symbols::batteryHalf                 50%                    37.5% -  62.5%
Symbols::batteryThreeQuarter         75%                    62.5% -  87.5%
Symbols::batteryFull                100%                    87.5% - 100.0%
```


Avamander:
This is unfortunately fixing the symptom, not the cause. 

The original battery percentage value should be calculated non-linear from voltage instead.

hassless:
I fully agree the battery percentage should be calculated non-linear from voltage. However, this PR is about mapping a given percentage (which is off right now, I agree) to the icon most appropriate for that percentage. Why not use the icon that best represents our current best-effort battery percentage, even if that battery percentage is not (yet) as accurate as we would like it to be?

Avamander:
> Why not use the icon that best represents our current best-effort battery percentage, even if that battery percentage is not (yet) as accurate as we would like it to be?

Because it's hiding the problem instead of fixing it. Has to be reverted the moment the actual issue is fixed.

hassless:
Thanks for explaining! Unfortunately, I don't think I follow. The function I modified simply takes the current battery percentage as input. It does not try to calculate, or change the estimate of, the current battery percentage.

This PR has the following effect:
"Given an inaccurate battery percentage, select the icon that most accurately represents that (inaccurate) battery percentage."

As soon as BatteryController is improved to calculate an accurate battery percentage, the way you suggested, then this PR has the same effect, namely:
"Given an accurate battery percentage, select the icon that most accurately represents that (accurate) battery percentage."

I don't see how this is hiding the fact that the battery percentage is inaccurate at the moment. Am I missing something?

Avamander:
> I don't see how this is hiding the fact that the battery percentage is inaccurate at the moment

What is it doing then?

> As soon as BatteryController is improved to calculate an accurate battery percentage...

Then it has the opposite effect - the displayed fractional battery UI won't display the left capacity of the battery accurately.

The PR solves the UI issue from the wrong end, but in the end it's up to JF to decide if it is worth it.

jonvmey:
There are two separate issues at play.

First the battery percentage is not being calculated in an ideal way. This PR does nothing to address that and isn't trying to.

Second the icon that's displayed on the UI for a given percentage (regardless of the correctness of that percentage to begin with) could use a slightly different mapping that @hassless feels is more accurate. That's what this PR is addressing.

Avamander:
The two issues are the one and same, the latter wouldn't "feel more accurate" if the former didn't exist.

hassless:
As an example, the old implementation would change the icon from batteryHalf to batteryOneQuarter the moment the battery percentage decreases from 50% to 49%. That did not feel accurate to me, considering that 49% is way closer to half-full than it is to one-quarter-full. That is why this new implementation retains the batteryHalf icon longer and only changes to the batteryOneQuarter icon when the battery percentage drops to 37%. At that point, the battery percentage is actually closer to one-quarter-full than it is to half-full, so I would say that is good point to switch to the batteryOneQuarter icon.

jonvmey:
Take the battery out of the equation entirely since it's irrelevant to this change. Suppose you were writing a function to convert a percentage value into an enum with states for 0%, 25%, 50%, 75% and 100%. The conversion that is most accurate will be the one that minimizes the total error through the range of inputs.

Now replace enum values with icons representing the same things and that's the change that @hassless is making here.

Riksu9000:
It took me a bit to understand this as well, so here is how I picture it. This PR simply spreads the battery icons across the percentage range evenly.

It can also be thought of as the accuracy being +25% before, whereas now the accuracy is +-13%. This is a good change.

difference

JF002:
I paid more attention to battery level recently, and I noticed that the battery would be displayed as "half full/half empty" where the actual battery level as 73%.
I agree that battery level measurement is not accurate for now, but even if it was, 73% being displayed as 50% does not seem right.

If I understand correctly, this PR better spreads the battery icons across the % range, and I think it's a good idea! It doesn't change the accuracy of the measurement, but it might look better for the end-user.

I'll flash this branch on a devkit with a battery to see how it looks on a real device. 

Thanks @Riksu9000  for your drawing!

Avamander:
In theory this is good, but it doesn't yield the "max 13% error" in reality and can have even worse locations because of the incorrect curve behind it.

For example with the new scheme the watch can display two bars at 10% battery (the end of 1 bar is at approx. at 10% charge) and 60% of battery!

quick doodle

Riksu9000:
> In theory this is good, but it doesn't yield the "max 13% error" in reality and can have even worse locations because of the incorrect curve behind it.
> 
> For example with the new scheme the watch can display two bars at 10% battery and 60% of battery!
> 
> quick doodle

But now you're bringing the voltage to the equation unnecessarily. The voltage conversion to percentage is separate from percentage conversion to symbol. Currently both of these conversions are wrong. This PR fixes one of those conversion. Even if there are worse locations, we would still be closer to the truth.

Avamander:
> But now you're bringing the voltage to the equation unnecessarily. 

It's **very** necessary if the goal is an accurate battery gauge from voltage.

jonvmey:
I think that the problem is that battery voltage non-linearity concerns are sneaking themselves into code which shouldn't care about them. The non-linearity definitely needs to be accounted for somewhere but this function is not the correct place. The correct place would be in the code that converts the measured battery voltage into a percentage capacity (ie. like what #443 attempts to do). But that's just my opinion.

JF002:
Let's say we are able to measure every bit of energy available in the battery with a perfect accuracy, we would still display (with the current implementation)
 - 3/4 battery at 76%
 - 1/2 battery at 74%

or
 - 1/2 battery at 51%
 - 1/4 battery at 49%

Which, even if it's mathematically correct, looks a bit pessimistic from the user point of view : there's very little difference between 51 and 49%, but the user will interpret this as if the battery was nearly empty while there's still nearly 50% of capacity available.

I think this PR is a UX/UI improvement : the battery will be displayed as half-full when the level is around 50%, a bit less, a bit more. So, instead of being **always** pessimistic (it displays a battery that is at most 25% lower than in reality), the display will be *sometimes* optimistic and *sometimes* pessimistic, with "only" 12.5% of max error.

Avamander:
@JF002 that is assuming the precentage calculation under it is correct, it is not. With this PR the max error is going to be 40%! too optimistic. It'll end up with watches dying unexpectedly, much more so on with older batteries or in the cold. Some people will have their watch die shortly after it displays two bars.

I'm not opposed to improving it, not at all, but this can't be done before fixing the underlying values used.

jonvmey:
Why not as a hold-over make this change to the icon function, and also add a conversion in the battery controller code that maps the voltage-to-percentage following the old values here? Then this code ends up improved and the overall voltage-to-icon mapping remains the same until battery measurement is improved.

Riksu9000:
>Some people will have their watch die shortly after it displays two bars.

Fair enough. Previously you were saying that this change was hiding the issue, which confused me because I think this PR is actually uncovering it.

I think there's a slight error in the graph, because in InfiniTime 3.2V is considered 0% rather than 3.1V, but this doesn't affect the results much.

The real question is do people rely on the battery icon? It doesn't provide much accuracy so I for one always check the percentage in BatteryInfo instead, which is unaffected by this PR.

JF002:
@Avamander As I said, I try to distringuish between the display of the level and its measurement. If we take them separately, I think the PR makes sense. But the side-effect you mentioned is interesting : as the discharge curve is not linear, the battery discharges faster from 40 to 0%, so that users won't have a lot of time to see the empty icon...

There are 2 other PRs related to the battery level : #483 and #443. Do you think they'll improve the measurement?

Avamander:
> There are 2 other PRs related to the battery level : #483 and #443. Do you think they'll improve the measurement?

Very likely.

JF002:
Here is the DFU i've just flashed on my unit. It's based on develop with #423, #443 and #483. I'll use it in the next few days and I'll try to keep it connected to gadgetbridge as much as possible!

[pinetime-mcuboot-app-dfu-1.3.99.zip](https://github.com/JF002/InfiniTime/files/6874102/pinetime-mcuboot-app-dfu-1.3.99.zip)

JF002:
Here's a new DFU, I fixed the issue : voltage was expressed in mV, where the LUT expects "centi-volts (?)" : 
[pinetime-mcuboot-app-dfu-1.3.99.zip](https://github.com/JF002/InfiniTime/files/6883960/pinetime-mcuboot-app-dfu-1.3.99.zip)

I ran it yesterday evening on my sealed unit, here's a screenshot of gadgetbridge : 
Screenshot_20210727-103058
The signal looks a bit noisy, with spikes at ~50% while the average values is around 80-90%. It may be caused by #483 ?

Riksu9000:
I wonder if there's an issue with the battery percentage notification, because smaller spikes are present in the 1.3.0 release and can be seen in your screenshot from the release as well.
I don't think it's just the voltage dropping, because if that were the case, the voltage should be higher during the night.
I also don't think the ADC can measure a significantly higher voltage due to noise because the ADC takes a sample of the voltage, and the sample can't be higher than the source voltage.
spikes

JF002:
Yes, it might be a bug in READ or NOTIFY functions of the battery service. I installed the new version of Gadgetbridge (I was using a custom build I did before the release) and lost the history... I re-charged my watch and I'll post the discharge curve in a few days ;)

JF002:
Here's the graph from gadgetbridge from my branch that merges  #423, #443 and #483. I worked a lot around the house so my pinetime was not connected as much as usual, but the discharge curve looks pretty good to me (except those strange spikes).

Screenshot_20210731-150732

I think we can fix the conflicts in those branches and merge them. We'll be able to continue to improve the feature later on if necessary :)

JF002:
What do everyone think about merging those 3 PRs about the battery level meaurement?

Riksu9000:
As I've been writing in #443, I think it requires more work.

I think a better stopgap would be to merge #423 and #483, and instead of #443, map the percentage linearly, but consider 3.5V as 0%. It's much closer to the real curve and discourages overdischarging which can damage the battery.

graph

PR #426, opened 2021-06-11
Fix TouchEvent::Tap

### LINKED ISSUE #418: Setting ReturnApp TouchEvents to Tap causes weird behaviour
If a SwipeDown event was changed to a Tap event in DisplayApp.cpp, the tap would "fall through" to the previous app. I found this problem when I tried to make the FlashLight app close on tap.

In this video
`ReturnApp(Apps::Settings, FullRefreshDirections::Down, TouchEvents::SwipeDown);`
was changed to
`ReturnApp(Apps::Settings, FullRefreshDirections::Down, TouchEvents::Tap);`
on line 314

https://user-images.githubusercontent.com/37774658/121019893-82edfc80-c7a8-11eb-892a-5bb6d97bf39a.mp4

Riksu9000:
Don't run `lvg.SetNewTapEvent` if the tap was processed elsewhere.
This fixes #418
Now `OnTouchEvent(TouchEvents event)` must return false if TouchEvent::Tap wasn't processed. QuickSettings and Flashlight misused this and they have been fixed.


JF002:
I've just merged this branch. By doing so, I fixed the conflicts and I also took the liberty to rename `DisplayApp::OnTouchEven()` to `Convert()` and move it to the anonymous namespace as it's now only converting a struct to another.

Thanks!

PR #428, opened 2021-06-13
Music service cleanup

jonvmey:
@Avamander here are the changes I mentioned in #359 plus one or two other small improvements.

1. Eliminated the need for `reinterpret_cast<ble_uuid_t*>`ing by passing pointers to the first member of the UUID structs, which is a `ble_uuid_t`, instead of the struct itself.
2. Replaced the macro definitions for the UUID values with `constexpr` functions that return UUID objects directly and used those to initialize them. Also moved these variables to static values in the source file since that's the only place they're referenced.
3. Moved `MusicCallback` into the anonymous namespace since it's also only referenced inside of `MusicService.cpp`
4. Moved initialization of member variables with fixed constants to class declaration in the header and added explicit initializers for some uninitialized members.

Another thing I was thinking of doing was adding a simple `operator==` for the UUID objects, which would just call `ble_uuid_cmp`. That would help make the if-chain in `MusicService::OnCommand` a lot cleaner to read.

### DIFF
--- a/src/components/ble/MusicService.cpp
+++ b/src/components/ble/MusicService.cpp
@@ -111,16 +111,6 @@ Pinetime::Controllers::MusicService::MusicService(Pinetime::System::SystemTask& 
   serviceDefinition[0] = {
     .type = BLE_GATT_SVC_TYPE_PRIMARY, .uuid = &msUuid.u, .characteristics = characteristicDefinition};
   serviceDefinition[1] = {0};
-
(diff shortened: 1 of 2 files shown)


Avamander:
I plan on testing this, but not sure if I'll manage it within this week.

JF002:
Nice :+1:  I like the way you replaced the macros by `constexpr` functions!

PR #430, opened 2021-06-13
Fix typo in variable names

jonvmey:


MysteriousLog6 (commented):
Yep Seems good , will make reading and understanding the code simpler in the future.

Will compile in a few minutes to see if it works.

JF002:
Thanks :+1:

PR #432, opened 2021-06-14
Improve stopwatch

### LINKED ISSUE #379: Stopwatch improvements
When using the stopwatch app for an "extended" amount of time, e.g. 1-2 minutes and wanting to do more accurate measurements there by hitting the stop button in time, its annoying to have the display turn off at its set timeout, meaning it needs to be turned on again first. So I'd suggest to just keep the display on while the app is in use, perhaps dimming it to the lowest brightness setting instead of turning off after the timeout.

Then, despite being in conflict with #303, it would be nice to repurpose the physical button to stop the measurement (and once stopped, use it to exit the app again), since its kinda impossible to use the stopwatch for diving for example (to see how long one can hold their breath). To solve this conflict, it may be a good idea to make the button behaviour a setting (either use the button to stop the running measurement, or use the button to exit the app while the measurement goes on).
To get back to the diving example, starting a new measurement with wet display and hands is another problem, but maybe someone has an idea regarding this .

Riksu9000:
Cleaned up the messy code.
While the stopwatch is running, sleep is disabled.
While the stopwatch is running, the button will stop the timer. In all other cases the button will exit the app. 
Fixes #379?

EDIT: Forgot to mention that removing the Events code also improves the responsiveness of the app, since it no longer waits for a screen refresh cycle.


Avamander:
Did you run Clang-Format on the new code? I'm also not sure getting rid of the state machine made the functionality clearer.

Riksu9000:
The state machine was inside the Refresh loop, which means it kept setting eveything multiple times per second. Now changes are made only when a button is pressed, and I think it is much easier to see what pressing a button does, whereas before some of the changes were in the button function, and some somewhere in the state machine.

Clang-format changes some lines so I'll commit that soon.

MysteriousLog6:
Seems to be using an older base?

JF002:
Looks good to me !
I agree with @Riksu9000 : The new code update what needs to be updated on response to an event instead of refreshing everything on each call to Refresh().

Issue #433, opened 2021-06-14
Fully charged text broken in battery option in settings


Itai-Nelken:
![20210614_192313.jpg](https://user-images.githubusercontent.com/70802936/121926788-ff538300-cd46-11eb-91ee-ad1a9be9a0e5.jpg)
![20210614_192327.jpg](https://user-images.githubusercontent.com/70802936/121926829-09758180-cd47-11eb-80f1-2644415e8d6d.jpg)

Itai-Nelken:
forgot to add - I'm on InfiniTime 1.1.0.

kieranc:
@JF002 This issues is resolved and can be closed.

PR #437, opened 2021-06-16
Modify status text in BatteryInfo so it fits on screen

kieranc:
See issue #433 
I've changed 'Battery is fully charged' to simply 'Battery charged' so it fits on screen. I also modified 'Battery is low' to 'Battery low' for consistency.

'Battery fully charged' would have been preferable IMO, but it doesn't fit. If anyone has a strong opinion on the matter I could make the text scroll instead, but for a small overflow it seems better to do it this way.


Itai-Nelken:
better solution than mine

JF002:
> 'Battery fully charged' would have been preferable IMO, but it doesn't fit. 

Maybe it could fit on 2 lines?

kieranc:
> > 'Battery fully charged' would have been preferable IMO, but it doesn't fit.
> 
> Maybe it could fit on 2 lines?

That might not fit between the charging bar and the voltage text below. I'm happy with this as a solution, but I can try it over 2 lines if you prefer.

JF002:
> That might not fit between the charging bar and the voltage text below, it'll be cramped if it does. I'm happy with this as a solution, but I can try it over 2 lines if you prefer.

Both solutions are ok for me, let's choose the one that looks best :)

kieranc:
batteryinfo1
batteryinfo2

Here are the 2 options in the simulator, colours are incorrect but the alignment is accurate. I.... don't really have strong feelings either way? Which do you prefer?

JF002:
That's a great use of the simulator, thanks for trying both options! I personally  like the 2nd one (2 lines). 
Does anyone else have an opinion about this?

kieranc:
> That's a great use of the simulator, thanks for trying both options! I personally like the 2nd one (2 lines).
> Does anyone else have an opinion about this?

Updated to the 2 line option - the only issue in my eyes is that all other messages are now a single line and fully charged is 2.

Itai-Nelken:
The single line looks better in my opinion

Itai-Nelken:
The single line looks better in my opinion, but the text in it isn't good. `battery charged` doesn't sound good to me, maybe simply put `charged` or `fully charged`?

kieranc:
I quite like the 'Fully charged' concept since 'Battery' is implied, given we're in the battery app, an there's nothing else on the device which can be charged. We have 4 text elements to define, so:

1, Current, Battery is fully charged doesn't fit on one line.
- Battery charging
- Battery is fully charged
- Battery is low
- Battery discharging

2, Not keen on the similarity between charged and charging
- Battery charging
- Battery charged
- Battery low
- Battery discharging

3, Too short
- Charging
- Fully charged
- Low
- Discharging

4, Also a bit short
- Charging
- Fully charged
- Battery low
- Discharging

5, Feels pretty good to me
- Battery charging
- Fully charged
- Battery low
- Battery discharging


I think I'd prefer to keep it all on one line, or have them all on 2 lines, and I don't think there's enough text to justify 2 lines.

Riksu9000:
I like the 4th option. I don't think it's too short, unless it looks weird in app? Maybe it could say "Almost empty" for battery low.

kieranc:
> I like the 4th option. I don't think it's too short, unless it looks weird in app? Maybe it could say "Almost empty" for battery low.

I kind of thought the same thing after my last comment... here's how it looks:

batopt4s1
batopt4s2
batopt4s3
batopt4s4

JF002:
If everyone agree with this last suggestion, let's do it ;)

kieranc:
Sounds good to me, ready to merge I think.

JF002:
Great, thanks!

PR #438, opened 2021-06-17
Using littlefs

joaquimorg:
This PR adds support for littlefs.

- New submodule littlefs
- Functions to read and write files in external flash
- Settings are now saved to a file in external flash
- Functions for LVGL to read from external flash


MysteriousLog6:
Size that this adds?

Avamander:
> Size that this adds?

Worth it. :D

src/components/fs/FS.cpp
```diff
@@ -0,0 +1,194 @@
...
+#endif
+
+}
+
+void FS::VerifyResource() {
+    // validate the resource metadata
+    resourcesValid = true;
+}
+
+void FS::FileOpen(lfs_file_t* file_p, const char* fileName, const int flags) {
```
Riksu9000, line 106: Shouldn't these functions return the error code that the lfs functions might return?

JF002:
Thanks @joaquimorg ! That's one PR I was waiting for :)

I'll have to delay my review a bit because I need to work on adding the support for the replacement motion sensor, but i'll revew it asap!

src/components/fs/FS.cpp
```diff
@@ -0,0 +1,194 @@
...
+        Pinetime::Controllers::FS& lfs = *(reinterpret_cast<Pinetime::Controllers::FS*>(c->context));
+        const size_t address = lfs.mStartAddress + (block * BLOCK_SIZE_BYTES);
+        lfs.mDriver.SectorErase(address);
+        return lfs.mDriver.EraseFailed() ? -1u : 0u;
+    }
+
+    static int sync(const struct lfs_config *c) {
+        // no hardware caching used
+        return 0u;
+    }
```
jonvmey, line 45: Returning unsigned literals in functions returning `int` looks a little weird.

src/components/fs/FS.cpp
```diff
@@ -0,0 +1,194 @@
...
+    .block_size = BLOCK_SIZE_BYTES,
+    .block_cycles = 1000u,
+
+    .cache_size = 16,
+    .lookahead_size = 16,
+
+    .name_max = 50,
+    .attr_max = 50
+
+};
```
jonvmey, line 65: All of the `static` things could be grouped together inside of the anonymous namespace. Also inside of the anonymous namespace you don't have to add `static` again.

src/components/fs/FS.cpp
```diff
@@ -0,0 +1,194 @@
...
+    static int prog(const struct lfs_config *c, lfs_block_t block,
+            lfs_off_t off, const void *buffer, lfs_size_t size) {
+        Pinetime::Controllers::FS& lfs = *(reinterpret_cast<Pinetime::Controllers::FS*>(c->context));
+        const size_t address = lfs.mStartAddress + (block * BLOCK_SIZE_BYTES) + off;
+        lfs.mDriver.Write(address, (uint8_t*)buffer, size);
+        return lfs.mDriver.ProgramFailed() ? -1u : 0u;
+    }
+
+    static int erase(const struct lfs_config *c, lfs_block_t block) {
+        Pinetime::Controllers::FS& lfs = *(reinterpret_cast<Pinetime::Controllers::FS*>(c->context));
```
jonvmey, line 36: I believe that these `reinterpret_cast`s only need to be `static_cast`s.

src/components/fs/FS.cpp
joaquimorg: Thank you, I will review the recommendations.

src/components/fs/FS.cpp
joaquimorg: Thank you, I will review the recommendations.

src/components/fs/FS.cpp
joaquimorg: Thank you, I will review the recommendations.

MysteriousLog6:
During compile I got this error:

/home/mysteriouslog6/InfiniTime2/src/libs/littlefs/lfs.c:429:13: warning: 'lfs_mlist_isopen' defined but not used [-Wunused-function]
  429 | static bool lfs_mlist_isopen(struct lfs_mlist *head,

JF002:
> During compile I got this error:
> 
> /home/mysteriouslog6/InfiniTime2/src/libs/littlefs/lfs.c:429:13: warning: 'lfs_mlist_isopen' defined but not used [-Wunused-function]
> 429 | static bool lfs_mlist_isopen(struct lfs_mlist *head,

I have the same warning. It's not a big deal, as it's just a method that is defined but not used. This warning is generated by a source file from littlefs, so I think we can ignore it. But it would be nice if we could find a way to prevent this warning from happening using some cmake trickery, for example. Is it possible to disable specific warning on specific target/lib ?

src/components/settings/Settings.cpp
```diff
@@ -4,108 +4,42 @@
...
-    if (FindHeader()) {
-      ReadSettingsData();
-    } else {
-      SaveSettingsToFlash();
-    }
-  } else {
-    // Read Settings from flash...
-    // never used :)
-    ReadSettingsData();
+  fs.FileRead(&settingsFile, (uint8_t *)&bufferSettings, sizeof(settings));
```
JF002, line 31: There are a few of these C-style casts that could be easily replaced by static_cast.

src/components/settings/Settings.h
```diff
@@ -95,9 +95,13 @@ namespace Pinetime {
       uint32_t GetStepsGoal() const { return settings.stepsGoal; };
 
     private:
-      Pinetime::Drivers::SpiNorFlash& spiNorFlash;
+      Pinetime::Controllers::FS& fs;
+
+      static constexpr uint32_t settingsVersion = 0x0001;
       struct SettingsData {
 
+        uint32_t version = settingsVersion;
```
JF002, line 103: I'm not sure this is needed, by we could save a few bytes here by ordering the fields of the struct by size (uint32_t first, then uint16_t and then uint8_t). This will reduce the number of padding bytes added by the compiler.
Also, all the enums used in this struct could be constrained to let's say uint8_t to ensure that the compiler won't use a larger storage for the enum, and won't implicitely change it from one version to another.

src/CMakeLists.txt
```diff
@@ -797,13 +806,25 @@ target_compile_options(lvgl PRIVATE
         $<$<COMPILE_LANGUAGE:ASM>: -MP -MD -x assembler-with-cpp>
         )
 
+# LITTLEFS_SRC
+add_library(littlefs STATIC ${LITTLEFS_SRC})
```
JF002, line 810: If possible : find a way to disable some warning generated by littlefs source code. Ex:

```
/home/jf/git/PineTime/src/libs/littlefs/lfs.c:429:13: warning: 'lfs_mlist_isopen' defined but not used [-Wunused-function]
  429 | static bool lfs_mlist_isopen(struct lfs_mlist *head,
```

I'm not sure if cmake provides a way to disable warnings coming from system/external libraries?

JF002 (commented):
Thanks for this great PR @joaquimorg ! This integration of a FS in InfiniTime will allow to add a lot of useful feature by freeing some space in the internal memory and by using this huuuuuge SPI memory!

The code is really good, I only have a few minor comments/questions you'll find in this review :)

src/CMakeLists.txt
Avamander: I tried adding to `-Wno-` + warning name under a specific subcomponents' CXX flags but that didn't have any effect.

Maybe reducing the amount of warnings enabled by the generic variable and then re-enabling in specific sub-blocks?

src/components/settings/Settings.cpp
Avamander: Casts away const, not allowed by reinterpret_cast I think, maybe there's a better solution though.

jonvmey: None of these cast away const, the two `(uint8_t*)` can just be replaced with `reinterpret_cast`s, and the two `(void*)`s on the `memcpy` are unnecessary. That being said the `memcpy` call could just be replaced with `settings = bufferSettings;`

src/components/settings/Settings.h
```diff
@@ -117,20 +121,10 @@ namespace Pinetime {
       uint8_t appMenu = 0;
       uint8_t settingsMenu = 0;
 
-      // There are 10 blocks of reserved flash to save settings
-      // to minimize wear, the recording is done in a rotating way by the 10 blocks
-      uint8_t settingsFlashBlock = 99; // default to indicate it needs to find the active block
-
-      static constexpr uint32_t settingsBaseAddr = 0x3F6000; // Flash Settings Location
-      static constexpr uint16_t settingsVersion = 0x0100;    // Flash Settings Version
+      lfs_file_t settingsFile;
```
jonvmey, line 124: `settingsFile` doesn't need to be a member variable, it's only used in `LoadSettingsFromFile` and `SaveSettingsToFile` and in each of the those it's used to open, read/write, then close the file and a local variable would suffice.

src/components/fs/FS.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+
+static constexpr size_t BLOCK_SIZE_BYTES = 4096;
+
+/*
+*   Interface between littlefs and SpiNorFlash
+*/
+
+namespace {
+    int read(const struct lfs_config *c, lfs_block_t block,
```
jonvmey, line 18: The indentation in this file isn't consistent with it's header (or the rest of the project) at 4 spaces instead of 2.

src/CMakeLists.txt
jonvmey: I added `-Wno-unused-function` after `${COMMON_FLAGS}` for the littlefs flags and that disabled the warning.

Beyond that it might be worth splitting out the warning related flags from `COMMON_FLAGS` into a separate list that can be omitted for compiling third party sources where we can't easily fix the warnings.

src/components/settings/Settings.cpp
Avamander: Hmm, okay. There certainly was a C-style cast that cast away `const` somewhere, might've confused it with some other line.

src/CMakeLists.txt
jwillikers: You can set the includes as `SYSTEM` in CMake to silence many of the warnings, too.

src/components/fs/FS.h
```diff
@@ -0,0 +1,64 @@
...
+#include "drivers/SpiNorFlash.h"
+#include <littlefs/lfs.h>
+
+namespace Pinetime {
+  namespace Controllers {
+    class FS {
+        public:
+            FS(Pinetime::Drivers::SpiNorFlash&);
+
+            Pinetime::Drivers::SpiNorFlash& mDriver;
```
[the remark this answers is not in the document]
joaquimorg: My idea was to simplify, putting it as private forced the creation of functions to call SpiNorFlash, on the other hand we have the possibility of calling SpiNorFlash where we are already using FS.

src/components/settings/Settings.cpp
joaquimorg: > None of these cast away const, the two `(uint8_t*)` can just be replaced with `reinterpret_cast`s, and the two `(void*)`s on the `memcpy` are unnecessary. That being said the `memcpy` call could just be replaced with `settings = bufferSettings;`

Thank you, I will review the recommendations.

src/components/settings/Settings.h
joaquimorg: Thank you, I will review the recommendations.

src/CMakeLists.txt
joaquimorg: > I added `-Wno-unused-function` after `${COMMON_FLAGS}` for the littlefs flags and that disabled the warning.
> 
> Beyond that it might be worth splitting out the warning related flags from `COMMON_FLAGS` into a separate list that can be omitted for compiling third party sources where we can't easily fix the warnings.

Thank you, I will review the recommendations.

src/components/settings/Settings.h
joaquimorg: Thank you, I will review the recommendations.

src/components/fs/FS.cpp
joaquimorg: Thank you, I will review the recommendations.

src/components/fs/FS.h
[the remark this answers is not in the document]
JF002: I agree that putting this field as private looks a bit cumbersome and adds an indirection that can be easily avoided.

However, my concern is about the encapsulation and the API exposed by this class FS.

When a field or a method is public, users of this class (`Controllers::Settings`, probably many others in the near future) should be able to use them without breaking the system. What if, in the future, someone writes some code that directly writes to the spi flash and corrupt the filesystem? Of course, we'll probably notice that during the review process, but it would be better if the API of the class did not allow that from the beginning.

Also in this case, `Settings` only needs to know it's saving the settings into a file system. It doesn't need to know this file system lives in an `SpiNorFlash`, in the internal flash or in RAM. If the underlying memory is exposed, `Settings` (or another Controller/App) might use this field for any reason, and would not be able to change the implementation of `FS` and `SpiNorFlash` that easily, because they are used at a lot more places than needed.

I know it's a bit theoretical, and there is very little point anyone would want to write these things right now, but you never know how the code will evolve. I always try to write my classes so that they expose a clear API and as few fields as possible, this always helps maintaining the code in the long term. It'll also help us the day we want to port the code to another platform with another memory configuration, for example.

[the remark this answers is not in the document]
joaquimorg: I got the idea and the concern, I'm going to do the functions to make the object private.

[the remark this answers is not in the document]
JF002: Thanks :+1:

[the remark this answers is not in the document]
joaquimorg: Changes made in FS to make SpiNorFlash private.

JF002:
Thanks for this PR @joaquimorg !

PR #440, opened 2021-06-18
Add support for BMA425 acceleration sensor.

JF002:
Pine factory cannot source the acceleration sensor (BMA421) anymore. However, they can source the BMA425, which is a very similar accelerometer (actually, it's a bit better as it provides with wrist raise detection in HW).

This PR add supports for the new sensor:
 - add chip id detection in BMA421 driver
 - add corresponding BLOB configuration file (+6KB :/)
 - displays the sensor type in SysInfo

photo5938170192758682733

This PR should support both BMA421 and BMA425 sensor.

DFU for testing : [pinetime-mcuboot-app-dfu-1.1.0.zip](https://github.com/JF002/InfiniTime/files/6678757/pinetime-mcuboot-app-dfu-1.1.0.zip)


src/displayapp/screens/SystemInfo.cpp
```diff
@@ -134,7 +137,13 @@ std::unique_ptr<Screen> SystemInfo::CreateScreen2() {
...
   batteryVoltageBytes[0] =
     static_cast<uint8_t>((batteryVoltage - batteryVoltageBytes[1]) * 100); // remove whole part of flt and shift 2 places over
-  //
+
+  char accel;
+  switch(motionController.DeviceType()) {
+    case Pinetime::Controllers::MotionController::DeviceTypes::BMA421: accel = '1'; break;
+    case Pinetime::Controllers::MotionController::DeviceTypes::BMA425: accel = '5'; break;
+    default: accel = '?'; break;
+  }
```
jonvmey, line 146: Totally up to you but for cases like these (declare variable then assign variable in each case of a switch statement) I like to use immediately invoked lambdas to directly initialize the variable. That has the added benefit of letting you make the variable const when desired.

kieranc:
I've just tested this on a device with BMA421, the accelerometer works as expected.

src/displayapp/screens/SystemInfo.cpp
Raupinger: To add another minor nitpick: this will make an unknown accelerometer display as BMA42?. It really shouldn't happen under normal circumstances and it allows using a char instead of a string, but if we add support for other devices in the future for example it might be incorrect.

src/drivers/Bma421_C/bma423.c
```diff
@@ -43,6 +43,521 @@
...
+  0x80, 0x2e, 0x18, 0x00, 0x80, 0x2e, 0x18, 0x00, 0x80, 0x2e, 0x18, 0x00,
+  0x80, 0x2e, 0x18, 0x00, 0x80, 0x2e, 0x18, 0x00, 0x80, 0x2e, 0x18, 0x00,
+  0x80, 0x2e, 0x18, 0x00, 0x80, 0x2e, 0x18, 0x00, 0x80, 0x2e, 0x18, 0x00,
+  0x80, 0x2e, 0x18, 0x00, 0x80, 0x2e, 0x18, 0x00, 0x80, 0x2e, 0x18, 0x00,
+  0x80, 0x2e, 0x18, 0x00, 0x80, 0x2e, 0x18, 0x00, 0x80, 0x2e, 0x18, 0x00,
+  0x80, 0x2e, 0x18, 0x00, 0x80, 0x2e, 0x18, 0x00, 0x80, 0x2e, 0x18, 0x00,
+  0x80, 0x2e, 0x18, 0x00, 0x80, 0x2e, 0x18, 0x00, 0x80, 0x2e, 0x18, 0x00,
+  0x80, 0x2e, 0x18, 0x00, 0x80, 0x2e, 0x18, 0x00, 0x80, 0x2e, 0x18, 0x00,
+  0x80, 0x2e, 0x18, 0x00, 0x80, 0x2e, 0x18, 0x00, 0x80, 0x2e, 0x18, 0x00
+};
```
Raupinger, line 559: this differs by 36 bytes from the dm241 config array. Maybe just store the difference and patch that to save on space?

Raupinger: nvm, its a lot more

Raupinger:
Works fine on my devkit

src/displayapp/screens/SystemInfo.cpp
JF002: Right! I moved the conversion to an anonymous function instead. It returns the whole device name instead of only the 1 or 5 :)

src/drivers/Bma421_C/bma423.c
JF002: Yep that was my first idea, but both blobs are very different...

Raupinger: A shame we have to add that much data to support the bm245. We have no idea how those blobs are build?

aykevl: @JF002 I was looking into writing my own driver for the BMA421 and BMA425. Where did you get these binary blobs? I can't (easily) find them in any public source.

PR #442, opened 2021-06-19
New look for settings and less duplicate code

Riksu9000:
I created a new CheckBoxes screen that apps can use to display options. It works similar to Tile and List.

Apps provide the title and an array of Options. This array is updated by the CheckBoxes screen when an option is picked. On exit, the previous app can check the array for changes, and handle them appropriately.

All of the settings screens which used checkboxes have been converted to use the new CheckBoxes screen. A lot of duplicate code has been removed.

EDIT: I was thinking if the 6 button layout should be separated from CheckBoxes. Soon we might have more than four watch faces for example, and since the names are quite long, they wouldn't fit in that layout. This however is easy to work around by simply adding scrolling like in List to CheckBoxes, and only providing four options for a single page. So in the end I don't think it should be separated to reduce duplicate code.

face
format
timeout
wakeup


Riksu9000:
Added support for multiple selected options, making SettingWakeUp compatible again.

kieranc:
I've tested this, it works, I like it. I think the validate/reset buttons in the Firmware app could do with enlarging and maybe the step goal buttons too?

MysteriousLog6:
A small little opinion! Is it possible to put a slider that allows for a custom screen timeout between 5-30seconds and then goes to a "never" option that sets the value to an insanely high value (1^100 seconds etc)

Riksu9000:
> A small little opinion! Is it possible to put a slider that allows for a custom screen timeout between 5-30seconds and then goes to a "never" option that sets the value to an insanely high value (1^100 seconds etc)

That's a good idea, but isn't directly related to this PR.

This PR might also end up getting scrapped if we end up sticking with LVGL checkboxes. And if what you propose was implemented, there would be even less of a need for this CheckBoxes screen.

PR #443, opened 2021-06-19
Improve battery charge percentage calculation by following observed discharge curve

hassless:
This PR modifies battery percentage calculation to use a lookup table to determine the battery percentage corresponding to the current battery voltage.

The contents of this voltage-to-percentage lookup table is determined by observing the current voltage of the battery at regular intervals for a full discharge cycle of the battery. To arrive at the lookup table, in summary, the following steps were taken:

1. Hack InfiniTime such that:
   - battery voltage is updated continually.
   - the digital watch face displays the battery voltage instead of the time.
   - the display never turns off.

2. Flash the modified InfiniTime to a sealed PineTime.

3. Fully charge the PineTime, turn it on, and place it under a webcam.

4. Let the webcam take a photo of the display every 20 seconds until battery is drained and the PineTime turns off.

modified-infinitime
_Modified InfiniTime displaying a voltage of 4.11 on the digital watch face._

5. Look at all the photos in the order in which they were taken and write down the voltages recorded on the photos alongside the time at which each photo was taken.

We now have a long list of voltages and timestamps, representing the discharge curve of the battery over time and under constant load. (I did not actually measure the current that was drawn, so 'constant load' is really just an assumption. I believe it to be valid though, since InfiniTime was doing the same thing continuously during the entire run: update and display battery voltage).

6. Plot the discharge curve (see image 'Observed discharge curve' below) and replace the label 'time' with 'battery charge percentage' on the x-axis (as was done in the image 'Voltage-percentage mapping'). Define the first recorded voltage (when the battery was fully charged) to correspond to 100% battery charge and define the last recorded voltage (right before the PineTime shuts down) to correspond to 0% battery charge. Linearly interpolate the remaining integer battery percentages (99% to 1%) between the defined 100% and 0%.

observed-discharge-curve
_Observed discharge curve: The observed data points are plotted in grey. The smooth curve created by making the observed data points non-increasing is plotted in blue._

7. Using the graph, pick a single integer battery percentage for each voltage (with two decimal digits). Pick the integer nearest to the average of the battery percentages corresponding to that 2-decimal voltage.

The result is a lookup table that translates voltages with two decimal digits to integer battery percentages. The image 'Voltage-percentage mapping' below plots the values from that table in dark blue.

voltage-percentage-mapping
_Voltage-percentage mapping: The values from the lookup table (dark blue) plotted against the observed data points (light blue). The old linear battery percentage calculation is plotted in orange._

8. Implement the lookup table as an array of bytes, with each entry representing the battery percentage for a single two-decimal-digit voltage. The entries are ordered by the voltages they represent, so the first entry represents the battery percentage corresponding to 2.98v (the lowest observed voltage) and the last entry represents the battery percentage at 4.18v (the highest observed voltage). Finally, to convert a voltage value to an index in the lookup table, multiply by 100 (to represent the same information as an integer) and subtract 298 (to let 2.98v correspond to index 0).

My C++ is rather rusty (its been 10+ years), so there may very well be a better way to implement this concept. Comments & suggestions for improvement are of course appreciated!


Riksu9000:
I don't like how the percentage can drop 5% at once, and due to fluctuations, jump 5% as well. This can't be helped though with the current ADC to voltage conversion precision. In #444, I improved the precision to almost double, which would decrease the stepping. The ADC is also running at 10bit depth, which I believe could be increased to 12, improving the precision 4x more.

If these changes were made, the lookup table would have to get much larger. In that case I think a better way to implement this would be to split the curve into multiple linear parts.
For example 3.75V < voltage < 4.0V would be mapped linearly, but different than 3.65V < voltage < 3.75V

hassless:
> I don't like how the percentage can drop 5% at once, and due to fluctuations, jump 5% as well.

I agree, the drops and jumps around the 60-25% range are much larger than I like.

> If these changes were made, the lookup table would have to get much larger. In that case I think a better way to implement this would be to split the curve into multiple linear parts.
> For example 3.75V < voltage < 4.0V would be mapped linearly, but different than 3.65V < voltage < 3.75V

That's a good idea! We probably don't even need that many linear line segments to get a really nice approximation of the discharge curve. It guess it also makes the implementation less dependent on the actual precision we use for the voltage, which makes the approach more robust. I'll look into that in a couple of days, when I have time.

jonvmey:
The fluctuations should be able to be reduced by applying some filtering to the incoming data to smooth it out.

Also lookup tables are often quite efficient. I wouldn't be surprised if a linear segment implementation would require more bytes of code than the ~120 bytes the LUT uses.

JF002:
Great work! I'm testing this PR right now (along with #423 and #483).
I was wondering : Does the lookup table implemented in the code correspond to the graph shown in the first comment of this PR?

Here is an example of my measurement : 3.67v is translated to 367. The index in the LUT is 367 - 298  =  67. The value at index 67 of the LUT is 22%.
However, according to the graph, 3.67 should be around 50%.

Is this difference expected?

Riksu9000:
Linearly mapped it's 50%, but according to the mapped value graph, it's 22%.

I also want to mention that this test might have to be redone because #444 and #483 change the reported voltage.

Riksu9000:
Maybe we should actually consider 3.5V as 0%. This would discourage users from discharging the battery too much, prolonging lifespan. We also wouldn't need to worry about the rapid voltage dropoff after this point.
One thing noticed is that the 3.3V regulator has about 200mV dropoff. This means that if the voltage is below 3.5V, the regulator cannot keep the 3.3V rail at 3.3 volts. All chips work at lower voltages, but it could cause inconsistent behaviour.

JF002:
@hassless What do you think of @Riksu9000's suggestion (more details [here](https://github.com/JF002/InfiniTime/pull/423#issuecomment-892966664)) ?

JF002:
@hassless Thanks for your feedback! So I guess we can start simple by changing [this constant](https://github.com/JF002/InfiniTime/blob/develop/src/components/battery/BatteryController.cpp#L55) as @Riksu9000  mentioned?

I'm also totally open to your suggestion (implementing the discharge curve using multiple linear segments).

If we do that (set min voltage to 3.5v and/or multiple segment), should we close this PR?

JF002:
> If we're going to do this, then maybe its a good idea to automatically turn the device off when the battery percentage reaches (the new) 0%

We have no way to cut the power to the MCU and the devices, but we can implement a "low power" mode, where all devices would be disabled, backlight set to minimum and where we would display a big "low battery" logo, for example.

hassless:
> So I guess we can start simple by changing [this constant](https://github.com/JF002/InfiniTime/blob/develop/src/components/battery/BatteryController.cpp#L55) as @Riksu9000 mentioned?

Yep, that should do it.

> I'm also totally open to your suggestion (implementing the discharge curve using multiple linear segments).
> 
> If we do that (set min voltage to 3.5v and/or multiple segment), should we close this PR?

Sure, the multiple-line-segments thing will have an implementation different from the one in this PR, so I guess that should go in a new PR. So yeah, I'd say this PR can be closed.

JF002:
I feel bad for closing this PR as I think it's a really good job (both analysis and code) :)

hassless:
> 
> 
> I feel bad for closing this PR as I think it's a really good job (both analysis and code) :)

Thanks :) But please don't; a first version of my line-segments-based implementation is almost done and I'll probably create a new PR for it tomorrow, so closing this PR will not mean that any of our work on this PR is lost!

JF002:
@hassless Thanks, I'm closing this one, then!

PR #444, opened 2021-06-20
Float voltage to int

Riksu9000:
In all uses of the voltage variable, the ADC value is first converted to float, and then back to integer.

This PR makes it so the voltage is now in millivolts. If an app requires float for any reason, it can easily be converted in the app.

Previously the precision was limited by roundf() to 10mV, now the precision is ~6mV

The only visible change is that the about page now shows the actual value of the voltage variable, so it is now in millivolts.

volts

In #258, the multiplication was changed from 2 to 2.04. I didn't find any explanation for this change. The *2 compensates for the hardware voltage divider. I think this might have been an error.
`voltage = (static_cast<float>(p_event->data.done.p_buffer[0]) * 2.04f) / (1024 / 3.0f);`


JF002:
Thanks for this PR! 
I have no idea where this 2.04 factor come from... Maybe some kind of trick to correct rounding (float<->int) issue?

@Riksu9000 Previous merges conflict with this branch, could you have a look at them so I can merge it?

JF002:
> Thanks for this PR!
> I have no idea where this 2.04 factor come from... Maybe some kind of trick to correct rounding (float<->int) issue?
> 
> ~~@Riksu9000 Previous merges conflict with this branch, could you have a look at them so I can merge it?~~ Nope, that OK :+1:

PR #448, opened 2021-06-21
Disable vibration while charging

### LINKED ISSUE #446: Enhancement: Turn off notifications while charging
Maybe it's just taste, but I think it would seem more clean if vibrations would be deactivated as a whole while the pinetime is charging.

Riksu9000:
I don't think having vibrations while charging is useful, and it makes unpleasant noises.

Fixes #446

Previously on OnChargingEvent, the watch would vibrate. I changed it so the watch wakes up instead. This is how most devices behave.

### DIFF
--- a/src/components/motor/MotorController.cpp
+++ b/src/components/motor/MotorController.cpp
@@ -7,7 +7,8 @@ APP_TIMER_DEF(vibTimer);
 
 using namespace Pinetime::Controllers;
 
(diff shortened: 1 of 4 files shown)


schneidr:
I rely on the smartwatch only for all notifications of my phone. I want to be able to notice them when the watch is charging as well.

Riksu9000:
> I rely on the smartwatch only for all notifications of my phone. I want to be able to notice them when the watch is charging as well.

Thanks for bringing this up. I'll have to make an option for it then.

PR #449, opened 2021-06-21
Removed an illogical comparison from SystemInfo and St7789 driver

Avamander:
This PR removes a few illogical lines from SystemInfo screen and the St7789 driver.

`batteryPercent` is cast to unsigned integer, the comparison is never true. The arguments in St7789 driver are both unsigned.


JF002:
Thanks :+1:

PR #452, opened 2021-06-23
Fix call notification button alignment

### LINKED ISSUE #305: Bug: Buttons in notifications are shifted on multiline caller name
The call control buttons are shifted down if the caller name is longer than one line.

bug_multiline_call_notification

Riksu9000:
This fixes #305, where the buttons can move if the caller name is too long. Now the buttons are static, and outside of the container.

call


JF002:
Thanks!

PR #458, opened 2021-06-25
Add color picker for PineTimeStyle watchface

kieranc:
I expected this to take a lot longer, but it works and I'm pretty happy with it so I'd like some input/review.
I'd like to make the buttons bigger/transparent as they're a bit hard to hit but that's about it.
The colors are stored in settings as an integer which is the index value of an array of colors stored as a variable in the class.
edit: The color picker adds about 3k to the DFU size


kieranc:
It looks like the default settings are not being applied and on first boot the colors are white/white/white, but I think I'm settings the default values in Settings.h at lines 133-135 ?

kieranc:
I added a check in case the settings are uninitialised - if all colours are white, it will automatically apply default colors. I've also improved the UI a bit and added reset button, please test!

JF002:
Looks like their is a potential issue on initial values of the colors. Here's a chat log:
image

User mentioned that these colors were the colors applied by default. 

We need to ensure that default colors will always be applied the first time the user uses a firmware with this feature, no matter what random values the memory might contain at that time.

kieranc:
> User mentioned that these colors were the colors applied by default.
> 
> We need to ensure that default colors will always be applied the first time the user uses a firmware with this feature, no matter what random values the memory might contain at that time.

I believe the most recent commit fixes this issue. I'll keep testing.

Riksu9000:
Would it make sense to open watchface settings by long pressing on the watch face? All watch faces could have their own settings menu and they wouldn't need to have their own entries taking up space in settings.

JF002:
I agree with @Riksu9000 : I think it would be more "scalable" if watchfaces and apps had their own settings "embedded" in the app instead of adding an entry to the global settings.

I'll have a look at this PR anyway. I think we can still merge it with color picker in the setting page, and move it in the app on a future PR.

kieranc:
I like the idea but I'm worried about accidental triggers, maybe an alternative would be to have a global watchface settings button which lets you change settings for each watchface below a single settings item?
If you're thinking about merging this today, I have a randomise colors button which should be ready to push this afternoon - so you're aware. In any case I'll keep working on extending this feature to other watchfaces.

JF002:
> I like the idea but I'm worried about accidental triggers, maybe an alternative would be to have a global watchface settings button which lets you change settings for each watchface below a single settings item?

Indeed, that's another possibility. I guess we could build prototypes of those solutions to see which one works best. But it's not easy to figure out what would be the best as this color picker is the first "app specific" setting in InfiniTime. 

> If you're thinking about merging this today, I have a randomise colors button which should be ready to push this afternoon - so you're aware. In any case I'll keep working on extending this feature to other watchfaces.

I'll probably start the review later today, but I can wait a bit before merging if you plan on adding improvements soon ;)

Riksu9000:
Random colors option seems a bit unnecessary. Also I haven't looked into this PR much, but does it conflict with the idea of having a selectable global accent color? I'm hoping some day there would be a single setting that can adjust colors in all screens, maybe including the watch face?

Avamander:
> I'm hoping some day there would be a single setting that can adjust colors in all screens, maybe including the watch face?

That's probably a separate issue and should be discussed there.

Riksu9000:
> That's probably a separate issue and should be discussed there.

Sure, I just opened a new issue for it. Still it might conflict with this and require changing this color picker afterwards if it isn't taken into consideration, which is why I mentioned it.

kieranc:
> Random colors option seems a bit unnecessary. Also I haven't looked into this PR much, but does it conflict with the idea of having a selectable global accent color? I'm hoping some day there would be a single setting that can adjust colors in all screens, maybe including the watch face?
Yes randomisation is unnecessary but it's kinda nice, and easy enough to implement. You should try the PR, while I like the default colors I also enjoy just picking some random colors. Eventually we could end up with a list of popular combinations and have the randomise button scroll through them instead of being totally random, but for now it's a nice way to change all the colors at once and hopefully find a combo you like.

Riksu9000:
You were right. I tried it and it really is a fun feature :D.

Maybe one of the color options should be to follow the accent color, and this would be the default. It might be weird if each time someone wanted to change the accent color it would have to be changed in multiple places.

Riksu9000:
The side bar can be set to black, but then the text isn't visible.

src/displayapp/screens/settings/SettingPineTimeStyle.cpp
```diff
@@ -0,0 +1,332 @@
...
+  }
+  if((object == btnRandom) && (event == LV_EVENT_PRESSED)) {
+    uint8_t randTime = rand() % 17;
+    uint8_t randBar = rand() % 17;
+    uint8_t randBG = rand() % 17;
+    // Check if the time color is the same as its background, or if the sidebar is black. If so, change them to more useful values.
+    if ((randTime == randBG) || (randBar == 3)) {
+      randBG += 1;
+      randBar = randTime;
+    }
```
Riksu9000, line 322: This should be two separate checks

src/displayapp/screens/settings/SettingPineTimeStyle.cpp
```diff
@@ -0,0 +1,332 @@
...
+bool SettingPineTimeStyle::Refresh() {
+  return running;
+}
+
+void SettingPineTimeStyle::UpdateSelected(lv_obj_t *object, lv_event_t event) {
+  uint8_t valueTime = settingsController.GetPTSColorTime();
+  uint8_t valueBar = settingsController.GetPTSColorBar();
+  uint8_t valueBG = settingsController.GetPTSColorBG();
+    
+  if((object == btnNextTime) && (event == LV_EVENT_PRESSED)) {
```
Riksu9000, line 230: `event == LV_EVENT_CLICKED` could be tested at the beginning of the function, so it wouldn't need to be duplicated in each if statement. CLICKED is preferred over PRESSED.

src/displayapp/screens/settings/SettingPineTimeStyle.cpp
```diff
@@ -0,0 +1,332 @@
...
+  lv_obj_set_click(backgroundLabel, true);
+  lv_label_set_long_mode(backgroundLabel, LV_LABEL_LONG_CROP);
+  lv_obj_set_size(backgroundLabel, 240, 240);
+  lv_obj_set_pos(backgroundLabel, 0, 0);
+  lv_label_set_text(backgroundLabel, "");
+
+  btnNextTime = lv_btn_create(lv_scr_act(), nullptr);
+  btnNextTime->user_data = this;
+  lv_obj_set_size(btnNextTime, 60, 60);
+  lv_obj_align(btnNextTime, lv_scr_act(), LV_ALIGN_IN_RIGHT_MID, -15, -80);
```
Riksu9000, line 154: Maybe some of these buttons should be aligned to the corners with `LV_ALIGN_IN_TOP_RIGHT` for example

src/displayapp/screens/settings/SettingPineTimeStyle.cpp
```diff
@@ -0,0 +1,332 @@
...
+  lv_obj_set_style_local_line_opa(stepGauge, LV_GAUGE_PART_NEEDLE, LV_STATE_DEFAULT, LV_OPA_COVER);
+  lv_obj_set_style_local_line_width(stepGauge, LV_GAUGE_PART_NEEDLE, LV_STATE_DEFAULT, 3);
+  lv_obj_set_style_local_pad_inner(stepGauge, LV_GAUGE_PART_NEEDLE, LV_STATE_DEFAULT, 4);
+  
+  backgroundLabel = lv_label_create(lv_scr_act(), nullptr);
+  lv_obj_set_click(backgroundLabel, true);
+  lv_label_set_long_mode(backgroundLabel, LV_LABEL_LONG_CROP);
+  lv_obj_set_size(backgroundLabel, 240, 240);
+  lv_obj_set_pos(backgroundLabel, 0, 0);
+  lv_label_set_text(backgroundLabel, "");
```
Riksu9000, line 149: Is it possible to somehow use some of the code from the real watchface?

src/displayapp/screens/settings/SettingPineTimeStyle.cpp
```diff
@@ -0,0 +1,332 @@
...
+    if ( valueBG > 0 ) {
+      valueBG -= 1;
+      settingsController.SetPTSColorBG(valueBG);
+      lv_obj_set_style_local_bg_color(timebar, LV_BTN_PART_MAIN, LV_STATE_DEFAULT, pts_colors[valueBG]);
+    } else {
+      valueBG = 16;
+      settingsController.SetPTSColorBG(valueBG);
+      lv_obj_set_style_local_bg_color(timebar, LV_BTN_PART_MAIN, LV_STATE_DEFAULT, pts_colors[valueBG]);
+    }
+  }
```
Riksu9000, line 303: In each of these inner if statements some of the same code is duplicated in both `if` and `else`. It could be moved outside of it.

Also `else if` can be used on the object checks.

src/displayapp/screens/settings/SettingPineTimeStyle.cpp
```diff
@@ -0,0 +1,332 @@
...
+  if((object == btnReset) && (event == LV_EVENT_PRESSED)) {
+      settingsController.SetPTSColorTime(11);
+      lv_obj_set_style_local_text_color(timeDD1, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, pts_colors[11]);
+      lv_obj_set_style_local_text_color(timeDD2, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, pts_colors[11]);
+      lv_obj_set_style_local_text_color(timeAMPM, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, pts_colors[11]);
+      settingsController.SetPTSColorBar(11);
+      lv_obj_set_style_local_bg_color(sidebar, LV_BTN_PART_MAIN, LV_STATE_DEFAULT, pts_colors[11]);
+      settingsController.SetPTSColorBG(3);
+      lv_obj_set_style_local_bg_color(timebar, LV_BTN_PART_MAIN, LV_STATE_DEFAULT, pts_colors[3]);
```
Riksu9000, line 312: There should probably be an enum for pts_colors[].

kieranc:
> The side bar can be set to black, but then the text and icons aren't visible.

Yes it can, but it won't be set to black by the randomiser, only manually. Having the default follow the accent color is a nice idea, just needs a system wide color picker implemented...

For now is everyone ok with it as it stands, with a plan of implementing the color picker in some fashion for all watchfaces and potentially the system accent color too?

Riksu9000 (commented):
The SettingPineTimeStyle.* files may need to be reformatted with clang-format

Riksu9000:
>>The side bar can be set to black, but then the text and icons aren't visible.
>
>Yes it can, but it won't be set to black by the randomiser, only manually.

I would still consider it a bug.

src/displayapp/screens/settings/SettingPineTimeStyle.cpp
kieranc: Thanks, resolved.

kieranc:
> > > The side bar can be set to black, but then the text and icons aren't visible.
> > 
> > 
> > Yes it can, but it won't be set to black by the randomiser, only manually.
> 
> I would still consider it a bug.

Ok, I added a check to avoid the sidebar being set to black

src/displayapp/screens/settings/SettingPineTimeStyle.cpp
kieranc: It would be nice but I don't see how I could. If the settings becomes integrated with the watchface it should be possible.

src/displayapp/screens/settings/SettingPineTimeStyle.cpp
kieranc: They could, but I'm happy with the current approach

src/displayapp/screens/settings/SettingPineTimeStyle.cpp
kieranc: I'm afraid I don't know what an enum is or what advantage it offers, but the main reason it's an array is because i only store the array index in settings rather than the color

lman0:
hi, @kieranc  would have been better to avoid the sidebard and main background color to be the same than the color of the text instead of a static avoiding?

kieranc:
The sidebar and text can happily be the same color, as can the sidebar and background. the only cases I want to avoid are the main time text being the same as the background, IE invisible, or the sidebar being black which again makes things invisible. in the future it might be possible to set the sidebar black and icons white, but not yet, so until then I think the checks implemented give the most freedom possible to the user while avoiding making things unreadable.

Anyway, I've been sat here tapping random and skipping around for ages, it works well.

lman0:
could you give a dfu here or on discord?
i would happy to test it

kieranc:
Here you go!
[pinetime-mcuboot-app-dfu-1.3.0.zip](https://github.com/JF002/InfiniTime/files/6988222/pinetime-mcuboot-app-dfu-1.3.0.zip)

Riksu9000:
> in the future it might be possible to set the sidebar black and icons white, but not yet

it should be pretty simple to add a check in PineTimeStyle `if barBG == black then font_color = white` if you'd like to do that in this PR.

kieranc:
The calendar icon would also need inverting which is less trivial, and the step count gauge....
I think I'd rather get it merged as just the color picker, let people play with it, then work on v2 with more features!

src/displayapp/screens/settings/SettingPineTimeStyle.cpp
Riksu9000: With an enum, this could be `pts_colors[colorBlack]` and `SetPTSColorBG(colorBlack)` for example, which is clearer than 3 or 11,  but there may be some even better way to do it.

JF002:
Thanks @kieranc  for this PR, and everyone for the review!

I pushed the [these changes](https://github.com/JF002/InfiniTime/commit/31bc47d1cb397f5de0275d0d95aac7ca29cc7392) just after merging the branch to add some improvements.

PR #459, opened 2021-06-25
Improved a bunch of format specifiers

Avamander:
I fixed a missing attribute in LVGL and this allowed me to detect and improve a bunch of format specifiers. 

Hopefully LVGL itself will contain the patch soon as well, but that's yet to be seen.


JF002:
Do you mean the format specifiers (%d and %i) were incorrect, or are they not supported by lvgl?

Avamander:
They were incorrect for the data type given.

JF002:
Too bad the compiler does not warn us about these type mismatches... or do we just need to add compiler options to enable those warnings?

Avamander:
> or do we just need to add compiler options to enable those warnings?

That, and the LVGL functions have to have an attribute added to them. For the latter I've made a PR to LVGL.

PR #464, opened 2021-07-03
Dim screen before sleep

Riksu9000:
Dims the screen to lowest brightness two seconds before going to sleep. This gives time for the user to refresh the idle timer by tapping the screen if the watch is about to go to sleep.


hubmartin:
Works fine, thanks.

PR #465, opened 2021-07-03
Condense firmware validation code

Riksu9000:
Version and shortref info were spread out in four different labels. Here I combined them into a single label.


JF002:
Thanks! :+1:

Issue #445, opened 2021-06-21
Stopwatch+timer bug


Itai-Nelken:
Setting a timer and then starting the stopwatch causes the image to shift up when the timer opens when it ends.
It even caused a bootloop on my brother's devkit.

Riksu9000:
Can confirm, pretty spooky. Stopwatch doesn't even need to be running, the app just needs to be open.

![corruption](https://user-images.githubusercontent.com/37774658/122755941-640a6280-d29e-11eb-929e-eb50beebf487.jpg)

Raupinger:
Honestly, I'm clueless on this one so far. The way the timer gets opened is mostly identical to how notification previews are. Those don't break.
I also don't see a pattern in which apps show the issue and which don't. 
Works: Clock, Screen List, Quick Settings, Steps, Meter, System Info, Display Timeouts, Wake up, Time Format
Doesn't work: Stopwatch, Music, Settings List, Heart Rate, Navigation, Paint, Pong
Additionally, some apps don't work on older branches but do on current develop.

kieranc:
I can also confirm this bug, the display remained shifted up until I rebooted.

JF002:
Wow that's not good! Is there an easy way to reproduce this issue? On which versions/commit does this happen?

hamblingreen:
Easy way to reproduce bug:
1. Start timer
2. Move to Pong app (testing what other apps do right now)
3. Wait for timer to go off
4. Bug is reproduced reliably (v1.2.0)

Itai-Nelken:
I confitm tgat this doesn't happen with the 2048 and accelerometer apps.

JF002:
@hamblingreen thanks for the procedure! I can indeed reproduce the issue!
This is kind of funny result :D We probably have a issue (data race?) around the hardware scrolling implementation (https://github.com/JF002/InfiniTime/blob/develop/src/displayapp/LittleVgl.cpp#L83)

Bloodaxe95:
Not sure if this is the right place, but this bug is still present in 1.3.0 of InfiniTime. At least I could reliably reproduce it on my sealed PineTime.

hamblingreen:
New way to reproduce something similar:
1. navigate to menu containing the Draw app
2. tap the draw icon twice quickly (this takes a few tries to get the timing down)
3. screen is slightly shifted up, and a restart fixes it.

[Video too big to upload](https://t.me/pinetime_dev/15990)

[...]

PR #466, opened 2021-07-04
Code cleanup

Riksu9000:
Remove unnecessary overrides, and clang-format some files thanks to the pre-commit hook.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "src/libs/lvgl"]
 	path = src/libs/lvgl
 	url = https://github.com/joaquimorg/lvgl.git
+[submodule "src/libs/littlefs"]
+	path = src/libs/littlefs
+	url = https://github.com/littlefs-project/littlefs.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.10)
-project(pinetime VERSION 1.2.0 LANGUAGES C CXX ASM)
+project(pinetime VERSION 1.3.0 LANGUAGES C CXX ASM)
 
 set(CMAKE_C_STANDARD 99)
 set(CMAKE_CXX_STANDARD 14)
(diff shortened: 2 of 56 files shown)


JF002:
Those background labels are there to ensure the whole screen is refreshed when launching a new app. This is needed for the hardware scrolling functionality to work correctly with LVGL. 

I've just tested this branch and... it doesn't seem to work properly : 
bad_hw_scroll

Did you check the code?

Riksu9000:
I thought I tested everything, but I totally forgot to test the digital watch face. Every other app seems to work fine without it though unless I've made another mistake..

JF002:
It may be an issue on my side, then... The display would be corrupted as soon as I scroll up or down... let me test again ;)

Riksu9000:
No I can replicate this as well. These issues only come up when using the digital watch face, all other faces and apps work just fine as far as I can tell.

Riksu9000:
I removed the backgroundLabel changes for now. I would still like to understand better why that happens as it doesn't seem reasonable that some labels are responsible for keeping the display in sync.

Also some more files were changed by the clang-format pre-commit hook. It fits the subject so I kept them.

hamblingreen:
The display corruption seems similar to that of issue #445, just a thought.

JF002:
@hamblingreen good find! So this issue could not be triggered by the modifications in this PR...

JF002:
> I removed the backgroundLabel changes for now. I would still like to understand better why that happens as it doesn't seem reasonable that some labels are responsible for keeping the display in sync.

@Riksu9000 The HW scrolling feature works correctly only when the whole display is updated at once (1 call to LittleVgl::FlushDisplay() with the `area` covering the whole screen. This is a kind of workaround we found to integrate that scrolling feature with LVGL (which doesn't support it natively). 

The analog clock is composed of a background image of 240*240px. When you switch from the analog clock to any other app, and update of the whole screen was needed as destroying the analog clock removes all the widget (and this full screen picture).

The digital clock, on the other side, is composed of many small widget, none of them covering the whole screen. If you switched from the digital clock to another app that doesn't have a widget covering the whole screen, the driver would probably enable the HW scrolling effect but only for a few lines, generating that strange corruption.

I agree, it might be a bit confusing and inefficient, but using these label covering the whole screen allows this mechanism to work correctly :)

Riksu9000:
This is taken from [LVGL 8 documentation](https://docs.lvgl.io/8/porting/display.html#draw-buffer): "In the display driver (lv_disp_drv_t) the full_refresh bit can be enabled to force LVGL to always redraw the whole screen. This works in both one buffer and two buffers modes."

So if there currently isn't a better way to do this, there might be if we update to LVGL 8, assuming this can be changed at runtime.

JF002:
Oooh that's interesting! And a nice incentive to update to LVGL8 !

PR #467, opened 2021-07-06
Update documentation

Riksu9000:
Fix some typos and remove some spaces.
Update readme. Removed mention of the Motion app as the app seems more like a debugging feature than something that people will use.


JF002:
Thanks :+1:

PR #468, opened 2021-07-06
PineTimeStyle step gauge bugfix

kieranc:
The step gauge seems to have issues if the range is over 100, so this change sets the gauge range to 100 and calculates the step count as a percentage of the step goal to show progress.

Edit for a bit more context. Previously I was scaling the step gauge by adjusting the range, so a step goal of 13000 would make the range of the gauge 0-130, which resulted in only 9 divisions appearing. I'm not sure why this happened, but this fix changes the approach by setting the gauge range to 0-100 always, then calculating the needle position/progress as a percentage of the step goal.


JF002:
Thanks!

PR #469, opened 2021-07-07
Make firmware updating more foolproof

### LINKED ISSUE #316: Bug: Repeatable firmware update failure in 1.0
Similar to that reported in #295, I have been able to repeatedly b0rk OTA updates with multiple pinetimes... 

Steps:
* Initiate the update
* Whilst updating, press the button to clear the firmware update progress screen.
* Short press the button to turn off the display
* Update progress almost immediately stops on the phone side, and approximately eight seconds later the watch will reboot (watchdog?). 

Perhaps the SPI flash is being switched off?

Two videos showing the bug in action. The second one I deliberately put off turning the display off, going into the menu somewhat, just to show that the update was still progressing without issue, and then turned the display. Unfortunately I stopped the video almost exactly as the watchdog fired, but it did reboot the pinetime without issue. 

https://youtu.be/T0CmH48ZbJU

https://youtu.be/bPcaRPkWoZc

Riksu9000:
Made the firmware update screen not exitable by the user. The screen closes automatically if an error occurs.
Added precautions to make sure the device doesn't sleep during firmware update.
Fixed bugs like State not resetting on error, and BleFirmwareUpdateFinished message being pushed before the Validated state was set.
Fixes #316

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -180,9 +180,13 @@ void DisplayApp::Refresh() {
         }
         break;
       case Messages::TouchEvent: {
-        if (state != States::Running)
+        if (state != States::Running) {
           break;
+        }
         auto gesture = OnTouchEvent();
+        if (gesture == TouchEvents::None) {
+          break;
+        }
         if (!currentScreen->OnTouchEvent(gesture)) {
           if (currentApp == Apps::Clock) {
             switch (gesture) {
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -141,9 +141,6 @@ void DisplayApp::Refresh() {
   if (messageReceived) {
     switch (msg) {
       case Messages::GoToSleep:
-        if (bleController.IsFirmwareUpdating()) {
-          break;
-        }
         brightnessController.Backup();
         while (brightnessController.Level() != Controllers::BrightnessController::Levels::Off) {
           brightnessController.Lower();
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -132,6 +132,9 @@ void DisplayApp::Refresh() {
   if (xQueueReceive(msgQueue, &msg, queueTimeout)) {
     switch (msg) {
       case Messages::GoToSleep:
+        if (bleController.IsFirmwareUpdating()) {
+          break;
+        }
         brightnessController.Backup();
         while (brightnessController.Level() != Controllers::BrightnessController::Levels::Off) {
           brightnessController.Lower();
@@ -277,6 +280,7 @@ void DisplayApp::LoadApp(Apps app, DisplayApp::FullRefreshDirections direction) 
       break;
     case Apps::FirmwareUpdate:
       currentScreen = std::make_unique<Screens::FirmwareUpdate>(this, bleController);
+      ReturnApp(Apps::Clock, FullRefreshDirections::Down, TouchEvents::None);
       break;
 
     case Apps::Notifications:
--- a/src/displayapp/screens/FirmwareUpdate.cpp
+++ b/src/displayapp/screens/FirmwareUpdate.cpp
@@ -5,21 +5,6 @@
 
 using namespace Pinetime::Applications::Screens;
 
-namespace {
-  // Copied from StopWatch.cpp
-  TickType_t calculateDelta(const TickType_t startTime, const TickType_t currentTime) {
-    TickType_t delta = 0;
-    // Take care of overflow
-    if (startTime > currentTime) {
-      delta = 0xffffffff - startTime;
-      delta += (currentTime + 1);
-    } else {
-      delta = currentTime - startTime;
-    }
-    return delta;
-  }
-}
-
 FirmwareUpdate::FirmwareUpdate(Pinetime::Applications::DisplayApp* app, Pinetime::Controllers::Ble& bleController)
   : Screen(app), bleController {bleController} {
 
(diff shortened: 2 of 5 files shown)


src/displayapp/screens/FirmwareUpdate.cpp
```diff
@@ -78,4 +94,9 @@ void FirmwareUpdate::UpdateValidated() {
 void FirmwareUpdate::UpdateError() {
   lv_label_set_recolor(percentLabel, true);
   lv_label_set_text(percentLabel, "#ff0000 Error!#");
+  startTime = xTaskGetTickCount();
+}
+
+bool FirmwareUpdate::OnButtonPushed() {
```
JF002, line 100: It is still possible to exit the app by swiping up and down. I think you can prevent that by overriding `bool Screen::OnTouchEvent(Pinetime::Applications::TouchEvents)`:

```
bool FirmwareUpdate::OnTouchEvent(Pinetime::Applications::TouchEvents) {
  return true;
}
```

src/displayapp/DisplayApp.cpp
```diff
@@ -141,6 +141,9 @@ void DisplayApp::Refresh() {
   if (messageReceived) {
     switch (msg) {
       case Messages::GoToSleep:
+        if (bleController.IsFirmwareUpdating()) {
```
JF002, line 144: This extra precaution should not be necessary, as DfuService sends the message `Messages::BleFirmwareUpdateStarted` to SystemTask when it's starting the update procedure. 
When receiving this message, SystemTask sets the flag `doNotGoToSleep` to true so that it doesn't go to sleep until it receives the message `Messages::BleFirmwareUpdateFinished`.

Unless there's a weakness in this implementation?

JF002 (commented):
Thanks for this PR! I've just tested the code, the only thing I would add is this small change to prevent the user from closing the app with a swipe on the display.

I've even been able to restart an OTA after a failed one, which was not working previously, awesome :+1 !

src/displayapp/DisplayApp.cpp
Riksu9000: Some apps enable sleep on exit. If the firmware update is started while an app like this is running, the watch will go to sleep during the update. If this was fixed, then the check probably won't be needed. I didn't intend on this to be just a workaround for that though but just extra protection for any future changes.

src/displayapp/screens/FirmwareUpdate.cpp
Riksu9000: It shouldn't be possible to exit because the return gesture is set to `TouchEvents::None`, but maybe it actually processes the None event which allows the app to exit. I'd rather fix this by adding a new check before loading the return app which would be `returnTouchEvent != TouchEvents::None`

JF002: Right! But when I tested it, it was still possible to close the app by swiping...
I think you need to add ` ReturnApp(Apps::Clock, FullRefreshDirections::Down, TouchEvents::None);` in the case `case Messages::BleFirmwareUpdateStarted:` of the eventloop of DisplayApp (line220), then.

src/displayapp/DisplayApp.cpp
JF002: Mhm ok, I see what you mean : DfuService sends the message, SystemTask receives it, set the flash `doNotGoToSleep` to true and then sends a message to DisplayApp that closes the current app and open the FirmwareUpdate app.
If the app that was running re-enable sleep, the firmware update app will go to sleep after the timeout, yes.

So yes, this is an easy workaround, but I was thinking : shouldn't we send the message 'disable sleep' from the firmware update app ? This way, sleep mode will be disabled/enabled in the correct order. What do you think?

Riksu9000: I thought about that, but the reason I didn't do it was because the app isn't actually responsible for the update.

Maybe this check should instead be in systemtask message EnableSleeping?
```
case Messages::EnableSleeping:
  if (!bleController.IsFirmwareUpdating()) {
    doNotGoToSleep = false;
  }
  break;
```

src/displayapp/screens/FirmwareUpdate.cpp
Riksu9000: I added this check to TouchEvent message to make sure that TouchEvents::None isn't processed in any way. Before this I got the app to exit just before the update was finished, but couldn't get it after this.
```
if (gesture == TouchEvents::None) {
  break;
}
```

ReturnApp() shouldn't be needed in BleFirmwareUpdateStarted message because it already gets run inside LoadApp()

src/displayapp/DisplayApp.cpp
JF002: Well, sleeping can be considered as mostly a UI think... But sleep also means that the SPI memory will also be put to sleep, and it'll make the OTA fail so... it's not that simple, indeed.
Yes, we can also put this check the EnableSleeping, I don't know which one would be the best...

src/displayapp/screens/FirmwareUpdate.cpp
JF002: You are right... Ok, I don't know what I did wrong... I'm sure I managed to close the update app with a swipe, but I cannot reproduce it anymore...

Riksu9000: Most likely the swipe gesture was what generated the TouchEvent message, but by the time displayapp read it, it had already changed to None in the touchscreen driver. The firmware update slows down processing, which probably added enough delay for this to happen.

src/displayapp/DisplayApp.cpp
Riksu9000: I made the change I proposed as it addresses the existing issue more directly.

JF002: Looks good to me, thanks! So this PR is ready to be merged?

Riksu9000: Yep, should be fine.

Issue #306, opened 2021-04-21
Enhancement: Battery app line fills too slow for my taste


qbinary:
I would like to see that the line fills much faster. I would propose a maximum time of a second until the bar should be filled completely.

Maybe one could add a `sin` function based animation so that the bar filling process is accelerated in the beginning and decelerated at the end.

pfeerick:
I would like that also, or perhaps for it to not to animate at all... I don't really see any point in it doing so, and it detracts from it's purpose as an alternate representation of percentage.

qbinary:
In my opinion the functionality should be more important than the UI but an aesthetic UI gives the user a better experience using it. So I think the animation should stay but it definitely needs to be accelerated.

pfeerick:
> but an aesthetic UI gives the user a better experience using it

I agree. But at the same time, wonder how the progress bar animation adds value to the user experience. It's not like it's an animation showing something filling or emptying. It's not like the animation at Windows applies to it's progress bars - which is in place... it animates the skin of the bar, not the bar moving to it's resting place. 

I just feel that a *moving* progress bar indicates something is changing... which it isn't... but I'm happy with whatever goes implemented.

qbinary:
Yeah I see your point and I think you are right. But it would be neat to have some kind of charging animation. The animation could just happen if the watch is charging. Otherwise it just appears filled. That would be a subtle indication which has an actual purpose. But the time has to be reduced one way or the other.

JF002:
Fixed in #470

PR #470, opened 2021-07-07
LVGL use system tick

### LINKED ISSUE #435: LVGL Tick gets incremented too fast
LVGL tick tells LVGL how much time has passed.
LVGL tick is incremented in: https://github.com/JF002/InfiniTime/blob/e90e8c7426f9a44e696b208a5b393a60476a3d43/src/main.cpp#L180
According to LVGL documentation, this should happen once every millisecond at most.
https://docs.lvgl.io/latest/en/html/porting/tick.html

However, examining the code of BatteryInfo.cpp shows that even though this task is set to happen once every 1000ms, it happens much more often. This task creates the animation by adding 1% to the bar, and it gets run many times per second.
https://github.com/JF002/InfiniTime/blob/e90e8c7426f9a44e696b208a5b393a60476a3d43/src/displayapp/screens/BatteryInfo.cpp#L69

I tried to move the increment function to vApplicationTickHook(), like the LVGL documentation says, but that seemed to increment much slower. Closer to once every 10ms, making the task run once every 10 seconds.

According to this line tick rate should have been 1024Hz, but maybe this is something else.
https://github.com/JF002/InfiniTime/blob/e90e8c7426f9a44e696b208a5b393a60476a3d43/src/FreeRTOSConfig.h#L62

Riksu9000:
Instead of ticking a separate lvgl counter, let LVGL use the system time.
For every 1000ms lvgl thinks 1024ms have passed, which is close enough, and much closer than before.

Fixes #435 and #306?

### DIFF
--- a/src/FreeRTOSConfig.h
+++ b/src/FreeRTOSConfig.h
@@ -77,7 +77,7 @@
 #define configENABLE_BACKWARD_COMPATIBILITY     1
 
 /* Hook function related definitions. */
-#define configUSE_IDLE_HOOK            1
+#define configUSE_IDLE_HOOK            0
 #define configUSE_TICK_HOOK            0
 #define configCHECK_FOR_STACK_OVERFLOW 0
 #define configUSE_MALLOC_FAILED_HOOK   0
--- a/src/displayapp/screens/BatteryInfo.cpp
+++ b/src/displayapp/screens/BatteryInfo.cpp
@@ -9,11 +9,6 @@ static void lv_update_task(struct _lv_task_t* task) {
   user_data->UpdateScreen();
 }
 
-static void lv_anim_task(struct _lv_task_t* task) {
-  auto user_data = static_cast<BatteryInfo*>(task->user_data);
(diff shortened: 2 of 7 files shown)


JF002:
Awesome! I didn't know LVGL provided a way to avoid that call to `lv_tick_inc()`! This solution is a lot more elegant and accurate than the original one! :1st_place_medal:

JF002:
Oops! I fixed the conflicts and merge the branch manually, and it looks like Github is not aware of that...

JF002:
@Riksu9000  Could you confirm all your changes have been correctly merged before I close this PR?

Riksu9000:
Everything seems correct and working.

JF002:
Thanks, and sorry for the confusion about this strange merge!

Riksu9000:
Actually there might be a new issue that comes up from the combination of this and the displayapp delay change. Paddle is now more laggy. This is odd as Paddle shouldn't be using lvgl time.

Issue #364, opened 2021-05-18
Set the time manually on the watch


km6thj:
Set the time using the watch only, without having to connect to a companion app.

jarbus:
This would be really nice. I'm having trouble with companion apps and this feature would at least make it usable without a companion

Riksu9000:
This feature has been merged and will be in the next release.

km6thj:
Thank you folks, that's awesome!

PR #472, opened 2021-07-09
Set date and time manually

timaios:
This PR adds two screens for setting date and time manually (solution for issue #364).

The screens are accessible through a third Settings page:
datetime-settings

Here's how the time is set:
You set the hours and the minutes, wait until the next full minute and press the Set button to set the time.
datetime-set-time

And this is how the date can be set:
datetime-set-date

In both screens there is visual feedback when the Set button has done its work. The button gets disabled (gray text) until you change a value on the screen.
datetime-set-disabled

### DIFF
--- a/src/displayapp/screens/settings/SettingSetDate.cpp
+++ b/src/displayapp/screens/settings/SettingSetDate.cpp
@@ -1,0 +1,272 @@
+#include "SettingSetDate.h"
+#include <lvgl/lvgl.h>
+#include <hal/nrf_rtc.h>
+#include <nrf_log.h>
+#include "displayapp/DisplayApp.h"
+#include "displayapp/screens/Symbols.h"
+
+#define POS_X_DAY   -72
+#define POS_X_MONTH   0
+#define POS_X_YEAR   72
+#define POS_Y_PLUS  -50
+#define POS_Y_TEXT   -6
+#define POS_Y_MINUS  40
+
+using namespace Pinetime::Applications::Screens;
+
+namespace {
+  static void event_handler(lv_obj_t * obj, lv_event_t event) {
+    SettingSetDate* screen = static_cast<SettingSetDate *>(obj->user_data);
+    screen->HandleButtonPress(obj, event);
+  }
+}
+
+SettingSetDate::SettingSetDate(
+  Pinetime::Applications::DisplayApp *app, Pinetime::Controllers::DateTime &dateTimeController) :
+  Screen(app),
+  dateTimeController {dateTimeController}
+{
+
+  lv_obj_t * container1 = lv_cont_create(lv_scr_act(), nullptr);
+
+  //lv_obj_set_style_local_bg_color(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(0x111111));
+  lv_obj_set_style_local_bg_opa(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, LV_OPA_TRANSP);
+  lv_obj_set_style_local_pad_all(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, 10);
+  lv_obj_set_style_local_pad_inner(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, 5);
+  lv_obj_set_style_local_border_width(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, 0);
+  lv_obj_set_pos(container1, 30, 60);
+  lv_obj_set_width(container1, LV_HOR_RES - 50);
+  lv_obj_set_height(container1, LV_VER_RES - 60);
+  lv_cont_set_layout(container1, LV_LAYOUT_COLUMN_LEFT);
+
+  lv_obj_t * title = lv_label_create(lv_scr_act(), NULL);  
+  lv_label_set_text_static(title, "Set current date");
+  lv_label_set_align(title, LV_LABEL_ALIGN_CENTER);
+  lv_obj_align(title, lv_scr_act(), LV_ALIGN_IN_TOP_MID, 15, 15);
+
+  lv_obj_t * icon = lv_label_create(lv_scr_act(), NULL);
+  lv_obj_set_style_local_text_color(icon, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_ORANGE);
+  
+  lv_label_set_text_static(icon, Symbols::clock);
+  lv_label_set_align(icon, LV_LABEL_ALIGN_CENTER);
+  lv_obj_align(icon, title, LV_ALIGN_OUT_LEFT_MID, -10, 0);
+
+  dayValue = static_cast<int>(dateTimeController.Day());
+  lblDay = lv_label_create(lv_scr_act(), NULL);
+  //lv_obj_set_style_local_text_font(lblDay, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, &jetbrains_mono_42);
+  lv_label_set_text_fmt(lblDay, "%d", dayValue);
+  lv_label_set_align(lblDay, LV_LABEL_ALIGN_CENTER);
+  lv_obj_align(lblDay, lv_scr_act(), LV_ALIGN_CENTER, POS_X_DAY, POS_Y_TEXT);
+
+  monthValue = static_cast<int>(dateTimeController.Month());
+  lblMonth = lv_label_create(lv_scr_act(), NULL);
+  //lv_obj_set_style_local_text_font(lblMonth, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, &jetbrains_mono_42);
+  UpdateMonthLabel();
+
+  yearValue = static_cast<int>(dateTimeController.Year());
+  if (yearValue < 2021)
+    yearValue = 2021;
+  lblYear = lv_label_create(lv_scr_act(), NULL);
+  //lv_obj_set_style_local_text_font(lblYear, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, &jetbrains_mono_42);
+  lv_label_set_text_fmt(lblYear, "%d", yearValue);
+  lv_label_set_align(lblYear, LV_LABEL_ALIGN_CENTER);
+  lv_obj_align(lblYear, lv_scr_act(), LV_ALIGN_CENTER, POS_X_YEAR, POS_Y_TEXT);
+
+  btnDayPlus = lv_btn_create(lv_scr_act(), NULL);
+  btnDayPlus->user_data = this;
+  lv_obj_set_size(btnDayPlus, 50, 40);
+  lv_obj_align(btnDayPlus, lv_scr_act(), LV_ALIGN_CENTER, POS_X_DAY, POS_Y_PLUS);
+  lv_obj_set_style_local_value_str(btnDayPlus, LV_BTN_PART_MAIN, LV_STATE_DEFAULT, "+");
+  lv_obj_set_event_cb(btnDayPlus, event_handler);
+
+  btnDayMinus = lv_btn_create(lv_scr_act(), NULL);
+  btnDayMinus->user_data = this;
+  lv_obj_set_size(btnDayMinus, 50, 40);
+  lv_obj_align(btnDayMinus, lv_scr_act(), LV_ALIGN_CENTER, POS_X_DAY, POS_Y_MINUS);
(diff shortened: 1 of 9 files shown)


src/displayapp/screens/settings/SettingSetDate.cpp
```diff
@@ -0,0 +1,272 @@
...
+                               static_cast<uint8_t>(monthValue),
+                               static_cast<uint8_t>(dayValue),
+                               0,
+                               dateTimeController.Hours(),
+                               dateTimeController.Minutes(),
+                               dateTimeController.Seconds(),
+                               nrf_rtc_counter_get(portNRF_RTC_REG));
+    lv_btn_set_state(btnSetTime, LV_BTN_STATE_DISABLED);
+  }
+}
```
Riksu9000, line 200: Add a check to the top to reduce duplicates. CLICKED event is preferred.
```
if (event != LV_EVENT_CLICKED) {
  return;
}
```
The if statements should be turned into a switch statement.

src/displayapp/screens/settings/SettingSetDate.cpp
```diff
@@ -0,0 +1,272 @@
...
+  btnYearMinus->user_data = this;
+  lv_obj_set_size(btnYearMinus, 50, 40);
+  lv_obj_align(btnYearMinus, lv_scr_act(), LV_ALIGN_CENTER, POS_X_YEAR, POS_Y_MINUS);
+  lv_obj_set_style_local_value_str(btnYearMinus, LV_BTN_PART_MAIN, LV_STATE_DEFAULT, "-");
+  lv_obj_set_event_cb(btnYearMinus, event_handler);
+
+  btnSetTime = lv_btn_create(lv_scr_act(), NULL);
+  btnSetTime->user_data = this;
+  lv_obj_set_size(btnSetTime, 70, 40);
+  lv_obj_align(btnSetTime, lv_scr_act(), LV_ALIGN_CENTER, 0, 90);
```
Riksu9000, line 120: Use LV_ALIGN_IN_BOTTOM_MID instead. Make the button wider (120px?) and taller if possible to make it easier to press.

src/displayapp/screens/settings/SettingSetDate.cpp
```diff
@@ -0,0 +1,272 @@
...
+    case 12:
+      lv_label_set_text_static(lblMonth, "Dec");
+      break;
+    default:
+      lv_label_set_text_static(lblMonth, "---");
+      break;
+  }
+  lv_label_set_align(lblMonth, LV_LABEL_ALIGN_CENTER);
+  lv_obj_align(lblMonth, lv_scr_act(), LV_ALIGN_CENTER, POS_X_MONTH, POS_Y_TEXT);
+}
```
Riksu9000, line 271: This already exists in https://github.com/JF002/InfiniTime/blob/61a4642221fc9fcab6889221e7bf9c29778589f2/src/components/datetime/DateTimeController.cpp#L119
Can you use that instead?

src/displayapp/screens/settings/SettingSetTime.cpp
```diff
@@ -0,0 +1,174 @@
...
+    dateTimeController.SetTime(dateTimeController.Year(),
+                               static_cast<uint8_t>(dateTimeController.Month()),
+                               dateTimeController.Day(),
+                               static_cast<uint8_t>(dateTimeController.DayOfWeek()),
+                               static_cast<uint8_t>(hoursValue),
+                               static_cast<uint8_t>(minutesValue),
+                               0,
+                               nrf_rtc_counter_get(portNRF_RTC_REG));
+    lv_btn_set_state(btnSetTime, LV_BTN_STATE_DISABLED);
+  }
```
Riksu9000, line 173: Same thing here.

src/displayapp/screens/settings/Settings.cpp
```diff
@@ -58,5 +61,15 @@ std::unique_ptr<Screen> Settings::CreateScreen2() {
...
 
-  return std::make_unique<Screens::List>(1, 2, app, settingsController, applications);
+  return std::make_unique<Screens::List>(1, 3, app, settingsController, applications);
+}
+
+std::unique_ptr<Screen> Settings::CreateScreen3() {
+
+  std::array<Screens::List::Applications, 4> applications {{
+    {Symbols::clock, "Set date", Apps::SettingSetDate},
+    {Symbols::clock, "Set time", Apps::SettingSetTime},
```
Riksu9000, line 71: Maybe the actual settings should be grouped together on top. So raise this to at least below Steps in the list?

src/displayapp/screens/settings/SettingSetTime.cpp
```diff
@@ -0,0 +1,174 @@
...
+  lv_obj_align(btnMinutesMinus, lv_scr_act(), LV_ALIGN_CENTER, 0, 40);
+  lv_obj_set_style_local_value_str(btnMinutesMinus, LV_BTN_PART_MAIN, LV_STATE_DEFAULT, "-");
+  lv_obj_set_event_cb(btnMinutesMinus, event_handler);
+
+  btnSetTime = lv_btn_create(lv_scr_act(), NULL);
+  btnSetTime->user_data = this;
+  lv_obj_set_size(btnSetTime, 70, 40);
+  lv_obj_align(btnSetTime, lv_scr_act(), LV_ALIGN_CENTER, 0, 90);
+  lv_obj_set_style_local_value_str(btnSetTime, LV_BTN_PART_MAIN, LV_STATE_DEFAULT, "Set");
+  lv_obj_set_event_cb(btnSetTime, event_handler);
```
Riksu9000, line 112: Same here.

src/displayapp/screens/settings/SettingSetDate.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+  //lv_obj_set_style_local_bg_color(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(0x111111));
+  lv_obj_set_style_local_bg_opa(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, LV_OPA_TRANSP);
+  lv_obj_set_style_local_pad_all(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, 10);
+  lv_obj_set_style_local_pad_inner(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, 5);
+  lv_obj_set_style_local_border_width(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, 0);
+  lv_obj_set_pos(container1, 30, 60);
+  lv_obj_set_width(container1, LV_HOR_RES - 50);
+  lv_obj_set_height(container1, LV_VER_RES - 60);
+  lv_cont_set_layout(container1, LV_LAYOUT_COLUMN_LEFT);
```
Riksu9000, line 40: Does this do anything?

src/displayapp/screens/settings/SettingSetTime.cpp
```diff
@@ -0,0 +1,174 @@
...
+  //lv_obj_set_style_local_bg_color(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(0x111111));
+  lv_obj_set_style_local_bg_opa(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, LV_OPA_TRANSP);
+  lv_obj_set_style_local_pad_all(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, 10);
+  lv_obj_set_style_local_pad_inner(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, 5);
+  lv_obj_set_style_local_border_width(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, 0);
+  lv_obj_set_pos(container1, 30, 60);
+  lv_obj_set_width(container1, LV_HOR_RES - 50);
+  lv_obj_set_height(container1, LV_VER_RES - 60);
+  lv_cont_set_layout(container1, LV_LAYOUT_COLUMN_LEFT);
+
```
Riksu9000, line 34: Same here.

Riksu9000 (commented):
This is great for making the watch usable independent of a phone! I haven't tested the code yet. Here are some possible optimizations to the code.

src/displayapp/screens/settings/SettingSetDate.cpp
timaios: Added the check on the method's top.
But I'm not sure if it's possible to use a switch statement for determining the clicked button - switch is only for comparisons with literals, right?

src/displayapp/screens/settings/SettingSetDate.cpp
timaios: Done.

src/displayapp/screens/settings/SettingSetDate.cpp
timaios: This was left from the code that I "stole" from SettingSteps. I've removed it and the code still works.

src/displayapp/screens/settings/SettingSetDate.cpp
Riksu9000: Oh whoops, well we can use else if instead to skip the other checks.

timaios: Ok, I did that when I removed the superfluous event checks in the `if` statements. :)

src/displayapp/screens/settings/SettingSetTime.cpp
timaios: Done.

src/displayapp/screens/settings/SettingSetTime.cpp
timaios: This was left from the code that I "stole" from SettingSteps. I've removed it and the code still works.

src/displayapp/screens/settings/SettingSetTime.cpp
timaios: Done.

src/displayapp/screens/settings/Settings.cpp
timaios: Sorry, I don't understand what you mean by grouping them together on top. Should I just put the two entries on the second page and move the succeeding entries to the third page? Or is there a way to actually group the two into a single entry?

Riksu9000: > Should I just put the two entries on the second page and move the succeeding entries to the third page?

Yes. I think that order would make more sense. It might also make sense to group them into a single entry, but that's a separate thing to think about.

src/displayapp/screens/settings/SettingSetDate.cpp
timaios: Done. I had to add some public static methods (overloaded variants of the existing methods) to `DateTime` to make the arrays accessible in the screens.

src/displayapp/screens/settings/Settings.cpp
timaios: Done. I've moved both entries just between Steps and Battery on the second page. Firmware and About dropped to the third page.

src/displayapp/screens/settings/SettingSetTime.cpp
```diff
@@ -0,0 +1,160 @@
...
+    lv_btn_set_state(btnSetTime, LV_BTN_STATE_RELEASED);
+  }
+  else if (object == btnMinutesMinus) {
+    minutesValue--;
+    if (minutesValue < 0)
+      minutesValue = 59;
+    lv_label_set_text_fmt(lblMinutes, "%02d", minutesValue);
+    lv_obj_align(lblMinutes, lv_scr_act(), LV_ALIGN_CENTER, 0, -6);
+    lv_btn_set_state(btnSetTime, LV_BTN_STATE_RELEASED);
+  }
```
Riksu9000, line 147: It should be possible to use `lv_obj_realign(obj)` or `lv_obj_set_auto_realign(obj, true)` here and in the other file, so the position doesn't need to be updated. Also in this file the positions aren't set by `#define`

timaios: `lv_obj_set_auto_realign` worked, so I changed the code in both files accordingly.
The positions in `SettingSetTime` are now determined by `#define`s.

Btw, thanks for your feedback. I usually code using Qt and this is my "first contact" with LVGL...

JF002:
Thanks for this PR and for the review! It'll probably take some time for me to review it as there are many other PRs to review, but it looks quite nice :+1:  !

ocdtrekkie:
As a note @JF002, the lack of this capability was super frustrating today as someone who only had an iOS device handy at the time.

I'm really glad someone has this fix, for new PineTime users in the future, this PR should be a priority in the hopes it can make it into a build that reaches new PineTime runs ASAP.

PR #473, opened 2021-07-09
Add task state info to SystemInfo

Riksu9000:
Show task state in systeminfo
R: Ready or running (Displayapp is probably always the one actually running)
B: Blocked
S: Suspended
D: Deleted
I: Invalid state

stateinfo


kieranc:
I've just tested this, it works as expected

JF002:
Thanks :+1:

PR #474, opened 2021-07-09
Suspend displayapp when it's not needed

Riksu9000:
Currently all FreeRTOS tasks have idle priority and they never suspend themselves other than for delay. AFAIK This wastes processing time as a task keeps running even if it does nothing until the kernel swaps out the task.

This PR makes displayapp loop once every 20ms, establishing a refresh rate of ~50Hz, and suspend as soon as it is done with all of it's work.

This has a visible effect on Paddle game, making it run smoother and faster.

EDIT: I think this behaviour is actually called blocking in FreeRTOS, but I also made displayapp actually suspend when the device goes to sleep.

EDIT2: The refresh rate probably ends up being 51.2Hz because 20ms gets rounded to 20 ticks and 1024Hz / 20 ticks = 51.2Hz

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -80,7 +80,7 @@ DisplayApp::DisplayApp(Drivers::St7789& lcd,
     timerController {timerController} {
 }
 
-void DisplayApp::Start() {
+TaskHandle_t DisplayApp::Start() {
   msgQueue = xQueueCreate(queueSize, itemSize);
 
   // Start clock when smartwatch boots
@@ -89,6 +89,7 @@ void DisplayApp::Start() {
   if (pdPASS != xTaskCreate(DisplayApp::Process, "displayapp", 800, this, 0, &taskHandle)) {
     APP_ERROR_HANDLER(NRF_ERROR_NO_MEM);
   }
+  return taskHandle;
 }
 
 void DisplayApp::Process(void* instance) {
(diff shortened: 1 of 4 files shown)


JF002:
> Currently all FreeRTOS tasks have idle priority and they never suspend themselves other than for delay. AFAIK This wastes processing time as a task keeps running even if it does nothing until the kernel swaps out the task.

All task are built around a message queue. When they are waiting for a message in the queue, the kernel put them in the WAITING state and should not waste any resources.

For example, DisplayApp uses a timeout of 20ms when the display is ON and *max value* when it is OFF. In theory, Display app should then run at least once every 20ms when the watch is awake and never when the watch is sleeping.

This is the theory... In practice, it looks like I'm wrong:

> This has a visible effect on Paddle game, making it run smoother and faster.

Any idea why these changes give better results than the initial implementation?

Riksu9000:
> All task are built around a message queue. When they are waiting for a message in the queue, the kernel put them in the WAITING state and should not waste any resources.

You're right. I may have been mistaken. I wonder if the difference then is because of the `vTaskDelayUntil()` function. If the loop takes 5ms to run, `xQueueReceive()` will wait for 20ms, making the entire loop take 25ms, whereas `vTaskDelayUntil()` only waits for 15ms, making the loop run at the intended frequency.

We could instead subtract the time spent from queueTimeout, which would probably achieve the same thing. I'll try that and if it works, I guess I'll scrap this idea.

Riksu9000:
That's what it was.. I've opened a new PR #482 for this new more reasonable way of achieving the same thing.

PR #475, opened 2021-07-09
Update contribute.md

Riksu9000:
Fixed and changed some spellings in contribute.md

I noticed that there is also this file. Do we need this separate one?
https://github.com/JF002/InfiniTime/blob/develop/CONTRIBUTING.md


JF002:
> Do we need this separate one?

I don't think so. CONTRIBUTING.md is suggested by github as community guideline, but it's not mandatory. Personally, I prefer to put all documentation files into */doc*, but github recommand CONTRIBUTING.md as community guideline.

We can either use only CONTRIBUTING.md, or link *doc/contributing.md* into CONTRIBUTING.md to make everyone happy.

What are your thoughts about this?

Riksu9000:
I think it should be linked. Do I make the changes here? I can add the info about autoformatting from CONTRIBUTING.md to doc/contributing.md, but I don't think there's anything else that needs to be added.

JF002:
> I think it should be linked. Do I make the changes here? I can add the info about autoformatting from CONTRIBUTING.md to doc/contributing.md, but I don't think there's anything else that needs to be added.

Yes, that would be nice if you could do that, thanks!

Riksu9000:
Looks like GitHub doesn't handle symbolic links very nicely. Is there a better way to link it or do I just move it anyway?

Avamander:
> Looks like GitHub doesn't handle symbolic links very nicely. Is there a better way to link it or do I just move it anyway?

Git supports symbolic links, I assume that's what you commited to the repository? If so then it's fine, the UI can't necessarily resolve it but in the end it's going to be humans who will know to check the right folder should they be using the web UI.

JF002:
When I talk about linking, I was thinking about adding a link (url) in one of the file to the second one :-)
But if symbolic links work, why not!

PR #476, opened 2021-07-10
Refactor trig functions into LUT

jonvmey:
Replaced the use of the standard library trig functions with a LUT-based
implementation instead. The standard library implementations produce
more accurate results but the usage here doesn't need that. This ends up
saving nearly 7kB of binary size.

### DIFF
--- a/src/displayapp/screens/WatchFaceAnalog.cpp
+++ b/src/displayapp/screens/WatchFaceAnalog.cpp
@@ -5,26 +5,43 @@
 #include "Symbols.h"
 #include "NotificationIcon.h"
 
-#include <cmath>
-
 LV_IMG_DECLARE(bg_clock);
 
 using namespace Pinetime::Applications::Screens;
 
-#define HOUR_LENGTH   70
-#define MINUTE_LENGTH 90
-#define SECOND_LENGTH 110
-#define PI            3.14159265358979323846
+namespace {
 
-// ##
-static int16_t coordinate_x_relocate(int16_t x) {
-  return ((x) + LV_HOR_RES / 2);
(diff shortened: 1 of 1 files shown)


jonvmey:
Final tally for binary size change with this diff is exactly 7000 bytes less. I still need to load this on my watch tomorrow to verify that everything's good but I think it's all correct.

Riksu9000:
There already seems to exist this in LVGL. Can you use this instead?
https://github.com/joaquimorg/lvgl/blob/23430cf20e32294549fff9b2879a9466dacc19bb/src/lv_misc/lv_math.c

Riksu9000:
There seem to be some issues with this.

https://user-images.githubusercontent.com/37774658/125158105-77ab3980-e177-11eb-9eb3-9ef15c415b20.mp4

jonvmey:
Yeah, that definitely looks wrong, I'll take another look to see what I messed up.

The implementation in LVGL doesn't appear to be part of its exposed API so I wasn't sure if we wanted to use that directly.

Avamander:
> The implementation in LVGL doesn't appear to be part of its exposed API so I wasn't sure if we wanted to use that directly.

If they remove it I guess we can think of a solution. For now I suspect it'd be fine. Quite a bit less to maintain.

jonvmey:
Updated the changes to use LVGL's _lv_trigo_sin function instead. The issue @Riksu9000 mentioned should also be fixed now.

Riksu9000:
Seems to work fine now.

text goes down 7160 and data goes up 4 compared to develop.

JF002:
Looks good to me, thanks for this PR, this nice review and these 7160 free bytes in memory ;)

PR #478, opened 2021-07-10
Feature add setting to configure the push button action

coxtor:
A new settings option to let the user decide how the push button works. For now the options are:

- go back one screen
- return to home screen

I hope you find this useful. I personally like it to be able to get back to the watch face quickly.
PXL_20210710_152622150


Riksu9000:
This seems like a good idea, but what if there was a button double press action. A single press could go back one screen and a double press could go back all the way. Then we wouldn't need a setting, unless we made that configurable as well.

kieranc:
> This seems like a good idea, but what if there was a button double press action. A single press could go back one screen and a double press could go back all the way. Then we wouldn't need a setting, unless we made that configurable as well.

I suspect a double press action would slow down navigation if you were using single presses to go back, or risk doing the wrong thing. Maybe a long press to go back to the watchface and single press to go back?

coxtor:
Great, I was also thinking about implementing multiway push button actions, awesome that #480 took care of that. 

In general I still think that giving the user the choice of what the button does is a good feature in terms of customizability. However, I'm not sure how problematic to much customizability and to many features would impact usability and the memory footprint.

I think most of the navigation should actually be done by swiping, swipe right to go back exclusively, thus leaving more options for the button.

PR #480, opened 2021-07-11
Add new button handler that can handle long press and double click

Riksu9000:
https://user-images.githubusercontent.com/37774658/125189055-0f278f80-e23f-11eb-95ca-4bd57fbdbf92.mp4

Basic implementation for detecting different button events.

I'm unsure what to do with the recovery source files list. I added the new files there. Is this fine?

There is a new issue in SystemInfo, because now there are eight tasks, and only seven can fit on screen.

EDIT: We could make waking up by pressing the button faster by uncommenting the piece of code, but then we can't make the watch handle different events while sleeping. For example if we wanted to enable flashlight by long pressing the button while the watch is sleeping, we would need to remove that code. Then again raise to wake would make this inconsistent.

EDIT2: I enabled the faster wakeup code. I also thought about making this task use a message queue, but that doesn't seem necessary, as the task needs to read the button state anyway and can't rely on messages only.

EDIT3: This probably fixes the issue with the screen sometimes turning on and off after a single button push.


kieranc:
I just commented on another PR that a long press function would be useful, and now I see this. You've been busy Riksu9000! 
I'll do some testing when I get back to my desk, but it won't be for about a week.

kieranc:
Tested, works well, I think it could do with a settings page to choose what single/long/double press does as different people probably want different things. I think the current single and long press behaviour are ideal for me, but if the double press could be configured to launch a specific app (music, navigation, timer...) that would be amazing.

Riksu9000:
> Tested, works well, I think it could do with a settings page to choose what single/long/double press does as different people probably want different things. I think the current single and long press behaviour are ideal for me, but if the double press could be configured to launch a specific app (music, navigation, timer...) that would be amazing.

Thanks for testing. The double press launching notifications is mainly just to demonstrate the ability to detect the event. Having it configurable might be a good idea, but that should be a separate PR after this.

There's also #487. We need to choose which shortcut should be used for favorite apps, or make them work together somehow.

hubmartin:
Works perfectly. Really makes browsing menus much more pleasant! Together with #586 the wake-up on press is instant 11 ms.
Thanks.

hubmartin:
Just noticed just a small issue. I have display timeout set to 5 seconds. When I press and hold button in different times like 2-3 seconds before screen sleep timeout, then the screen dims a bit (this is a feature few seconds before going to sleep) and then it goes to the About screen.
Sometimes screen goes off for a fraction of second, sometimes it turns off.
I guess you need to add display timeout reset to some parts. Preferably every time user just presses and holds the button.
Otherwise works perfectly.

Riksu9000:
Good find. Fixed by calling ReloadIdleTimer on press and release. It can still dim if you hold the button for more than three seconds, but I think that's acceptable as nothing happens after two seconds and this avoids calling ReloadIdleTimer continuously.

I also noticed that the way SystemTask processed the button events was a bit weird, so I moved everything to messages, which is futureproof in case we want SystemTask to process the button events in some way.

hubmartin:
Thanks, tested and works fine.

hubmartin:
The last improvement fd39ba3 brings a bug. The previous code worked fine without this bug.

When the watch is turned on and brightness is max, then I press button to go to sleep. But during the fading down (or maybe a while after screen is off) I press the button again. The watch freezes and resets.

Video
https://photos.app.goo.gl/fzrYU1qYeXzUXLeH8

Riksu9000:
When I moved the functions to messages, I left out a check for `isGoingToSleep`. Turns out it is necessary so I added it back.

hubmartin:
That fixed it, thanks.

kieranc:
It seems like wake-sleep-wake quickly (either button or double tap) misses the second wake event - if i press the button 3 times fairly quickly, the device wakes, then sleeps, but doesn't wake again. Half a second later it will wake just fine. Could this be the double press behaviour interfering with the wake behaviour?

Riksu9000:
> It seems like wake-sleep-wake quickly (either button or double tap) misses the second wake event - if i press the button 3 times fairly quickly, the device wakes, then sleeps, but doesn't wake again. Half a second later it will wake just fine. Could this be the double press behaviour interfering with the wake behaviour?

I think that happens regardless of this PR. There is for example this code that stops execution for half a second each time the screen is turned off.
```
void St7789::DisplayOff() {
  WriteCommand(static_cast<uint8_t>(Commands::DisplayOff));
  nrf_delay_ms(500);
}
```

kieranc:
> I think that happens regardless of this PR. There is for example this code that stops execution for half a second each time the screen is turned off.
You are correct!

geekbozu:
> 
> 
> > It seems like wake-sleep-wake quickly (either button or double tap) misses the second wake event - if i press the button 3 times fairly quickly, the device wakes, then sleeps, but doesn't wake again. Half a second later it will wake just fine. Could this be the double press behaviour interfering with the wake behaviour?
> 
> I think that happens regardless of this PR. There is for example this code that stops execution for half a second each time the screen is turned off.
> 
> ```
> void St7789::DisplayOff() {
>   WriteCommand(static_cast<uint8_t>(Commands::DisplayOff));
>   nrf_delay_ms(500);
> }
> ```

Why is this even doing this? This feels...wrong. does nrf_delay even yeild to the freertos?

Riksu9000:
> > > It seems like wake-sleep-wake quickly (either button or double tap) misses the second wake event - if i press the button 3 times fairly quickly, the device wakes, then sleeps, but doesn't wake again. Half a second later it will wake just fine. Could this be the double press behaviour interfering with the wake behaviour?
> > 
> > 
> > I think that happens regardless of this PR. There is for example this code that stops execution for half a second each time the screen is turned off.
> > ```
> > void St7789::DisplayOff() {
> >   WriteCommand(static_cast<uint8_t>(Commands::DisplayOff));
> >   nrf_delay_ms(500);
> > }
> > ```
> 
> Why is this even doing this? This feels...wrong. does nrf_delay even yield to RreeRTOS? That feels...bad.

Yeah I doubt that's necessary. Someone should look into that..

hubmartin:
This PR is working perfectly fine and I would add it to the next release pleaase, 'cause I got used to it :)
I really like the wakeup/debounce speed improvement and doubleclick to the home screen.
(configuraton of button long press could be done i another PR).

hubmartin:
Just an idea when I see #708 App chooser made by @coxtor.

Since we have a settings page to choose Favorite app, and this button handler will sooner or later also need some App chooser for doubleclick and looong buton press - wouldn't be the best solution to create a "shared app/settings chooser" which could be used by other apps? That way we could use single code/dialog to set:
- Favorite app
- Button doubleclick
- Button looong press

It might not display only apps but also settings. Now I'm thinking - wouldn't be possible to use current menu for apps and settings? It might work the same way, but if called with some flag, it might display some text in the top bar like "Choose a app" and when selected - it will return to previous app instead of running the app. The same may apply for Settings dialog.
All this should be done in separate PR so this one gets merged ASAP.

This might deduplicate lot of things. What do you think?

trman:
Hi @Riksu9000 , i think that this pull can be either closed or be put as draft since the new is now on 1.7.0.
Unless @JF002 has not yet chooses which pul to merge?

Riksu9000:
> Hi @Riksu9000 , i think that this pull can be either closed or be put as draft since the new is now on 1.7.0. Unless @JF002 has not yet chooses which pul to merge?

It's likely that the new one will be merged instead of this, but both of these PRs are functional, and as one is merged, the other one will be closed.

JF002:
I'm closing this one as we've just merged #782.

PR #481, opened 2021-07-11
Notify battery level on BLE

JF002:
Andreas from Gadgetbridge added support for notifications for battery level : https://codeberg.org/Freeyourgadget/Gadgetbridge/commit/603050b7cc30f6ef11ffaf73f3f88bf17a6922b6.
This will allow gadgetbridge (and other companion apps that supports it) to display the current battery level and update it with up to date values. Gadgetbridge also provides a graph of the battery level, low battery notification,...

InfiniTime just needs to send a BLE NOTIFY message when the battery level changes.

- [x] Notify battery level every 10 minutes when connected to a BLE host
- [x] Refactor battery percent : only use uint8_t to store the battery % remaining.

### DIFF
--- a/src/components/battery/BatteryController.h
+++ b/src/components/battery/BatteryController.h
@@ -19,7 +19,7 @@ namespace Pinetime {
      insert member function overwrites the next data to the current
     HEAD and moves the HEAD to the newly inserted value.
     */
-      void insert(const int num) {
+      void Insert(const uint8_t num) {
         head %= cap;
         arr[head++] = num;
         if (sz != cap) {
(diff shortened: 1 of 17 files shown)


JF002:
Here's the result after ~24 of usage (ble was disabled during the night)

Screenshot_20210712-192830

src/systemtask/SystemTask.h
```diff
@@ -135,6 +135,8 @@ namespace Pinetime {
       void GoToRunning();
       void UpdateMotion();
       bool stepCounterMustBeReset = false;
+      static constexpr TickType_t batteryNotificationPeriod = 1000 * 60 * 10; // 1 tick ~= 1ms. 1ms * 60 * 10 = 10 minutes
```
Riksu9000, line 138: This should use `pdMS_TO_TICKS()`

src/components/battery/BatteryController.h
```diff
@@ -46,8 +46,11 @@ namespace Pinetime {
       void Init();
       void Update();
 
-      int PercentRemaining() const {
-        return percentRemainingBuffer.GetAverage();
+      uint8_t PercentRemaining() const {
+        auto avg = percentRemainingBuffer.GetAverage();
+        avg = std::min(avg, static_cast<uint8_t>(100));
+        avg = std::max(avg, static_cast<uint8_t>(0));
```
Riksu9000, line 52: The values in the buffer are already limited before they're inserted, so these shouldn't be needed.

PR #482, opened 2021-07-11
Adjust displayapp delay to compensate time spent

Riksu9000:
Makes displayapp run every 20ms more consistently by subtracting time spent executing from the next delay.

The simpler way to achieve the same results as #474. Whoops.


JF002:
Nice! :+1: 

> The simpler way to achieve the same results as #474. Whoops.
Don't worry, that's what the review are for : challenge each others to find the best solution!

Paddle game looks indeed a bit faster and smoother!

PR #483, opened 2021-07-12
Fix misconfigured ADC and remove now unnecessary filtering

Riksu9000:
There's this table in the SoC datasheet. The battery is connected to the SoC through a 1M resistor, so we must use the longest acquisition time.

After this change, the reported battery voltage is about 100mV higher! The raw measurements are now equally as low noise as the filtered output used before, so the filtering is now unnecessary.

EDIT: If there is still too much noise, hardware oversampling should be used instead of manually averaging samples.


hatmajster:
Tested this today, indeed it seems to be working as You say. It was even the first time I was able to charge pinetime to 100%! (usually it went up to around 87%)

src/components/battery/BatteryController.cpp
```diff
@@ -60,22 +59,21 @@ void Battery::SaadcEventHandler(nrfx_saadc_evt_t const* p_event) {
...
-    percentRemaining = (voltage - battery_min) * 100 / (battery_max - battery_min);
-    percentRemaining = std::max(percentRemaining, 0);
-    percentRemaining = std::min(percentRemaining, 100);
-    percentRemainingBuffer.Insert(percentRemaining);
-
-    samples++;
-    if (samples > percentRemainingSamples) {
-      nrfx_saadc_uninit();
-      isReading = false;
+    voltage = p_event->data.done.p_buffer[0] * (0.6 * 8 * 1000) / 1024;
```
deztructor, line 66: Is there a real need in FP calculations here? `6 * 8 * 100` should be equivalent to `(0.6 * 8 * 1000)`, while usage of `0.6` makes code to use (soft) FP calculations (`__floatsidf, __muldf3...`), as also more code is generated.

Avamander: Did you check the disassembly that the compiler doesn't optimize that away?

deztructor: Yes, I did.

deztructor:
```
Dump of assembler code for function Pinetime::Controllers::Battery::SaadcEventHandler(nrfx_saadc_evt_t const*):                                                                                                    
   0x0000b060 <+0>:     push    {r3, r4, r5, lr}                                                                                                                                                                   
   0x0000b062 <+2>:     ldrb    r3, [r1, #0]                                                                                                                                                                       
   0x0000b064 <+4>:     mov     r4, r0                                                                                                                                                                             
   0x0000b066 <+6>:     mov     r5, r1                                                                                                                                                                             
   0x0000b068 <+8>:     cmp     r3, #0                                                                                                                                                                             
   0x0000b06a <+10>:    bne.n   0xb0f4 <Pinetime::Controllers::Battery::SaadcEventHandler(nrfx_saadc_evt_t const*)+148>                                                                                            
   0x0000b06c <+12>:    movs    r1, #1                                                                                                                                                                             
   0x0000b06e <+14>:    bl      0x22878 <nrfx_saadc_buffer_convert>                                                                                                                                                
   0x0000b072 <+18>:    cbz     r0, 0xb078 <Pinetime::Controllers::Battery::SaadcEventHandler(nrfx_saadc_evt_t const*)+24>                                                                                         
   0x0000b074 <+20>:    bl      0x243a0 <app_error_handler_bare>                                                                                                                                                   
   0x0000b078 <+24>:    ldr     r3, [r5, #4]                                                                                                                                                                       
   0x0000b07a <+26>:    ldrsh.w r0, [r3]                                                                                                                                                                           
   0x0000b07e <+30>:    bl      0x41dcc <__floatsidf>                                                                                                                                                              
   0x0000b082 <+34>:    ldr     r3, [pc, #124]  ; (0xb100 <Pinetime::Controllers::Battery::SaadcEventHandler(nrfx_saadc_evt_t const*)+160>)                                                                        
   0x0000b084 <+36>:    movs    r2, #0                                                                                                                                                                             
   0x0000b086 <+38>:    bl      0x41ea0 <__muldf3>                                                                                                                                                                 
   0x0000b08a <+42>:    movs    r2, #0                                                                                                                                                                             
   0x0000b08c <+44>:    ldr     r3, [pc, #116]  ; (0xb104 <Pinetime::Controllers::Battery::SaadcEventHandler(nrfx_saadc_evt_t const*)+164>)                                                                        
   0x0000b08e <+46>:    bl      0x41ea0 <__muldf3>                                                                                                                                                                 
   0x0000b092 <+50>:    bl      0x423d4 <__fixunsdfsi>                                                                                                                                                             
   0x0000b096 <+54>:    uxth    r0, r0                                                                                                                                                                             
   0x0000b098 <+56>:    strh    r0, [r4, #10]                                                                                                                                                                      
...
```

deztructor: I'd not expect that compiler optimizes FP operations because in this case results could differ from results of integer calculations.

deztructor: BTW, I wonder why e.g. `bl      0x41ea0 <__muldf3> ` is generated instead of the FP operation while `-mfloat-abi=hard -mfpu=fpv4-sp-d16 ` is passed...

Riksu9000: Interesting. The reason I changed it is because (0.6 * 8 * 1000) says more than 4800. I did expect it to get optimized.

deztructor: @Riksu9000 Yes, I understood your intent :+1: It is more readable (while, unfortunately, compiler can't do this optimization, most probably due to the reason I supposed). BTW, fractoin could be also a possible way to describe the intent in optimize-able integers: `1000 * 8 * 6 / 10`.

deztructor: And your 600mV solution looks good! :+1:

Avamander: Does it get optimized if `-ffast-math` is given?

deztructor: @Avamander nope

Avamander: Very interesting.

deztructor: BTW, calls to soft-fp functions are made because `0,6` is `double` and nrf52 FPU supports only `float`. Using `0.6f` results in `vmul.f32        s15, s15, s14`. Option `-fsingle-precision-constant` could help to avoid fall-back to soft-fp for cases with accidental `double` constants (while theoretically could result in unexpected results for cases with higher precision (are there any? :thinking: )).

JF002: If you want to keep the details of the calculation (which is a good thing, imo), maybe you can try to declare a ìnt` variable as contexpr
```
static constexpr int  conversionFactor = 0.6 * 8 * 1000 
```
and use this constant in the code ? That way, the code still explicitely details the calculation while ensuring that the variable is an `int` and won't trig floating point processing.

PR #484, opened 2021-07-13
Add new unique icons for meter, wakeup and watch face

Riksu9000:
New icons for meter, wakeup, and **home** screen/watch face.
meter
settings


JF002:
Thanks! :+1:

Issue #462, opened 2021-07-01
Settings list code isn't dynamic.


Riksu9000:
https://github.com/JF002/InfiniTime/blob/develop/src/displayapp/screens/settings/Settings.cpp

In Settings.cpp, there are separate arrays of applications for each screen. This is easy to fix by creating one array of all apps in Settings.h and replacing the old CreateScreenN functions with a single new function that uses this array.

The real issue is when we add a new app and have to create a new screen. This always requires changes in multiple places in the code. It would be better if the code could automatically create enough screens for all apps. Then it would be much easier to add, remove and move around options and maybe even change them at runtime.

I wanted to add a debug menu to the settings for developers. This menu would have to be hidden somehow for non developers. This is much harder to do with the current code reasonably.

All of this applies to ApplicationList as well, so that can also benefit from solving this.

JF002:
The idea behind this implementation is to be simple, and to avoid memory allocation at runtime (dynamic memory allocation). But I agree with you, when you want to add a new app/setting app, you have to manually add the new app , the CreateScreenN function, increase the size of the array, add the symbol in the font, add the app in Apps.h,...

This could certainly be refactored and improved, indeed.

Avamander:
If this is still something that really needs attention, let's reopen.

Riksu9000:
I think it does if we want to separate apps from the firmware. The code would need to be able to dynamically create enough screens to fit any number of apps. Updated the title as this applies to applist as well.

Avamander:
Let's create a clean new issue that lists the potential steps towards fully dynamic watchapps?

In theory you could create a dynamic app menu that's separate from the system's built-ins, which sounds reasonable.

JF002:
If this is a subject some contributors want to work on, I can create a "project" dedicated to the implementation of dynamic apps. That would give a space to think all the steps needed to reach that goal. What do you think?

PR #487, opened 2021-07-13
[POC] Feature Favorite Apps on swipe left

coxtor:
This is more a POC instead of pretty code,  but this PR adds a favorite option when swiping left, including the configurable options.
Issues with this currently:

- No Scrolling, if #442  is merged a whole lot would change
- We should have a central way to query installed apps with icons and a textual description
- There still needs to be a handle that checks whether the return app should be the watch face or the previous menu, however this would currently require a change in every app so I didn't do it to keep this PR more as an idea for now.  [or #478, #480

https://user-images.githubusercontent.com/7712562/125523594-aeaddcfb-67fc-45bf-8246-db844c3d1c76.mp4

]


Avamander:
Just FYI the commit message is not very accurate, it should describe what the commit actually does.

src/displayapp/screens/settings/SettingFavoriteApp.h
```diff
@@ -0,0 +1,31 @@
...
+      private:
+        Controllers::Settings& settingsController;
+        uint8_t optionsTotal;
+        lv_obj_t* cbOption[13];
+        Apps favoriteOptions[13] = {Apps::Music,Apps::Navigation,Apps::StopWatch,Apps::Timer,Apps::None,Apps::FlashLight,Apps::Paint,Apps::Paddle,Apps::Twos,Apps::HeartRate,Apps::Metronome,Apps::Motion,Apps::Steps};   
+        char favoriteNames[13][12] = { "Music","Navigation","Stop Watch","Timer","None","Flash Light","Paint","Paddle","Twos","Heart Rate","Metronome","Motion","Steps"};
+      };
+    }
+  }
+}
```
Avamander, line 31: This file hasn't been formatted with clang-format I think

Avamander (changes requested):
On a first glance this looks nice, but there's a few inconsistencies.

Riksu9000:
>We should have a central way to query installed apps with icons and a textual description

I've mentioned something similar to this in #462. It should be easy to combine the [separate arrays in ApplicationList.cpp](https://github.com/JF002/InfiniTime/blob/57b339707861c5688f5d432f1506a99df6bb0fce/src/displayapp/screens/ApplicationList.cpp#L48) and move them to [ApplicationList.h](https://github.com/JF002/InfiniTime/blob/develop/src/displayapp/screens/ApplicationList.h), where it could be easily queried, without the description though.

>There still needs to be a handle that checks whether the return app should be the watch face or the previous menu

I think this can be done by creating separate entries in the [LoadApp function in DisplayApp.cpp](https://github.com/JF002/InfiniTime/blob/57b339707861c5688f5d432f1506a99df6bb0fce/src/displayapp/DisplayApp.cpp#L259) for when the app is opened from watchface and for when it is opened from settings and they'd have different ReturnApps.

Riksu9000:
Potentially if #492 gets merged, the favorite app could be selected by long pressing on an app in the app list. The selected favorite app could be a different color, or have a small star.

>>There still needs to be a handle that checks whether the return app should be the watch face or the previous menu

>I think this can be done by creating separate entries in the LoadApp function in DisplayApp.cpp

I was thinking of just the settings menu. I realise now that this would also be required for all apps to know whether to return to the app list or watch face, which obviously isn't a great solution. However could `  ReturnApp(Apps::Clock, ...);
` simply be run after this line?
https://github.com/JF002/InfiniTime/blob/3e0974eacdad15a47c2f359cf7f02304806f5513/src/displayapp/DisplayApp.cpp#L192

Izaic:
This would be a really great feature to have for Version 1.4

coxtor:
Hi all, I'm not quite sure what went wrong that this PR has so many commits - however I got to implement the "long tap to select favorite" option. However I need help here - for some reason the watch crashes when I don't use a static Application when saving a favorite (tile.cpp -> apps[buttonId] ) . Can someone help me out here?

kieranc:
Any chance you can clean up the PR so we can have a look more easily? I tested this yesterday and it seems to work well but it's a bit of a pain to merge currently

kieranc:
I hope you'll bring this back, I was testing it because people were asking for it, and it works nicely!

coxtor:
I did some cleaning up and we can continue the discussion on PR #708

PR #488, opened 2021-07-13
Set correct refresh times for LVGL

Riksu9000:
After merging both #470 and #482, a new issue came up. LVGL would refuse to refresh in sync with displayapp, causing lag in some cases.

Here I fixed this issue by setting the refresh times in LVGL to 20ms.

The behaviour of the arc in Metronome has also changed with all these changes, ~~but it should be as intended now.~~

EDIT: There is still one more regression. In metronome, the dropdown menu can't be scrolled. Every value can still be reached by reopening the menu.

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -43,6 +43,8 @@
 #include "displayapp/screens/settings/SettingDisplay.h"
 #include "displayapp/screens/settings/SettingSteps.h"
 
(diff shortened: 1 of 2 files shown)


JF002:
This issue with the dropdown in the metronome app is weird... None of the changes from these PRs we done in this app... It's probably caused by a side effect of these changes...

I'll merge this PR, but feel free to open a new one if you (or someone else) find a solution for the Metronome!

PR #489, opened 2021-07-14
Update touchpad driver

Riksu9000:
Cleaned up unused parts.
`isTouch` was buggy. Replaced with `isValid` and `touching`.
Read information about the chip and display it in SystemInfo in format `ChipId.VendorId.FwVersion`.

info

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -336,7 +336,7 @@ void DisplayApp::LoadApp(Apps app, DisplayApp::FullRefreshDirections direction) 
       break;
     case Apps::SysInfo:
       currentScreen =
-        std::make_unique<Screens::SystemInfo>(this, dateTimeController, batteryController, brightnessController, bleController, watchdog, motionController);
+        std::make_unique<Screens::SystemInfo>(this, dateTimeController, batteryController, brightnessController, bleController, watchdog, motionController, touchPanel);
       ReturnApp(Apps::Settings, FullRefreshDirections::Down, TouchEvents::SwipeDown);
       break;
     case Apps::FlashLight:
@@ -395,7 +395,7 @@ void DisplayApp::PushMessage(Messages msg) {
 
 TouchEvents DisplayApp::OnTouchEvent() {
   auto info = touchPanel.GetTouchInfo();
-  if (info.isTouch) {
+  if (info.isValid) {
     switch (info.gesture) {
       case Pinetime::Drivers::Cst816S::Gestures::SingleTap:
         if (touchMode == TouchModes::Gestures) {
--- a/src/displayapp/screens/SystemInfo.cpp
+++ b/src/displayapp/screens/SystemInfo.cpp
@@ -31,7 +31,8 @@ SystemInfo::SystemInfo(Pinetime::Applications::DisplayApp* app,
                        Pinetime::Controllers::BrightnessController& brightnessController,
                        Pinetime::Controllers::Ble& bleController,
                        Pinetime::Drivers::WatchdogView& watchdog,
-                       Pinetime::Controllers::MotionController& motionController)
+                       Pinetime::Controllers::MotionController& motionController,
+                       Pinetime::Drivers::Cst816S& touchPanel)
   : Screen(app),
     dateTimeController {dateTimeController},
     batteryController {batteryController},
@@ -39,6 +40,7 @@ SystemInfo::SystemInfo(Pinetime::Applications::DisplayApp* app,
     bleController {bleController},
     watchdog {watchdog},
     motionController{motionController},
+    touchPanel{touchPanel},
     screens {app,
              0,
              {[this]() -> std::unique_ptr<Screen> {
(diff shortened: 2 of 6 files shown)


JF002:
> UPDATE: Another PR might replace this soon.
Is this #492  ? If so, can we close this one?

Riksu9000:
#492 doesn't show the information about the chip. I'll update this after #492 is handled in case we want to show this information. If we don't need to show this information, then this can be closed.

JF002:
There's no actual need to display the info about the chip as the Pinetime always use the same chip, but it could help to debug issues with the touchpanel (factory installs another chip, chip fails to initialize,... -> display "???" or "Error" in that case, for example).

Riksu9000:
It seems that this information can be useful to detect faulty hardware #590.

Riksu9000:
I've added error checking on boot. In case the touch controller doesn't respond as expected, this screen will be shown.

error

ildar:
Would you resolve conflicts here or make another PR?

lman0:
> I've added error checking on boot. In case the touch controller doesn't respond as expected, this screen will be shown.
> 
> error

@Riksu9000 i don't think it's useful to add a button in order to "proceed"  because the user will be unable to click on it since the touch controller don't work to begin with.

it's better to remove the button (that's make the user angry since he can't click on it how much he want/try )
and replace it by a 15-30s timer , then proceed to main screen automatically

Riksu9000:
> i don't think it's useful to add a rectangular button in order to "proceed" because the user will be unable to click on it since the touch controller don't work to begin with.

The screen can be closed by pressing the physical button.

lman0:
it's still redundant then because , as user , i will try  first to press the button "proceed "
and then after some angry try , will try to click on the physical button.
the "porceed" button feel like a nagging button and that's not good for someone that already have a bad product

Riksu9000:
> @Riksu9000 it's still redundant then because , as user , i will try first to press the button "proceed "
> and then after some angry try ,
> i will , at last, try to click on the physical button as last try....
> 
> the "proceed" button feel like a nagging button and that's not good for someone that already have a bad product...

This screen can be used to warn about other errors as well, in which case the proceed button makes more sense. Also touch could still just work if the warning is a false positive.

The user might miss the warning entirely if there was a timeout, so I think it should require confirmation from the user.

hubmartin:
As an owner of PineTime and Colmi P8. Acording to this message, this warning screen appears on Colmi P8 everytime after boot.
https://github.com/JF002/InfiniTime/pull/492#issuecomment-929092213 

I see InfiniTime project should be more open to be ported on other devices. I feel like this might hurt/annoy more people developing on other platforms or nrf52 kits.

If this was created because of one person was having issues, then if anyone else will ever ask, we can guide him to this thread where you created firmware showing those 3 registers on screen. https://github.com/JF002/InfiniTime/issues/590#issuecomment-900273637

Does this check really needs to be so strict on reading 3 internal registers (if I checked it right). Wouln't be enought to check whether the touch screen I2C address sends ACK so the sensor is there and correctly connected to the I2C?

Riksu9000:
InfiniTime officially only supports the PineTime and the Cst816 controller currently. The controller used in the P8 isn't compatible with the new touch handler currently, so showing a warning because of an unsupported controller is appropriate isn't it?

If official support for the Cst716 controller was added, then that controller could be whitelisted by allowing fwVersion to be either 1 or 2. 1 is for Cst816 and 2 is for Cst716.

JF002:
@Riksu9000 & @hubmartin 
Indeed, for now, InfiniTime officially only supports the PineTime. This is a deliberate choice because I cannot support multiple device on my own (more about this just below :) ).
Also, I know that Pine64 have already had to refuse a batch of PineTime coming from the factory because the wrong touch sensor was mounted on the PCB (a cheaper version of the CST816S). I have also received 2 engineering samples from the factory that had the wrong touch sensor. Now, I don't know if devices with the wrong sensor were actually shipped to the end-users, but I can only say that it's a possibility that the factory might install the wrong sensor by accident.
In this context, I think this PR makes sense, and will considerably help the user and the support if this would happen.

Now, I know that @hubmartin is very involved in a port of InfiniTime on the ColmiP8, and I really appreciate that! I would like to explore solutions that would allow us to continue working on InfiniTime for the PineTime while allowing some contributors to maintain ports for other targets. Should these port be forks of InfiniTime, completely independent from this project, or should InfiniTime be slightly refactored to be more easily ported across different devices? My long term dream is to port InfiniTime to the BL60x MCU, so porting to the P8 (which is a lot more similar to the PineTime than any BL60x board) might be a first step to this goal.
I'm not promising anything, but I think we should be able to find a nice compromise here ;)

hubmartin:
After thinking, I don't know I'll find time and motivation to fix cst718 driver for P8. It was nice bonus, that it worked before touch driver update. I understand your primary focus on your PineTime hardware.
I guess that refactoring InfiniTime for more generic drivers, which are not tied to the chips, will be needed as a first step to make it more universal. Forking is not good way in my opinion. But this driver refactoring is beyond my ability. So I'll keep P8 aside for now and will hope to find another PineTime somewhere in EU to buy second piece for debugging, instead of P8.

Issue #471, opened 2021-07-08
Improving touch event handling could improve responsiveness greatly.


Riksu9000:
https://user-images.githubusercontent.com/37774658/124918631-ee242c00-dffd-11eb-9d7c-7e595b22e3b8.mp4

I was working on improving the touch handling and found some potential big improvements.

By adjusting registers on the touchscreen controller, we can get an interrupt on touch state change. I believe currently we only get interrupts from gestures. The single tap **gesture** has a long delay, which is why currently the watch responds sluggishly.

```
/*
[7] EnTest - Interrupt pin to test, enable automatic periodic issued after a low pulse.
[6] EnTouch - When a touch is detected, a periodic pulsed Low.
[5] EnChange - Upon detecting a touch state changes, pulsed Low.
[4] EnMotion - When the detected gesture is pulsed Low.
[0] OnceWLP - Press gesture only issue a pulse signal is low.
*/
static constexpr uint8_t irqCtl = 0b00110000;
twiMaster.Write(twiAddress, 0xFA, &irqCtl, 1);
```
Here's where I'm at currently.
https://github.com/Riksu9000/InfiniTime/tree/improved_touch_handling
Most things work fine, but with some apps touch seems inconsistent. Sometimes taps will "buffer", when a tap seems to get ignored, but tapping elsewhere will apply the missing tap. The first touch is more consistent than subsequent touches. Arc in metronome might flicker between max and min positions.

I'm starting to run out of ideas, so any help would be appreciated.

This potential can easily be tapped into in some cases by simply enabling TouchModes::Polling. This works on Timer on develop branch and it improves usability greatly.

EDIT: I was studying FreeRTOS and I thought that touch events should really be handled in a separate task. Then we could also enable EnTouch, and the handling could be really simple. I'll look into this more.

JF002:
On your video, the reactivity of the touchscreen seems to be really improved!

Your discovery about the configuration of the irq pin is really interesting because that'll also a finer control of the touch panel!

As you noticed, for now, InfiniTime uses the touch IRQ only to detect gestures : tap, swipes,... But some apps needed to know the absolute position of the finger when it's moving. That's why we implemented these TouchModes : gestures or polling. Polling is a bit more expensive because it needs to poll the touch controller periodically. 

> This potential can easily be tapped into in some cases by simply enabling TouchModes::Polling. This works on Timer on develop branch and it improves usability greatly.

[...]

PR #492, opened 2021-07-16
WIP New touch handler

### LINKED ISSUE #491: Metronome arc is laggy
On latest commit.

https://user-images.githubusercontent.com/70802936/125686791-d97e00e6-2d97-4c81-8f7a-e96605eeab57.mp4

Riksu9000:
- Greatly improved responsiveness. Demonstration in #471.
- TouchModes are no longer needed.
- LVGL is now correctly told about the touch positions and times. Dragging, holding and long pressing etc. should now always work correctly. LVGL isn't aware if apps or screens have changed, so CancelTap() is used to stop the current touch from activating buttons.
- Cleaned up touchdriver code. Events (0 = Down, 1 = Up, 2 = contact) seem inconsistent, so they're replaced with a simple boolean touching.
- For a single touch, only one swipe gesture can be generated. This fixes double gesture in Twos and when swiping up during the animation going from QuickSettings to watch face.

This code probably requires further stress testing to make sure it always works correctly.

Implements #471 and fixes #491.


src/touchhandler/TouchHandler.cpp
```diff
@@ -0,0 +1,76 @@
...
+  if (info.touching) {
+    isCancelled = true;
+    lvgl.SetNewTouchPoint(-1, -1, true);
+  }
+}
+
+Pinetime::Drivers::Cst816S::Gestures TouchHandler::GestureGet() {
+  auto returnGesture = gesture;
+  gesture = Drivers::Cst816S::Gestures::None;
+  return returnGesture;
```
jonvmey, line 18: This could be simplified as `return std::exchange(gesture, Drivers::Cst816S::Gestures::None);`

Riksu9000: It could be argued whether this is simpler or not. Anyone can read the current code, but I must not be the only one who's never heard of `std::exchange()` before. I can change it though if this is important to comply with the "Modern C++" requirement.

jonvmey: Up to you ultimately, just a suggestion. I just like to mention things like these because as you said yourself there's lots of useful things in the standard library that many people aren't familiar with. I personally think that `std::exchange` is nicer/simpler than the equivalent code to do it manually but YMMV.

Riksu9000:
For some reason the watch freezes if we swipe the screen multiple times while the watch is sleeping.

Riksu9000:
Removing every `twiMaster.Sleep()` after Init seems to fix the issue. What should we do about this? It's not great that currently it's woken up and put to sleep in many places, making the real state hard to keep track of.

JF002:
I haven't look at the code, but it looks like you are improving the touch driver a lot!

Regarding your last comment

> Removing every twiMaster.Sleep() after Init seems to fix the issue. What should we do about this? It's not great that currently it's woken up and put to sleep in many places, making the real state hard to keep track of.

Originally, the goal was to handle sleep/wakeup at one place. It was easy when all devices were put to sleep when the watch is going to sleep. 
The motion sensor made this a bit more complicated because it's being polled even when the watch is asleep (needed for wake on wrist rotation feature).
Another call to twimaster.Sleep() is needed when a tap is detected on the touch panel : we have to check if it's a single/double tap and wake the watch if it's configured to wake on single/double tap.

Which calls to `twimaster.Sleep` did you remove? Is it this one https://github.com/JF002/InfiniTime/blob/develop/src/systemtask/SystemTask.cpp#L129 ?

Riksu9000:
>Which calls to twimaster.Sleep did you remove? Is it this one https://github.com/JF002/InfiniTime/blob/develop/src/systemtask/SystemTask.cpp#L129 ?

No. That's the only one i left. I just removed the three other ones in SystemTask.cpp. I don't know if just one of them was causing the problem.

I was thinking maybe twiMaster could just be woken up in the Read() and Write() functions, so there wouldn't be any need to keep track of the state. After my first experiment, I got a bootloop, so I don't really wan't to mess with it any more right now since I don't have a devkit. But does this sound like a good idea?

geekbozu:
This seems to queue the last gesture given. 
Turn off screen -> Swipe down -> turn on screen -> Jumps to notifications. 
Otherwise much nicer input feels much more natural and consistent

kieranc:
I've been testing this today and it's certainly a lot faster than the old setup, but I've had a number of false inputs, taps interpreted as swipes and the inverse, also the watch locked up and rebooted after a game of 2048. I wonder if it needs some sort of debouncing or filtering, IE if something is interpreted as a swipe but moved only 10px, it's probably meant to be a tap...
Double tap to wake seems less reliable too.

Riksu9000:
> I've been testing this today and it's certainly a lot faster than the old setup, but I've had a number of false inputs, taps interpreted as swipes and the inverse, also the watch locked up and rebooted after a game of 2048. I wonder if it needs some sort of debouncing or filtering, IE if something is interpreted as a swipe but moved only 10px, it's probably meant to be a tap...
> Double tap to wake seems less reliable too.

I remember it working quite well for me when I was testing it. I guess I'll have to test it some more.

kieranc:
> I remember it working quite well for me when I was testing it. I guess I'll have to test it some more.
I'll test it again tomorrow, it could have been something I did.

kieranc:
> > I remember it working quite well for me when I was testing it. I guess I'll have to test it some more.
> > I'll test it again tomorrow, it could have been something I did.

I'm testing it again with the ui/settings update and it seems to be behaving fine. It might have been that my screen was dirty!

kieranc:
> I'm testing it again with the ui/settings update and it seems to be behaving fine. It might have been that my screen was dirty!

Further testing makes me think double tap to wake is not working correctly still, I have this PR plus UI/settings updates applied and double tap to wake is very unreliable.

I'm going to test with just this PR, and with a stock build, but could you see if double tap to wake works correctly for you? It seems like it works fine at first but after a few hours it will not wake by double tap.

Riksu9000:
I was testing the double tap to wake yesterday, and even after leaving the watch on a table overnight, the double tap worked first try in the morning. Maybe you're experiencing something related to the freezing issue, or the queueing issue I haven't fixed yet.

I did find one issue though, which is probably what you were experiencing as well. This is what I wrote in #471: 
>When we press the screen outside of a button, slide our finger on top of a button and release, a click event is sent. This can even be seen in this online LVGL example.

If we do a very short and quick swipe in the app menu for example, LVGL doesn't care that we moved our finger, it sends a click event anyway where the screen was last touched. CancelTap() is used to combat this, but the swipe gesture arrives too late, and LVGL has already sent the event. I wish there was a way to disable this behaviour in LVGL, but we must work around it for now.

kieranc:
> I was testing the double tap to wake yesterday, and even after leaving the watch on a table overnight, the double tap worked first try in the morning. Maybe you're experiencing something related to the freezing issue, or the queueing issue I haven't fixed yet.

I tested overnight with a stock build and double tap worked reliably. I've just flashed a build with only this PR applied, I should know if it's reliable by the end of the day. I can't see how the UI updates PRs could have caused this behaviour, but we will see.

kieranc:
Double tap wake is still being unreliable for me with this PR I'm afraid. I think we need someone else to test it.

raulgool:
I have been trying this for about 24 hours now. This is a great improvement over the released image.

As kieranc and others have mentioned before. Double-tap to wake is not very reliable.

Double-tap worked well for me when I turn off the screen using the button.
But when the screen goes to sleep after a timeout, the double-tap doesn't work well or at all.

Riksu9000:
I would like the freezing issue to be solved before i tackle the wakeup issue, as it could be related since both happen during sleep. I'm hoping someone could help me with this issue in case it requires reworking twiMaster.

hatmajster:
Not sure I get it, because I've just started testing it: the "freezing issue" is happening only when `twimaster.Sleep` aren't removed? Because after swiping, pressing, buttoning, wake-up timeouting I managed to freeze the watch, though no idea how exactly right now.
I would like to say that for me it works cool, brightness works perfectly, though timer for some reason not (which I was hoping for :/). 
For the wake up issue, its that as @geekbozu said, OS seems to queue gestures/presses. When I swipe few times on turned off screen, wake up with double tap doesn't work. I have to press a button.
I do seem to have a problem with timeout as @raulgool described, however not always, couldn't test it enough yet.
Waiting for the frozen watch to wake up;) Probably I am finally forced to unsolder the battery from devkit;)

Riksu9000:
>The code looks good to me! I'm just wondering what was the reason for adding a new task to handle the touch panel? Handling touch in System/Display task was adding too much latency? It does offer a better "de-coupling" (is that a word?) between the touch management and the rest of the system, which is nice.

Mainly just to decouple it. Looking at it now, I don't think a separate task is necessary. Maybe at some point I thought it was necessary. It could easily be changed back to pushing a message to systemtask and having it call the Work function. If I did this, should I move around the files too?

src/main.cpp
```diff
@@ -158,11 +161,13 @@ Pinetime::System::SystemTask systemTask(spi,
...
                                         displayApp,
                                         heartRateApp,
-                                        fs);
+                                        fs,
+                                        touchHandler);
 
 void nrfx_gpiote_evt_handler(nrfx_gpiote_pin_t pin, nrf_gpiote_polarity_t action) {
   if (pin == pinTouchIrq) {
-    systemTask.OnTouchEvent();
+    twiMaster.Wakeup();
```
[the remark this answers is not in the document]
Riksu9000: That's true. Maybe there is a way to keep the wakeups and sleeps in order if they happen from multiple places at once, but I think it gets unnecessarily complicated. #568 fixes the freezing issue, and this.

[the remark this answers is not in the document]
JF002: I was also thinking about something that would ensure the device is put to sleep/woken up when necessary, but I don't think we need that complexity right now :)

JF002:
TODO

- [ ] Merge #568

Riksu9000:
I just tried making it not a task, but for some reason it started registering double swipes very frequently. I didn't change any of the logic, so it might be latency and/or order related. It would require some more work to convert it after all.

Riksu9000:
I think I managed to fix every issue mentioned. TouchHandler is now not a task, making TouchHandler nice and simple. Doubletap wakeup has been fixed and double swipes were fixed by testing for touching. It seems the touchpanel can generate a swipe event event after release which is kind of weird. I brought back the TouchEvent message in DisplayApp, so it doesn't have to read the gesture every cycle. This even works without reworking twiMaster.

Riksu9000:
Nope, there's still some weird behaviour. If I swipe from QuickSettings to watch face and doubletap during the animation, the watch will go to sleep. If I then try to wake it up by double tapping, it will sometimes freeze. With #568, the firmware won't freeze, but some graphical glitches can appear on screen, or the screen will be black. The issues disappear after sleep and wakeup or swiping to a different screen. Why would this affect the screen?

Riksu9000:
> > If I swipe from QuickSettings to watch face and doubletap during the animation, the watch will go to sleep.
> 
> I like the double-tap to sleep ~bug~ feature ;) .

That part is fine and can be considered working "as intended", but the next paragraph is the issue.

> But I came across some other bug playing Pong. The speed of the ball increases when you touch the screen. I built the DFU image using the most recent commit `fe33c75`.

I mentioned this in the comment above. It is a side effect of the way displayapp works currently and #497 would fix it. Alternatively it could be made so that reading messages doesn't make the full cycle run more often than intended. The current way it works is too fragile IMO.

JF002:
@Riksu9000 So, if I understand correctly, we should also merge #497 (in addition to #568) to improve the touch management? 

Are there any other PR related to the touch management that we should merge?

Riksu9000:
Even if those PRs were merged there are still issues with this one, and there are other ways to get around some of the issues other than merging those PRs. I would rather just delay this and consider those PRs separately.

So basically if you like #497 and #568, then merge them, but don't merge them just because of this PR.

JF002:
Ok, I'll have a loo at #497 and 568 and consider this one as WIP. I might try to help when I have some time ;)

In the meantime, keep us up to date of your progress, you're doing great work in this project!

Riksu9000:
> Nope, there's still some weird behaviour. If I swipe from QuickSettings to watch face and doubletap during the animation, the watch will go to sleep. If I then try to wake it up by double tapping, it will sometimes freeze. With #568, the firmware won't freeze, but some graphical glitches can appear on screen, or the screen will be black. The issues disappear after sleep and wakeup or swiping to a different screen. Why would this affect the screen?

Oh this actually happens on develop as well, so its a separate issue. This PR itself might already be working correctly.

JF002:
> Nope, there's still some weird behaviour. If I swipe from QuickSettings to watch face and doubletap during the animation, the watch will go to sleep. If I then try to wake it up by double tapping, it will sometimes freeze. With #568, the firmware won't freeze, but some graphical glitches can appear on screen, or the screen will be black. The issues disappear after sleep and wakeup or swiping to a different screen. Why would this affect the screen?

Looks like a race condition between the touch IRQ and the I²C bus going to sleep.

> Oh this actually happens on develop as well, so its a separate issue. This PR itself might already be working correctly, other than the DisplayApp issue, but thats easy to fix once we decide which way we want to solve it.

What's this DisplayApp issue?

Riksu9000:
>What's this DisplayApp issue?

I meant the one where sending messages to DisplayApp makes it run too frequently. Originally mentioned in https://github.com/JF002/InfiniTime/pull/492#issuecomment-896257887

JF002:
Thanks!
But I think touch event were already sent to display task before these change, right? How come that we notice that only in this branch?

Riksu9000:
>But I think touch event were already sent to display task before these change, right? How come that we notice that only in this branch?

Previously it sent events on gestures only. Now it sends events each time the controller reads the panel and detects touch.

So actually this code can also be moved to the event handling.
```
  if (touchHandler.IsTouching()) {
    currentScreen->OnTouchEvent(touchHandler.GetX(), touchHandler.GetY());
  }
```

JF002:
Ok, I understand! Out of my mind, an easy solution would be to throttle the refresh of the UI so that it's not refresh more often than 20ms, for example. 
Do you think of another solution?

Riksu9000:
>It is a side effect of the way displayapp works currently and #497 would fix it. Alternatively it could be made so that reading messages doesn't make the full cycle run more often than intended. The current way it works is too fragile IMO.

https://github.com/JF002/InfiniTime/pull/492#issuecomment-897922074

JF002:
Great :+1 Thanks! 

So, in the end, if we merge #497, this PR is not WIP anymore, right?

Riksu9000:
> So, in the end, if we merge #497, this PR is not WIP anymore, right?

I just tested it with both PRs and it seems to work correctly as far as I can tell.

JF002:
Awesome! I added #568, #492 and #497 to 1.4, and I'll test/review/merge them soon :)

kieranc:
> Awesome! I added #568, #492 and #497 to 1.4, and I'll test/review/merge them soon :)

DFU with #492 #497 and #568 
[pinetime-dfu-1.4.0-rc1-492-497-568.zip](https://github.com/JF002/InfiniTime/files/7014961/pinetime-dfu-1.4.0-rc1-492-497-568.zip)

DFU with the above plus #458 #480 and #586 
[pinetime-dfu-1.4.0-rc1-492-497-568-458-480-586.zip](https://github.com/JF002/InfiniTime/files/7014971/pinetime-dfu-1.4.0-rc1-492-497-568-458-480-586.zip)

JF002:
LGTM! :+1:

hubmartin:
Hi, I tried to merge this to my P8 and watches react just to the first click/gesture. I'm sure there is different touch controller cst716 in the P8. I'm in the beginings of debugging it. **If you have any idea what to check and touch datasheet with registers, that would help me a lot.**

it seems like a new configuration is sent to the driver, if I set 0x00 to this register it behaved the same way - only first touch gesture is detected after boot.
```
  static constexpr uint8_t irqCtl = 0b01110000;
  twiMaster.Write(twiAddress, 0xFA, &irqCtl, 1);
```

Then the `GetTouchInfo` is different, I'm didn't get further yet. However when debugging, every touch is detected in the GetTouchInfo function and I see proper X and Y coordinates. Also not sure if the issue could be around `isCancelled`. 

Thank you.

Riksu9000:
https://smarterwat.ch/soc-list/nordic-semiconductor/p8.html

>Alternatively, swipe behaviour can determine which you have. If you swipe and keep your finger on the display, the CST716 will not register the swipe until after your finger has been removed, whereas the CST816S will register the swipe before your finger has been removed.

Because of the difference in behaviour, info.touching will be false and the gesture won't be registered. The code would need to be tweaked slightly for the different controller.
```
      if (info.gesture == Pinetime::Drivers::Cst816S::Gestures::SlideDown ||
          info.gesture == Pinetime::Drivers::Cst816S::Gestures::SlideLeft ||
          info.gesture == Pinetime::Drivers::Cst816S::Gestures::SlideUp ||
          info.gesture == Pinetime::Drivers::Cst816S::Gestures::SlideRight ||
          info.gesture == Pinetime::Drivers::Cst816S::Gestures::LongPress) {
        if (info.touching) {
          gesture = info.gesture;
          gestureReleased = false;
        }
```

>To see which version you have on the stock firmware, you can look at the about page. There will be a string that looks like 00 B4 00 02. Ending in 01 means you have the CST816S, and ending in 02 means you have the CST716S.

This value could be used to detect the controller and adjust the touch handler accordingly. I've added reading this value in #489.

hubmartin:
Thanks for tips.
After reboot on the digital watchface I could click and every time the touch point seems to be registered and `lvgl.SetNewTouchPoint(info.x, info.y, true);` is called and coordinates seems to be right.
But since on digital watchface there is nothing to click, I have to swipe.
After swipe the `isCancelled` is `true` and calls to `lvgl.SetNewTouchPoint` are not passed further because of this condition:

image

And I'm not sure what this variable exactly does yet.

I also did this hack and it seems that I could swipe many times now. However I still have issues with touch because isCancelled is false,

image

Riksu9000:
CancelTap() is also related to this swipe behaviour in Cst816. If we swipe, the screen will switch while we're still touching the screen, and when we finally release, it may activate a button on the new screen. CancelTap() and isCancelled set the touch point to somewhere outside the screen, so when we release, it won't activate anything. So in practice, the touch has been cancelled and won't activate anything in LVGL.

Releasing a touch generates an interrupt because of "EnChange" in irqCtl register, so UpdateLvglTouchPoints() gets run with info.touching being false, which should set isCancelled to false

```
        case Messages::OnTouchEvent:  //handle release interrupt
          if (touchHandler.GetNewTouchInfo()) {  //info.touching = false
            touchHandler.UpdateLvglTouchPoint();  //isCancelled = false
          }
```

hubmartin:
Do you have a proper datasheet with registers and bits explained? Thanks

Dne so 11. zář 2021 19:21 uživatel Riku Isokoski ***@***.***>
napsal:

Riksu9000:
The PineTime wiki has datasheets for the components
https://wiki.pine64.org/index.php/PineTime
https://wiki.pine64.org/images/2/2f/CST816S.zip

This one has the registers. 
[CST816S寄存器说明-20190508.zh-CN.en.pdf](https://github.com/JF002/InfiniTime/files/7148317/CST816S.-20190508.zh-CN.en.pdf)

ildar:
It's a pity P8 users got stuck with 0.13 because of this pull.
> This value could be used to detect the controller and adjust the touch
handler accordingly. I've added reading this value in #489
<https://github.com/JF002/InfiniTime/pull/489>.

Tried that but it doesn't help much only displaying a nice error message.

StarGate01:
I added support for the CST716 back in my fork, see https://github.com/StarGate01/InfiniTime/commit/6ecb9a183febee3d14675624f967e3ce2cfffe8a . For more info, see https://github.com/InfiniTimeOrg/InfiniTime/issues/62#issuecomment-1073943874 . It rarely detects a swipe as a touch, apart from that it works great.

PR #494, opened 2021-07-17
Replace app_timer with FreeRTOS timers in TimerController

Riksu9000:
I read that this would be preferred, as app_timer is an nrf specific feature.

### DIFF
--- a/src/systemtask/SystemTask.cpp
+++ b/src/systemtask/SystemTask.cpp
@@ -2,7 +2,7 @@
 #include <hal/nrf_rtc.h>
 #include <libraries/gpiote/app_gpiote.h>
 #include <libraries/log/nrf_log.h>
-
+#include <app_timer.h>
 #include "BootloaderVersion.h"
 #include "components/battery/BatteryController.h"
 #include "components/ble/BleController.h"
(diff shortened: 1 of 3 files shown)


Riksu9000:
After #544, converting MotorController to use FreeRTOS Timers would require more work, so I removed the changes for now. For now only TimerController is converted.

Riksu9000:
@NeroBurner The simulator doesn't have these two functions `xTimerGetExpiryTime()` and `xTimerIsTimerActive()`. If this ever gets merged, they will need to be added.

NeroBurner:
@Riksu9000 thanks for the notification, please try with the current `main` branch of the simulator (commit https://github.com/InfiniTimeOrg/InfiniSim/commit/3cc6e74ea14a8d4f6ee4f5f6339f4aa8a9c05111)  where https://github.com/InfiniTimeOrg/InfiniSim/issues/17 is fixed and the two missing functions implemented

NeroBurner (approved):

src/systemtask/SystemTask.cpp
```diff
@@ -2,7 +2,7 @@
 #include <hal/nrf_rtc.h>
 #include <libraries/gpiote/app_gpiote.h>
 #include <libraries/log/nrf_log.h>
-
+#include <app_timer.h>
```
JF002, line 5: Is this include (and the call to app_timer_init()) still needed? Maybe because the MotorController still use NRF timers?

Riksu9000: If I recall correctly, I had to add this include after I removed it from TimerController. App_timers are still used elsewhere, so the initialization is necessary. After this PR, we can take a look at #1095, which seems to remove the rest of the app_timers.

JF002 (approved):

PR #495, opened 2021-07-17
Reduce duplication in CmakeLists.txt

jonvmey:
Factored out a couple of repeated values into their own variables and simplified some of the generator expressions used in assigning compile options.


Riksu9000:
These changes have been merged in #1488. Thanks!

PR #497, opened 2021-07-19
Rework Refresh

Riksu9000:
Now that LVGL has more accurate timing, we can use LVGL tasks to refresh apps instead of having DisplayApp do it. Using LVGL tasks to take care of refreshing lets each app choose the refresh rate, or not have it at all. Also it's better when less things are dependent on a task loop.

LVGL for some reason runs all tasks with higher priority again after a task with a lower priority has finished. The screen draw task is `LV_TASK_PRIO_MID`, so if the refresh task priority is lower than mid, it will redraw the screen. If the screen draw takes longer than the refresh period, the refresh task will be run again, which will run the screen draw again, resulting in an infinite loop and `lv_task_handler()` never returns. Because of this, the refresh task must be at least LV_TASK_PRIO_MID.


Avamander:
I think it would be better and a smaller patch to simply stop calling some DisplayApp's `Refresh()` on our own and let LVGL do it. Meaning that instead of a `static void refresh(struct _lv_task_t* task)` in every watchapp we could have a single one and just call the correct `Refresh()`.

Riksu9000:
> I think it would be better and a smaller patch to simply stop calling some DisplayApp's `Refresh()` on our own and let LVGL do it. Meaning that instead of a `static void refresh(struct _lv_task_t* task)` in every watchapp we could have a single one and just call the correct `bool Refresh()`.

Good idea. I've now created `RefreshTaskCallback(lv_task_t*)` in the Screen class. I still wanted to make the Refresh functions void though as the return value doesn't mattery any more.

JF002:
@Riksu9000 Thanks for the fast conflict fixing!

PR #507, opened 2021-07-21
Fit more tasks in SystemInfo

Riksu9000:
Make room for two more tasks, which could be #480 and #492

tasks


JF002:
@Riksu9000 Thanks for the PR!
And @Avamander & @geekbozu thanks for the review ;)

PR #510, opened 2021-07-22
Ticker

wilsonjwco:
Feature add. Added a scrolling ticker on the digital watch face. Ticker has 3 modes controlled by a settings screen.

1. Off: Do not display the ticker.
2. Update: Display the most recent message. As a convenience, the last message can be seen at a glance without the need to swipe up.
3. Keep: Keep the currently displayed message, and do not update it when receiving new messages. Useful for persistent reminders or to help with learning/memorizing facts, Bible verses, quotes, etc. Message can be set manually using GadgetBridge's Debug screen.

Compiled, downloaded OTA to my watch, and tested. 

Pics can be seen on this google shared album: [https://photos.app.goo.gl/HSDs2YWnEf5c1Xk48](https://photos.app.goo.gl/HSDs2YWnEf5c1Xk48)

Ideas for future functionality:

- display news/RSS items
- stock market ticker
- user-programmable filter to show only high priority or certain types of messages on the watch face
- display status from other apps (notify when timer is done, or when step goal met)
- debug/log messages for developers trying to catch rare or hard to repeat bugs

### DIFF
--- a/src/components/settings/Settings.h
+++ b/src/components/settings/Settings.h
@@ -12,6 +12,7 @@ namespace Pinetime {
       enum class ClockType { H24, H12 };
       enum class Vibration { ON, OFF };
       enum class WakeUpMode { None, SingleTap, DoubleTap, RaiseWrist };
+      enum class TickerMode { Off, Update, Keep };
 
       Settings(Pinetime::Drivers::SpiNorFlash& spiNorFlash);
 
@@ -77,6 +78,15 @@ namespace Pinetime {
         return settings.wakeUpMode;
       };
 
+      void setTickerMode(TickerMode ticker) {
+        if (ticker != settings.tickerMode)
+          settingsChanged = true;
+	settings.tickerMode = ticker;
+      }
+      TickerMode getTickerMode() const {
+        return settings.tickerMode;
+      }
+
       void SetBrightness(Controllers::BrightnessController::Levels level) {
         if (level != settings.brightLevel)
           settingsChanged = true;
@@ -107,6 +117,7 @@ namespace Pinetime {
         uint32_t screenTimeOut = 15000;
 
         WakeUpMode wakeUpMode = WakeUpMode::None;
+	TickerMode tickerMode = TickerMode::Update;
 
         Controllers::BrightnessController::Levels brightLevel = Controllers::BrightnessController::Levels::Medium;
       };
--- a/src/components/settings/Settings.h
+++ b/src/components/settings/Settings.h
@@ -133,7 +133,7 @@ namespace Pinetime {
       uint8_t settingsFlashBlock = 99; // default to indicate it needs to find the active block
 
       static constexpr uint32_t settingsBaseAddr = 0x3F6000; // Flash Settings Location
-      static constexpr uint16_t settingsVersion = 0x0100;    // Flash Settings Version
+      static constexpr uint16_t settingsVersion = 0x0101;    // Flash Settings Version
 
       bool FindHeader();
       void ReadSettingsData();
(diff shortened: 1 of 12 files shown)


src/components/settings/Settings.h
```diff
@@ -77,6 +78,15 @@ namespace Pinetime {
         return settings.wakeUpMode;
       };
 
+      void setTickerMode(TickerMode ticker) {
+        if (ticker != settings.tickerMode)
```
Avamander, line 82: Missing braces

src/displayapp/screens/WatchFaceDigital.cpp
```diff
@@ -100,6 +101,33 @@ WatchFaceDigital::WatchFaceDigital(DisplayApp* app,
...
+
+  /*
+  if (title == nullptr)
+    title = "Notification";
+  char* pchar;
+  pchar = strchr(title, '\n');
+  while (pchar != nullptr) {
+    *pchar = ' ';
+    pchar = strchr(pchar + 1, '\n');
+  }
```
Avamander, line 113: Unnecessary comment

src/displayapp/screens/WatchFaceDigital.cpp
```diff
@@ -129,12 +157,13 @@ bool WatchFaceDigital::Refresh() {
 
   notificationState = notificatioManager.AreNewNotificationsAvailable();
   if (notificationState.IsUpdated()) {
-    if (notificationState.Get() == true)
+    if (notificationState.Get() == true) {
       lv_label_set_text(notificationIcon, NotificationIcon::GetIcon(true));
-    else
+    } else
```
Avamander, line 162: Missing brace

src/displayapp/screens/WatchFaceDigital.h
```diff
@@ -57,6 +60,7 @@ namespace Pinetime {
         lv_obj_t* label_time;
         lv_obj_t* label_time_ampm;
         lv_obj_t* label_date;
+	lv_obj_t* label_ticker;
```
Avamander, line 63: Run clang-format on these files.

src/displayapp/screens/settings/SettingTicker.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+        if (i == 0) {
+          settingsController.setTickerMode(Pinetime::Controllers::Settings::TickerMode::Off);
+        };
+        if (i == 1) {
+          settingsController.setTickerMode(Pinetime::Controllers::Settings::TickerMode::Update);
+        };
+        if (i == 2) {
+          settingsController.setTickerMode(Pinetime::Controllers::Settings::TickerMode::Keep);
+        };
```
Avamander, line 95: Unnecessary semicolons

Avamander (changes requested):

Riksu9000:
Maybe instead of having modes, we could simply pin a notification to show on the watchface.

0x1a8510f2:
Just so you're aware @wilsonjwco, you haven't removed geolocation data from the photos in the album, meaning that anyone can see where the photo was taken (likely where you live). Would recommend you delete the album and re-create it without geolocation data (there should be a setting in Google photos).

wilsonjwco:
> Maybe instead of having modes, we could simply pin a notification to show on the watchface.

Do you mean we could have a button on the watch face that would pin the current notification?

wilsonjwco:
Thanks everyone for the comments, I'll work through them. Also noticed the message get corrupted after a few wake cycles, so it looks like I've got more debugging to do before it's merged.

Riksu9000:
> > Maybe instead of having modes, we could simply pin a notification to show on the watchface.
> 
> @Riksu9000 do you mean we could have a button on the watch face that would pin the current notification?

No. I was thinking something like opening the notification that we want to pin and long pressing on it. Then when we return to the watch face, it would be shown there. This way we could easily choose any notification to keep.

wilsonjwco:
> I was thinking something like opening the notification that we want to pin and long pressing on it.

Great idea-- I'll get rid of the settings and make it like you suggested. Will follow up with this discussion once I've done this, fixed the above mentioned problems, and am ready for the merge to be considered again.

lman0:
any new on this @wilsonjwco ?

Riksu9000:
I'll close this since it's very outdated. Please create a new PR if you want to keep working on the notification pinning feature.

PR #513, opened 2021-07-22
Reduce repeated code for watchface list render

sigseg5:
I reduced the amount of code in SettingWatchFace.cpp to improve code readability and convenience, so now when adding a new watchface, you won't have to copy many lines of code.

### DIFF
--- a/src/displayapp/screens/settings/SettingWatchFace.cpp
+++ b/src/displayapp/screens/settings/SettingWatchFace.cpp
@@ -41,33 +41,50 @@ SettingWatchFace::SettingWatchFace(Pinetime::Applications::DisplayApp* app, Pine
   lv_obj_align(icon, title, LV_ALIGN_OUT_LEFT_MID, -10, 0);
 
   optionsTotal = 0;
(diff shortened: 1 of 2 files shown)


hatmajster:
I think You can also remove `optionsTotal` and just use `i` in Your loop

sigseg5:
> I think You can also remove ~optionsTotal~ `watchFaceCount` and just use `optionsTotal`. Also just use `i` in Your loop

Updated, I think it's OK due to some limitation like `Variable-sized object may not be initialized` etc.

Avamander:
It's okay and probably safer to use `std:string` instead of C-style string/array.

sigseg5:
> probably safer to use `std:string` instead of C-style string/array

I think that `const * char watchFaceNames[watchFaceCount]` is safe enough in this context. Indeed I could use std::string, so the code would be more uniform (and probably safer in the future). But on the other hand, using std::string does bring a small performance degradation, you may want to take a look at this [asm](https://godbolt.org/z/fW319PYbo). What ideas do you have about it that would be more appropriate to use?

Riksu9000:
A similar enhancement has been merged.

PR #520, opened 2021-07-24
Code cleaning in Clocks (analog, digital & PineTimeStyle)

JF002:
PR #232 by @nscooling is too old to be easily mergeable. I took most of the changes from this PR and applied them to analog and PineTimeStyle watchface too : 

- Replaced unnecessary sprintf statements with the build-in lvgl lv_label_set_text_fmt function (designed for that exact purpose)
- Cleaned up Clock construction and member-initialisation list, where the NSDMI can be used in preference (e.g. defaults zero values and nullptr)
 - Cleaned up some logic code
 - Change some instances of || to or (this is a C++ keyword, as is and and not and has been since the original C++98 standard)
 - Ran Clock.* through clang-tidy based on .clang-format file at project root
 - Replaced any remaining NULL with nullptr
 - Replaced C-style casts with C++ casts

My apologies to @nscooling for not merging the original PR sooner.

### DIFF
--- a/src/components/datetime/DateTimeController.cpp
+++ b/src/components/datetime/DateTimeController.cpp
@@ -55,9 +55,9 @@ void DateTime::UpdateTime(uint32_t systickCounter) {
   auto time = date::make_time(currentDateTime - dp);
   auto yearMonthDay = date::year_month_day(dp);
 
-  year = (int) yearMonthDay.year();
-  month = static_cast<Months>((unsigned) yearMonthDay.month());
-  day = (unsigned) yearMonthDay.day();
+  year = static_cast<int>(yearMonthDay.year());
+  month = static_cast<Months>(static_cast<unsigned>(yearMonthDay.month()));
+  day = static_cast<unsigned>(yearMonthDay.day());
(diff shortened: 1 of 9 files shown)


Riksu9000:
> Change some instances of || to or (this is a C++ keyword, as is and and not and has been since the original C++98 standard)

Do we really want to do this? Using `!` `||` and `&&` we can clearly separate them from variables and functions by just taking a quick glance. By using `not`, `and`, `or`, we need to take a closer look at the context to see which words are variables and which aren't.

Avamander:
I would also concur that `!`, `||` and `&&` is what people are more accustomed to and I don't get the sense that `and`, `or` improves readability.

JF002:
I guess it's a matter of taste... These are valid keyword in C and C++ so I see no reason to forbid them. I'm also accustomed to `!`, `&&` and `||` and I was a bit surprised when I learned `not`, `and` and `or` existed, but now I get used to them, I sometimes use them too...
Do they hurt the readability in this case?

Avamander:
I suspect its effect is neutral, but I have written a fair amount of Python, it might be negative for some.

In the end what you choose is what others should use, for consistency.

PR #521, opened 2021-07-24
Replace lv_label_set_text with better options where possible

jonvmey:
If the text being set is static use lv_label_set_text_static to avoid
unnecessary memory allocations. Otherwise if a temporary char buffer is
being sprintf-ed and then set to the label used lv_label_set_text_fmt to
avoid the intermediary data copies.

### DIFF
--- a/src/displayapp/screens/Music.cpp
+++ b/src/displayapp/screens/Music.cpp
@@ -62,7 +62,7 @@ Music::Music(Pinetime::Applications::DisplayApp* app, Pinetime::Controllers::Mus
   lv_obj_align(btnVolDown, nullptr, LV_ALIGN_IN_BOTTOM_LEFT, 0, 0);
   lv_obj_add_style(btnVolDown, LV_STATE_DEFAULT, &btn_style);
   label = lv_label_create(btnVolDown, nullptr);
-  lv_label_set_text(label, Symbols::volumDown);
+  lv_label_set_text_static(label, Symbols::volumDown);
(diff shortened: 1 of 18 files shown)


src/displayapp/screens/Music.cpp
```diff
@@ -212,33 +212,30 @@ bool Music::Refresh() {
...
-            ((totalLength % (60 * 60)) / 60) % 100);
-    lv_label_set_text(txtTrackDuration, timer);
+    lv_label_set_text_fmt(txtTrackDuration,
+                          "%02d:%02d/%02d:%02d",
+                          (currentLength / (60 * 60)) % 100,
+                          ((currentLength % (60 * 60)) / 60) % 100,
+                          (totalLength / (60 * 60)) % 100,
+                          ((totalLength % (60 * 60)) / 60) % 100);
   } else {
     char timer[12];
```
hatmajster, line 232: this is probably also obsolete

hatmajster:
I was wondering how it will impact internationalization: https://github.com/JF002/InfiniTime/pull/417 but as all text is static it shouldn't be an issue at any point... I hope

src/displayapp/screens/Music.cpp
jonvmey: Yes, thanks for pointing that out. I'll remove it too.

Riksu9000:
@jonvmey Hi. Can you fix the conflicts, so we could merge this?

Riksu9000:
These improvements have been added in #1044. Thanks!

PR #524, opened 2021-07-24
Remove unnecessary C-style casts with BLE UUIDs

jonvmey:
Instead of casting the UUID object to the ble_uuid_t* used throughout
the NimBLE API just pass the address of the ble_uuid_t member that's at
the start of each of the UUID structs.

### DIFF
--- a/src/components/ble/AlertNotificationClient.cpp
+++ b/src/components/ble/AlertNotificationClient.cpp
@@ -55,7 +55,7 @@ bool AlertNotificationClient::OnDiscoveryEvent(uint16_t connectionHandle, const 
     return true;
   }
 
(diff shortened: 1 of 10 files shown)


geekbozu:
I kind of dislike how its just .u ...it feels ambiguous. Any reason to not change it to .uuid? Otherwise i do feel this is much clearer to read compared to before

jonvmey:
These types are from NimBLE so you'd have to take that up with them.

Avamander:
Very likely that if you'd make a PR that refactors it for niceness it will be merged, @geekbozu. The name has misled more than one person by now so why not try.

Avamander (approved):
I think it's a good change, both style and safety wise. The current casts are not great.

JF002:
@Avamander 
> I think it's a good change, both style and safety wise. The current casts are not great.

I totally agree!

@jonvmey Thanks for this PR, and sorry for the delay to review it !

PR #526, opened 2021-07-25
New appmenu layout

Riksu9000:
A more pleasant layout with circular buttons, but the buttons end up being quite small. Is this a worthwhile tradeoff?

A single screen doesn't need to have so many apps, but fewer apps can't be laid out as nicely, unless we have just four apps per page. Seven apps can also produce a nice pattern, but the buttons couldn't be made any larger, so we might as well have eight.

appmenu

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -418,6 +418,7 @@ list(APPEND SOURCE_FILES
         displayapp/screens/BatteryInfo.cpp
         displayapp/screens/Steps.cpp
         displayapp/screens/Timer.cpp
+        displayapp/screens/HoneyComb.cpp
 
         ## Settings
         displayapp/screens/settings/QuickSettings.cpp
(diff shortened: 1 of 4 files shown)


Avamander:
I would personally like this, I don't have issues tapping these things, but some might want it even larger than the current one (there's one issue about this, *somewhere*). 

How about making it a configuration option 4 - 6 - 7 - 8 - 9? :sweat_smile:

MysteriousLog6:
It could be made so that the number of apps per page is an adjustable slider , people can change it according to there liking.

trman:
this pull is very outdated  , you should close it like this draft https://github.com/InfiniTimeOrg/InfiniTime/pull/510 @Riksu9000 !

Riksu9000:
It's not just about whether it's old or not, but I suppose this isn't something that we'll be changing in the near future.

PR #533, opened 2021-07-28
Tilt to reload idle timer

Riksu9000:
Reload the idle timer with the same wrist gesture as wake up. This way it can be refreshed single handed if the screen dimmed.
Simplified and adjusted `ShouldWakeUp()` to work the same way regardless of whether the watch was sleeping or not.

### DIFF
--- a/src/components/motion/MotionController.cpp
+++ b/src/components/motion/MotionController.cpp
@@ -9,17 +9,8 @@ void MotionController::Update(int16_t x, int16_t y, int16_t z, uint32_t nbSteps)
   this->nbSteps = nbSteps;
 }
 
-bool MotionController::ShouldWakeUp(bool isSleeping) {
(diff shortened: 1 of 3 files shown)


geekbozu:
My concern with this is how aggressive the wake on raise currently is, The feature it self sounds great but I feel like this will aggressively effect battery life in its current state. 
Just some thoughts!

Riksu9000:
> My concern with this is how aggressive the wake on raise currently is, The feature it self sounds great but I feel like this will aggressively effect battery life in its current state.
> Just some thoughts!

I don't think this will affect the idle battery life noticably. This patch could extend the awake duration, but without this it would just wake up anyway.

Another thing this actually fixes is if the watch is already somehow woken up before we raise it. Before it would go to sleep soon, but with this patch the idle timer will be refreshed after it is raised.

PR #535, opened 2021-07-29
Small refactoring

deztructor:
Refactoring to improve readability and get rid of clang warnings.

(Flashed and verified that it it is working)

### DIFF
--- a/src/displayapp/screens/SystemInfo.cpp
+++ b/src/displayapp/screens/SystemInfo.cpp
@@ -12,15 +12,16 @@
 #include "drivers/Watchdog.h"
 
 using namespace Pinetime::Applications::Screens;
+using DeviceTypes = Pinetime::Controllers::MotionController::DeviceTypes;
 
 namespace {
-  const char* ToString(const Pinetime::Controllers::MotionController::DeviceTypes deviceType) {
+  const char* ToString(const DeviceTypes deviceType) {
     switch (deviceType) {
-      case Pinetime::Controllers::MotionController::DeviceTypes::BMA421:
+      case DeviceTypes::BMA421:
         return "BMA421";
-      case Pinetime::Controllers::MotionController::DeviceTypes::BMA425:
+      case DeviceTypes::BMA425:
         return "BMA425";
-      case Pinetime::Controllers::MotionController::DeviceTypes::Unknown:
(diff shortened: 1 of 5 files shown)


src/displayapp/screens/SystemInfo.cpp
```diff
@@ -12,15 +12,16 @@
 #include "drivers/Watchdog.h"
 
 using namespace Pinetime::Applications::Screens;
+using DeviceTypes = Pinetime::Controllers::MotionController::DeviceTypes;
```
jonvmey, line 15: If you're doing `using Foo = Some::Namespaces::Foo;` you can just do `using Some::Namespaces::Foo;` instead.

deztructor: I can't because `Pinetime::Controllers::MotionController` is not a namespace or unscoped enum.

Riksu9000:
Maybe instead of adding `WorkLoopCycle()`, the code that gets run once should be moved to `Process()` and a while loop should be added around `app->Work()` instead similar to how it is done in DisplayApp.

deztructor:
> Maybe instead of adding `WorkLoopCycle()`, the code that gets run once should be moved to `Process()` and a while loop should be added around `app->Work()` instead similar to how it is done in DisplayApp.

It could be  done as a next step: the goal of this refactoring commit is to extract the code to decrease nesting, so it is easier to modify it in future. I wanted to make an obvious and easy to review change. Because as you see, this is the hot spot: conflicts appear while the review is pending, I already rebased it with conflicts when I did refactoring (for another feature I'd like to add), now there is a conflict again.

Riksu9000:
This PR will be hard to merge, as it contains different opinionated changes and is outdated. The SystemTask change should be done by itself and in a single step, if it's worth doing. Will you update this PR accordingly, or should this be closed in favor of newer PRs @deztructor ?

deztructor:
> This PR will be hard to merge, as it contains different opinionated changes and is outdated. The SystemTask change should be done by itself and in a single step, if it's worth doing. Will you update this PR accordingly, or should this be closed in favor of newer PRs @deztructor ?

Yeah, things moved on, so it would be better to close it. And things could be done in a new PR.

PR #537, opened 2021-07-30
Simplify systemtask states

Riksu9000:
Instead of having three separate booleans, have just a single state variable.
Now we don't need to worry about checking if it's going to sleep or waking up. If the state is sleeping, it is sleeping and if the state is running, it is running.
This simplified many checks and made the code easier to read.

### DIFF
--- a/src/systemtask/SystemTask.cpp
+++ b/src/systemtask/SystemTask.cpp
@@ -265,8 +265,7 @@ void SystemTask::Work() {
             nimbleController.RestartFastAdv();
           }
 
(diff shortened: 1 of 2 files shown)


geekbozu:
This is much appreciated I'll give it a good review in the next few days :) 🚀

trman:
is this pull still relevenant @Riksu9000 ?
isthat the case can @geekbozu  , or @JF002  take a view at it?

JF002 (approved):

PR #539, opened 2021-07-31
Change notification timeout with display

evergreen22:
This PR changes the notification (preview) timeout to be equal to the screenTimeOut value. The vibration is set to be proportional to the display timeout

This is one of two proposals to allow the user to determine how long a notification is displayed. The other PR is "notification-timeout-setting" (#538). Only one of the two should be merged.


Riksu9000:
If someone is using the watch and a notification pops up they have to wait for up to 30 seconds until they can continue what they were doing. Why not make the notification exitable with swipe down and set the timeout to maybe 10 seconds?

The vibration is often too short, but making it proportional to the timeout doesn't really make sense.

evergreen22:
Thanks for the feedback @Riksu9000 . I'll work on updating the PR with your suggestions.

> Why not make the notification exitable with swipe down and set the timeout to maybe 10 seconds?

Wouldn't swipe up make more sense since that is how we currently exit the (normal) notification screen?

> The vibration is often too short, but making it proportional to the timeout doesn't really make sense.

OK. my thinking was that if you liked longer display times, you'd like longer vibrations. It is easier to make it a fixed value and I don't have strong feelings about it other than the current vibration is too short.

Riksu9000:
>>Why not make the notification exitable with swipe down and set the timeout to maybe 10 seconds?
>
>Wouldn't swipe up make more sense since that is how we currently exit the (normal) notification screen?

When opening apps, the apps rise from the bottom, and are closed by pushing them back down. For some reason the notification is animated like it's falling from above, but I think this should be changed so it also rises from the bottom and is closed like all other apps in InfiniTime.

The reason the other menu is exited by swipe up is because it is opened by swipe down. It's more like scrolling up and down than entering and exiting screens.

evergreen22:
> When opening apps, the apps rise from the bottom, and are closed by pushing them back down. For some reason the notification is animated like it's falling from above, but I think this should be changed so it also rises from the bottom and is closed like all other apps in InfiniTime.
> 
> The reason the other menu is exited by swipe up is because it is opened by swipe down. It's more like scrolling up and down than entering and exiting screens.

Thanks for the clarification @Riksu9000 .

Yes, the notification UI is inconsistent with the rest of InfiniTime. As you note, all other screens/apps open by popping up and are closed by pushing them down. I'm fine with either swipe up or down to close the preview notification and muscle memory favors swipe up as you suggested.

evergreen22:
> ... and muscle memory favors swipe up as you suggested.

[EDIT] swipe down as you suggested.

evergreen22:
Thanks for the comments

PR #542, opened 2021-07-31
Adjust touch panel autosleep

Riksu9000:
By default the touch panel goes to a low power mode after 2 seconds. In low power mode the touch panel is less responsive, so it may miss some touches. The touch panel is automatically put to sleep when the watch goes to sleep, so this feature isn't needed.

Set autosleep time to 30 seconds, which is the longest display timeout setting.

Move code from Init to Wakeup so the panel won't be configured on each wakeup

The register is called AutoSleepTime in the datasheet, but really it's low power mode.


tmilburn:
This is probably a good time to clean up all the register constants and link to the correct documentation.
[CST816S.en.pdf](https://github.com/JF002/InfiniTime/files/7017988/CST816S.en.pdf)

Riksu9000:
> This is probably a good time to clean up all the register constants and link to the correct documentation.
> [CST816S.en.pdf](https://github.com/JF002/InfiniTime/files/7017988/CST816S.en.pdf)

Good idea, but I think I would rather do that in a separate PR instead.

kieranc:
I'm testing this one as well, seems to make sense, no issues yet.

lman0:
hi 
why not make the auto sleep things , not to 30s but the selected choice for the screen time out?
or do you taken care of it?

Riksu9000:
> hi
> why not make the auto sleep things , not to 30s but the selected choice for the screen time out?
> or do you taken care of it?

The value doesn't matter as long as it is longer than the timeout, as InfiniTime puts the touch panel to sleep itself. I didn't disable it entirely just so it isn't entirely reliant on InfiniTime in case of an error. I guess the timeout could be reached by pressing the button to refresh idle timer though.

tmilburn:
Is there any data for power consumption in normal mode vs low power mode?

Riksu9000:
These are translated from the revision 1.1 datasheet

>Typical power consumption in dynamic mode <2.5mA
Typical power consumption in standby mode <10uA
Typical power consumption in sleep mode <5uA

In revision 1.0 there is also a monitoring mode?

>Typical power consumption in dynamic mode <3.0mA;
Typical power consumption in monitoring mode <100uA;
Typical power consumption in standby mode <30uA;
Typical power consumption in sleep mode <5uA;

It seems that the low power mode is this monitoring mode. I wonder why it was removed from the new revision.

Upon further reading I'm seeing that this change probably isn't necessary.

The issues I was attempting to fix were touch being missed when the watch was woken up and the paddle getting stuck in the paddle game. Testing this now this doesn't actually fix the issues, so I'll close this and look for other solutions.

tmilburn:
@Riksu9000 Where is the revision 1.1 and revision 1.0 datasheet?

Riksu9000:
The 1.1 is on the PineTime wiki
https://files.pine64.org/doc/datasheet/pinetime/CST816S%E6%95%B0%E6%8D%AE%E6%89%8B%E5%86%8CV1.1.pdf

1.0 is inside this zip also from the wiki
https://wiki.pine64.org/images/2/2f/CST816S.zip
[CST816S数据手册V1.0.pdf](https://github.com/JF002/InfiniTime/files/7243595/CST816S.V1.0.pdf)

tmilburn:
@Riksu9000 Thanks, I presume there isn't an English translation available.

PR #549, opened 2021-08-02
Put all duplicated GPIO pin definitions to a single file

hubmartin:
Hello, I use InfiniTime firmware on my P8. I see that people use InfiniTime also on their P8 and maybe other watches. I build firmware myself and it was not easy to understand what to change to get firmware running. But after all it is just 3 GPIOs that needs to be changed. 

As I see InfiniTime evolving faster and better I also hope that this opensource firmware could be used on many other watches.
So this is my kick-off to separate pin definitions to single place, where it could be later easily configured for different hardware variantions in this, or others PRs.
Please have in mind that I'm more low-level bare-bone/FreeRTOS programmer in C. So InfiniTime with C++ and OOP is something I'm still learning. So any suggestions and language-constructs are welcome and I will try to learn more C++.

Right now I separated only 3 necessary P8 GPIOs. But based on your suggestions I would continue to append all the GPIO definitions from the different places in the code.

1) I would also welcome suggestions how to change `CMakeLists.txt` file to propagate `WATCH_P8` define to the `PinMap.h` file.
I would like to have something like
`cmake ... -DWATCH=P8 ../` or `cmake .... -DWATCH_P8 ../`

2) Not sure whether `#define` is the best way to define pins. I guess that in C++ are ways to make it more complicated with zero overhead, but define is what I use.

3) Is needed `drivers/PinMap.h` definition in `INCLUDE_FILES` CmakeLists.txt?

4) Should I use `#include <drivers/PinMap.h>` in every header, or rather use `#include "drivers/PinMap.h"`?

5) In some modules there is `pinIrq` and somwhere `chargingPin`. word `pin` is in begining or the end. I could also unify this. I guess that variable should start with `pin`. Any suggestions to use different name for my defines `PINMAP_BUTTON_PIN` are welcome. Now I would say that `PINMAP_BUTTON` would be ok.


Avamander:
> I guess that in C++ are ways to make it more complicated with zero overhead, but define is what I use.

`constexpr` variables are basically the same but type-safe version of the thing.

Avamander:
> Should I use #include <drivers/PinMap.h> in every header, or rather use #include "drivers/PinMap.h"?

It's a project header so `""`.

Avamander:
> Is needed drivers/PinMap.h definition in INCLUDE_FILES CmakeLists.txt?

Explicit is better than implicit, style- and tooling-wise (some things might not pick up on the file otherwise).

jonvmey:
> > I guess that in C++ are ways to make it more complicated with zero overhead, but define is what I use.
> 
> `constexpr` variables are basically the same but type-safe version of the thing.

Using `constexpr` variables also lets you put them in a namespace, ie. `PinMap` and then you don't need to prefix each symbol with `PINMAP` to avoid conflicts.

hubmartin:
@jonvmey I've found constexpr and namespaces a good approach.

Now my `PinMap.h` file has
```
namespace Pinetime {
  namespace PinMap {
      static constexpr uint8_t Charging = 19;
      static constexpr uint8_t Button = 19;
  }
}
```
And in the code I use
```
nrfx_gpiote_in_init(PinMap::Button, &pinConfig, nrfx_gpiote_evt_handler);
```
I will use it this way.

hubmartin:
Converted all pins I could find. Removed duplicate constexpr for SPI in different places.
I've tested it with PineTime and Colmi P8.

hubmartin:
When I put PineTime and P8 with this FW it seems to me that PineTime is reacting slower on touch. I press button or + symbol on Stopwatch and reaction is not immediate. Not sure if that was before since this is first day I have PineTime hardware. The only difference in P8/PineTime builds are 3 GPIO pins which does not have anything in common to the touch..
```
    #ifdef WATCH_P8
      // COLMI P8
      static constexpr uint8_t Charging = 19;
      static constexpr uint8_t Cst816sReset = 13;
      static constexpr uint8_t Button = 17;
    #else
      // Pinetime
      static constexpr uint8_t Charging = 12;
      static constexpr uint8_t Cst816sReset = 10;
      static constexpr uint8_t Button = 13;      
    #endif
```
Any testing and ideas welcome.
I'm going to study commit changes once again.

src/drivers/PinMap.h
```diff
@@ -0,0 +1,38 @@
...
+    #endif
+
+    static constexpr uint8_t Cst816sIrq = 28;
+    static constexpr uint8_t PowerPresent = 19;
+
+    static constexpr uint8_t Motor = 16;
+
+    static constexpr uint8_t LcdBacklight1 = 14;
+    static constexpr uint8_t LcdBacklight2 = 22;
+    static constexpr uint8_t LcdBacklight3 = 23;
```
Riksu9000, line 25: Maybe these pins should be renamed to BacklightLow, BacklightMedium and BacklightHigh.

hubmartin: The naming is consistent with schematics, I would keep that.
https://files.pine64.org/doc/PineTime/PineTime%20Schematic-V1.0a-20191103.pdf

However this opens another conversation as I'm looking at this brightness schematics part first time. Lower brightness actually doesn't save power - it burns the rest energy in the resistor. So it might be optimized with PWM and we will use only Light1 wit 30 ohm resistor which burns least power. I might investigate it later in different PR.
image

hubmartin: [sorry I had to delete my previous answer]
The naming is really not consistent with schematics, I would keep 1 2 3, but swap them, so pin 14 is backlight3
https://files.pine64.org/doc/PineTime/PineTime%20Schematic-V1.0a-20191103.pdf

However this opens another conversation as I'm looking at this brightness schematics part first time. Lower brightness actually doesn't save power - it burns the rest energy in the resistor. So it might be optimized with PWM and we will use only Light1 wit 30 ohm resistor which burns least power. I might investigate it later in different PR.

image

Riksu9000: >Lower brightness actually doesn't save power - it burns the rest energy in the resistor.

That's not quite how resistors work. Higher resistance lowers current, which lowers power consumption.

>Even without PWM there might be more levels using variants of those 3 signals.

I noticed this too and tested this some time ago. There is one brightness setting that could be achieved between high and medium, the rest aren't noticable. Still this level is just slightly dimmer than max, and if we included this, the step from low to mid would be much larger than the other steps.

>This PWM would allow gradient change of brightness. Not sure there is a reason why the hardware does not use PWM.. because of EMI, not fast/efficient FET transistor?

PWM could be used to smooth transitions and achieve an even dimmer minimum brightness. I don't know if it would be useful to replace the existing steps with it.

>The naming is really not consistent with schematics, I would keep 1 2 3, but swap them, so pin 14 is backlight3

Having actual words describing the pins would still be clearer than numbers, since different pins affect the brightness different amounts.

hubmartin:
Pins LcdBacklight1,2,3 renamed to
```
    static constexpr uint8_t LcdBacklightLow = 14;
    static constexpr uint8_t LcdBacklightMedium = 22;
    static constexpr uint8_t LcdBacklightHigh = 23;
```

hubmartin:
Now I have also PineTime so I've tested this and pin mapping is fine.

hubmartin:
@JF002 Can you please merge it? With this I could test every firmware also on my P8 in paralel. Thanks

hatmajster:
I've been testing this for two days on pinetime, everything was working correct.

JF002:
Thanks for this PR, and sorry for taking so long for reviewing it!

> I would also welcome suggestions how to change CMakeLists.txt file to propagate WATCH_P8 define to the PinMap.h file.

```
Index: CMakeLists.txt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/CMakeLists.txt b/CMakeLists.txt
--- a/CMakeLists.txt	(revision 8390d0ef7243261675aab3a5c19862eb3fc53e68)
+++ b/CMakeLists.txt	(date 1631388837351)
@@ -51,6 +51,14 @@
   set(BUILD_DFU true)
 endif()
 
+option(WATCH_COLMI_P8 "Build for the Colmi P8" OFF)
+set(TARGET_DEVICE "PineTime")
+
+if(WATCH_COLMI_P8)
+  set(TARGET_DEVICE "Colmi P8")
+  add_definitions(-DWATCH_P8)
+endif()
+
 set(PROJECT_GIT_COMMIT_HASH "")
 
 execute_process(COMMAND git rev-parse --short HEAD
@@ -68,6 +76,7 @@
 message("    * Toolchain : " ${ARM_NONE_EABI_TOOLCHAIN_PATH})
 message("    * GitRef(S) : " ${PROJECT_GIT_COMMIT_HASH})
 message("    * NRF52 SDK : " ${NRF5_SDK_PATH})
+message("    * Target device : " ${TARGET_DEVICE})
 set(PROGRAMMER "???")
 if(USE_JLINK)
   message("    * Programmer/debugger : JLINK")
```
With these changes, the firmware will always be built for the PineTime by default, and for the P8 if you add `-DWATCH_COLMI_P8=1` to your CMake command line.

Feel free to apply to changes to your branch if you like them. If so, could you also update the [build documentation](https://github.com/JF002/InfiniTime/blob/develop/doc/buildAndProgram.md#project-generation-using-cmake) with this new parameter?

> Not sure whether `#define` is the best way to define pins.
I'm not a huge fan of `#define' but I think it does the job in this case. We'll probably be able to get rid of them once we port InfiniTime on different MCU than the NRF52, where we'll need to better encapsulate the hardware and low level functions.

> Is needed drivers/PinMap.h definition in INCLUDE_FILES CmakeLists.txt?

As @Avamander said, I think it's better to add all files explicitly. The cherry on the cake is that it makes CLion happy (it complains that I'm using files from outside of the project otherwise) ;)

hubmartin:
@JF002 Thanks a lot for Cmake tip. I've added new `WATCH_COLMI_P8` to the Cmake & docs. Now its ready to merge.

PR #552, opened 2021-08-03
Make README more friendly for new users.

Riksu9000:
Development documentation should be separated from user instructions.

Move getting started links higher and to a separate section so they're easier to find.
Replace section about PineTime with a link to pine64.org. Less information makes finding relevant information easier.

The "Using the firmware" (better name Companion apps) section should also be moved somewhere else, but where? Should they be in the same section as getting started?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,20 +1,15 @@
-# PineTime
+## New to InfiniTime?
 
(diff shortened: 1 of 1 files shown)


geekbozu:
All looks good to me

Issue #398, opened 2021-05-23
Screen Timeout handling issue


Batcastle:
Screen time out is reset only when initial contact is made with the touch screen. In situations like the Pong game, this means the screen times out in the middle of the game because the finger may not have been lifted from the screen. I think that either:

 A) screen time out should be outright disabled in some situations.

or

B) the screen should not time out while contact is being made with the screen.

I can see the benefit of either approach, but think the first one would be better from a battery life stand point.

Raupinger:
there are already instances where sleeping gets disabled, (the flashlight for example) Its just a matter of identifying where the screen should stay on and adding that line of code there.

Batcastle:
@Raupinger Awesome. I'm glad it's already been handled then. I noticed this issue while playing with the Pong game so could this be fixed in that app then?

MysteriousLog6:
> @Raupinger Awesome. I'm glad it's already been handled then. I noticed this issue while playing with the Pong game so could this be fixed in that app then?

I guess so , will be checking about this and maybe making a PR.

Riksu9000:
The screen won't time out while touching the screen anymore after #492, which is in InfiniTime 1.4.0, so this can be closed.

PR #553, opened 2021-08-03
Disable sleep while in Paddle app

stoehraj:
Paddle.cpp: Added systemTask to disable sleeping in the constructor and
re-enable sleeping in the destructor.  Reformatted constructor function
header to match that of FlashLight.cpp due to added length.

Paddle.h: Added include for SystemTask.h.  Added SystemTask parameter to
constructor function prototype and added a private SystemTask variable.

DisplayApp.cpp: Added systemTask pointer to make_unique() function call.

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -401,7 +401,7 @@ void DisplayApp::LoadApp(Apps app, DisplayApp::FullRefreshDirections direction) 
       currentScreen = std::make_unique<Screens::InfiniPaint>(this, lvgl);
       break;
     case Apps::Paddle:
-      currentScreen = std::make_unique<Screens::Paddle>(this, lvgl);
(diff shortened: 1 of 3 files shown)


stoehraj:
Problem mentioned in issue #398

Riksu9000:
InfiniTime won't sleep while touching the screen anymore after #492 was merged, so this isn't necessarily needed anymore.

Dudemanjude:
I would still find this useful, as the screen will often timeout for me in the Paddle app, because I like my screen timeout set to 5 seconds.

Riksu9000:
It is true that the timeout should be longer than 5s while in the paddle game, but by disabling sleep altogether, there is the danger of forgetting to close it and the battery draining quickly. This is also an issue with Flashlight currently, and it should be fixed. There needs to be a more complete solution to this.

Closing due to inactivity.

PR #554, opened 2021-08-04
Timer code cleanup

Riksu9000:
Reduced duplication.
Hide buttons instead of creating and deleting them.
Make the active area of buttons larger than what they look like. This makes the small buttons easier to press.
Fix alignment.
Small tweaks like minutes going up to 60 instead of 59.

### DIFF
--- a/src/displayapp/screens/Timer.cpp
+++ b/src/displayapp/screens/Timer.cpp
@@ -1,173 +1,123 @@
 #include "Timer.h"
-
-#include "Screen.h"
 #include "Symbols.h"
-#include "lvgl/lvgl.h"
-
 
 using namespace Pinetime::Applications::Screens;
 
-
 static void btnEventHandler(lv_obj_t* obj, lv_event_t event) {
-  Timer* screen = static_cast<Timer*>(obj->user_data);
+  auto* screen = static_cast<Timer*>(obj->user_data);
   screen->OnButtonEvent(obj, event);
 }
 
-void Timer::createButtons() {
-  btnMinutesUp = lv_btn_create(lv_scr_act(), nullptr);
-  btnMinutesUp->user_data = this;
-  lv_obj_set_event_cb(btnMinutesUp, btnEventHandler);
-  lv_obj_align(btnMinutesUp, lv_scr_act(), LV_ALIGN_IN_LEFT_MID, 20, -80);
-  lv_obj_set_height(btnMinutesUp, 40);
-  lv_obj_set_width(btnMinutesUp, 60);
-  txtMUp = lv_label_create(btnMinutesUp, nullptr);
-  lv_label_set_text(txtMUp, "+");
-  
-  btnMinutesDown = lv_btn_create(lv_scr_act(), nullptr);
-  btnMinutesDown->user_data = this;
-  lv_obj_set_event_cb(btnMinutesDown, btnEventHandler);
-  lv_obj_align(btnMinutesDown, lv_scr_act(), LV_ALIGN_IN_LEFT_MID, 20, +40);
-  lv_obj_set_height(btnMinutesDown, 40);
-  lv_obj_set_width(btnMinutesDown, 60);
-  txtMDown = lv_label_create(btnMinutesDown, nullptr);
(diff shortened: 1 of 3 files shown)


hubmartin:
Good work. I didn't know you can extend "clickable" area around some object. Seems like perfect fit for watches. This should be used almost everywhere in InfiniTime UI.

https://github.com/JF002/InfiniTime/pull/554/commits/d7c05f84f088cfe99527204d39672d491cafffca#diff-81ce0b0481912540170470032e08a92cb2353cd2dbe0acfdf49c0c19ef5191c8R32

Riksu9000:
> Good work. I didn't know you can extend "clickable" area around some object. Seems like perfect fit for watches. This should be used almost everywhere in InfiniTime UI.
> 
> [d7c05f8#diff-81ce0b0481912540170470032e08a92cb2353cd2dbe0acfdf49c0c19ef5191c8R32](https://github.com/JF002/InfiniTime/commit/d7c05f84f088cfe99527204d39672d491cafffca#diff-81ce0b0481912540170470032e08a92cb2353cd2dbe0acfdf49c0c19ef5191c8R32)

Usually the buttons should be made as big as possible. In this case it might look weird to have the buttons be larger than the time, so I just extended the click area instead. I think this UI could with a redesign in the future.

src/displayapp/screens/Timer.cpp
```diff
@@ -1,173 +1,123 @@
...
+      SetButtonsHidden(false);
+    } else if (seconds + minutes > 0) {
+      lv_obj_set_style_local_value_str(btnPlayPause, LV_BTN_PART_MAIN, LV_STATE_DEFAULT, Symbols::pause);
+      timerController.StartTimer((seconds + minutes * 60) * 1000);
+      SetButtonsHidden(true);
+    }
+  } else if (obj == buttons[minUp]) {
+    minutes = (minutes + 1) % 61;
+  } else if (obj == buttons[minDown]) {
+    if (minutes == 0) {
```
hatmajster, line 100: I think the same formula as for minUp could be used here: `minutes = (minutes - 1) % 61;`
Also, I think timer can go without any memory/display/count issue up to 99:59. Though maybe its better this way, because its easy to set one hour timer which is probably used quite often.

src/displayapp/screens/Timer.cpp
```diff
@@ -1,173 +1,123 @@
...
-      running{true},
-      timerController{timerController} {
-  
   time = lv_label_create(lv_scr_act(), nullptr);
   lv_obj_set_style_local_text_font(time, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, &jetbrains_mono_76);
   lv_obj_set_style_local_text_color(time, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_GRAY);
-  
-  uint32_t seconds = timerController.GetTimeRemaining() / 1000;
-  lv_label_set_text_fmt(time, "%02lu:%02lu", seconds / 60, seconds % 60);
+  lv_label_set_text_static(time, "00:00");
```
hatmajster, line 16: Don't know when first refresh was done (probably it was right at the beginning), but previous implementation was setting current remaining time if timer was already going.

src/displayapp/screens/Timer.cpp
Riksu9000: (- 1 % 61) on an unsigned 8 bit int is 11 and not 60. I don't want to change the behaviour much in this PR, but I remember trying to make it count higher some time ago, and having issues with it going above ~70 minutes for some reason.

src/displayapp/screens/Timer.cpp
Riksu9000: The Refresh() function is called at the end of the constructor, which will set the correct time right at the beginning. This placeholder value is to correctly align the label.

Riksu9000:
Added long press repeat. Since the button is above the number, it is hard to see, but I'd rather rework the UI separately.

https://user-images.githubusercontent.com/37774658/131247815-00ee8b43-835b-4ce3-ac0c-0f605cde0af6.mp4

PR #555, opened 2021-08-04
UI size setting

### LINKED ISSUE #540: Watchface with bigger font for Date and Health/Fitness Data
I wanted to give a Pine time to my father as a gift but I'm afraid of the fonts on the Display being too small to view without reading glasses. Would it be possible to change the digital Watchface or create a new one. The time on digital watchface is pretty readable but I think the date, HRM, step count and status symbols have to be bigger.

Thanks in advance.

Riksu9000:
Add option for larger UI elements to improve accessibility.

I had to add alphabets to the 42px font. The ttf for JetBrainsMono-Regular was missing, so I added it as well.

I added the font size changes from #564 here. 42px font was changed to 40px to fit 10 characters on screen.

TODO:
- [ ] Pick setting icon
- [x] Improve layout so double digit day can fit (or slightly smaller font?)
- [ ] Make more screens adaptive?
- [x] Update font conversion readme

Fixes #540

digi24
digi12
uisize
setting


hatmajster:
This looks great, though I wonder if this shouldn't just be a default setting... I think if everything will need to have two size modes (or more) it can over complicate code. Also, even new larger fonts are needed...

Riksu9000:
Maybe you're right considering how small the screen is, so there's not that much more that can be done at least currently. I think I'll adjust the watch face and open a new PR and see what happens.

PR #556, opened 2021-08-04
Set BLE icon on WatchFaceDigital to brighter color to better see on sunlight

hubmartin:
When you're on a bright sunlight, then the blue BLE icon is not easily visible on black background. This is more important when I ride a bike and phone is in my backpack so I could easily check whether I'm connected and don't miss notifications.

image

I like colorful LCD display but in practice the white text is the most readable.

Similar ui PRs/issues:
#540 
#555

### DIFF
--- a/src/displayapp/screens/WatchFaceDigital.cpp
+++ b/src/displayapp/screens/WatchFaceDigital.cpp
@@ -52,7 +52,7 @@ WatchFaceDigital::WatchFaceDigital(DisplayApp* app,
   lv_obj_align(batteryPlug, batteryIcon, LV_ALIGN_OUT_LEFT_MID, -5, 0);
 
   bleIcon = lv_label_create(lv_scr_act(), nullptr);
-  lv_obj_set_style_local_text_color(bleIcon, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(0x000000));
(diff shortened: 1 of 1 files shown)


Avamander:
Configurable please.

Riksu9000:
The official color for bluetooth is R: 0 G: 130 B: 252 (#0082FC) according to [bluetooth brand guide](https://www.bluetooth.com/develop-with-bluetooth/marketing-branding). Maybe with that it would be bright enough since it is be brighter than the #0000FF used currently.

hubmartin:
@Avamander Someone else would need to do that configurable.
@Riksu9000 good point, I set #0082FC. It is practically same, but looks beter and it is on-spec.
I would call this a day, forget the white and keep #0082FC.
Let me know if I should delete this PR in case you apply this fix in your #555 

Can you tell the difference? :)
image

hatmajster:
It looks a lot nicer on my pinetime. Though I can't tell if its really better for sunlight...
Its such a simple change that maybe JF002 will just merge this, so I would've left this PR. Maybe there are bluetooth icons in other places where You can also change this?

jonvmey:
The bluetooth icon is also displayed on the PineTimeStyle watchface if you wanted to change the colour there too.

Riksu9000 (approved):

JF002:
Colors are a matter of personal taste, but in this is a case of spec compliance, so... I merge :+1:

PR #558, opened 2021-08-04
Stopwatch - Extend button touch area

hubmartin:
Now you can touch up to 50 px above the button and it still gets recognized as a touch.


Riksu9000:
50 might be too much. If people get used to this extremely oversized active area, and it turns out it has to be shrinked, people wouldn't like that. This is why I made it a safe 10px in the other PR. This app could also be redesigned to start and stop on tapping the background for example.

Riksu9000:
This may cause accidental presses with #612

Riksu9000:
The stopwatch UI has been updated, obsoleting this change.

PR #560, opened 2021-08-06
BLE NUS - text console over BLE

hubmartin:
I used this BLE NUS uart console on other embedded NRF52 projects and it is really handy. You can log errors over BLE and debug your app.
To display data from NRF chip you need an app, or web browser.

Right now the firmware advertises has BLE NUS characteristic, can receive packet and replies the same data back. So low-level functionality is there, now I need to help with integration. More info under screenshots.

Console might be a way for not so skilled users to do some minor watch configuration which might not be possible on small screen. For example size, color and positions of watchface elements. Editing config, list and edit files in LittleFS. All that without reflashing, which is somehow hard for beginners to create a new watchface.

(sorry for big pictures, please see text in between them)

You can use NRF Toolbox app on phone (select UART, swipe right to see console, at the end of the command add newline before pressing SEND, see https://devzone.nordicsemi.com/f/nordic-q-a/33687/nrf-toolbox-2-6-0-uart-does-not-send-lf-cr-or-cr-lf-as-eol)

Screenshot_20210806_085652_no nordicsemi android nrftoolbox

Bluefruit app from Adafruit
Screenshot_20210806_085611_com adafruit bluefruit le connect

or use some WEBBLE (Web BLE) terminal console https://github.com/makerdiary/web-device-cli (not supported anymore, this sends single characters with remote echo, too much complex).

Screenshot from 2021-08-06 08-57-39
Screenshot from 2021-08-06 08-58-23

WEB BLE terminal working in any chromium-based browser
https://terminal.hardwario.com/

Whan now needs to be done is some object oriented C++ which is not my cup of tea. I'm struggling how to divide this because I see that in the code there is dependency injection in constructors. This is too high level for me so I'll be glad to hear how to partition BleNus class and create some Console class. Some code-prototypes would be better then explain me in C++ lingo what to do, I'm still learning C++.

Console class could be created in SystemTask and be somehow connected to BleNus which will send received characters to Console class over some function. These cahracters will be buffered (what OS primitives use to this?) until new line is received and command is parsed, executed and response send back to client).

Console will also pass filesystem `fs` and `lvgl` dependency so later console could list files, change color or position of LVGL objects on screen.

# Bonus points
Also Console might have some static (hope I use this term right) class which will send also all debug info from all parts of the firmware where logs are used. Right now all goes to the NRF_LOG_INFO macro, but we might create new function `log_info` which will send data to RTT `NRF_LOG_INFO` and also to some `Console.printf()` function. So you can see logs also over BLE. And I'm not sure how all this glue together. So any tips are welcome. I do not have plenty of time so the easisest solution would win. I would like to close this PR in the state of working console with some basic commandds. The NRF_LOG_INFO redirection might be part of another PR in future.

Thanks


src/components/ble/BleNus.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+BleNus::BleNus()
+  : characteristicDefinition {{
+                                .uuid = (ble_uuid_t*) &rxCharacteristicUuid,
+                                .access_cb = BleNusCallback,
+                                .arg = this,
+                                .flags = BLE_GATT_CHR_F_WRITE | BLE_GATT_CHR_F_WRITE_NO_RSP,
+                              },
+                              {
+                                .uuid = (ble_uuid_t*) &txCharacteristicUuid,
```
Avamander, line 78: C-style casts are discouraged and also unnecessary in this case. See either one of the PRs or one of the modernized services to see the best practice.

src/components/ble/BleNus.cpp
```diff
@@ -0,0 +1,92 @@
+#include "BleNus.h"
+
+using namespace Pinetime::Controllers;
+
+constexpr ble_uuid128_t BleNus::nusServiceUuid;
+constexpr ble_uuid128_t BleNus::rxCharacteristicUuid;
+constexpr ble_uuid128_t BleNus::txCharacteristicUuid;
+uint16_t BleNus::attr_read_handle;
+//static uint16_t conn_handle;
```
Avamander, line 9: If this is unused it should probably be removed.

Avamander (changes requested):

src/components/ble/BleNus.cpp
hubmartin: This is a PR draft to brainstorm, not a PR review. We can deal with details that later. I would appreciate some help/ideas how to integrate this, we can iron all other things later when I chage draft do review.

hubmartin:
What C++ feature I should use to buffer received characters from BLE? Nimble sends bytes in mbuf. In C I use char arrays, but I'm sure C++ has some stdlib feature in its sleeve. If you can point me to some example code or similar feature already somewhere in the firmware it would be great.

hubmartin:
Other question. The data arrives in Nimble thread context? I see that different classes uses its Process() function in SystemTask loop. I guess it is to save RAM resources from too many tasks and stacks. So I will process the console also in SystemTask.

hubmartin:
Can anyone from the experienced C++ developers help me please? @JF002 @Avamander @Riksu9000 @kieranc
I'm stuck for many hours with some stupid class definition mistake and cannot compile code. I'm getting really frustrated, and I'm sure for someone skilled in C++ it could take few minutes to find the issue. I'm a C developer and I'm learning C++ with this project and there's nobody else around me to help. :(
I've added BLE console which works, now I'd like to add the Console class, but even I followed the same way the `NimbleController` class is defined and added new source files to CmakeLists.txt, it looks like I'm missing something
```
In file included from /home/martin/Desktop/dev/watches/InfiniTime/src/components/console/Console.h:7,
                 from /home/martin/Desktop/dev/watches/InfiniTime/src/components/console/Console.cpp:2:
/home/martin/Desktop/dev/watches/InfiniTime/src/systemtask/SystemTask.h:118:29: error: 'Console' in namespace 'Pinetime::Components' does not name a type
  118 |       Pinetime::Components::Console console;
      |                             ^~~~~~~
```
The latest push has this issue.
Thanks for your help.

image

Riksu9000:
It's because of this line in Console.h
`#include "systemtask/SystemTask.h"`

Console requires systemtask but systemtask requires Console.

hubmartin:
@Riksu9000 Thanks a lot (I wish I asked few deays earlier). It makes perfect sense now. Now I can move a bit. Thanks.

hubmartin:
Ugly concept is working, added `console.Print()` command to `SystemTask` and getting results in Web BLE terminal.

TODO:
- [ ] transmit buffering, BLE NUS buffer might not send longer strings
- [ ] receiving buffering to receive incoming command and parse it
- [ ] Figure out how to integrate this console thing so you can call it from anywhere.
- [ ] Find out how and add option for routing `NRF_LOG_INFO` also to `console.Print`
- [ ] Find or create better terminal than https://wiki.makerdiary.com/web-device-cli/. We do not need sending every character separately and also remote echo is not efficient. Single input line with sending on enter and textarea with response will be best.

image

Avamander:
You can turn it into a logging backend now. There is a nice example there already.

hubmartin:
> You can turn it into a logging backend now. There is a nice example there already.

Can you be more specific please? Do you mean NRF logging and adding it by `nrf_log_backend_add` ? I keep that for later or someone else. Seems big bite for me now.

Avamander:
Check the `logging` folder and the relevant classes. It's really not a big bite :D.

hubmartin:
@Avamander I did, but then I run into error. What I understand is that NRF lib have different order of items in structure than in initializer? Is this issue of the NRF or InfiniTime project which has set errors for some warnings like this?
Seems like bad idea to fix it in NRF lib, is there some workaround?

I shoved parts of nrf backend registration to System task to see if I at least could it compile. Not sure I would be sucesfull.

image

hubmartin:
Hmm, I'm stuck again. I can send data to bleNus to send text to computer, but I have no idea how to properly connect it in the oposite way, so received characters over BLE goes to console class.

In the image below I'm passing to `console` class also `nimbleController` which contains `BleNus` class responsible for communication. So when I call `console.Print` it goes to `nimbleController.Print()` which calls `bleNus.Print` (this is also ugly, ready for suggestions).

And in the oposite way I have no idea how to properly with namespaces and classes connect so `BleNus` class which receives a character over BLE will pass it down through parent `nimbleController` to the `systemTask` whic contains `console.Received()` which adds bytes to RX buffer.

In C I would pass pointer to the function, but I'm sure there are different patterns in C++ how to deal with this.

`SystemTask.cpp` constructor
image

I drafted how terrible the situation is now with Print going over `nimbleController` which is also ugly. But how to connect it the other way from BleNus to the Console?

Thanks for help  @JF002 @Avamander @Riksu9000.

image

hubmartin:
It somehow works :)
https://youtu.be/pN13IN-AXJ4

hubmartin:
I've added ACC command to print X,Y,Z accelerometer data. Justa concept so we could for example create more robust wake on wrist
image

hatmajster (commented):
I've been testing this with bluefruit connect, and it works really great. This is a feature I really appreciate, as its possible to do some simple scripting/communication between phone/pc. However, I don't know if this shouldn't be a developer only thing, maybe it should be only compiled as an option. At least in current state. But maybe others will find appreciation for this too;)

src/components/ble/BleNus.h
```diff
@@ -0,0 +1,58 @@
...
+#include <host/ble_gap.h>
+#undef max
+#undef min
+#include "Version.h"
+#include <functional>
+
+
+// Service UUID
+// 6E400001-B5A3-F393-E0A9-E50E24DCCA9E
+#define NUS_SERVICE_UUID_BASE \
```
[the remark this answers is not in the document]
hubmartin: Moved directly to class

src/components/ble/BleNus.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+  res = ble_gatts_add_svcs(serviceDefinition);
+  ASSERT(res == 0);
+}
+
+void BleNus::SetConnectionHandle(uint16_t connection_handle) {
+    conn_handle = connection_handle;
+}
+
+void BleNus::Print(char *str)
```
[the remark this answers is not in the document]
hubmartin: Fixed on few places

src/components/ble/NimbleController.cpp
```diff
@@ -143,6 +147,8 @@ int NimbleController::OnGAPEvent(ble_gap_event* event) {
       /* A new connection was established or a connection attempt failed. */
       NRF_LOG_INFO("connection %s; status=%d ", event->connect.status == 0 ? "established" : "failed", event->connect.status);
 
+      bleNusService.SetConnectionHandle(event->connect.conn_handle);
```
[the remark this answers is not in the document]
hubmartin: Moved to the else section.
However not sure how to deal with disconnect. It seems to work ok after disconnect. Here I'm not so familiar with nimble stack so any suggestions are welcome.

src/components/console/Console.cpp
```diff
@@ -0,0 +1,151 @@
...
+    static uint32_t accCount = 0;
+
+    // Simple stupid comparison, later would be nice to add commands lookup table with argument parsing
+    if(hasCommandFlag)
+    {
+        hasCommandFlag = false;
+
+        // This AT > OK needs to be there, because https://terminal.hardwario.com/ waits for the answer
+        // When we use or create better webpage terminal, this can go out
+        if(strncmp(rxBuffer, "AT", 2) == 0)
```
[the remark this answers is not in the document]
hubmartin: Updated to function cmdCmp which returns true when compare is match.

src/components/console/Console.cpp
```diff
@@ -0,0 +1,151 @@
...
+        nimbleController.bleNus().Print(accBuf);
+    }
+}
+
+void Console::Received(char* str, int length)
+{
+    //char b[128];
+
+    // Wrap if input is too long without CR/LN
+    if(rxPos == bufferSize - 1)
```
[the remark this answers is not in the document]
hubmartin: Comparison moved inside the for loop, so every added character is checked whether fits to the buffer. Additional change to std::stringstrem could be done by someone experienced in C++ later.

src/systemtask/SystemTask.cpp
```diff
@@ -295,10 +347,12 @@ void SystemTask::Work() {
           xTimerStart(dimTimer, 0);
           break;
         case Messages::OnTouchEvent:
+          console.Print((char*)"Touch event\r\n");
```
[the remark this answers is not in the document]
hubmartin: Fixed by `const std::string`.

hubmartin:
I've added parsing parameters.

Commands (https://github.com/JF002/InfiniTime/pull/560/files#diff-dd8468434c59d9e2a95dbc4af16a663df286d51cc2703d68285cbecfefdbf08cR130)

LVGL - listst LVGL stats, number of elements on current screen and position and width, height of every element on current screen
COLOR #112233 - changes color of all elements on current screen. This is just a demo, not much userful. On digital watchface it changes color of all elements, on other screens just some
VIBRATE - vibrates a motor
WKUP - wake up display
SLEEP - sleep display
ACC - samples and sends 50 values X,Y,Z from accelerometer.

Riksu9000:
>I hope, that console would be also in release, not just debug. The reason is that users could configure colors, fonts and positions of watchfaces without need to recompile firmware.

We'd have to implement all the configuration options separately anyway, and at that point why not let it be configured in the firmware. I don't think we should make the end users need to use the console for anything.

src/systemtask/SystemTask.cpp
```diff
@@ -351,6 +363,9 @@ void SystemTask::Work() {
       batteryNotificationTick = xTaskGetTickCount();
     }
 
+    
+    console.Process();
+
```
Riksu9000, line 368: Is the accelerometer values reporting the only reason this is called periodically? Is there a better way to do it?

hubmartin: `Process()` was primarily added so the execution of the buffered command (and stack) is in the systemtask context and not the nimble callback context.
Then it was handy for accelerometer values cause it is called in a reasonable ~10Hz pace?

Riksu9000: You could move it to a SystemTask Message so it only gets run when needed, but the accelerometer reporting would need another solution. The accelerometer part should probably be a separate function.

hubmartin: Current commands are really to just show off the posibilities of console. If I do just the console then not all people might understand what it is good for. We can delete all the commands if that means faster merge with develop.

hubmartin: I got it now :) You mean separate function, but within current Console class? I'll do that.

hubmartin: Moved lvgl command and acceleromter debug to functions in f15ce01

Riksu9000: What about moving the Process() function to a new SystemTask Message? When a command is received, it could just push a message to systemtask to process it. The reason for separating AccelerometerDebug() was so it could be called from somewhere else than Process() since that is the only part that needs to be called periodically right?

hubmartin: I don't feel that it shuld be event driven. I can imagine that we might need some  polling in case Console.Print() will have too long message to fit into a single BLE packet, then you need to buffer and divide that and this should be also handled in Console Process polling function.
There is not any overhead in Process, since its checking just a bool and exits immediatelly.

src/components/console/Console.cpp
```diff
@@ -0,0 +1,241 @@
...
+        // Wrap if input is too long without CR/LN
+        if(rxPos == bufferSize - 1)
+        {
+            rxPos = 0;
+        }
+
+        rxBuffer[rxPos++] = str[i];
+        rxBuffer[rxPos] = '\0'; // terminate for debug print 
+
+        if(str[i] == 13 || str[i] == 10)
```
Riksu9000, line 234: Using `'\n'` and `'\r'` would make this clearer.

src/systemtask/SystemTask.cpp
Riksu9000: Couldn't the function be made to not return before the entire message is sent, requiring only one call to Process() even if the message needs to be split? Process() primarily just reads the received command, so having to call it multiple times for a single command doesn't really make sense and I don't see why it shouldn't be event based. Maybe you're trying to do too many things in one function.

src/components/console/Console.cpp
hubmartin: Not sure why I used numbers :) And I looked for them in ASCII table to have them right :facepalm:  a3317d8

src/systemtask/SystemTask.cpp
hubmartin: Changed console parsing to the new event Messages::ConsoleProcess. e94d0fd
Removed accelerometer completely from this PR to focus on main functionality.

hubmartin:
Added Python tool console https://github.com/hubmartin/InfiniTime-tools
You can still use WebBLE console https://terminal.hardwario.com/

JF002:
Thanks @hubmartin for this PR! 
I can see how useful the serial console over BLE could be for development and debugging purposes! 

>> I hope, that console would be also in release, not just debug. The reason is that users could configure colors, fonts and positions of watchfaces without need to recompile firmware.
> 
> We'd have to implement all the configuration options separately anyway, and at that point why not let it be configured in the firmware. I don't think we should make the end users need to use the console for anything.

I can't think of any user's use-case for this text console right know, except for some diagnostic purposes, maybe. But if this PR is lightweight (in RAM and ROM), we could still add it in the release build, any maybe enable it via a "developer mode" in the settings. What do you think?

---
@hubmartin I tried to use the console using [https://terminal.hardwario.com/](https://terminal.hardwario.com/) (on chromium on my Linux desktop) and nRF UART v2.0 and NRFToolbox (on Android). While them seem to connect to the watch and send commands, I receive no resonse. I'll try to debug the code, to see what's going wrong, but in the mean time, did you notice this behavior?

hubmartin:
@JF002 Thanks for testing and review.
The most commands are empty placeholders because I've noticed that this PR did not compiled properly with recovery firmware (which doesn't have filesystem, maybe LVGL..). So the only command that is replying now is "AT". See

https://github.com/InfiniTimeOrg/InfiniTime/pull/560/files#diff-dd8468434c59d9e2a95dbc4af16a663df286d51cc2703d68285cbecfefdbf08cR169

I don't know how to properly enable this only in the non-recovery application build, it seems like I would need to add lot of `#ifdef`s to the constructors. Don't know if that is the cleanest way and how to format it. Suggestions are welcome.

hubmartin:
BLE-NUS is using 424 bytes of RAM (0.74%) and 1456 byte of FLASH (0.29%).
The RAM is mainly 256 byte receive buffer which can be made smaller: 32-64 bytes

I would say that this is really small footprint. The question is how it starts to increase when we come with the first commands. I would personally try to solve that when resources becomes an issue.

ble-nus
```
           FLASH:      361728 B       480 KB     73.59%
             RAM:       54160 B        64 KB     82.64%
```
             
develop
```
           FLASH:      360272 B       480 KB     73.30%
             RAM:       53736 B        64 KB     81.99%
```

JF002:
@hubmartin regarding memory usage : this is not too bad, indeed :) I think this will be a nice tool for developers and hacker, even if we don't add more commands : they could be added locally by some devs when needed, but not necessarily pushed and merged.

I did try the "AT" command, but did not receive the response "OK". I'll try again later and keep you up to date !

hubmartin:
@JF002 No reply on "AT" is weird. This "AT">"OK" is needed for https://terminal.hardwario.com/ because it activates the input textbox. Without OK response it wouldn't allow you to type anything.

After connection it should show you AT and OK right after you select BT device:
image

It would be perfect to keep this console in the standard firmware so developers could use it to print something or create some temporary commands without commiting them. I would be curious how it will evolve.

JF002:
I did some more test with this NUS service:
I can easily connect using BLEFRUIT app and [Martin's tools](https://github.com/hubmartin/InfiniTime-tools).

But Chromium (on 2 x68 laptops and a PBP) and Chrome on Android do not work : they connect on BLE but send nothing to the NUS service (I checked with the debugger, `BleNus::OnDeviceInfoRequested()` is never called.

In the developer console, I found this error:
image

As if it couldn't find the NUS service (I triple checked : the service is correctly initialized and discovered during the discovery procedure).

So, I don't think this is an issue for this PR, but I probably won't be the only one to have this issue...

---
Regarding this PR, I think the code could be clean (remove unused comments and commented code, fix indentation, remove unneeded empty lines,...).

Then, I'm wondering : as this console will be targeted to developers first, is it OK to enable it all the time, or should we add a 'developer mode' to enable it?

geekbozu:
Some conjecture on the 256bytes, odds are that is based off the max MTU size, so the most data we can RX/TX at once. 

I would personally like to see this gated behind a debug build. Not sure if its needed for "release" code but just having it there when working in debug might be really nice :) then its almost the best of both worlds!

hubmartin:
Hi @JF002, I did all the suggestions from @Avamander. Let me know if it could be merged in current state. Thanks.

hubmartin:
Hi, another month passed with no response from anyone. Since I'm not so active now in InfiniTime development I guess that I woulnd't be able to babysit this PR to be without conflicts all the time, since it connects to SystemTask.cpp which is changed quite often.
Let me please know if this PR will be merged or not. I did all the clang formatting and cleanup.
Currently this BLE console does not emit any messages, has small footprint to be in main build, but it is ready for developers to build on it and debug stuff.
Looking for answer from anyone @JF002 @Avamander @geekbozu 
Thanks.

Riksu9000:
I'm not sure we really have a need for this. Since this wouldn't be used by end users, It's basically just for devs who don't want to open their watch or buy a devkit, but also want logs. I don't think this is many people. InfiniSim has logs. I've developed on a sealed watch, but it's slow and we don't recommend developing on a sealed watch anyway, so why would we provide a feature specifically to make that easier?

JF002:
@hubmartin I'm really sorry for not providing any feedback for so much time. I don't have any excuse except that I'm completely overwhelmed by the number of PRs to review :/

I'm 100% convinced that this feature could be really useful fore many "alternative/niche" use-cases : debug logs, data logging, integrated command line interpreter, integration with 3rd party apps,... It could also be uses to extract more data from the motion and HR sensor to help design better algorithm to process the data, and probably many more things.

However, most of those use-cases are targeted towards developers, hacker and tinkerers and do not fit well with our (new) [project vision](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/doc/InfiniTimeVision.md). Since we want to focus on this vision, I guess we'll have to do without this feature, unfortunately. But this could be the basis of an InfiniTime variant that target specifically those use-cases, of course!

PR #564, opened 2021-08-08
Larger font on digital watch face

### LINKED ISSUE #540: Watchface with bigger font for Date and Health/Fitness Data

Riksu9000:
At first I thought that the original watchface should be kept in case people prefer it. This is why I originally created #555. This PR instead modifies the existing watchface.

I shrunk the 42px font to 40px so it can fit 10 characters on screen. I added the alphabet to the font and updated the font readme.

Fixes #540

This is the difference between 40px and 42px.

12h
24h


stoehraj:
I'll be honest, I much prefer the current font sizes on the digital face -- with this change I find the date and step count to be a bit distracting from the actual time, which is the main reason I look at the watch in most cases.

I agree that not everything should be a setting, but perhaps this should be -- I know many devices offer settings to use larger fonts for accessibility purposes.

Just my two cents as somebody who recently got a PineTime.

Riksu9000:
> I'll be honest, I much prefer the current font sizes on the digital face -- with this change I find the date and step count to be a bit distracting from the actual time, which is the main reason I look at the watch in most cases.
> 
> I agree that not everything should be a setting, but perhaps this should be -- I know many devices offer settings to use larger fonts for accessibility purposes.

Do you prefer #555 or can I do something to improve this?

stoehraj:
I think that PR looks great!

PR #565, opened 2021-08-09
Add dice rolling app

stoehraj:
### Preview Video

https://user-images.githubusercontent.com/31697004/128671489-f3a153a2-a103-4d30-8fe4-19bd4dd53d4a.mp4

### Description
This PR adds a new dice rolling app.  I play D&D and thought it would be fun to be able to roll dice on my PineTime, so here we are.  I thought others might enjoy it too.

This is what the basic UI looks like:

dice_ui

There are buttons to select the number and size of dice.  The number of dice can be between 1 and 99, and the supported dice sizes are d2, d3, d4, d6, d8, d12, d20, and d100.

When the desired number and size of dice have been configured, the "Roll" button will, well, roll the dice.  Rolls can also be made by slowly rolling the wrist, as shown below:

https://user-images.githubusercontent.com/31697004/128671511-c4d74205-6235-47f4-ac0b-bc59e43df45a.mp4

If the roll is made using the motion control, the watch will vibrate, giving feedback that the roll occurred.

After rolling, the first line of the output box at the top of the screens shows the total result, and, if multiple dice are rolled, the second line shows each individual die roll.


src/displayapp/screens/Dice.cpp
```diff
@@ -0,0 +1,231 @@
...
+  lv_label_set_long_mode(resultIndRollsLabel, LV_LABEL_LONG_SROLL_CIRC);
+  lv_obj_set_size(resultIndRollsLabel, 220, 30);
+  lv_obj_set_style_local_text_color(resultIndRollsLabel, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_BLACK);
+  lv_label_set_text(resultIndRollsLabel, "");
+
+  configCont = lv_cont_create(lv_scr_act(), nullptr);
+  lv_obj_set_size(configCont, 240, 160);
+  lv_obj_set_pos(configCont, 0, 80);
+  lv_obj_set_style_local_bg_color(configCont, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_BLACK);
+
```
Riksu9000, line 57: I don't think this is necessarily needed. Does it make the code cleaner?

src/displayapp/screens/Dice.cpp
```diff
@@ -0,0 +1,231 @@
...
+  rollBtn = lv_btn_create(configCont, nullptr);
+  rollBtn->user_data = this;
+  lv_obj_set_event_cb(rollBtn, roll_event_handler);
+  lv_obj_set_size(rollBtn, 70, 150);
+  lv_obj_set_pos(rollBtn, 165, 5);
+  lv_obj_set_ext_click_area(rollBtn, 5, 5, 5, 5);
+  rollBtnLabel = lv_label_create(rollBtn, nullptr);
+  lv_label_set_text(rollBtnLabel, "Roll");
+
```
Riksu9000, line 66: You should use `lv_obj_align()` to automatically align all buttons to edges. Is there a reason to leave 5px on the edges? The buttons are already large so do we need to extend the click area? I personally wouldn't use it unless it's necessary. You could also use `lv_obj_set_style_local_value_str()` instead of labels to reduce some code.

src/displayapp/screens/Dice.cpp
```diff
@@ -0,0 +1,231 @@
...
+    motorController.SetDuration(60);
+    rollDice();
+  }
+  return running;
+}
+
+bool Dice::OnTouchEvent(Pinetime::Applications::TouchEvents event) {
+  return true;
+}
+
```
Riksu9000, line 180: This makes it so the app can't be exited by swiping. Any reason to do so?

src/displayapp/screens/Dice.cpp
```diff
@@ -0,0 +1,231 @@
...
+           Controllers::MotorController& motorController,
+           System::SystemTask& systemTask)
+  : Screen(app),
+    systemTask {systemTask},
+    motionController {motionController},
+    motorController {motorController},
+    numDice(1),
+    diceSizeIndex(3) {
+
+  app->SetTouchMode(DisplayApp::TouchModes::Polling);
```
Riksu9000, line 160: This improves responsiveness, but is kind of hacky. I think it's fine though.

Riksu9000 (commented):
A very nice looking app and well written code.

hubmartin:
I might overlook something - but if you use `rand()` and do not set seed by `srand()`, do you get the same pseudorandom numbers again after InfiniTime reboots?
I cannot see `srand` anywhere in the InfiniTime code, so either it should be set system-wide after boot based on NRF52832 RNG peripheral, or in the app where seed could be for example uptime in milliseconds.
Otherwise, nice app. Thanks for adding this feature!

src/components/motion/MotionController.cpp
```diff
@@ -31,13 +31,43 @@ bool MotionController::ShouldWakeUp(bool isSleeping) {
   }
   return false;
 }
+
+bool MotionController::ShouldRollDice() {
+  if (z > 0) {
+    diceRollFirstTrigger = false;
+    return false;
+  }
```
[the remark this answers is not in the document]
stoehraj: Sounds good -- definitely can change that.  I'm thinking `WristRolled()` or something similar.

src/displayapp/fonts/jetbrains_mono_bold_20.c
```diff
@@ -1,17 +1,17 @@
 /*******************************************************************************
  * Size: 20 px
  * Bpp: 1
- * Opts: 
+ * Opts:
```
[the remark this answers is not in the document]
stoehraj: I had to regenerate this file because I added the "dice" symbol, and I avoided it when running clang-format, but I just realized that there's a git hook (hooks/pre-commit) that must have done this when I made my commit...  I'll fix this and look into filtering out certain files from that git hook

src/displayapp/screens/Dice.cpp
stoehraj: I suppose it's not necessary, but is it hurting anything?  It just helped me conceptually get everything organized during development.  I can rework it and remove it if that would be preferable.

src/displayapp/screens/Dice.cpp
stoehraj: I was originally playing around with using Rollers from LVGL instead of buttons for selecting the number and size of dice, and the swiping gestures got in the way of that.  I ended up reworking it to use buttons because functionality and performance with the rollers wasn't ideal, and I must have left this in.  I'll take it out.

stoehraj:
> I might overlook something - but if you use `rand()` and do not set seed by `srand()`, do you get the same pseudorandom numbers again after InfiniTime reboots?
> I cannot see `srand` anywhere in the InfiniTime code, so either it should be set system-wide after boot based on NRF52832 RNG peripheral, or in the app where seed could be for example uptime in milliseconds.
> Otherwise, nice app. Thanks for adding this feature!

I saw that another app was using `rand()` (Twos) without anything additional, so I figured everything was all set up for it.  I just tested and do indeed get the same output if I do the same roll after a reboot.  I'll take care of that.

src/displayapp/screens/Dice.cpp
stoehraj: I saw that this was something that other apps had done so I included it as well.  If it's fine as-is I will just leave it.

src/displayapp/screens/Dice.cpp
stoehraj: The 5px space outside the buttons was primarily aesthetic.  I did not have it originally and the screen looked a bit cramped with all of the buttons touching, so I added the space and extended the click area to compensate for that.  It does function without extending the click area, but I found that it felt better with it included.  As far as aligning, I found it easier to pack everything in by manually placing things -- I'm worried about getting everything to look right just using `align`.

Would `lv_obj_set_style_local_value_str()` still apply even though it's two separate lines of text?  I found it easier to break the text out into two labels so that both pieces could easily be updated separately.

Riksu9000: `lv_obj_set_style_local_value_str` can be used for buttons. The results label is fine.

I totally agree that buttons shouldn't be touching, but what i meant was that the buttons aren't actually touching the edge of the screen.

stoehraj: Oh -- totally missed that this was for the label on the button.  I apologize.  Yeah, I can give that a shot.  And I can move the buttons out to touch the edge of the screen, I just found that the uniform padding made it a bit easier to keep everything aligned/even.

src/displayapp/screens/Dice.h
```diff
@@ -0,0 +1,50 @@
...
+    void diceSizeIncBtnEventHandler(lv_event_t event);
+    void diceSizeDecBtnEventHandler(lv_event_t event);
+    void rollBtnEventHandler(lv_event_t event);
+
+  private:
+    Pinetime::System::SystemTask& systemTask;
+    Controllers::MotionController& motionController;
+    Controllers::MotorController& motorController;
+
+    static const uint8_t diceSizes[];
```
hatmajster, line 32: I think You could use `std::array` here

src/displayapp/screens/Dice.cpp
```diff
@@ -0,0 +1,231 @@
...
+  lv_obj_set_ext_click_area(diceSizeDecBtn, 5, 5, 10, 5);
+  diceSizeDecBtnLabel = lv_label_create(diceSizeDecBtn, nullptr);
+  lv_label_set_text(diceSizeDecBtnLabel, "-");
+}
+
+void Dice::rollDice() {
+  uint8_t roll = 0;
+  uint16_t result = 0;
+
+  lv_label_set_text(resultTotalLabel, "");
```
hatmajster, line 124: Is this necessary? Its set at the end of function...

src/components/motion/MotionController.cpp
```diff
@@ -31,13 +31,43 @@ bool MotionController::ShouldWakeUp(bool isSleeping) {
...
+
+  if (y < -240 && diceRollFirstTrigger) {
+    diceRollFirstTrigger = false;
+    return true;
+  }
+
+  return false;
+}
 void MotionController::IsSensorOk(bool isOk) {
```
hatmajster, line 58: Maybe add line breaks here?

src/components/motion/MotionController.cpp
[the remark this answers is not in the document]
hatmajster: I think that `ShouldRollDice` and `ShouldWakeUp` could have its logic somehow combined...

src/displayapp/screens/Dice.cpp
```diff
@@ -0,0 +1,231 @@
...
+bool Dice::OnTouchEvent(Pinetime::Applications::TouchEvents event) {
+  return true;
+}
+
+void Dice::numDiceIncBtnEventHandler(lv_event_t event) {
+  if (event != LV_EVENT_CLICKED) {
+    return;
+  }
+
+  if (numDice < 99) {
```
hatmajster, line 186: Such high number is nice, but can You see all 99 dice results in the app? I think something like 10 would be enough

src/components/motion/MotionController.cpp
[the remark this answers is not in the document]
Riksu9000: That would make sense. Maybe for now this app could use `ShouldWakeUp()`, and in another PR that could be improved?

src/displayapp/screens/Dice.cpp
stoehraj: You can certainly see the total -- and the individual rolls label uses the circular scroll long mode, so *technically* you could see all of the individual rolls, though it's unlikely somebody would wait all the way through that.

10 is definitely a bit low, but I'd be happy to lower it to something else if needed.  At 99 100-sided dice it's noticeably slower on the display, but it still handles it mostly fine.

src/displayapp/screens/Dice.cpp
stoehraj: You're right, it shouldn't be necessary.  Was trying to keep it consistent with the individual rolls label that does need to be cleared in the beginning, but there's not really a purpose to it.  I'll remove it.

src/displayapp/screens/Dice.h
stoehraj: I'll switch it over to `std::array` if that would be preferable.

src/components/motion/MotionController.cpp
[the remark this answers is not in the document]
stoehraj: I actually did try using the `ShouldWakeUp` logic and it was way too sensitive for the dice roll.  It was rolling 3-4 times in one motion.

While they're somewhat similar motions, I think they're different enough to warrant separate functions.  The one I made for the dice roll is intended to be much more restrictive to avoid unintentional rolls.

src/components/motion/MotionController.cpp
stoehraj: I guess I'm not sure what you mean -- these lines don't look too long to me.  They fit with the clang-format, at least.

src/displayapp/fonts/jetbrains_mono_bold_20.c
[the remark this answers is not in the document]
stoehraj: I resolved this in the latest changes.

src/displayapp/screens/Dice.cpp
```diff
@@ -0,0 +1,231 @@
...
+  resultCont = lv_cont_create(lv_scr_act(), nullptr);
+  lv_obj_set_size(resultCont, 240, 80);
+  lv_obj_set_pos(resultCont, 0, 0);
+  lv_cont_set_layout(resultCont, LV_LAYOUT_COLUMN_LEFT);
+  lv_obj_set_style_local_pad_all(resultCont, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, 10);
+
+  resultTotalLabel = lv_label_create(resultCont, nullptr);
+  lv_obj_set_size(resultTotalLabel, 220, 30);
+  lv_obj_set_style_local_text_color(resultTotalLabel, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_BLACK);
+  lv_label_set_text(resultTotalLabel, "");
```
jonvmey, line 45: `lv_label_set_text` and `lv_label_set_text_fmt` both need to allocate memory to copy the text into. `lv_label_set_text_static` doesn't allocate and should be preferred when setting the text to a static string literal.

src/displayapp/screens/Dice.cpp
```diff
@@ -0,0 +1,231 @@
...
+  numDiceIncBtnLabel = lv_label_create(numDiceIncBtn, nullptr);
+  lv_label_set_text(numDiceIncBtnLabel, "+");
+
+  numDiceLabelCont = lv_cont_create(configCont, nullptr);
+  lv_obj_set_size(numDiceLabelCont, 80, 40);
+  lv_obj_set_pos(numDiceLabelCont, 0, 60);
+  lv_obj_set_style_local_bg_color(numDiceLabelCont, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_BLACK);
+  lv_cont_set_layout(numDiceLabelCont, LV_LAYOUT_CENTER);
+  numDiceLabel = lv_label_create(numDiceLabelCont, nullptr);
+  lv_label_set_text_fmt(numDiceLabel, "%d", 1);
```
jonvmey, line 82: If you're formatting with a constant value you might as well just `lv_label_set_text_static(numDiceLabel, "1");` instead and avoid the cost of formatting. Same applies to setting `diceSizeLabel` below.

src/displayapp/screens/Dice.cpp
```diff
@@ -0,0 +1,231 @@
+#include "Dice.h"
+
+#include "Screen.h"
+#include "Symbols.h"
+#include "lvgl/lvgl.h"
+
+using namespace Pinetime::Applications::Screens;
+const uint8_t Dice::diceSizes[] = {2, 3, 4, 6, 8, 12, 20, 100};
```
jonvmey, line 8: The "standard" dice set also includes d10.

src/displayapp/screens/Dice.h
jonvmey: I would also suggest adding a constant for the number of dice sizes supported (which is non-optional if using `std::array` anyways). Otherwise it's implicitly defined by the definition of `diceSizes` which can cause issues if it ever changes in the future and the other places where it's currently explicitly repeated (ie. `diceSizeIncBtnEventHandler`) don't get updated to match.

src/displayapp/screens/Dice.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+void Dice::rollDice() {
+  uint8_t roll = 0;
+  uint16_t result = 0;
+
+  lv_label_set_text(resultTotalLabel, "");
+  lv_label_set_text(resultIndRollsLabel, "");
+
+  for (int i = 0; i < numDice; i++) {
+    roll = rand() % diceSizes[diceSizeIndex] + 1;
```
jonvmey, line 128: Don't think that anyone will particularly care but this implementation creates a slightly biased distribution of results for any dice sizes that are not a factor of `RAND_MAX` (which is all of the current ones).

src/displayapp/screens/Dice.cpp
stoehraj: Oh, yep, you're absolutely right -- missed the d10 somehow.  I'll make sure that gets included with future changes.

src/displayapp/screens/Dice.cpp
stoehraj: Interesting.  Have any advice for a different implementation?  Or do you think it's good enough as-is?

src/displayapp/screens/Dice.cpp
stoehraj: @jonvmey Fixed this in the latest changes.

src/displayapp/screens/Dice.cpp
stoehraj: @jonvmey Fixed this in the latest changes.

src/displayapp/screens/Dice.cpp
stoehraj: @jonvmey Fixed this in https://github.com/JF002/InfiniTime/pull/565/commits/42b048d648c6f2aad3064e1760771321f0d08e19

src/displayapp/screens/Dice.cpp
stoehraj: @hatmajster Fixed this in https://github.com/JF002/InfiniTime/pull/565/commits/42b048d648c6f2aad3064e1760771321f0d08e19

src/displayapp/screens/Dice.h
stoehraj: @hatmajster I was able to implement this using `std::array` but when I did, the application would fail to load unless I opened a couple other apps before attempting to open the dice app.  I looked around a bit and other places in the code (like `DateTimeController`) use the built-in array for a similar purpose, so I think I'm just going to leave it as-is.

@jonvmey Added a constant for the number of dice sizes in https://github.com/JF002/InfiniTime/pull/565/commits/42b048d648c6f2aad3064e1760771321f0d08e19

src/displayapp/screens/Dice.cpp
stoehraj: @Riksu9000 Fixed this in https://github.com/JF002/InfiniTime/pull/565/commits/42b048d648c6f2aad3064e1760771321f0d08e19

src/displayapp/screens/Dice.cpp
stoehraj: @Riksu9000 I played around with the buttons and could not get it to look nice without the padding.  I think the math just works out better with the 5px padding.  I did add the same 5px padding to result label in https://github.com/JF002/InfiniTime/pull/565/commits/42b048d648c6f2aad3064e1760771321f0d08e19, which makes the edges of it line up with the edges of the buttons -- it looks much nicer with this.  Thoughts?

Also, I did remove the rollBtnLabel and replace it with the local value string style in the same commit.

src/components/motion/MotionController.cpp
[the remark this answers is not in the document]
stoehraj: @Riksu9000 I updated the name of the function in https://github.com/JF002/InfiniTime/pull/565/commits/42b048d648c6f2aad3064e1760771321f0d08e19

stoehraj:
@hubmartin I could use some help with the `srand` call for seeding the random number generator.  I tried using `srand(time(nullptr));` in my object's constructor and also tried doing `srand(std::chrono::system_clock::now().time_since_epoch().count());` and neither seemed to have an impact -- that's to say, I could still reproduce getting the same roll each time after a reboot.  Would it be better to get the system time using the `DateTimeController` functions?  Also would it be better to have the seed set system-wide as you mentioned?  What would that look like?

src/displayapp/screens/Dice.cpp
Riksu9000: By using 76px buttons and 6px space between buttons everything fits perfectly. #452 and music app in #429 for reference. They also use `lv_obj_align()`.

src/displayapp/screens/Dice.h
Riksu9000: The constant doesn't completely fix the issue. Still if just the array was changed, the code would break. If you're not going to use std::array, sizeof would probably be better.

src/displayapp/screens/Dice.cpp
Riksu9000: Couldn't the value be initialized in the header instead?

src/displayapp/screens/Dice.cpp
Riksu9000: You could use `lv_obj_set_style_local_value_str()` to replace the increment and decrement button labels as well.

hubmartin:
My guess is that constructors are run immediatelly after start and they are called cycle-perfect every time, so you seed the random, but it is same every time. You have to call srand somewhere where the called function is run based on user's input. It might be `Dice::createScreen()` which if i'm not mistaken is called when you run application

Then I would suggest to seed rather some milliseconds counter/timer. But your `srand(std::chrono::system_clock::now().time_since_epoch().count());` might work, or I see also `dateTimeController.Uptime().count()` value. Its seconds counter so it might be ok, but I would suggest to find millisecond counter somewhere. Not sure whether it has some function or it needs to be added, because it seems like lot of them returns just a seconds.

virtualbeehive:
This is a cool idea

geekbozu:
With @hubmartin suggestion of using the screen constructor using the current time should be much more effective, ```dateTimeController.CurrentDateTime().count()``` should work well as long as the clock is set.

src/displayapp/screens/Dice.cpp
stoehraj: @Riksu9000 Added this in https://github.com/JF002/InfiniTime/pull/565/commits/7c0174284033c67c6e11b61d2829570715e35cb0

src/displayapp/screens/Dice.h
stoehraj: @Riksu9000 Fixed this in https://github.com/JF002/InfiniTime/pull/565/commits/7c0174284033c67c6e11b61d2829570715e35cb0

src/displayapp/screens/Dice.cpp
stoehraj: @Riksu9000 Fixed the button size/spacing and switched to `lv_obj_set_style_local_value_str()` for all buttons in https://github.com/JF002/InfiniTime/pull/565/commits/7c0174284033c67c6e11b61d2829570715e35cb0

stoehraj:
@hubmartin @geekbozu I did a bit more looking at this -- I ended up using `srand(dateTimeController.CurrentDateTime().time_since_epoch().count());` in the constructor, and that seemed to do the trick.  No repeat rolls after rebooting and trying the same roll (10 six-sided dice) three times.

I believe the constructor is actually called every time the app is opened, rather than once every boot, though with this implementation (using current time rather than uptime) it shouldn't really matter either way.

geekbozu:
I gave this a try the other day and it worked really well I will try the new changes and see if it works out well. Never can go wrong with more features. 

Is there any plans to actually make apps configurable on compile time so we can remove stuff we don't want when we finally get to the point of running out of room. Not necessarily something you need to figure out just more of a curiosity.

src/displayapp/screens/Dice.h
```diff
@@ -0,0 +1,51 @@
+#pragma once
+
+#include "systemtask/SystemTask.h"
+#include "components/motion/MotionController.h"
+#include "components/motor/MotorController.h"
+
+#include <array>
+
+namespace Pinetime::Applications::Screens {
```
Riksu9000, line 9: This shows a warning `nested namespace definition is a C++17 extension`

src/displayapp/screens/Dice.h
```diff
@@ -0,0 +1,51 @@
+#pragma once
+
+#include "systemtask/SystemTask.h"
+#include "components/motion/MotionController.h"
+#include "components/motor/MotorController.h"
+
+#include <array>
+
```
Riksu9000, line 8: This isn't needed if std::array isn't used

src/displayapp/screens/Dice.cpp
```diff
@@ -0,0 +1,231 @@
...
+Dice::Dice(DisplayApp* app,
+           Controllers::MotionController& motionController,
+           Controllers::MotorController& motorController,
+           System::SystemTask& systemTask)
+  : Screen(app),
+    systemTask {systemTask},
+    motionController {motionController},
+    motorController {motorController},
+    numDice(1),
+    diceSizeIndex(3) {
```
Riksu9000, line 149: Redundant initializers warning

src/displayapp/screens/Dice.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+#include "Screen.h"
+#include "Symbols.h"
+#include "lvgl/lvgl.h"
+
+using namespace Pinetime::Applications::Screens;
+const uint8_t Dice::diceSizes[] = {2, 3, 4, 6, 8, 12, 20, 100};
+
+static void num_dice_inc_event_handler(lv_obj_t* obj, lv_event_t event) {
+  auto dice = static_cast<Dice*>(obj->user_data);
```
Riksu9000, line 11: multiple `auto dice can be declared as auto* dice` warnings

src/displayapp/screens/Dice.cpp
```diff
@@ -0,0 +1,219 @@
...
+  numDiceIncBtn = lv_btn_create(configCont, nullptr);
+  numDiceIncBtn->user_data = this;
+  lv_obj_set_event_cb(numDiceIncBtn, num_dice_inc_event_handler);
+  lv_obj_set_size(numDiceIncBtn, 76, 52);
+  lv_obj_align(numDiceIncBtn, NULL, LV_ALIGN_IN_TOP_LEFT, 0, 0);
+  lv_obj_set_style_local_value_str(numDiceIncBtn, LV_BTN_PART_MAIN, LV_STATE_DEFAULT, "+");
+
+  numDiceLabelCont = lv_cont_create(configCont, nullptr);
+  lv_obj_set_size(numDiceLabelCont, 76, 50);
+  lv_obj_align(numDiceLabelCont, NULL, LV_ALIGN_IN_LEFT_MID, 0, 0);
```
Riksu9000, line 67: multiple `Use nullptr` warnings

Riksu9000 (commented):
This code generates some warnings and I copied them here. Other than that this PR looks good.

src/components/motion/MotionController.cpp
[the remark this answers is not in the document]
Riksu9000: Maybe the function could return the "strength" of the roll, and some apps could just have a higher threshold. Then just one function would be enough. It doesn't need to be changed in this PR, but just an idea.

src/displayapp/screens/Dice.h
```diff
@@ -29,8 +28,9 @@ namespace Pinetime::Applications::Screens {
     Pinetime::System::SystemTask& systemTask;
     Controllers::MotionController& motionController;
     Controllers::MotorController& motorController;
+    Controllers::DateTime& dateTimeController;
```
jonvmey, line 31: There's no need to store a reference to `dateTimeController` inside of this object. It's only used once inside of the constructor.

On that note `Dice` doesn't need to know anything about date/time at all. It simply needs a seed value which in this case coincidentally gets supplied with the current time. I'd personally suggest changing the constructor parameter to be an `unsigned seed` instead and passing in `dateTimeController.CurrentDateTime().time_since_epoch().count()` to it.

toitoinou:
For the seed part for the random number generation, in an embedded software project I participate with we used ADC values on unused pins or very fluctuant voltage in order to have something different each time. I don't know if we have an ADC on the nrf but we can try to use something proportional to the battery voltage for example before filtering if possible.

hubmartin:
@toitoinou NRF52 has special and certified RNG peripheral exactly for this reason. I did some quick tests and had something wrong, the numbers did not come. If anyone could look into that it would be great to have system-wide random seed right after boot for the whole system.

tmilburn:
Using the RNG is pretty simple on the NRF52 https://devzone.nordicsemi.com/f/nordic-q-a/19429/random-number-without-softdevice. I suggest seeding using srand on boot-up.

hubmartin:
@geekbozu we don't need to use RNG for `rand`. Single RNG number on boot and applying it to the random seed `srand` should be enough. Then `srand` gives us number immediatelly. RNG is "slow" if you need many numbers in a row. Maybe could be an issue, maybe not.

jp-bennett:
This is such a cool idea. Bump, to see this finished and included in the next build.

geekbozu:
> 
> 
> @geekbozu we don't need to use RNG for `rand`. Single RNG number on boot and applying it to the random seed `srand` should be enough. Then `srand` gives us number immediatelly. RNG is "slow" if you need many numbers in a row. Maybe could be an issue, maybe not.

Yeah this makes a ton of sense. Even more so for applications where randomness is only partially critical :P.

stoehraj:
> This is such a cool idea. Bump, to see this finished and included in the next build.

Hey -- thanks for the support.  I'd love to finish this up and get it included in the next build.

Unfortunately life has gotten in the way a bit -- I hopefully should have some time in the next couple weeks to come back to it.

MysteriousLog6:
A small recommendation.

When the icon is pressed change the icon to the number that has been rolled , for other things long pressing it can open the full app.

stoehraj:
I just pushed new changes that make the dice app compatible with 1.6.0.  I likely still have to revisit the `tidy` warnings, and also the wrist-motion rolling seems to not work on the new version.  I'll try to look into these items over the next few days.

The app itself runs fine otherwise.

Riksu9000:
There are actually two shake detection algorithms in the works. #691 #660. Seems like the roll detection is quite sensitive, so would a shake work better here too?

stoehraj:
> There are actually two shake detection algorithms in the works. #691 #660. Seems like the roll detection is quite sensitive, so would a shake work better here too?

Oh, that's awesome -- that would be perfect for this application.  I'll keep my eyes on that.

src/displayapp/screens/Dice.h
stoehraj: Fixed this in https://github.com/JF002/InfiniTime/pull/565/commits/75c7e280842f08b91be91de121d600ec5f87defd

src/displayapp/screens/Dice.h
stoehraj: @Riksu9000 Fixed this in https://github.com/JF002/InfiniTime/pull/565/commits/75c7e280842f08b91be91de121d600ec5f87defd

src/displayapp/screens/Dice.cpp
stoehraj: @Riksu9000 Fixed this in https://github.com/JF002/InfiniTime/pull/565/commits/75c7e280842f08b91be91de121d600ec5f87defd

src/displayapp/screens/Dice.cpp
stoehraj: @Riksu9000 Fixed this in https://github.com/JF002/InfiniTime/pull/565/commits/75c7e280842f08b91be91de121d600ec5f87defd

src/displayapp/screens/Dice.cpp
stoehraj: @Riksu9000 Fixed this in https://github.com/JF002/InfiniTime/pull/565/commits/75c7e280842f08b91be91de121d600ec5f87defd

src/displayapp/screens/Dice.h
stoehraj: @jonvmey You're right -- this is a much better way to do it.  Implemented in https://github.com/JF002/InfiniTime/pull/565/commits/75c7e280842f08b91be91de121d600ec5f87defd

stoehraj:
@Riksu9000 -- I believe I have all format/tidy issues worked out.  At least I didn't see anymore on my end, though I'm not entirely sure I was using `clang-tidy` correctly.  In the meantime, I might just leave this PR open to see if progress gets made on the shake algorithm in the near future.  Or do you think it would be better to just strip out my attempt at the motion controls, get the PR pulled in without them, and add motion controls back in later once #691 is pulled in?

Either way, I'd like to squash my commits once the code is finalized and before completing the PR just to keep the git history a bit more clean.

Riksu9000:
The reason that the roll gesture isn't working is because `Refresh()` now requires an LVGL task.

This will call the `Refresh()` function every 20 ticks through the callback function, which is defined in Screen.cpp
```
taskRefresh = lv_task_create(RefreshTaskCallback, LV_DISP_DEF_REFR_PERIOD, LV_TASK_PRIO_MID, this);
```

But maybe it's best to remove the gesture and add it separately, as this app works fine without it.

One more thing I noticed while testing is that maybe on pressing the roll button the watch should vibrate or give some other feedback, so we can tell that the tap was actually registered if it ends up rolling the same number.

geekbozu:
I would recommend removing the roll and adding it back later. There should be some changes coming down the pipeline on how the BMA interface is defined...one less thing to update with that PR would be nice ;)

hubmartin:
@stoehraj Hi, was working on other PR and discovered that Nimble is already using RNG peripheral and it has nice driver and wrapper for `rand()`. All I had to do was:

`#include <controller/ble_ll.h>`
and then call `uint32_t r = ble_ll_rand();` to get the value.

It seems like Nimble has some buffer of pre-generated values so it seems like you can call `ble_ll_rand` instead of `rand` directly without any waiting for RNG to generate numbers.
This also explains why I was not able to use RNG, since Nimble reconfigured registers for its own use.

Riksu9000:
@stoehraj Hi. Could you update this PR, so we can merge it hopefully?

medeyko:
I think that in the case when n1, d2 the indicator also should write (heads) or (tails) for those who wishes to flip a coin,  see #919

medeyko:
@Riksu9000
> update this PR, so we can merge it hopefully

As far as I see, @stoehraj doesn't respond. What should be updated in this PR? Maybe it could be done w/o @stoehraj?

stoehraj:
> @stoehraj Hi. Could you update this PR, so we can merge it hopefully?

@Riksu9000 Hey -- apologies for the extremely late response and lack of activity.  I see you've changed some stuff over the past few days.  I just flashed the latest stuff onto my unit to play around with a bit.

I do want to try out what @hubmartin suggested with using `ble_ll_rand` rather than just `rand` for getting the random values.  I'll report back.

Also, if nothing else, I want to squash all of my prior commits to clean up the git history a bit before this gets merged in.

stoehraj:
@hubmartin -- the `ble_ll_rand()` call seems to work really nicely.  Just pushed up changes to move to that over the normal `rand()`.

stoehraj:
@Riksu9000 -- do you care if I squash your commits in with mine?  Or would you prefer to keep yours separate?

Riksu9000:
@stoehraj Nah I don't care. Do what you want.

stoehraj:
@Riksu9000 -- just took care of all of the squashing.  Got everything into one nice commit and (hopefully) ready to merge in.

medeyko:
It is a miniscule moment, but I think it might be a right thing to handle it:
`roll = ble_ll_rand() % diceSizes[diceSizeIndex] + 1;` introduces so called "modulo bias".
See f.e. https://funloop.org/post/2015-02-27-removing-modulo-bias-redux.html

stoehraj:
> It is a miniscule issue, but I think it might be a right thing to handle it: `roll = ble_ll_rand() % diceSizes[diceSizeIndex] + 1;` introduces so called "modulo bias". See f.e. https://funloop.org/post/2015-02-27-removing-modulo-bias-redux.html

@medeyko -- I just implemented this in the latest changes.

src/displayapp/screens/Dice.cpp
stoehraj: @jonvmey @tmilburn -- just implemented a fix for this in the latest changes!

medeyko:
Sorry for persisting, but I have a question again: are there any issues left with this app?
It looks well-written and compact, but still useful.

Is it possible to do anything in order to help to include it in the firmware?

Avamander:
@medeyko It's difficult to say, the amount of available RAM and storage is getting very very close. It's difficult to choose which new features should be merged first, if there's even space available.

How to keep track of features that depend on additional resources, or do not change the memory footprint and could be instantly merged, is currently WIP. 

Hopefully we'll be able to more clearly communicate what's the status of each PR soon.

w4tsn:
@Avamander a bit off-topic to this PR but as a reply to your thoughts: maybe, just maybe, there need to be some way of selecting which apps to include in the build or even have "editions/presets"

JF002:
> @Avamander a bit off-topic to this PR but as a reply to your thoughts: maybe, just maybe, there need to be some straight-forward way of selecting which apps to include in the build or even have "editions/presets"

This is something I have in mind : adding the possibility to select which apps will be installed at build time. This is something that is definitely possible to do. However, I think it'll be very difficult to maintain and test : we'll have to maintain and test multiple 'editions', and/or ensure that every combination of apps will work well, ensure that there's enough RAM memory available for all the apps,...

medeyko:
Sorry for asking again, but could anyone please write what is the state of this app?
Is there anything left to do with it before inlcusion?
How much resources does it really need?
How much resources are relly left?
I guess this app is one of the smallest possible apps, as it doesn't use any unusual library functions, which may significantly increase code size. It is compact itself.
But it is still usable and also useful.

Ertain:
Just a little neat addition: I think it would be cool if, when the user has set up the "dice" (that is, the dice number and such), they can shake the PineTime, and it "rolls" the dice.

stoehraj:
@Ertain hey, thanks for the comment -- if you look at the original description and some of the earlier comments, that actually is a feature I had initially, but we ended up pulling it out when there was some investigation on standardizing gestures a bit more across apps.  I think it's a really cool idea, though!  Makes the rolling a bit more tangible, especially combined with vibration.

Either way, looks like there's a PR for a different dice app that seems more likely to get merged in than this one (https://github.com/InfiniTimeOrg/InfiniTime/pull/1326) so may be worth it to request that feature over there as well.

NeroBurner:
closing in favor of merged https://github.com/InfiniTimeOrg/InfiniTime/pull/1326

PR #568, opened 2021-08-10
TwiMaster rework

Riksu9000:
Made using TwiMaster simpler. The bus is simply enabled on each read/write and disabled when done. There's no need to think about or control the state outside of TwiMaster. I don't know if this is more expensive, but it seems to work just fine.

Internal pullups were enabled. They're unnecessary because external pullups are used.

~~This will fix the freezing issue in #492 so I can finally move forwards with it.~~

Sleep() and Wakeup() can almost be made private, but there's still this piece of code in SystemTask I'm not brave enough to touch.

```
  // Reset the TWI device because the motion sensor chip most probably crashed it...
  twiMaster.Sleep();
  twiMaster.Init();
```

### DIFF
--- a/src/drivers/TwiMaster.cpp
+++ b/src/drivers/TwiMaster.cpp
@@ -72,8 +72,10 @@ void TwiMaster::Init() {
 
 TwiMaster::ErrorCodes TwiMaster::Read(uint8_t deviceAddress, uint8_t registerAddress, uint8_t* data, size_t size) {
   xSemaphoreTake(mutex, portMAX_DELAY);
+  Wakeup();
   auto ret = Write(deviceAddress, &registerAddress, 1, false);
   ret = Read(deviceAddress, data, size, true);
+  Sleep();
   xSemaphoreGive(mutex);
   return ret;
 }
@@ -81,9 +83,11 @@ TwiMaster::ErrorCodes TwiMaster::Read(uint8_t deviceAddress, uint8_t registerAdd
 TwiMaster::ErrorCodes TwiMaster::Write(uint8_t deviceAddress, uint8_t registerAddress, const uint8_t* data, size_t size) {
   ASSERT(size <= maxDataSize);
   xSemaphoreTake(mutex, portMAX_DELAY);
+  Wakeup();
   internalBuffer[0] = registerAddress;
   std::memcpy(internalBuffer + 1, data, size);
   auto ret = Write(deviceAddress, internalBuffer, size + 1, true);
+  Sleep();
   xSemaphoreGive(mutex);
   return ret;
 }
@@ -179,13 +183,11 @@ void TwiMaster::Sleep() {
   }
   nrf_gpio_cfg_default(6);
   nrf_gpio_cfg_default(7);
-  NRF_LOG_INFO("[TWIMASTER] Sleep");
 }
 
 void TwiMaster::Wakeup() {
   ConfigurePins();
   twiBaseAddress->ENABLE = (TWIM_ENABLE_ENABLE_Enabled << TWIM_ENABLE_ENABLE_Pos);
-  NRF_LOG_INFO("[TWIMASTER] Wakeup");
 }
 
 /* Sometimes, the TWIM device just freeze and never set the event EVENTS_LASTTX.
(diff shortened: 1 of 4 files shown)


Riksu9000:
Does disabling the bus or changing pin configuration actually save power?

JF002:
> Does disabling the bus or changing pin configuration actually save power?

I think I found some info in datasheet/nrf devzone saying that the peripherals (spi, twi,...) would consume a bit of current when enabled (even with no communication happening). They also keep the HFClock enabled. That's why I tried to disabled them as soon as the watch goes to sleep mode.
I'll try to find these info again ;)

src/drivers/TwiMaster.cpp
```diff
@@ -9,21 +9,27 @@ using namespace Pinetime::Drivers;
 // TODO use DMA/IRQ
 
 TwiMaster::TwiMaster(const Modules module, const Parameters& params) : module {module}, params {params} {
+  mutex = xSemaphoreCreateBinary();
```
JF002, line 12: The initialization of the mutex was previously done in the ctor(), but I moved it to Init() in 6d524ebea2c97e309633d5e01c3a1e37c182f27d. The goal was to avoid [Static Initialization Order Fiasco](https://en.cppreference.com/w/cpp/language/siof) (see [this](https://github.com/JF002/InfiniTime/pull/415#issuecomment-859004238). 
I also wanted to ensure FreeRTOS was correctly initialized before instantiating mutexes and other objects. 

Can we ensure this won't cause any issue?

Riksu9000: There is a separate function called `xSemaphoreCreateBinaryStatic()`, so I'd assume that one isn't static and works correctly.

Riksu9000: I can't say I completely understand the issue, but this doesn't access global state so it doesn't apply right? I dont think we shoud be afraid of using constructors either because they make the classes safe to use right away, and I would even like to move more stuff to the constructors to reduce the Init functions in code that can be avoided.

Riksu9000:
I found something weird.

`static constexpr uint32_t MaxTwiFrequencyWithoutHardwareBug {0x06200000};`
This is set as the frequency parameter

But then it is cast to Frequencies. What will happen here?
```
enum class Frequencies { Khz100, Khz250, Khz400 };

  switch (static_cast<Frequencies>(params.frequency)) {
    case Frequencies::Khz100:
      twiBaseAddress->FREQUENCY = TWIM_FREQUENCY_FREQUENCY_K100;
      break;
    case Frequencies::Khz250:
      twiBaseAddress->FREQUENCY = TWIM_FREQUENCY_FREQUENCY_K250;
      break;
    case Frequencies::Khz400:
      twiBaseAddress->FREQUENCY = TWIM_FREQUENCY_FREQUENCY_K400;
      break;
  }
```

The attempted frequency parameters value is slightly smaller than the K400 value, but can we be sure its safe because its not documented?
```
#define TWIM_FREQUENCY_FREQUENCY_K100 (0x01980000UL) /*!< 100 kbps */
#define TWIM_FREQUENCY_FREQUENCY_K250 (0x04000000UL) /*!< 250 kbps */
#define TWIM_FREQUENCY_FREQUENCY_K400 (0x06400000UL) /*!< 400 kbps */
```

src/drivers/TwiMaster.cpp
JF002: Yes, of course, most of the time, we shouldn't need an Init() function and use the ctor() to initialise the whole object. However, in this case, the ctor is called by the Startup code, before main() is called.
When you look at the code of `xSemaphoreCreateBinary()` you see it access to variables from FreeRTOS and even calls `pvPortMalloc()`.

I'm not sure completely sure that it's ok that call freertos function before the scheduler is init'd and running.
It might work, as previously, all mutexes were initialized in the ctro(), but I moved them when I was chasing memory corruption issue (which, in the end, were not caused by those initializations in the ctor() so...).

JF002:
Indeed, this casting for the frequency does not look right.
Iirc, I had to use that `MaxTwiFrequencyWithoutHardwareBug` constant to fix an issue reference in the [errata sheet of the mcu](https://infocenter.nordicsemi.com/index.jsp?topic=%2Fstruct_nrf52%2Fstruct%2Fnrf52832_errata.html).

src/drivers/TwiMaster.cpp
Riksu9000: I moved it back so we don't need to worry about it now.

Riksu9000:
>Iirc, I had to use that MaxTwiFrequencyWithoutHardwareBug constant to fix an issue reference in the errata sheet of the mcu.

Okay I found the same issue mentioned there and the same workaround so it seems fine to use.

JF002:
But I think this cast and `switch/case` should be fixed, because, in the end, the workaround value is not written to FREQUENCY.

Riksu9000:
> But I think this cast and `switch/case` should be fixed, because, in the end, the workaround value is not written to FREQUENCY.

I already fixed it :)

JF002:
> I already fixed it :)

Awesome, you're too fast for me ! :-D

PR #574, opened 2021-08-12
Add post_build.sh to .gitignore

MysteriousLog6:
Add post_build.sh to .gitignore


JF002:
I might be wrong, but I don't think this is needed:  `post_build.sh` is generated by cmake in the build folder, which should not be tracked by git.
In which situation did you notice that git would add this file?

Avamander:
> In which situation did you notice that git would add this file?

When you don't follow best-practice and run `cmake` in the project's root folder (instead of `build`). It doesn't get instantly added, but people have accidentally added it to git themselves.

src/main.cpp
```diff
@@ -125,7 +125,6 @@ Pinetime::Controllers::FS fs {spiNorFlash};
 Pinetime::Controllers::Settings settingsController {fs};
 Pinetime::Controllers::MotorController motorController {settingsController};
 
-
```
Avamander, line 128: This change is not needed

Avamander (changes requested):
This doesn't hurt and can avoid someone accidentally adding these files, even if they aren't running `cmake` in the right folder. If that one line is reverted it's fine to merge.

Avamander:
I'll close this as it's a counter to cmake misuse, which might have future unintended consequences we don't want to deal with.

Issue #571, opened 2021-08-11
Metronome not keeping up speed


arjanvlek:
Metronome is not accurate compared to a decent metronome app (Tempo). 

See the videos. my PineTime Watch is on InfiniTime 1.3.0. 

https://imgur.com/6id3yX2

https://imgur.com/eYVFjS2

Riksu9000:
A change was merged that improves Metronome accuracy. This will be in the next release, or you could already test it by building the firmware to see if it is good enough now.

The high bpm count issue is known and it has been discussed in #491, #492 and #541

PR #577, opened 2021-08-13
Improve metronome

Riksu9000:
Clean up code.
Fix warnings.
Improve accuracy. It can still drift but much less than before. #571
Larger buttons.

metronome


JF002:
Thanks ! :+1:

PR #602, opened 2021-08-19
Remove unused variables

Riksu9000:
Remove unused variables in a couple of files.


evergreen22 (approved):
I couldn't figure out why these existed. I have a WIP I'll PR that sets these callbacks.

void ble_manager_set_ble_connection_callback(void (*connection)());
void ble_manager_set_ble_disconnection_callback(void (*disconnection)());

PR #605, opened 2021-08-20
add different virbrations tunes/patterns support as bzzzt for 50 or 35ms is boring

akito45:
Things attempted

- refactor, so that the different vibration types can be declared in one place
- support different tunes/patterns,  not only durations
- try to conserve memory
- preserve motorController.RunForDuration(duration) functionality for manual access if needed

### DIFF
--- a/src/components/motor/MotorController.cpp
+++ b/src/components/motor/MotorController.cpp
@@ -8,15 +8,25 @@ APP_TIMER_DEF(longVibTimer);
 
 using namespace Pinetime::Controllers;
 
+constexpr MotorController::Tune MotorController::tunes[];
+
+
 MotorController::MotorController(Controllers::Settings& settingsController) : settingsController {settingsController} {
 }
+
+
+
+uint8_t MotorController::step = 0;
+MotorController::TuneType MotorController::runningTune = MotorController::TuneType::STOP; 
+
+
 
 void MotorController::Init() {
   nrf_gpio_cfg_output(pinMotor);
   nrf_gpio_pin_set(pinMotor);
   app_timer_init();
 
-  app_timer_create(&shortVibTimer, APP_TIMER_MODE_SINGLE_SHOT, StopMotor);
+  app_timer_create(&shortVibTimer, APP_TIMER_MODE_SINGLE_SHOT, Vibrate);
   app_timer_create(&longVibTimer, APP_TIMER_MODE_REPEATED, Ring);
 }
 
@@ -26,13 +36,26 @@ void MotorController::Ring(void* p_context) {
 }
 
 void MotorController::RunForDuration(uint8_t motorDuration) {
+  nrf_gpio_pin_set(pinMotor);
   if (settingsController.GetVibrationStatus() == Controllers::Settings::Vibration::OFF) {
     return;
   }
-
+  step = 0;
+  runningTune = TuneType::STOP;
   nrf_gpio_pin_clear(pinMotor);
   app_timer_start(shortVibTimer, APP_TIMER_TICKS(motorDuration), nullptr);
+
 }
+
+void MotorController::VibrateTune(TuneType tune) {
+  nrf_gpio_pin_set(pinMotor);
+  if (settingsController.GetVibrationStatus() == Controllers::Settings::Vibration::OFF)
+    return;
+  step = 0;
+  runningTune = tune;
+  Vibrate(nullptr);
+}
+
 
 void MotorController::StartRinging() {
   if (settingsController.GetVibrationStatus() == Controllers::Settings::Vibration::OFF) {
@@ -47,6 +70,22 @@ void MotorController::StopRinging() {
   nrf_gpio_pin_set(pinMotor);
 }
 
-void MotorController::StopMotor(void* p_context) {
-  nrf_gpio_pin_set(pinMotor);
+
+void MotorController::Vibrate(void* p_context) {
+  
+  if (step >= tunes[runningTune].length || step >= 8) { //end of tune turn off vibration
+    nrf_gpio_pin_set(pinMotor);
+    return; 
+  }  
+ 
+  if (((1 << step) & tunes[runningTune].tune) > 0) {
+    nrf_gpio_pin_clear(pinMotor);
+  } else {
+    nrf_gpio_pin_set(pinMotor);
+  }
+
+  ++step;     
+  /* Start timer for the next cycle */
+  app_timer_start(shortVibTimer, APP_TIMER_TICKS(tunes[runningTune].tempo), NULL);
 }
+
--- a/src/components/motor/MotorController.h
+++ b/src/components/motor/MotorController.h
@@ -7,19 +7,48 @@
 namespace Pinetime {
   namespace Controllers {
     static constexpr uint8_t pinMotor = 16;
-
+    
+    
     class MotorController {
+    
+
     public:
+      enum TuneType : uint8_t {
+        NOTIFICATION,
+        SHORT,
+        RING,
+        STOP
+      };
+
       MotorController(Controllers::Settings& settingsController);
       void Init();
       void RunForDuration(uint8_t motorDuration);
       void StartRinging();
       static void StopRinging();
+      void VibrateTune(TuneType tune);
 
     private:
-      static void Ring(void* p_context);
+
+    private:
+      struct Tune { 
+        uint8_t tune;
+        uint8_t length;
+        uint8_t tempo;
+      };
       Controllers::Settings& settingsController;
-      static void StopMotor(void* p_context);
+      static TuneType runningTune;
+      static uint8_t step;
+
+      static constexpr Tune tunes[] =  {
+        [TuneType::NOTIFICATION] = {.tune = 0x29, .length = 6, .tempo = 50},
+        [TuneType::SHORT] = {.tune = 0x01, .length = 2, .tempo = 35},
+        [TuneType::RING] = {.tune = 0x0f, .length = 8, .tempo = 50},
+        [TuneType::STOP] = {.tune = 0x00, .length = 0, .tempo = 0},
+      };
+
+      static void Vibrate(void* p_context);
+      static void Ring(void* p_context);
+    
     };
   }
 }
--- a/src/displayapp/screens/Notifications.cpp
+++ b/src/displayapp/screens/Notifications.cpp
@@ -40,7 +40,7 @@ Notifications::Notifications(DisplayApp* app,
     if (notification.category == Controllers::NotificationManager::Categories::IncomingCall) {
       motorController.StartRinging();
     } else {
-      motorController.RunForDuration(35);
+      motorController.VibrateTune(Controllers::MotorController::TuneType::NOTIFICATION);
       timeoutLine = lv_line_create(lv_scr_act(), nullptr);
 
       lv_obj_set_style_local_line_width(timeoutLine, LV_LINE_PART_MAIN, LV_STATE_DEFAULT, 3);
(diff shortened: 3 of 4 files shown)


src/components/motor/MotorController.cpp
```diff
@@ -47,6 +70,22 @@ void MotorController::StopRinging() {
...
+ 
+  if (((1 << step) & tunes[runningTune].tune) > 0) {
+    nrf_gpio_pin_clear(pinMotor);
+  } else {
+    nrf_gpio_pin_set(pinMotor);
+  }
+
+  ++step;     
+  /* Start timer for the next cycle */
+  app_timer_start(shortVibTimer, APP_TIMER_TICKS(tunes[runningTune].tempo), NULL);
```
Riksu9000, line 89: This makes this function loop forever even when there is no vibration.

src/components/motor/MotorController.h
```diff
@@ -7,19 +7,48 @@
...
       Controllers::Settings& settingsController;
-      static void StopMotor(void* p_context);
+      static TuneType runningTune;
+      static uint8_t step;
+
+      static constexpr Tune tunes[] =  {
+        [TuneType::NOTIFICATION] = {.tune = 0x29, .length = 6, .tempo = 50},
+        [TuneType::SHORT] = {.tune = 0x01, .length = 2, .tempo = 35},
+        [TuneType::RING] = {.tune = 0x0f, .length = 8, .tempo = 50},
+        [TuneType::STOP] = {.tune = 0x00, .length = 0, .tempo = 0},
```
Riksu9000, line 46: These should use binary notation instead of hexadecimal because you're reading the bits in Vibrate().

src/components/motor/MotorController.h
```diff
@@ -7,19 +7,48 @@
...
+    
     class MotorController {
+    
+
     public:
+      enum TuneType : uint8_t {
+        NOTIFICATION,
+        SHORT,
+        RING,
+        STOP
```
Riksu9000, line 20: STOP should probably be first if it is needed at all.

src/components/motor/MotorController.cpp
```diff
@@ -8,32 +8,55 @@ APP_TIMER_DEF(longVibTimer);
...
 MotorController::MotorController(Controllers::Settings& settingsController) : settingsController {settingsController} {
 }
 
+
+
+uint8_t MotorController::step = 0;
+MotorController::TuneType MotorController::runningTune = MotorController::TuneType::STOP; 
+
+
+
```
Riksu9000, line 23: These variables only need to be declared in the header file.

src/components/motor/MotorController.cpp
```diff
@@ -47,6 +70,22 @@ void MotorController::StopRinging() {
...
+  if (step >= tunes[runningTune].length || step >= 8) { //end of tune turn off vibration
+    nrf_gpio_pin_set(pinMotor);
+    return; 
+  }  
+ 
+  if (((1 << step) & tunes[runningTune].tune) > 0) {
+    nrf_gpio_pin_clear(pinMotor);
+  } else {
+    nrf_gpio_pin_set(pinMotor);
+  }
```
Riksu9000, line 85: Any reason not to use RunForDuration() for the patterns as well? It would provide better abstraction.

Riksu9000 (changes requested):

src/components/motor/MotorController.h
```diff
@@ -7,19 +7,48 @@
...
       void Init();
       void RunForDuration(uint8_t motorDuration);
       void StartRinging();
       static void StopRinging();
+      void VibrateTune(TuneType tune);
 
     private:
-      static void Ring(void* p_context);
+
+    private:
```
Riksu9000, line 32: There are now two private sections.

src/components/motor/MotorController.h
akito45: STOP is needed for the RunForDuration functionality. RunForDuration turns on the motor and then schedules a STOP event at the appropriate time.

Riksu9000: I know but with some rework it might not be necessary. Or maybe it is. Just saying.

src/components/motor/MotorController.cpp
akito45: RunForDuration() is public facing API call, that is currently implemented as "just another tune to play". Would like to not add any more logic to the public facing API. Adding enableMotor() and disableMotor() calls would maybe make it more readable.

Riksu9000: >RunForDuration() is public facing API call, that is currently implemented as "just another tune to play".

That's because you changed it that way. That function didn't necessarily have to be changed at all.

src/components/motor/MotorController.cpp
akito45: I get undefined reference to Pinetime::Controllers::MotorController::step, etc If I do not initialize it.

src/components/motor/MotorController.h
akito45: Fixed. Added some manual formatting to make it more readable.

src/components/motor/MotorController.h
akito45: fixed

src/components/motor/MotorController.cpp
Riksu9000: Ah that's because these functions and variables are static. Usually in this project something like this is used to get around cases where a callback requires a static function.
```
void MotorController::Ring(void* p_context) {
  auto* motorController = static_cast<MotorController*>(p_context);
  motorController->RunForDuration(50);
}
```

src/components/motor/MotorController.h
akito45: removed STOP state

src/components/motor/MotorController.cpp
akito45: moved step and runningingTune from static scope. 
Would like to leave tunes[] as static. The probability is miniscule, that motorController will have multiple instances, but still.

src/components/motor/MotorController.cpp
akito45: Refactored the approach to use a unified ScheduleVibrationTimer call.

Riksu9000:
Would it be better to encode patterns as a simple list of delays between motor toggles?

A pattern could be `pattern[] = {50, 30, 50, 0}`. It would first enable the motor. After 50ms it would toggle the pin and wait for 30ms, effectively adding a silent period. Then after the 30ms the pin would again be toggled, this time enabling the motor, and the next delay would be set to 50ms.

A zero could be used to detect the end of the pattern and stop the motor.

I think doing it this way could be more flexible and simple.

hubmartin:
I find original pattern format (bit pattern, length & tempo) flexible and memory efficient enough. I can imagine that composing will be easier to change bit patterns than copypaste different numbers in an array. I would just increase the `uint8_t tune;` to `uint32_t` to have more bits for longer variantions if someone will use 16 or 32 bit pattern instead of just 8.

In case of longer tune the array representation will become worse to read I guess. Also it would need to have every "note" on a different line to see what is length and what is a pause.

```
pattern[] = {
50, 50, 
150, 50,
50, 50,
150, 50,
50, 50,
150, 50,
}
```

Now changing the speed will need to change all the values, or at lease 6 of them.

The original approach might look like this and feels cleaner to me

```
[TuneType::NOTIFICATION] = {.tune = 0b11101011101011101, .length = 17, .tempo = 50},
```

@akito45 Maybe I would suggest to read the bit pattern from the left (MSB) first. We read text from the left, cursor moves to the right when I press 0 or 1, notes are read from the left to the right, so for composing it will be much better.

akito45:
The bit pattern is less wasteful on memory. If made MSB first, the .length could be eliminated, as the length can be calculated on runtime, by just ignoring the preceding 0's. 

I would side with the VIBRATE(X, Y) format, if we would have access to some kind of frequency component. The format would make sense for VIBRATE(freq, duration), but not so much for on/off. 

Not that familiar with the NRF platform, but can that pin output PWM. In that case VIBRATE(intensity, duration) would make some sense. Pauses could be encoded as VIBRATE(0, duration). 

If the majority decides to go with the VIBRATE(X, Y) format, I would advocate for VIBRATE(intencity, duration) for future proofing, not for VIBRATE (on_duration, off_duration) as it seems needlessly custom, for no seeable benefit.

Riksu9000:
I don't think PWM is necessary because the length of the pulse is practically the same as intensity, and more often than not we would like the vibration to be more noticable than less.

I'm also not sure if we should use a macro to combine the on and off period. We could just decide that one unit is 10ms for example and 10 is easy to work with.

hubmartin:
Right, PWM is possible on any gpio pin but it really does not make any effect on the motor that you would fell. Specially in these short vibrating pulses.

10ms resolution and no macro coulde be fine. However for code readability the macro would be better and self explanatory (with division by 10ms), rather than array of deci-millisecond values.

akito45:
Either 
uint32_t MSB first binary format
`[TuneType::NOTIFICATION] = {.tune = 0b11101011101011101, .tempo = 50},`
or alternating on / off timing pattern of 10ms 
`uint8_t pattern[] = {5, 5, 15, 5, 5, 5, 15, 5, 5, 5, 15, 5}`

I vote for the first, for memory conservation. It also is more intuitively readable for me. Altho I have 0 musical talent so it is not saying much. Maybe the stakeholders could cast their votes and I could implement the more popular version.

Riksu9000:
Benefits of using an array of delays are that the code that plays the pattern would be simpler, more efficient and more flexible.

Having a fixed tempo is more limiting, because each vibration has to land on a beat. This could be compensated by increasing the speed, but then the 32 bit limit forces the pattern to be short. Maybe it could be increased to 64 bits, but now it's no longer very memory efficient. The array of delays doesn't have this issue. Patterns can be any length and any precision, making it much more flexible.

The array of delays is harder to read for sure, but splitting the pattern on multiple lines makes it much more readable. Now I can easily tell what this pattern is going to be like.

```
uint8_t pattern[] = {
	5, 5,
	15, 5,
	5, 5,
	15, 5,
	5, 5,
	15, 0
}
```

hubmartin:
I'm with riksu with 10ms granuality, so 5 = 50 ms if I got it right.
Every single vibration on a separate line.
```
uint8_t pattern[] = {
	5, 5,
	15, 5,
	5, 5,
	15, 5,
	5, 5,
	15, 0
}
```

kieranc:
Any chance of bringing this PR up to date and implementing the requested change to the pattern format @akito45 ?

CCranney:
I'd also be interested in this PR being updated - I've subscribed to changes, but thought I'd throw out that there's another interested party

agittins:
Two thoughts...

Another option for specifying the pattern could be to approach it like a step-sequencer (old school drum machines etc), and have an int to define step-interval (tempo, if you like) and then 16 or 32bits to define on or off for that step. Some experimentation would be required to see if that many bits is sufficient to give enough pattern length for a given number of intensity durations.

Has anyone *actually* experimented with pwm on the pinetime motor pin? Varying the pwm frequency (esp at lower freqs) might give usable options. Also, I haven't done the math on it but the pwm controller onboard looks like it might do all the pattern stuff we might want, as it can read step patterns via DMA without loading the cpu - https://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.nrf52832.ps.v1.1%2Fpwm.html
To be honest I haven't grokked how it all works so I don't know if it can run patterns slowly enough for this purpose, but it does mention LED display patterns in a special mode - the description of which I could not grok - is it perhaps further sub-dividing the clock using the other three patterns?

PR #606, opened 2021-08-21
docs: Fix a few typos

timgates42:
There are small typos in:
- doc/versioning.md
- src/components/ble/NimbleController.cpp
- src/systemtask/SystemTask.cpp

Fixes:
- Should read `unnecessary` rather than `uncesseray`.
- Should read `target` rather than `tharget`.
- Should read `functioning` rather than `functionning`.
- Should read `recommended` rather than `recomended`.

Semi-automated pull request generated by
https://github.com/timgates42/meticulous/blob/master/docs/NOTE.md


geekbozu:
Nothing wrong fixing typos all are in comments. So +1

evergreen22 (approved):
Thanks

JF002:
Thanks, that's a nice automation!

I noticed that it detected typos in  `src/libs/mynewt-nimble`. As this is an external library, can we avoid modifying them? Those changes will be lost next time we update the library, and also when we'll use submodule for these libs.

evergreen22:
Good catch! I didn't see the nimble ones.

Agreed we should *not* modify upstream libraries in our repo.

I suggest the PR author submit a PR to nimble for the typos in their 
codebase.

timgates42:
Thanks @evergreen22 have removed the nimble changes.

PR #611, opened 2021-08-25
Ignore vibration status in metronome

ColinKinloch:
I feel the metronome should either ignore the motor status or notify the user if the motor is disabled.

This MR adds a force option to `RunForDuration` which causes it to ignore the vibration status, and forces vibration for the metronome.


Riksu9000:
I'm not sure if a force option is the correct solution. I think the button should just disable notifications (silent mode), but it has been implemented such that it disables all vibrations. Is there a case where someone would actually want to disable all vibrations, considering the PineTime has no other way to notify the user of anything?

PR #612, opened 2021-08-25
Improved StopWatch laps

### LINKED ISSUE #392: Allow scrolling through lap times in the stopwatch
I would like to be able to scroll the lap times to look at them. I also wish more were shown at once (three instead of two previous laps at the least), similar to the Pebble stopwatch app.
This might interfere with swiping to change apps, which I only just realized is a thing, and I don't think is necessary since you can back out to the main screen. Being able to scroll in every app to add more functionality seems much more valuable.

Riksu9000:
Now StopWatch can store up to 20 laps and they can be viewed while paused.

While showing laps, the centisecond counter is hidden. If only four laps were on screen at once, it could fit. Which would be better? EDIT: The time label could separately be made to automatically adapt to size requirements, which would make showing hours possible as well.

Fixes #392

https://user-images.githubusercontent.com/37774658/130828623-3f1d189e-e7b3-4e0a-9f53-a40ec681290d.mp4

### DIFF
--- a/src/displayapp/screens/StopWatch.cpp
+++ b/src/displayapp/screens/StopWatch.cpp
@@ -5,11 +5,11 @@ using namespace Pinetime::Applications::Screens;
 
 namespace {
   TimeSeparated_t convertTicksToTimeSegments(const TickType_t timeElapsed) {
-    const int timeElapsedMillis = (static_cast<float>(timeElapsed) / static_cast<float>(configTICK_RATE_HZ)) * 1000;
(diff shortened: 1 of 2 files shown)


trman:
@Riksu9000  it's mare that 6 month already !
maybe it' s time to close this pull , like the other you have closed that have not updated...
so people could improve it , since you since don't have the time anymore for this

Riksu9000:
@trman This feature has been complete, so there's been no need to update it. I can fix the conflicts soon.

Riksu9000:
I'll be rewriting this feature beginning with #1244.

PR #615, opened 2021-08-26
Infini-iOS

Utopiah:
added as experimental companion app


ajack2001my:
Would it be a good idea to port Infini-iOS to Android? I'm not sure about the relationship to Gadgetbridge, but having an independent and not bound by any restrictions (I know the Pebble support was not as good as Pebble's own app). Just asking...

JF002:
@ajack2001my That's probably a question to ask on Infini-iOS repo.

Avamander:
@ajack2001my Generally there are no restrictions, there's a different path. The proposed port or an alternative is much more time-consuming and difficult than writing a few ContentProviders.

PR #617, opened 2021-08-26
Correct change test for wake up mode setting

ColinKinloch:
This seems like it should have caused a bug, but I haven't noticed any.


Riksu9000 (approved):

JF002:
Thanks! :+1:

PR #619, opened 2021-08-27
Silent mode enabled rather than vibration enabled

ColinKinloch:
This might just be my personal preference but I think this is more intuitive as a silent mode toggle, rather than a vibration toggle.

Silence disabled / Vibration enabled Silence enabled / Vibration disabled

left: silent mode is off (watch will vibrate), right: silent mode is on (watch will not vibrate)

### DIFF
--- a/src/displayapp/screens/settings/QuickSettings.cpp
+++ b/src/displayapp/screens/settings/QuickSettings.cpp
@@ -81,18 +81,18 @@ QuickSettings::QuickSettings(Pinetime::Applications::DisplayApp* app,
   lv_obj_set_event_cb(btn3, ButtonEventHandler);
   lv_btn_set_checkable(btn3, true);
   lv_obj_add_style(btn3, LV_BTN_PART_MAIN, &btn_style);
-  lv_obj_set_style_local_bg_color(btn3, LV_BTN_PART_MAIN, LV_STATE_CHECKED, LV_COLOR_GREEN);
+  lv_obj_set_style_local_bg_color(btn3, LV_BTN_PART_MAIN, LV_STATE_CHECKED, LV_COLOR_RED);
   lv_obj_set_size(btn3, buttonWidth, buttonHeight);
   lv_obj_align(btn3, nullptr, LV_ALIGN_IN_BOTTOM_LEFT, buttonXOffset, 0);
 
   btn3_lvl = lv_label_create(btn3, nullptr);
   lv_obj_set_style_local_text_font(btn3_lvl, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, &lv_font_sys_48);
 
-  if (settingsController.GetVibrationStatus() == Controllers::Settings::Vibration::ON) {
+  if (settingsController.GetVibrationStatus() == Controllers::Settings::Vibration::OFF) {
     lv_obj_add_state(btn3, LV_STATE_CHECKED);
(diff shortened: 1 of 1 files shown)


lman0:
> I don't think we should change this. It will throw everyone off at first and I think this is less intuitive. Mainly because it's matching disabled button color with enabled icon and enabled button color with disabled icon.
> 
> If we look at just the left photo, are notifications enabled because of the icon, or are they disabled because of they gray background?
> 
> When the background is green AND the icon is enabled, it's obvious that notifications are enabled, and when the background is gray AND the icon is disabled, it's obvious that notifications are disabled.

@Riksu9000  , i'm again what you said , since the red is known for most that it mean stop (unlike gray that is neutral, and the same color of the enabled icon for screen light which is disturbing)

As user it thinks that the red color for the disabled notification is a good things since it show easily that it's disabled 
and the enabled state can be either like other color on setting (aka gray)  for coherence or be like before green but for activated.

but since the  default state will be enabled then the gray is more understandable and readable for an user like.

by the way ,  sensibility are  personal ,  but once theme are created for the infinitme screens maybe you will be be able to have what you want

Riksu9000:
>unlike the original gray colour for the disabled notification state, that was the same colour of the enabled icon for screen light which is disturbing. (doesn't know if notification is disabled with color alone since otherwise the screen light level , would have be to be green as well)

The backlight isn't toggleable, so in that case gray means it's just a button. It's not perfect, but I don't think this PR helps with this issue anyway.

>but since the default state will be enabled then the gray is more understandable and readable for an user like.

A new user could easily think that the default state is disabled because of the gray color.

>but red and green colour as commonly used for disabled and enabled state
where as gray not much as disabled

Yes, green and red are used for enabled and disabled, but when a toggling element is either colored or gray, gray is always the disabled state. Usually enabled state is the accent color, which can be anything, even red.

If this needs to be changed, then I think it should be green when enabled and red when disabled, but what would we do if we wanted to add user selectable accent color? Having all other UI elements purple, but that one button be green or red wouldn't fit very well.

ColinKinloch:
I was thinking of it more along the lines of mute or "do not disturb"

Riksu9000:
>if one user want everything in purple then whatever colour was the setting , it would be purple , so it's not a problem.

What if the accent color was green? Then notification disabled would be green and that would definitely be strange.

>I was thinking of it more along the lines of mute or "do not disturb", as opposed to the default state.

The issue is that the icon implies that it controls ringing, but the color shows the state of mute, not ringing.

lman0:
>What if the accent color was green? Then notification disabled would be green and that would definitely be strange.

like i said before : its would be possible to change the individual icons colours when themes/colours accents is implemented
and then the green color then could be changed by whatever the user have selected as choice of color

by the way there people that are blind to green  so it's not like green color is good for everthing....

ColinKinloch:
> The issue is that the icon implies that it controls ringing, but the color shows the state of mute, not ringing.

Great point, this switch doesn't disable notifications.
Should the icon be the [mute](https://fontawesome.com/v5.15/icons/volume-mute?style=solid) and [volume-up](https://fontawesome.com/v5.15/icons/volume-up?style=solid)? ([volume](https://fontawesome.com/v5.15/icons/volume?style=solid) appears to be "pro")

Riksu9000:
>Great point, this switch doesn't disable notifications.

That might also change soon with #664, which makes the switch actually toggle notifications.

ColinKinloch:
#664 seems like a good resolution.

PR #622, opened 2021-08-29
Simplify touch event handling

Riksu9000:
(In this PR) main.cpp only pushes an `OnTouchEvent` message to SystemTask, spending less time in ISR. The logic was moved to `OnTouchEvent`.
`TouchWakeUp` and `OnTouchEvent()` were combined to `OnTouchEvent` message.


hubmartin:
Tested, works fine. Thanks Riksu.

PR #623, opened 2021-08-29
Let LVGL control queueTimeout

Riksu9000:
`lv_task_handler()` returns the time in ticks to the next task it needs to run.

This can now be used now that DisplayApp isn't locked to a certain refresh period to refresh apps consistently. #497

Using this time as the queueTimeout is more accurate, as messages can still cause slight delays. This also makes it possible to have tasks that run even more often than the refresh period.

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -141,19 +141,15 @@ uint32_t count = 0;
 bool toggle = true;
 void DisplayApp::Refresh() {
   TickType_t queueTimeout;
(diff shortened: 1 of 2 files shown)


geekbozu (approved):
Much cleaner and uses the proper FreeRTOS workflows. 
Good job as always riksu9000!

hubmartin:
Tested for few minutes, tried all apps and no issues. Great work Riksu.

JF002:
LVGL integration becomes more and more clean and efficient, thanks, @Riksu9000 !

Issue #562, opened 2021-08-07
Feature request: mqtt and smart home integration


drzacek:
I know this might be an ambitious one, but this is a smart watch and I'd love to see some actual smart features beside notifications and step counting. 

So the way I see it, the sensor data like battery life, steps, hr etc, should be possible to export to mqtt broker, so it could be then used in systems like Home Assistant and similar. That would require fiddling with the companion app on the phone more than with the pinetime itself. 

But second part, a dedicated app/screen with maybe 2 buttons and 2 displayed values, pretty minimalistic. 
It should be configurable through companion app (widget type/mqtt topic), and should provide simple means to control favorite lamp at home, or for starting sprinklers when at work etc. 

Not sure how we stand with free memory for additional app, but I think it wouldn't be too complex and should be doable.

hubmartin:
Hi, thanks for bringing this up. This feature is what I wan't also to achieve with InfiniTime. I have dozens sensor/actuators in my home with MQTT, like doorbell, sensors, lights and garage control. However I also use Node-red on my server to interconnect different protocols, so I might be fine with any protocol control from my watch.

### Companion app
To achieve this also companion app has to support it. Monhts ago I compiled gadgetbridge and tried to create separate simle "hello world" android app to connect and send mqtt packet. I have no experience with Android programming (Gadgetbridge) so this was just my test. On the BLE level i guess there has to be some characteristic which will be neede to invent to do one of the protocols like HTTP, MQTT or others. I googled if there is some standard like TCP over BLE but with no luck, also it will be too much complicated to put HTTP/MQTT client stack on InfiniTime.
Now how thic BLE communication will work. Should companion app be set to some HTTP/MQTT broker? Or shout it behave just like a bridge and all the IP addresses and ports will be in the watch? It needs to be discussed what will be the best.
(then there is the "security", should bridge also act as a firewall? should Watch FW have access to the who HTTP internet? :) )

### InfiniTime firmware
I'm an embedded C programmer still struggling to understand C++ code in this project, but it is definitely easier for me than to learn Android development. So from my point of view I would like to put all the configuration and settings to the watches, and companion app will be just dump bridge with practically no UI.
With all the logic in InfiniTime, there should be some user configuration, because no one likes to recompile firmware just to change IP. Now we have filesystem LittleFS which can have configurations for example for MQTT message buttons and IP address.
[...]

PR #624, opened 2021-08-29
WIP: SmartControl, UI is semi-ready

drzacek:
This is a displayapp for the "Smart Home control", with simple UI - 3 displayed sensor values and 3 buttons. 
The widgets are with dummy values now, there is no bluetooth communication at all yet.


hubmartin:
Thanks, my plan is to add a BLE service and do a prototype Python BLE-MQTT gateway which could be later integrated to GadgetBridge by someone who knows how.

hubmartin:
This PR is based on ideas in #562

hubmartin:
Just an update on my side. I have concept of python MQTT client bridge between Watches and MQTT broker.
Right now the code is using my BLE NUS #560 service and in the end it will use antoher service (but it will be practically same like the current BLE NUS, since it will use text commands divided by some divide symbol like `;`)
https://github.com/hubmartin/InfiniTime-tools/blob/master/ble-mqtt.py

Riksu9000:
This PR is WIP, but hasn't been updated. Without the bluetooth part, this isn't very useful for us. Closing as stale.

PR #621, opened 2021-08-29
Change BLE appearance setting to "smart watch"

evergreen22:
Set the appearance value to 'smart watch' as documented in the bluetooth specification.

https://specificationrefs.bluetooth.com/assigned-values/Appearance%20Values.pdf

PR #625, opened 2021-08-29
Refactor ble advertising logic

evergreen22:
This is my WIP attempt to resolve the ble re-connect issue and get more reliable ble behavior. Comments, discussion, and suggestions are welcome. I'm testing on both my watch and dev-kit. The watch has been connected now for 3+ days. The dev-kit still fails periodically. I'm using both nRFConnect and GB for testing. I am (still) waiting on the breakout to connect my Segger JLink and gaze into the dumpster fire that is ble.

- Refactor ble advertising based on ble standards and conventions.
- Changes are based on the bleprph example code, bluetooth docs, and nimble docs.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -92,6 +92,9 @@ set(SDK_SOURCE_FILES
 set(TINYCRYPT_SRC
         libs/mynewt-nimble/ext/tinycrypt/src/aes_encrypt.c
         libs/mynewt-nimble/ext/tinycrypt/src/utils.c
+        libs/mynewt-nimble/ext/tinycrypt/src/cmac_mode.c
+        libs/mynewt-nimble/ext/tinycrypt/src/ecc.c
+        libs/mynewt-nimble/ext/tinycrypt/src/ecc_dh.c
         )
 
 set(NIMBLE_SRC
@@ -104,6 +107,10 @@ set(NIMBLE_SRC
         libs/mynewt-nimble/nimble/host/src/ble_l2cap.c
         libs/mynewt-nimble/nimble/host/src/ble_hs_mbuf.c
         libs/mynewt-nimble/nimble/host/src/ble_sm.c
+        libs/mynewt-nimble/nimble/host/src/ble_sm_cmd.c
+        libs/mynewt-nimble/nimble/host/src/ble_sm_lgcy.c
+        libs/mynewt-nimble/nimble/host/src/ble_sm_alg.c
+        libs/mynewt-nimble/nimble/host/src/ble_sm_sc.c
         libs/mynewt-nimble/nimble/host/src/ble_gap.c
         libs/mynewt-nimble/nimble/host/src/ble_gatts.c
         libs/mynewt-nimble/nimble/host/src/ble_gattc.c
@@ -127,10 +134,6 @@ set(NIMBLE_SRC
         libs/mynewt-nimble/nimble/host/src/ble_hs_atomic.c
         libs/mynewt-nimble/nimble/host/src/ble_hs_adv.c
         libs/mynewt-nimble/nimble/host/src/ble_hs_flow.c
-        libs/mynewt-nimble/nimble/host/src/ble_sm.c
-        libs/mynewt-nimble/nimble/host/src/ble_sm_cmd.c
-        libs/mynewt-nimble/nimble/host/src/ble_sm_lgcy.c
-        libs/mynewt-nimble/nimble/host/src/ble_sm_alg.c
         libs/mynewt-nimble/nimble/host/src/ble_hs_mqueue.c
         libs/mynewt-nimble/nimble/host/src/ble_hs_stop.c
         libs/mynewt-nimble/nimble/host/src/ble_hs_startup.c
--- a/src/components/ble/NimbleController.cpp
+++ b/src/components/ble/NimbleController.cpp
@@ -159,6 +159,9 @@ int NimbleController::OnGAPEvent(ble_gap_event* event) {
 
       if (event->connect.status != 0) {
         /* Connection failed; resume advertising. */
+        currentTimeClient.Reset();
+        alertNotificationClient.Reset();
+        connectionHandle = BLE_HS_CONN_HANDLE_NONE;
         bleController.Disconnect();
         StartAdvertising();
       } else {
--- a/src/components/ble/NimbleController.cpp
+++ b/src/components/ble/NimbleController.cpp
@@ -151,7 +151,7 @@ int NimbleController::OnGAPEvent(ble_gap_event* event) {
       StartAdvertising();
       break;
 
-    case BLE_GAP_EVENT_CONNECT: {
+    case BLE_GAP_EVENT_CONNECT:
       NRF_LOG_INFO("Advertising event : BLE_GAP_EVENT_CONNECT");
 
       /* A new connection was established or a connection attempt failed. */
@@ -170,7 +170,7 @@ int NimbleController::OnGAPEvent(ble_gap_event* event) {
         systemTask.PushMessage(Pinetime::System::Messages::BleConnected);
         // Service discovery is deferred via systemtask
       }
-    } break;
+      break;
 
     case BLE_GAP_EVENT_DISCONNECT:
       NRF_LOG_INFO("Advertising event : BLE_GAP_EVENT_DISCONNECT");
@@ -207,7 +207,8 @@ int NimbleController::OnGAPEvent(ble_gap_event* event) {
       break;
 
     case BLE_GAP_EVENT_MTU:
-      NRF_LOG_INFO("mtu update event; conn_handle=%d cid=%d mtu=%d\n", event->mtu.conn_handle, event->mtu.channel_id, event->mtu.value);
+      NRF_LOG_INFO("mtu update event; conn_handle=%d cid=%d mtu=%d\n",
+        event->mtu.conn_handle, event->mtu.channel_id, event->mtu.value);
       break;
 
     case BLE_GAP_EVENT_REPEAT_PAIRING: {
@@ -224,8 +225,7 @@ int NimbleController::OnGAPEvent(ble_gap_event* event) {
       /* Return BLE_GAP_REPEAT_PAIRING_RETRY to indicate that the host should
        * continue with the pairing operation.
        */
-    }
-      return BLE_GAP_REPEAT_PAIRING_RETRY;
+    } return BLE_GAP_REPEAT_PAIRING_RETRY;
 
     case BLE_GAP_EVENT_NOTIFY_RX: {
       /* Peer sent us a notification or indication. */
@@ -260,7 +260,7 @@ uint16_t NimbleController::connHandle() {
 }
 
 void NimbleController::NotifyBatteryLevel(uint8_t level) {
-  if(connectionHandle != BLE_HS_CONN_HANDLE_NONE) {
+  if (connectionHandle != BLE_HS_CONN_HANDLE_NONE) {
     batteryInformationService.NotifyBatteryLevel(connectionHandle, level);
   }
 }
--- a/src/components/ble/NimbleController.cpp
+++ b/src/components/ble/NimbleController.cpp
@@ -122,6 +122,15 @@ void NimbleController::StartAdvertising() {
 
   adv_params.conn_mode = BLE_GAP_CONN_MODE_UND;
   adv_params.disc_mode = BLE_GAP_DISC_MODE_GEN;
+  /* fast advertise for 30 sec */
+  if (fastAdvCount < 15) {
+    adv_params.itvl_min = 32;
+    adv_params.itvl_max = 47;
+    fastAdvCount++;
+  } else {
+    adv_params.itvl_min = 1636;
+    adv_params.itvl_max = 1651;
+  }
 
   fields.flags = BLE_HS_ADV_F_DISC_GEN | BLE_HS_ADV_F_BREDR_UNSUP;
   fields.uuids128 = &dfuServiceUuid;
@@ -139,7 +148,7 @@ void NimbleController::StartAdvertising() {
   rc = ble_gap_adv_rsp_set_fields(&rsp_fields);
   ASSERT(rc == 0);
 
-  rc = ble_gap_adv_start(addrType, NULL, 5000, &adv_params, GAPEventCallback, this);
+  rc = ble_gap_adv_start(addrType, NULL, 2000, &adv_params, GAPEventCallback, this);
   ASSERT(rc == 0);
 }
 
@@ -163,6 +172,7 @@ int NimbleController::OnGAPEvent(ble_gap_event* event) {
         alertNotificationClient.Reset();
         connectionHandle = BLE_HS_CONN_HANDLE_NONE;
         bleController.Disconnect();
+        fastAdvCount = 0;
         StartAdvertising();
       } else {
         connectionHandle = event->connect.conn_handle;
@@ -181,6 +191,7 @@ int NimbleController::OnGAPEvent(ble_gap_event* event) {
       alertNotificationClient.Reset();
       connectionHandle = BLE_HS_CONN_HANDLE_NONE;
       bleController.Disconnect();
+      fastAdvCount = 0;
       StartAdvertising();
       break;
 
--- a/src/components/ble/NimbleController.cpp
+++ b/src/components/ble/NimbleController.cpp
@@ -42,6 +42,19 @@ NimbleController::NimbleController(Pinetime::System::SystemTask& systemTask,
     serviceDiscovery({&currentTimeClient, &alertNotificationClient}) {
 }
 
+void nimble_on_reset(int reason) {
+  NRF_LOG_INFO("Resetting state; reason=%d\n", reason);
+}
+
+void nimble_on_sync(void) {
+    int rc;
+
+    rc = ble_hs_util_ensure_addr(0);
+    ASSERT(rc == 0);
+
+    nptr->StartAdvertising();
+}
+
 int GAPEventCallback(struct ble_gap_event* event, void* arg) {
   auto nimbleController = static_cast<NimbleController*>(arg);
   return nimbleController->OnGAPEvent(event);
@@ -50,6 +63,10 @@ int GAPEventCallback(struct ble_gap_event* event, void* arg) {
 void NimbleController::Init() {
   while (!ble_hs_synced()) {
   }
+
+  nptr = this;
+  ble_hs_cfg.reset_cb = nimble_on_reset;
+  ble_hs_cfg.sync_cb = nimble_on_sync;
 
   ble_svc_gap_init();
   ble_svc_gatt_init();
@@ -64,28 +81,31 @@ void NimbleController::Init() {
   batteryInformationService.Init();
   immediateAlertService.Init();
   heartRateService.Init();
-  int res;
-  res = ble_hs_util_ensure_addr(0);
-  ASSERT(res == 0);
-  res = ble_hs_id_infer_auto(0, &addrType);
-  ASSERT(res == 0);
-  res = ble_svc_gap_device_name_set(deviceName);
-  ASSERT(res == 0);
+
+  int rc;
+  rc = ble_hs_util_ensure_addr(0);
+  ASSERT(rc == 0);
+  rc = ble_hs_id_infer_auto(0, &addrType);
+  ASSERT(rc == 0);
+  rc = ble_svc_gap_device_name_set(deviceName);
+  ASSERT(rc == 0);
+  rc = ble_svc_gap_device_appearance_set(0xC2);
+  ASSERT(rc == 0);
   Pinetime::Controllers::Ble::BleAddress address;
-  res = ble_hs_id_copy_addr(addrType, address.data(), nullptr);
-  ASSERT(res == 0);
+  rc = ble_hs_id_copy_addr(addrType, address.data(), nullptr);
+  ASSERT(rc == 0);
   bleController.AddressType((addrType == 0) ? Ble::AddressTypes::Public : Ble::AddressTypes::Random);
   bleController.Address(std::move(address));
 
-  res = ble_gatts_start();
-  ASSERT(res == 0);
+  rc = ble_gatts_start();
+  ASSERT(rc == 0);
+
+  if (!ble_gap_adv_active() && !bleController.IsConnected())
+    StartAdvertising();
 }
 
 void NimbleController::StartAdvertising() {
-  if (bleController.IsConnected() || ble_gap_conn_active() || ble_gap_adv_active())
-    return;
-
-  ble_svc_gap_device_name_set(deviceName);
+  int rc;
 
   /* set adv parameters */
   struct ble_gap_adv_params adv_params;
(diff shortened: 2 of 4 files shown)


evergreen22:
I re-enabled the assert statements that have been commented out in the advertising function and I have not observed them fire.

My nRFConnect shows the watch advertising in 64ms intervals. I set the interval in the call to ble_gap_adv_start to 5000 based on my nRFConnect logs which showed 5000 as the agreed upon parameters. I;m not clear on why the agreed connection parameters are 5000 ms and the actual interval on the watch is 64 ms.

evergreen22:
I implemented bonding with a passkey in my local branch and tested it to see if that impacted the re-connection issue. The only change I observed was that everytime the watch restarted advertising (i.g. central out or range) I would have to delete the bond to successfully re-connect. I believe this can be fixed by persisting the keys on the watch.

geekbozu:
> 
> On Sun, 29 Aug 2021, Tim Keller wrote: Not sure if this is the PR to do it but after some time spent advertising at 64ms, the spec suggests dropping to a much lower advertisement interval. Maybe take a timestamp on the last disconnect and clear it on connect. Setting adv_params.itvl_min is the parameter to change that i believe
> I see where that is suggested in the docs. First, I'm trying to resolve the re-connect issue then I thought we could design an appropriate algorithm for backing off on the adv interval.

Yeah getting proper bonding working would be great! Feel free to nag me on the discord as well I'm almost always around :)

evergreen22:
On Sun, 29 Aug 2021, Tim Keller wrote:

> Yeah getting proper bonding working would be great! Feel free to nag me 
> on the discord as well I'm almost always around :)
>
>

I can force a bonded connection with or without a pass key on every 
connect message. But that is super annoying if you are not attached to 
your phone. If you get out of range (of the central) you have to go 
through the whole connect sequence again.

The good news if that with a secure connection my android will recognize 
pinetime as a trusted device. I loved that feature on my Pebble and 
_really_ want that working properly on pinetime.

Thanks for the feedback.

evergreen22:
After 4 days, 18 hours, and some minutes of continous bluetooth connectivity my watch and GB disconnected. I noticed the bluetooth icon was gone on the watch while looking at the time and pulled up GB to see "waiting for reconnect". I opened NRFConnect and the watch was not advertising. :-(

I'm confident the watch received a disconnect event from nimble (because the ble icon was off) and it should have been advertising (the code says so). This is very frustrating.

evergreen22:
Another unexplained disconnect today that could only be fixed by rebooting the watch.

I started this WIP to discover the root cause of the disconnects and now I do not believe this PR will resolve that issue. However, based on the specs, docs, and example nimble apps, I do feel that this is the right way to manage advertising.

Hopefully @JF002 will discover a useful clue with the logic probe and sniffer. I am in need of some inspiration ;-|

geekbozu:
I'm in agreement, 
Sad that its not solving the issue and that it lies elsewhere...however the changes are still very much appreciated and needed. Just means we have to keep looking!

Do you think its in the scope of this PR to limit the advertising speed after a few attempts? To "save" power when its just idling advertising.

evergreen22:
Great suggestion! I'll dive back into the standard to see what is 
prescribed (if anything) and add to this PR.

lman0:
hi @evergreen22 ,

here https://github.com/espruino/Espruino/tree/master/targetlibs/nrf5x_12/components/ble is the bangle.js  source code related to Bluetooth ,
It might help you since pinetime and banglejs smartwatch share nearly the same component .

it's based on  open source nordic source code in order to achieve proper ble stack and may help to understand what pinetime lack.

@JF002 since we use nordic , why we don't use their stack? since from i can see it's open source (otherwise esprino would not be able to modifiy it and share it) , and would resolve  on ble problem.
otherwise i think it should be possible to use the nnordic source code used/modified by esprino for nimble in pinetime since it's not forbidden by the license.

best regards

evergreen22:
The "linear decrease..." commit [0275f86](https://github.com/JF002/InfiniTime/pull/625/commits/0275f868a3e4cff77eef32d131ff0a5b534feb58) slows down the rate of advertisements to conserve battery when no connection is available. See the commit message for details on how it works. This chart visually describes the advertisement intervals versus time (smaller interval is faster advertising).

interval_chart

I've tested this on my dev-kit and sealed unit with numerous connect-disconnect and power cycles and it is working.

evergreen22:
Thanks @lman0.

geekbozu:
I feel like we should be doing the 62.5 Advertising interval for about 30 seconds every-time the screen turns on since its a "priority" style thing. And then back off either linearly or in a jump after that. I'm not sure of the benefit of the linear drop...Some rough numbers are ~1.35mA -> 1.18mA after it fully scales down. 
Which is ~5days and 6 hours to 6 days or about 18 hours extra life compared to full bore advertising! 

Any input on why you picked a linear drop? thanks for putting this effort in :)

evergreen22:
Note to @geekbozu, somehow I've broken my client, so I'm recreating my response to you here (assuming you got the response in the first place).

Thanks for the feedback and the discussion. I do appreciate it.

> I feel like we should be doing the 62.5 Advertising interval for about 30 seconds every-time the
> screen turns on since its a "priority" style thing. And then back off either linearly or in a jump
> after that.

I don't see the point of advertising when we are already connected. We can only have one connection at a time on the watch.

What do you mean by "priority style thing"?

> Some rough numbers are ~1.35mA -> 1.18mA after it fully scales down.
> Which is ~5days and 6 hours to 6 days or about 18 hours extra life compared to full bore
> advertising!

Thanks for measuring this! 18 hours is nice and more than I expected. If anything, this confirms my intution that advertising is fairly energy efficient on this hardware.

> Any input on why you picked a linear drop?

Linear is simple. Simple code is easier for humans to understand and less likely to contain defects.

My first approach was based on Ethernet's exponential backoff algorithm, However, that was too fast of a decrease imo for this application (and unnecessarily complex). This specific linear curve gives folks a full minute of fairly quick advertising to get connected but still feels responsive even after the 75 seconds. The worst case, after the advertising reaches the slowest interval, is that a connection will take 4.43 seconds. The average case will be 2.2 seconds and both feel pretty snappy imo.

geekbozu:
Ahh I explained poorly then. 

So all of this only applies when not connected and we should be advertising.

So when not connected, We initially advertise at the 62.5 Interval for about 30 seconds. After those 30 seconds we back off to a more power friendly speed. Say 1.2 seconds like recommended by the spec. Or less if we determine its snappy enough. 

Now while not connected, When the screen turns on. We could consider going back up to the 62.5interval for responsiveness and compatibility. (Thats what I meant by "priority" )

Yeah 18 hours is quite a bit id say its a good trade off for the watch always advertising.

evergreen22:
> So when not connected, We initially advertise at the 62.5 Interval for 
> about 30 seconds. After those 30 seconds we back off to a more power 
> friendly speed. Say 1.2 seconds like recommended by the spec. Or less if 
> we determine its snappy enough.
>

OK, I think I'm following. You're suggesting that instead of the linear 
slow down in 15 steps use two steps.

The code would do this:

* When powered on, advertise at 62.5ms interval for 30 seconds then 
change the interval to 1.2 seconds until connected.

* When a disconnect event is received, restart the two step sequence.


> Now while not connected, When the screen turns on. We could consider 
> going back up to the 62.5interval for responsiveness and compatibility. 
> (Thats what I meant by "priority" )
>

I don't understand why we would advertise when the screen turns on.

evergreen22:
> Now while not connected, When the screen turns on. We could consider going back up to the
> 62.5interval for responsiveness and compatibility. (Thats what I meant by "priority" )
> 

**While not connected** is the key part I missed above.

I think I see what you're saying now.

* When powered on, advertise at 62.5ms interval for 30 seconds then
change the interval to 1.2 seconds until connected.

* When a disconnect event is received, restart the two step sequence.

* When _NOT_ connected _AND_ the screen is on, restart the sequence.

evergreen22:
I've pushed the reworked advertising backoff scheme to PR #625. It now advertises agressively for at least 30 seconds then backs down to about one advertisement a second. It also restarts aggressive advertising under selected conditions. The chosen advertising intervals are consistent with Apple's developers guidelines for BLE and since we have a iOS app now that seems right.

I'm ready to remove the 'WIP' designation if a dev is willing to give it an 'approve' review.

Thanks to @geekbozu and @hubmartin for the feedback.

evergreen22:
Thanks for the feedback. I do really appreciate the discussion.

> I feel like we should be doing the 62.5 Advertising interval for about 
> 30 seconds every-time the screen turns on since its a "priority" style 
> thing. And then back off either linearly or in a jump after that.
>

I don't see the point of advertising when we are already connected. We can 
only have one connection at a time on the watch so there doesn't appear to 
be any need to advertise when connected.

What do mean by "priority style thing"?

> I'm not sure of the benefit of the linear drop...Some rough numbers are 
> ~1.35mA -> 1.18mA after it fully scales down. Which is ~5days and 6 
> hours to 6 days or about 18 hours extra life compared to full bore 
> advertising!
>

18 hours of extra battery life seems nice and more than I expected. Thanks 
for measuring this btw. If anything, this confirms my intuition that 
advertising is fairly energy efficient in the first place.

> Any input on why you picked a linear drop? thanks for putting this effort in :)
>

Linear is simple. Simple code is easier for humans to understand, and less 
likely to contain defects.

My first approach was something similar to Ethernet's exponential backoff 
algorithm, but that slows down too fast imo for this application. I picked 
this specific linear curve because it gives the user enough time to get 
their device connected before it levels off and after it levels off the 
connection time still feels responsive. It should connect reasonably 
quickly when you tap (re)connect in GB even after the 75 seconds. The 
worst case (re)connect would be 4.43 seconds.

evergreen22:
> Now while not connected, When the screen turns on. We could consider 
> going back up to the 62.5interval for responsiveness and compatibility. 
> (Thats what I meant by "priority" )
>

"While not connected" is the key part I missed above.

I think I see what you're saying now.

* When powered on, advertise at 62.5ms interval for 30 seconds then
change the interval to 1.2 seconds until connected.

* When a disconnect event is received, restart the two step sequence.

* When _NOT_ connected _AND_ the screen is on restart the sequence.

kieranc:
> I'm ready to remove the 'WIP' designation if a dev is willing to give it an 'approve' review.

I'm not qualified to provide a review however I'm testing this PR (plus #621) and will report back once I have > 24h uptime

evergreen22:
Thanks. More testing would be super helpful.

I've been running this for a couple of weeks on both my dev-kit and watch 
and have not noticed any regressions. I have seen several days of 
continuous connection, several days of advertising, and the mystery 
disconnects.

I set out to remedy the disconnect problem, but sadly this has not fixed 
it. However, this PR does get our ble advertising in spec with the ble 
standards, Apple developer guidelines, and best practices for bluetooth.

geekbozu:
Code looks great! I will actually pull some Power plots of this shortly so we can calculate a new "average" battery life!

geekbozu:
ppk-20210907T022214
Fast advertising, Quite power hungry compared to what it used to be...but man it connects fast/nicely.

ppk-20210907T021548
Slow advertising. Nominally what we used to have as advertising power draw. 

ppk-20210907T022347
Connected.
Little more then slow power draw ironically.

geekbozu:
So i have been running this as a "daily driver' for a very small sample size of 2 days and its working great. Pairing is way more consistent and much more responsive. I use a self compiled version of gadget-bridge with the Auto Reconnect on Out of range enabled and I have yet to notice BT going down/not coming back after those events either now. Where before it would sometimes take a good minute to actually connect. 

Does not solve the 18Hour bug as i have come to call it, But hey still super nice!

geekbozu (approved):

lman0:
hi, @evergreen22 

i have tested your  pull and it do advertising faster 
but only once i reset the smartwatch , 
because some time after be connected , if it disconnect
 and try for example to connect the pinetime to an Bluetooth hr app
 the adversiting is not working anymore ( i can't see the infinitime )
 
can you add an icon like
((-))   
when it advertising? it
 would show that it truly do advertising 
and when connected it would show the bluethooth icon instead .

thanks for your understanding

geekbozu:
lman0 there is a bug with InfiniTime where ALL BLE stops working after about 18 hours of runtime.
There are a few of us actively trying to hunt this bug down as its quite frustrating, as far as InfiniTime is concerned it is still advertising. But the radio just stops actually sending data out.

lman0:
> lman0 there is a bug with InfiniTime where ALL BLE stops working after about 18 hours of runtime.

, @geekbozu, in my case , it's not in 18 h but
in mere 1-2h after i reset the pinetime , that the c stop working
(and since there is no icon that show that it's do advertising , i think there is a bug lurking in this pull)

that's why i wish that @evergreen22  add an icon when it's advertising

so i can reliably said that the icon is not here when it should
allowing him to help resolve the bug

lman0:
>  Currently you can assume, If BLE is not broken, and you are not connected. We are advertising with this pr. As that is the goal of it!

then i can say that , sadly , the pull is not working
and it my word against yours 
since
there is no visual way to prove me otherwise that the advertising a restarted after a disconnect

lman0:
by the way , @evergreen22 , if like @geekbozu  say , the bluetooth stop working after 18h of runtime , 
you still can show the icon if the uptime < 18 of runtime in code 
so we can still be able to check the adversting is restarting properly in the time the ble is working

evergreen22:
You can't determine if the radio is advertising with an 'icon'.

geekbozu:
Yeah...there is no way to know when we are advertising. In the code, on any connection drop or advertisement finish event finish we start advertising again.

https://github.com/evergreen22/InfiniTime/commit/22571d4b384e40d647cd994202956f08ed32d925#diff-c69ced976eada3d4c4a15fd5ce542651ff381e21e2bb8cee746ac198c8847f89R152
https://github.com/evergreen22/InfiniTime/blob/22571d4b384e40d647cd994202956f08ed32d925/src/components/ble/NimbleController.cpp#L160
https://github.com/evergreen22/InfiniTime/blob/22571d4b384e40d647cd994202956f08ed32d925/src/components/ble/NimbleController.cpp#L176

We check if the call to advertise works, When it does we continue. If it does not we just reset the watch completely. So we have no way of knowing if we are actually advertising or not. Because as far as InfiniTime is concerned we are advertising. So an indicator will show as advertising even when its not. Since we actually have no way of knowing what the radio it self is actually doing.

lman0:
at very least  @evergreen22 , you could make a icon showing when the "startAdvetising()" , is  launched , right?
it should help regard less , to pin point use case that you may overlooked...
because in my case the pull seem not working propely

lman0:
by the way , @geekbozu what you say should be true then with the blutooth icon when we are connected ,
yet it's still shown , so i think @evergreen22  that an advertising icon is still legit

evergreen22:
To continue the reply from @geekbozu

The assert will fire if nimble indicates that advertising failed to 
start/restart in the advertising function. Neither of my test units are 
restarting. No one else has reported their watch is restarting in the 
advertising function. Therefore, we can conclude that nimble has told the 
watch it is advertising. What you are suggesting is assert does not work. 
I see no evidence to support your claim. I'm not saying you are wrong. I 
am just unwilling to accept your claim without some concrete proof that 
the assert does not work.

Everyone agrees there is a problem with intermittent disconnects and this 
refactor is a step forward in identifying the cause because it conforms to 
the ble standards and conventions. It allows us to eliminate 
'unconventional or nonconforming' usage as a cause of the disconnects. We 
are just as disappointed as you that the root cause has yet to be 
discovered.

lman0:
how i can prove my , claim if i have no way to prove that the the advertising have been started?
again it's my word against yours .... and i have no visual way way to  prove otherwise...
it's should not that hard of a dev of your calibre  to add such icon ....

if you don't do it , so be it , but don't dare to say your test are perfect because even big corporation made mistake 
and none is perfect

evergreen22:
I don't accept your claim without proof because you are arguing that the 
compiler or build process is broken. My reluctance has nothing to do with 
my PR code.

> how i can prove my , claim

Can you identify code in InfiniTime that disables the assert macro or 
otherwise prevents it from working?

kieranc:
> at very least @evergreen22 , you could make a icon showing when the "startAdvetising()" is launched , right?
> (and not showing when it connected or stopped),

You seem to understand this pretty well, it should be fairly trivial to display an icon, why don't you do it?

lman0:
i'm an user @evergreen22 , not a dev , and i feel like talking to support that say their program is 'perfect"'
and when i say  to but an icon when it connected they say 
and i have t prove them ithe code (that i don't understand) to found where is the bug
of why it not connected if i want the icon to be dev

lman0:
> > at very least @evergreen22 , you could make a icon showing when the "startAdvetising()" is launched , right?
> > (and not showing when it connected or stopped),
> 
> You seem to understand this pretty well, it should be fairly trivial to display an icon, why don't you do it?

i think they are good dev since they tackle a problem know to be very difficult to resolve
that why i think that they could be able to do it

kieranc:
> even more when they say that their unit test / use case are perfect ...
Who said that?

lman0:
when he say that his assert are working properly 
and say i have to prove that  my use case , the advertising is not working
it's like he mean that his assert are perfect , and i have to prove him otherwise....

kieranc:
> when he say that his assert are working properly
> and say i have to prove that my use case , the advertising is not working
> it's like he mean that his assert are perfect , and i have to prove him otherwise....

Its not 'his' assert, its an assert, one of hundreds in the code, and if it wasn't working as expected then many many other things would be broken.

> 
> by the way , @evergreen22 say "Neither of my test units are restarting. No one else has reported their watch is restarting in the advertising function. "
> when i didn't say , even once , that the watch is restarting by itself
> but only that the advertising stop working after 1 -2 hour and i need to reset the watch to see it again

He means he has 2 watches running code which should reboot the watch if advertising is not running (ie, the same approach as display an icon if its running/not) and no watches are rebooting, so as far as the code is concerned, the watch is advertising. We know the advertising is not working correctly, and while this pr doesn't fix it, it fixes other things which will hopefully make finding the issue with advertising easier.

Riksu9000:
I don't really know how this works, but obviously StartAdvertising() must have succeeded because of the asserts. The only other thing I can think of is if StartAdvertising() somehow doesn't get called. This could stop the advertising from working until reset if there are no new events that call it right? Previously waking the watch up would restart advertising. However with how `BLE_GAP_EVENT_DISCONNECT` and `BLE_GAP_EVENT_ADV_COMPLETE` are set up, it would at least seem like it wouldn't fail right?

>when he say that his assert are working properly
and say i have to prove that my use case , the advertising is not working
it's like he mean that his assert are perfect , and i have to prove him otherwise....

StartAdvertising() hasn't changed significantly. It still works basically the same way as before, but with more asserts.

>by the way , @evergreen22 say "Neither of my test units are restarting. No one else has reported their watch is restarting in the advertising function. "
when i didn't say , even once , that the watch is restarting by itself
but only that the advertising stop working after 1 -2 hour and i need to reset the watch to see it again

If it isn't restarting, that means StartAdvertising() is working.

evergreen22:
> You seem to understand this pretty well, it should be fairly trivial to display an icon, why don't you do it?

As explained previously, the 'icon' would be on all the time.

@lman0 is taking this personally and I don't mean it that way. I assumed they were a programmer and would understand my ask, However, since @lman0 is not a programmer, I concede it is an unreasonable request and I apologize.

Thank you @lman0. You have inspired a new line of thought about troubleshooting the disconnects. My test builds enable the assertions. They are not enabled in the production build. You have caused me to wonder if they are enabled in nimble - I shall investigate - thank you.

lman0:
thanks @evergreen22 i wish you good luck , in your investigation

evergreen22:
> thanks @evergreen22 i wish you good luck , in your investigation
>

You are most welcome.

For those who wish to build a debug version against 1.4.0 , here is what 
you need:

diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 67bbc83..c651352 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -777,8 +777,8 @@ add_definitions(-D__STACK_SIZE=1024)
  add_definitions(-D__HEAP_SIZE=4096)

  # NOTE : Add the following defines to enable debug mode of the NRF SDK:
-#add_definitions(-DDEBUG)
-#add_definitions(-DDEBUG_NRF_USER)
+add_definitions(-DDEBUG)
+add_definitions(-DDEBUG_NRF_USER)

  if (NOT CMAKE_BUILD_TYPE)
      set(CMAKE_BUILD_TYPE "Release")

lman0:
i will try to understand code , in order to be able one day be able to add functianlity , 
but as user ,such path is not easy one

JF002:
@evergreen22 Thank you so much for this PR! Even if it doesn't fix all our BLE issues, it'll probably improve the reliability of the (re)connection to the phone :+1: 

I'll try to answer to some of the comments I've read in this review.

@lman0 
> @JF002 since we use nordic , why we don't use their stack? since from i can see it's open source (otherwise esprino would not be able to modifiy it and share it) , and would resolve some ble problem.

Nordic stack (SoftDevice) is **not** open source, but the SDK is. The SoftDevice is provided as a binary file (<NRF SDK>/components/softdevice/s132/hex/s132_nrf52_6.1.1_softdevice.hex) you have to flash at a specific location in flash. Then, the software (using the open source SDK) is able to call functions from this binary file using some linker script magic. This is a smart way to allow open source software to use a proprietary stack, as the software does not link directly with the proprietary code.

At the beginning, InfiniTime used the NRF SoftDevice as BLE stack, but I decided to swtich to NimBLE because I wanted InfiniTime to be a fully FOSS firmware ;) The switch was done in [InfiniTime 0.5](https://github.com/JF002/InfiniTime/releases/tag/0.5.0).

@evergreen22  @geekbozu  Regarding **fast/slow advertising**, I think you chose the best compromise (fast advertising for 30s, then slow advertising). :+1: 

@lman0 Regarding the **advertising icon**, it could probably be useful to tell the user the watch is actively advertising and trying to the connect to a host device. Maybe some users will find this additional feedback useful, maybe others will just find it confusing... I don't really know. However, in this case, it won't be useful at all as our BLE issues cause inconsistencies between InfiniTime and the radio : InfiniTime (and NimBLE) really think that the watch is advertising while the radio outputs nothing into the air. Which mean that the advertising icon will be displayed even when advertising stopped working for any reason.
Anyway, I think this subject is closed, and I thank everyone for keeping a cordial and respectful tone in this discussion!

It's a bit late, I'll continue the review tomorrow !

kieranc:
> when he say that his assert are working properly
> and say i have to prove that my use case , the advertising is not working
> it's like he mean that his assert are perfect , and i have to prove him otherwise....

Its not 'his' assert, its an assert, one of hundreds in the code, and if it wasn't working as expected then many many other things would be broken.

> 
> by the way , @evergreen22 say "Neither of my test units are restarting. No one else has reported their watch is restarting in the advertising function. "
> when i didn't say , even once , that the watch is restarting by itself
> but only that the advertising stop working after 1 -2 hour and i need to reset the watch to see it again

He means he has 2 watches running code which should reboot the watch if advertising is not running (ie, the same approach as display an icon if its running/not) and no watches are rebooting, so as far as the code is concerned, the watch is advertising. We know the advertising is not working correctly, and while this pr doesn't fix it, it fixes other things which will hopefully make finding the issue with advertising easier. 


> > I'm ready to remove the 'WIP' designation if a dev is willing to give it an 'approve' review.
> 
> I'm not qualified to provide a review however I'm testing this PR (plus #621) and will report back once I have > 24h uptime

FWIW, I've been running this PR for nearly a week and haven't noticed any regressions, it seems to reconnect automatically more reliably than before

JF002:
Here's a DFU file for testing this PR (built on https://github.com/JF002/InfiniTime/commit/6356c7f276d09ab32d6f1c1338dac0b00a602232).
[pinetime-mcuboot-app-dfu-1.4.99.zip](https://github.com/JF002/InfiniTime/files/7156729/pinetime-mcuboot-app-dfu-1.4.99.zip)

PR #629, opened 2021-08-31
WIP: Score app to track your score during your sport match

toitoinou:
I don't know if it is too soon to create the PR, it's in progress so very basic for now.

I play a bit of Tennis and other racket sports and thought it will be interesting to use a watch to keep track of the score you always forgot ^^'

Only 2 counters are working, with underflow and overflow protection plus a reset button. That's all I got for now.

But I'm interested in your code review, I based my app on the Timer app but I tried to remove the more duplicate code I could (I hate duplicate code).

I plan to implement several racket sport rules : tennis (the most difficult), badminton, squash, table tennis.. and even magic the gathering (for fun a phone will be much more practical for mtg).

### DIFF
--- a/src/displayapp/screens/ScoreApp.cpp
+++ b/src/displayapp/screens/ScoreApp.cpp
@@ -56,12 +56,15 @@ ScoreApp::ScoreApp(DisplayApp* app, Controllers::MotorController& motorControlle
   lv_dropdown_set_selected(modeDropdownWdg, 0);
   lv_dropdown_set_show_selected(modeDropdownWdg, true);
 
+  taskRefresh = lv_task_create(RefreshTaskCallback, LV_DISP_DEF_REFR_PERIOD, LV_TASK_PRIO_MID, this);
+
   // lv_dropdown_set_text(modeDropdownWdg, "mode");
 
   // TODO(toitoinou): add active mode icon in empty area
 }
 
 ScoreApp::~ScoreApp() {
+  lv_task_del(taskRefresh);
   lv_obj_clean(lv_scr_act());
 }
 
--- a/src/displayapp/screens/ScoreApp.cpp
+++ b/src/displayapp/screens/ScoreApp.cpp
@@ -1,0 +1,86 @@
+#include "Screen.h"
+#include "Symbols.h"
+#include "lvgl/lvgl.h"
+
+#define SCREEN_WIDTH 240
+#define SCREEN_HEIGHT 240
+
+#define SCORE_WIDTH (SCREEN_WIDTH*60)/100
+#define SCORE_HEIGHT SCREEN_HEIGHT/4
+
+#define MINUS_WIDTH SCREEN_WIDTH-SCORE_WIDTH
+#define MINUS_HEIGHT SCREEN_HEIGHT/4
+
+#define RESET_WIDTH SCREEN_WIDTH-SCORE_WIDTH
+#define RESET_HEIGHT SCREEN_HEIGHT/4
+
+using namespace Pinetime::Applications::Screens;
+
+
+static void btnEventHandler(lv_obj_t* obj, lv_event_t event) {
+  ScoreApp* screen = static_cast<ScoreApp*>(obj->user_data);
+  screen->OnButtonEvent(obj, event);
+}
+
+lv_obj_t* ScoreApp::_createButton(lv_align_t alignment, uint8_t x, uint8_t y, uint8_t width, uint8_t height, char text[]){
+  lv_obj_t *button;
+
+  button = lv_btn_create(lv_scr_act(), nullptr);
+  button->user_data = this;
+  lv_obj_set_event_cb(button, btnEventHandler);
+  lv_obj_align(button, lv_scr_act(), alignment, x, y);
+  lv_obj_set_width(button, width);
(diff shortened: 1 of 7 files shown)


src/displayapp/screens/ScoreApp.h
```diff
@@ -0,0 +1,51 @@
...
+
+    ~ScoreApp() override;
+
+    void Refresh() override;
+
+    void OnButtonEvent(lv_obj_t* obj, lv_event_t event);
+
+  private:
+    static constexpr uint16_t displayWidth = 240;
+    static constexpr uint16_t displayHeight = 240;
```
Riksu9000, line 25: `LV_HOR_RES` and `LV_VER_RES` already have these values, so they can be used in place of these.

src/displayapp/screens/ScoreApp.h
```diff
@@ -0,0 +1,51 @@
+#pragma once
+
+#include "Screen.h"
+#include "systemtask/SystemTask.h"
+#include "../LittleVgl.h"
+#include <lvgl/src/lv_core/lv_obj.h>
+
+namespace Pinetime::Applications::Screens {
```
Riksu9000, line 8: This shows a warning, "Nested namespace definition is a C++17 extension"

src/displayapp/screens/ScoreApp.cpp
```diff
@@ -0,0 +1,80 @@
...
+ScoreApp::~ScoreApp() {
+  lv_task_del(taskRefresh);
+  lv_obj_clean(lv_scr_act());
+}
+
+void ScoreApp::Refresh() {
+  lv_label_set_text_fmt(btnMyScore.label, "%d", myScore);
+  lv_label_set_text_fmt(btnYourScore.label, "%d", yourScore);
+}
+
```
Riksu9000, line 60: This app doesn't require `Refresh()` at all at least currently. You can remove the task and update the labels in `OnButtonEvent()` instead

src/displayapp/screens/ScoreApp.h
```diff
@@ -0,0 +1,51 @@
...
+
+    struct widget_t {
+      lv_obj_t* button;
+      lv_obj_t* label;
+    };
+
+    uint8_t myScore = 0;
+    uint8_t yourScore = 0;
+
+    widget_t _createButton(lv_align_t alignment, uint8_t x, uint8_t y, uint8_t width, uint8_t height, const char text[]);
```
Riksu9000, line 44: I think this should be PascalCase and without the underscore.

src/displayapp/screens/ScoreApp.cpp
```diff
@@ -0,0 +1,80 @@
...
+ScoreApp::widget_t ScoreApp::_createButton(lv_align_t alignment, uint8_t x, uint8_t y, uint8_t width, uint8_t height, const char text[]){
+
+  widget_t button;
+
+  button.button = lv_btn_create(lv_scr_act(), nullptr);
+  button.button->user_data = this;
+  lv_obj_set_event_cb(button.button, btnEventHandler);
+  lv_obj_align(button.button, lv_scr_act(), alignment, x, y);
+  lv_obj_set_width(button.button, width);
+  lv_obj_set_height(button.button, height);
```
Riksu9000, line 23: You can use `lv_obj_set_size(object, width,  height)`

Riksu9000 (commented):
I haven't tested this but it seems good. Maybe it could be used as a generic counter/tracker as well.

toitoinou:
> I haven't tested this but it seems good. Maybe it could be used as a generic counter/tracker as well.

Why not, for points in table top game for example or whatever. OK I'll let a "generic" mode.

toitoinou:
Uploading VID_20210831_105204798.mp4…

A little video of the very basic app.

src/displayapp/screens/ScoreApp.cpp
toitoinou: Yes, I did not understood that refresh is used mainly for synchronous/external event and asynchronous event like buttons and touch can be handled directly.
Corrected in https://github.com/toitoinou/Pinetime/commit/b5a762a1de2682963eb98bda42a06ee710741935

src/displayapp/screens/ScoreApp.h
toitoinou: Bad python habit ^^
Corrected in https://github.com/toitoinou/Pinetime/commit/b5a762a1de2682963eb98bda42a06ee710741935

src/displayapp/screens/ScoreApp.cpp
toitoinou: Corrected in https://github.com/toitoinou/Pinetime/commit/b5a762a1de2682963eb98bda42a06ee710741935

OrgansWithoutBodies:
I was thinking that something like this would be useful tracking workouts too - if it could be integrated (along w the footstep counter) as something serializable over BLE that would be a great generically useful feature

Riksu9000:
@toitoinou Hi. Are you still working on this?

toitoinou:
Hi, sorry, lots of work and a moving, prevent me from working on it, but I would like to continue my little app...

Riksu9000:
This PR is WIP, but hasn't been updated. Closing as stale.

PR #631, opened 2021-08-31
Add battery percentage & colouring

Sykomaniac:
Definitely a WIP but submitting to see if there is any appetite for me to continue this work?

From what I can see it looks like each watchface would need to add the battery percentage - not sure if this is desired or if the battery icon should be reworked to include percentage functionality. Could also add the option to enable/disable in settings.

IMG_20210831_145529


evergreen22:
I'd rather see the battery icon change colors for different ranges in state of charge. For example, green for SOC > 75%, yellow for 50% - 74%, and red for SOC < 50%.

Adding a number to the left of icon just feels like clutter to me.

Riksu9000:
There are people who want this and there are people like us that think its clutter. I would personally even like to replace the battery icon on the watch face entirely with a low battery notification as I don't see a reason to stress about the battery until it is time to charge.

hubmartin:
@evergreen22 I would not suggest to use different colors for battery level. Different watchfaces use different colors and now you can choose even your own colors. This might break consistency across different watchfaces. Also red color icon is almost invisible in daylight so you would't spot that.

Sykomaniac:
I decided to implement both suggestions - I have only changed the style for the digital watch face. I did see that there were some discussion about refactoring settings for watchfaces so it might be better to revisit this after then?

I've attached some screenshots (sorry for the poor quality). Also I've fixed some casing and renamed a few things since I took them to maintain consistency with other settings pages.

IMG_20210907_223400
IMG_20210907_223332
IMG_20210907_223347

Sykomaniac:
Any interest in this PR or shall I close it?

Avamander:
I think people have been plenty interested.

Riksu9000:
I'm not sure the colorful icon needs to be an option. For most of the range white makes sense and maybe when it's under 15% it can be red or orange. I'm sure the color can be adjusted so that it is visible in daylight. #638 would make the battery icon green only when it's fully charged.

Sykomaniac:
Updated PR to resolve merge conflicts and add styling to analog watchface
IMG_20210922_225048

Sykomaniac:
I don't plan on making any more changes to this PR unless people would like things moving/features removing. I won't update the pinetime style watchface to adhere to these settings unless its again something people want (as that watchface already has a number of style options)

src/displayapp/screens/WatchFaceDigital.cpp
```diff
@@ -106,6 +110,19 @@ void WatchFaceDigital::Refresh() {
...
     lv_label_set_text(batteryIcon, BatteryIcon::GetBatteryIcon(batteryPercent));
+
+    if (settingsController.GetBatteryColorStatus() == Controllers::Settings::BatteryColor::ON) {
+      lv_obj_set_style_local_text_color(batteryIcon, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, BatteryIcon::GetBatteryColor(batteryPercent));
+    } else {
+      lv_obj_set_style_local_text_color(batteryIcon, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, BatteryIcon::GetDefaultBatteryColor());
+    }
+    
+    if (settingsController.GetBatteryPercentageStatus() == Controllers::Settings::BatteryPercentage::ON) {
+      lv_label_set_text_fmt(batteryValue, "%lu", batteryPercent);
```
AlexXZero, line 121: Would it be better to print percentage sign?
```suggestion
      lv_label_set_text_fmt(batteryValue, "%lu%%", batteryPercent);
```

Sykomaniac:
Re-implemented this in: #1028

Riksu9000:
I'll close this in favor of #1028

PR #633, opened 2021-09-01
Flashlight color settings

wilsonjholmes:
My first attempt at a contribution, so forgive me if I have done anything non-standard for this process. I am open to advice.

I often need a light at night that doesn't mess up my night vision (amateur astronomy), and I thought I could make some settings for the flashlight to allow the switch between white and red. I ended up adding green and blue as well. I guess the only thing that is being weird for me is the `Symbols::highlight` as it doesn't seem to be working in my added settings.

settingsPage
whiteSelected
redSelected
redFlashlightOn
redFlashlightOff


Riksu9000:
Is this practical? If so, would changing the colors with a long press or a swipe work, so no setting would be required?

Avamander:
> If so, would changing the colors with a long press or a swipe work, so no setting would be required?

Only if it remembers past state, it turning white each launch would negate its usefulness.

wilsonjholmes:
> Is this practical?

Are you asking if this pr is feature creep... Or, are you asking if having a whole menu just for the flashlight color makes sense when you could change it another way (i.e. while in the app itself)

Riksu9000:
> > Is this practical?
> 
> Are you asking if this pr is feature creep... Or, are you asking if having a whole menu just for the flashlight color makes sense when you could change it another way (i.e. while in the app itself)

I can't think of many use cases for this feature and the light isn't very bright so I'm not sure how useful this is, however making it adjustable in app results in less clutter, which might make it fine.

tmilburn:
For astronomy do you want the whole UI to use only red in which case there could be a setting which modifies the display driver to set the blue and green components to 0.

wilsonjholmes:
That is a good point. Would that be implemented in `St7789.cpp`? Thinking about adding it in `DrawPixel`, but not sure if that is the right place... Would setting blue and green to 0 make everything display as a shade of red?

geekbozu:
I think this is a great contribution...however yeah it probably should be a swipe to change color that is saved in settings. For astronomy stuff however maybe a double swipe or something would be better? Mostly so it does not accidentally change...can take quite some time to get that night vision back. And the back light on this watch is damaging enough to that. 

Flip side there, what @wilsonjholmes said would be nifty. Adding in some driver level color muxing for Astronomy/BW/and Color Correction for those with color blindness would be neat. Albeit out of the scope of this original PR :)

JF002:
Astronomy is a nice use-case! Did you try it once it's red? Is it usable in your use-case?

Regarding the setting, I asked a similar question [here](https://github.com/JF002/InfiniTime/pull/636#issuecomment-912052329) : should the settings of the app be handled in a separate setting app or directly into the app.

minacode:
This is definitel out of scope for this PR, but maybe there could be a general redshift-like setting for the whole UI that will also be used by the flashlight to dim it towards red?

Riksu9000:
I'll close this because it's outdated and we still aren't sure if these colors are useful additions. If there's enough reason to add red, please create a new PR and make it the leftmost option in the current flashlight app.

PR #636, opened 2021-09-02
Move PineTimeStyle settings up

Itai-Nelken:
It's now over the 'Firmware' entry. It make sense to have 'Firmware' and 'About' the 2 last entries as they are the least used in my opinion.

### DIFF
--- a/src/displayapp/screens/settings/Settings.cpp
+++ b/src/displayapp/screens/settings/Settings.cpp
@@ -50,8 +50,8 @@ std::unique_ptr<Screen> Settings::CreateScreen2() {
   std::array<Screens::List::Applications, 4> applications {{
     {Symbols::shoe, "Steps", Apps::SettingSteps},
     {Symbols::batteryHalf, "Battery", Apps::BatteryInfo},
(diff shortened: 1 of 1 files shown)


JF002:
Yup, that makes sense!

We could probably use this opportunity to think about how we'll handle the settings of other apps/watchface : should we add a new setting app for each app and watchface, or should the apps/watchface handle their settings by themselves?

Itai-Nelken:
I think that a single 'Watchface settings' entry that contains settings for the current watchface is a good idea. That way settings won't get cluttered with entries.

0x1a8510f2:
For the time being, I think the PineTimeStyle settings should probably be just under the watch face settings so it's closer to related settings.

kieranc:
I'm going to try integrate the settings with the watchface in a way that can be extended to the other watchfaces - the current idea is that a long press on the watchface will make a settings icon appear for a few seconds, if you tap the icon it opens the settings. I don't want to just make long press open settings directly as I'm concerned it'll be triggered by accident.

JF002:
@kieranc That might be a good idea! It would be nice if you could do a quick PoC to seet what it looks like!

PR #638, opened 2021-09-03
Detect full charge and improve watchface battery display

Riksu9000:
Even after reaching the "maximum" voltage, the battery will keep charging for a while. We can use the CHG pin to see when the battery is actually fully charged.

I made it so it will show at most 99% while it isn't full. This makes it so the last percentage takes quite a bit longer to reach, but this can't be helped if we only use the voltage to calculate percentage. This could be improved by using time to estimate percentage while charging.

Analog watch face will now show a plug icon instead of a battery icon while charging and a green full battery icon when fully charged.

Digital watch face will now ~~hide the plug icon and~~ show a green full battery icon when fully charged.

PineTimeStyle will show a full battery icon when fully charged.


hubmartin:
I would keep the plug symbol when the charger is plugged.

I'm sure the charging chip will eventually start to charge battery again if I keep it on the charger for a week and battery looses some voltage.

Also the pogo pins sometimes don't make proper contact, so I found it very handy to make sure it is connected properly.

Riksu9000:
> I would keep the plug symbol when the charger is plugged.
>
> Also the pogo pins sometimes don't make proper contact, so I found it very handy to make sure it is connected properly.

It will always show the plug icon first when it is plugged in, so this won't be an issue. It will only be hidden when the battery is full and it has stopped charging. On the digital watch face it could show both the battery and the plug icons when it is full, but the question becomes whether the plug icon should imply that it is charging, or that it is plugged in. It could be changed to a lightning bolt to better suggest charging if that is preferred.

> I'm sure the charging chip will eventually start to charge battery again if I keep it on the charger for a week and battery looses some voltage.

It does, but the user doesn't need to think about that. It will just keep showing it as full while it is plugged in.

hubmartin:
I understood that initially the plug will be showed everytime and disappear later. But it is plug symbol and watch is plugged.

When my notebook, phone is on the chager and charged, I always see the lightning bolt next to it no matter what the state of charge of the device is. **Even when its fully charged**. On PineTime the lightning bolt is the red plug icon. The lightning bold would be mabe better, but the idea is the same, that plug/lightning symbol should be there when device is connected to the power.

Riksu9000:
> When my notebook, phone is on the chager and charged, I always see the lightning bolt next to it no matter what the state of charge of the device is. Even when its fully charged.

2/4 devices I tested hide the lightning bolt when full, and all of them worked slightly differently from each other, so I don't think there's a standard way to do it. Are you still talking about the digital watch face, as all the watch faces work a bit differently as well?

hubmartin:
Yes, I refer to digital watchface only.

geekbozu (approved):
Not tested but code looks great. 
Changes to IsCharging are very welcome and will help with the lots of questions of "Why does this not fully charge"
Looks like a worthwhile merge as is to me. Watch face code being mutable as we determine what we want our charging behavior to actually be.

kieranc:
I'm running a build with this PR, it seems to behave as expected so far, I'll update if anything goes wrong!

Riksu9000:
I changed it so that it keeps showing the plug icon even while fully charged to match the current behaviour (with fixes). Also fixed an issue so it can show 100% while not plugged in.

hubmartin:
Tested with digital face, works as expected. :+1:

JF002:
Looks good to me, thanks!

PR #639, opened 2021-09-03
Add charging icon to Analog watchface

Itai-Nelken:
it replaces the battery icon while charging.


Itai-Nelken:
#638 already implements this

PR #645, opened 2021-09-04
specify where to change env variables for VSCode

lectrician1:
I was stuck with stupid problems building all because I didn't know where to set envs, so this should help other users who might run into the problem.


geekbozu:
So that is one place to specify these however, those files are versioned which will most likely cause some issues later on with people submitting their own paths and such to the repository. 
Really this should be updated to reflect specifying to launch VSCode with those variables set. Which is what the intention was. However it was never tested or proven on windows so all of the instructions are currently Linux specific.

PR #648, opened 2021-09-06
Improve sleep time

Riksu9000:
DisplayOn() and DisplayOff() are direct lcd commands, it's better to use only Sleep() and Wakeup().

The lcd is already controlled in SystemTask, so these aren't necessary in DisplayApp. SleepIn() puts the lcd into the lowest power mode so DisplayOff() doesn't do anything as far as I can tell.

EDIT: lcd can't be removed from DisplayApp, because DisplayAppRecovery needs it, and DisplayAppRecovery also has a lot of unused parameters. I'll look at this separately.

This eliminates the mysterious 500ms delay that exists in DisplayOff(). https://github.com/JF002/InfiniTime/blob/6f9f0e8b0e42a5526d47ca664534fb6b0ccb6ace/src/drivers/St7789.cpp#L119

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -182,12 +182,10 @@ void DisplayApp::Refresh() {
           brightnessController.Lower();
           vTaskDelay(100);
         }
-        lcd.DisplayOff();
         PushMessageToSystemTask(Pinetime::System::Messages::OnDisplayTaskSleeping);
         state = States::Idle;
         break;
       case Messages::GoToRunning:
-        lcd.DisplayOn();
         brightnessController.Restore();
         state = States::Running;
         break;
--- a/src/drivers/St7789.h
+++ b/src/drivers/St7789.h
@@ -22,9 +22,6 @@ namespace Pinetime {
 
       void DrawBuffer(uint16_t x, uint16_t y, uint16_t width, uint16_t height, const uint8_t* data, size_t size);
 
-      void DisplayOn();
-      void DisplayOff();
-
       void Sleep();
       void Wakeup();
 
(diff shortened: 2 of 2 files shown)


geekbozu:
I like these changes they definitely feel more appropriate. 
Do we know why this was included to begin with or are we just assuming it was a previous choice that is now just Tech-Debt?

src/displayapp/DisplayApp.h
```diff
@@ -44,8 +44,7 @@ namespace Pinetime {
       enum class States { Idle, Running };
       enum class FullRefreshDirections { None, Up, Down, Left, Right, LeftAnim, RightAnim };
 
-      DisplayApp(Drivers::St7789& lcd,
-                 Components::LittleVgl& lvgl,
+      DisplayApp(Components::LittleVgl& lvgl,
```
terencode, line 49: Shouldn't `Drivers::St7789& lcd` also be removed from https://github.com/JF002/InfiniTime/blob/develop/src/displayapp/DisplayAppRecovery.h#L45 ?

Riksu9000: Lcd is actually used in DisplayAppRecovery, but it looks like DisplayApp and DisplayAppRecovery must use all the same parameters even if they aren't needed. I would like to improve this as well, but for now I'll just put lcd back in DisplayApp even though it isn't used, just to keep this PR simple.

hubmartin:
Work fine. I've tested it for a few minutes together with my lcd wakeup optimizations and yours better button handler. I did not noticed any issues sleeping or when pressing button many times.

lman0:
any news on reviewing this @geekbozu ?

geekbozu:
I want to do a comparison on battery draw given these changes still. Life has been busy, PR is in my todo list to review/data log :)

JF002:
The [datasheet](https://wiki.pine64.org/images/5/54/ST7789V_v1.6.pdf) does not provide much details about DISPLAY ON and OFF modes, unfortunately, except that the display is in DISPLAY OFF mode after reset/sleep.
So the command DISPLAY ON is probably needed at startup and after sleep, but I don't know if we really need to send the command DISPLAY OFF before sleep.

IIRC, I wrote the display driver with the help of other community member at that time, and maybe by reading the sources of other similar driver found online. So I might have copied that displayon/off command just to be sure the display will work properly.

If there's no visible side effect of that command, let's remove it and spare 500ms each time the display goes to sleep :)

JF002 (approved):

PR #655, opened 2021-09-09
Paddle: compare ballX coordinate with horizontal resolution

NeroBurner:
Fix the comment as the comparison checks if the ball is at the right
side of the screen.

Compare the x coordinate of the ball with the horizontal resolution of
the screen, instead of the vertical resolution. On the PinePhone this
does make no difference as we have square 240x240 screen. Change it
anyways to be completely correct.


Riksu9000 (approved):

NeroBurner:
rebased on current develop branch

JF002:
Thanks!

PR #658, opened 2021-09-10
Run CI on master and develop branch

NeroBurner:
The documentation https://github.com/JF002/InfiniTime/blob/develop/doc/contribute.md#how-to-submit-a-pull-request-
tells us to make pull requests against `develop` branch.
Update the CI to run on those pull requests to have a check if the PR breaks something unintentionally.

Like https://github.com/JF002/InfiniTime/pull/126 but specifically for the develop branch, because a CI (Continuous Integration) should check the code continuously

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,13 +9,13 @@ name: Build PineTime Firmware
 # When to run this Workflow...
 on:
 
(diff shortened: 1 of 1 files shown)


JF002:
Yep, that's a good idea, thanks for the suggestion and for the PR!

PR #662, opened 2021-09-11
Basic alarm app

mruss77:
I added a basic alarm app that allows you to set one alarm to go off once, every day, or on weekdays (M-F).

I know there is discussion in [issue 366](https://github.com/JF002/InfiniTime/issues/366) about a full-featured alarm app with multiple alarms, which looks great. This basic alarm meets my simple need to wake up on time in the morning.

The UI is one screen based off the Timer app UI:

Basic Alarm app demo thumbnail

- The plus/minus buttons set the hours and minutes.
- The bottom-left button turns the alarm off/on, and becomes the "stop vibrating" button when the alarm goes off.
- The bottom right button toggles between repeating alarm settings.
- The "i" button will pop up a message showing how long until the alarm will go off.

Video of the alarm in action: https://www.youtube.com/watch?v=9U8QRy7Ixpw 

I added a method to MotorController to activate the "ringer" (repeating vibrations) even if notifications are turned off in settings. I like to turn off notifications when I go to sleep so I'm not woken up by some random 3 AM email, but I still need the alarm to go off in the morning.

Any comments or suggestions welcome.

Issue #610, opened 2021-08-25
[Feature request] don't show notifications if in a app. 


Itai-Nelken:
Its annoying that for example I'm playing one of the games and a notification arrives going back home after the timeout.

I think there is a notification icon in the watchfaces, it could be used to show there is a unread notification. Maybe turning off vibration can also make notifications not show.

ptman:
> A possible alternative is that when an open app is interrupted (eg. by a notification, or a timer going off) that the currently open app's state is saved and returned to when the interruption is complete/exited.

I've had lots of problems with the stopwatch since it resets when there is a notification. It could continue running in the background. Or I would event be happy to navigate back from the main screen. But losing the state is bad.

PR #664, opened 2021-09-12
Toggle notifications only, keep vibrations.

### LINKED ISSUE #637: Metronome does not work when vibrate-on-notification is disabled
To reproduce:
1. Disable vibrate-on-notification (bell icon on left screen)
2. Open metronome app
3. Start metronome

What I expect: periodic vibrations in the given tempo.
What actually happens: nothing!

Riksu9000:
This makes the button with a bell icon disable notifications instead of disabling vibrations. This makes Timer and Metronome work correctly when notifications are disabled. When disabled, the watch won't wake up when receiving notifications and will show an icon on the watch face indicating a new notification. This was always present, but wasn't ever shown.

notif

Fixes #637
Related to #610, because now the notification screen won't pop up when notifications have been disabled.
Inspred by #662 so that we wouldn't need a `StartRingingDisregardSettings()` function.

### DIFF
--- a/src/components/ble/NotificationManager.cpp
+++ b/src/components/ble/NotificationManager.cpp
@@ -79,14 +79,6 @@ bool NotificationManager::AreNewNotificationsAvailable() {
   return newNotification;
 }
 
-bool NotificationManager::IsVibrationEnabled() {
-  return vibrationEnabled;
-}
-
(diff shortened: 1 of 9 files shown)


JF002:
Looks good to me, thanks!

PR #667, opened 2021-09-13
Theme and color cleanup

Riksu9000:
Use LVGL defined colors where possible.
Add some custom colors to Colors.h.
Swap arc background and foreground colors in theme, so that background is darker than foreground/filled part. In Metronome the colors get swapped and in Steps the background will be darker.
Commented shadow and outline styles as they're disabled in [lv_conf.h](https://github.com/JF002/InfiniTime/blob/develop/src/libs/lv_conf.h)
Set disabled colors for button and label in theme, so they don't have to be set in app.

Some grays have changed very slightly, like 0x88 to 0x80, which probably isn't noticable.

This ends up using 1368 less bytes of flash. I assume some colors were calculated at runtime as `lv_color_hex()` couldn't be made constexpr and changing some `lv_color_hex()` to `LV_COLOR_MAKE()` saved some bytes.

metronome
steps


kieranc:
Nice work, much needed. I've merged this with my current build, will monitor for any abnormalities...

lman0:
hi, 
@kieranc is this pr make infinitime faster? last longer?

lman0:
hi @Avamander  , since you have accepted/approved https://github.com/InfiniTimeOrg/InfiniTime/pull/792
you could do the same for this one as well and 
infintime would gain 2kb more memory

Avamander:
It has merge conflicts at the moment and a review gets outdated after a commit, something might accidentally slip past easier.

lman0:
@Riksu9000 could you resolve the conflicts so @Avamander could review this pull?

Riksu9000:
Thanks for reminding me.

Something new has come up however. With the LVGL8 upgrade #734, the default color palette will change, which would make this PR obsolete. With LVGL8 we would instead need to stop using LVGL defined colors and define all our colors in Colors.h preferrably.

So if we'll upgrade to LVGL8 soon, it might not make sense to merge this.

PR #680, opened 2021-09-20
Create PineTimeStyleBase to reduce duplication

Riksu9000:
Created PineTimeStyleBase to create the LVGL objects, so they don't have to be created in PineTimeStyle and SettingPineTimeStyle separately.

Uses 1456 bytes less memory.


hubmartin:
Works fine. Checked PT Settings and TP clockwatch and everything works like expected. Thanks

PR #681, opened 2021-09-20
Motion: keep screen awake while showing motion app

NeroBurner:
Instead of having to manually periodically tap on the phone to keep the
screen awake, just prevent the screen to go to sleep while showing the
motion sensor graph.

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -421,7 +421,7 @@ void DisplayApp::LoadApp(Apps app, DisplayApp::FullRefreshDirections direction) 
       ReturnApp(Apps::Launcher, FullRefreshDirections::Down, TouchEvents::None);
       break;
     case Apps::Motion:
-      currentScreen = std::make_unique<Screens::Motion>(this, motionController);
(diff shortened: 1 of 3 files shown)


Riksu9000:
It seems that we're slowly disabling sleep in every app. Maybe instead the screen timeout should be longer in apps and shorter on the watchface.

Avamander:
@Riksu9000 
Though disabling sleep altogether is a bad idea IMHO, can end up with accidentally opening an app and the battery being dead by the morning for example. I really think an upper limit instead of disabling altogether would be a good idea.

mjnck:
@Avamander noone wants to disable sleep, just make timeout for example always two times longer than timeout on the watchface

lman0:
@Riksu9000  @Avamander maybe it would be good to have 2 "kind"  of settings for timeout :
-one for the watchface 
-one for the rest of apps

tmilburn:
Garmin watches have the following settings:

> Normal
> Sets the device to enter low-power watch mode after 5 minutes of inactivity.
> Extended
> Sets the device to enter low-power watch mode after 25 minutes of inactivity.

Riksu9000:
We will need a more complete solution to this, because simply disabling sleep isn't a good idea.

Issue #302, opened 2021-04-20
Watch does not automatically reconnect to GadgetBridge after turning airplane mode on and off


GithubUser5462:
InfiniTime version: 1.0.0 RC3
Gadgetbridge version: 0.56.1

Steps to reproduce:

1. Pair InfiniTime with Gadgetbridge, enable the "Reconnect automatically" setting in GB
2. While the two are connected turn  airplane mode on and then off after the watch is disconnected
3. GB tries to reconnect, first with status "Connecting" and then gets stuck with "Waiting for reconnect"

The workaround I have found is to fully disconnect in GB - status "Not connected" and then pair the watch again.
Restarting the watch or Gadgetbridge did not seem to help.

I don't know if the problem is with GB or IT but it would be nice if the two could automatically reconnect in this scenario if possible.

petterhs:
hmm I'm not able to reproduce this issue running same versions. Do you experience the same if you toggle Bluetooth  off/on on the phone?

GithubUser5462:
Yes, exact same result.

JF002:
This kind of issue is really hard to debug, and is probably caused by a mix of PineTime HW, InfiniTime, phone HW and software stack. 
InfiniTime does advertise itself automatically as soon as it's disconnected during 3 minutes. It also restarts advertisting every time the display is switched on. Then, it's up to the companion app to receive these advertising messages and reconnect to the watch.

pfeerick:
Not knowing that behaviour is most likely the 'cause' of the problem... i.e. I would probably have had a similar complaint if I didn't know that turning the screen on wakes up advertising. 

> InfiniTime does advertise itself automatically as soon as it's disconnected during 3 minutes. It also restarts advertisting every time the display is switched on. Then, it's up to the companion app to receive these advertising messages and reconnect to the watch.

Maybe a future thought could be have periodic advertising... say once every five minutes or so, for 10 seconds or whatever (I'm sure there's some convention that can be followed - or not!) ... i.e. try and make itself available of the phone wants to reconnect.

GithubUser5462:
I made a couple more tries to try to understand what is happening.
It seems that it is not limited to Gadgetbridge since it also happens with nRF Connect.

In nRF Connect when I disable/enable bt the "bonded" tab in nRF Connect gets wiped, but the scanner does detect watch - as not bonded.
[...]

PR #688, opened 2021-09-25
Fix advertising packets stopping after 18.2 hours

danielgjackson:
The previous code always returned 0 (when `MYNEWT_VAL_BLE_LL_RFMGMT_ENABLE_TIME` is not defined), rather than a time near to the current tick.  This caused an issue in at least one place: `ble_ll_adv_sm_start()`, where the calculation of `delta` overflows when the system timer is at 0x80000000 or above -- causing an incorrect, huge adjustment to be made to the scheduled time, ultimately stopping adverts from being sent.

This returns the current time as the time at which the RF will be fully enabled.  

This fixes the advertising issue described in the comments under Issue #302 (although the issue itself might be about broader connectivity problems).

### DIFF
--- a/src/libs/mynewt-nimble/nimble/controller/include/controller/ble_ll_rfmgmt.h
+++ b/src/libs/mynewt-nimble/nimble/controller/include/controller/ble_ll_rfmgmt.h
@@ -51,7 +51,7 @@ static inline void ble_ll_rfmgmt_reset(void) { }
 static inline void ble_ll_rfmgmt_scan_changed(bool e, uint32_t n) { }
 static inline void ble_ll_rfmgmt_sched_changed(struct ble_ll_sched_item *f) { }
 static inline void ble_ll_rfmgmt_release(void) { }
-static inline uint32_t ble_ll_rfmgmt_enable_now(void) { return 0; }
+static inline uint32_t ble_ll_rfmgmt_enable_now(void) { return os_cputime_get32(); }
 static inline bool ble_ll_rfmgmt_is_enabled(void) { return true; }
 
 #endif


geekbozu:
I will give this a good once over tonight or tomorrow....I've been hunting this for a few weeks my self.... I hope this is it!!!

evergreen22:
I'm pulling this into my setup for testing. Nice work!

geekbozu:
So this is a fix but I do not think it actually fixes the bug in the code. 
It seems in the original commit of this rfmgmt code some bugs were written. 
Mostly its a casting issue.

geekbozu:
[pinetime-mcuboot-app-dfu-1.4.99.zip](https://github.com/JF002/InfiniTime/files/7230552/pinetime-mcuboot-app-dfu-1.4.99.zip)
Test Build for those wanting to test while we wait for the workflow :)

tmilburn:
I think the proper fix would be to check `ble_ll_rfmgmt_is_enabled` and not call `ble_ll_rfmgmt_enable_now` to do a comparison if the RF is currently enabled.

geekbozu:
> 
> 
> I think the proper fix would be to check `ble_ll_rfmgmt_is_enabled` and not call `ble_ll_rfmgmt_enable_now` to do a comparison if the RF is currently enabled.

That would mostly work, assuming  `ble_ll_rfmgmt_is_enabled` except its hardcoded to return true when you have an RF enabled time of 0. Which is what we have. 

https://github.com/JF002/InfiniTime/blob/d7ca217872307357200e32ad679786d8f3396ca0/src/libs/mynewt-nimble/nimble/controller/include/controller/ble_ll_rfmgmt.h#L55

The case we are running into has to do with the fact that our XTAL for our radio never sleeps so we do not have a "warm up time" for the radio. Which is what rfmgmt is meant to handle. So we are falling onto the "always available" case. Which did not handle the `ble_ll_rfmgmt_enable_now` correctly and was returning 0 Instead of the next earliest time the radio would be available.

JF002:
@danielgjackson Thank you so much for helping us on these issues! What you found is really interesting, and I guess many people (me included) will test your PR soon ;)
Finger crossed that it'll fix at least one of our BLE issues :+1: 

I'm really curious about the feedback from nimble team [here](https://github.com/apache/mynewt-nimble/pull/1027)!

hubmartin:
I've loaded that to the PineTime and Colmi P8. I'll check it tomorrow. Thanks a lot!

danielgjackson:
@JF002 Thanks for such a fantastic project -- I'm happy to help.  This was quite a tricky issue to track down, so I was glad to finally spot the issue.

> I'm really curious about the feedback from nimble team [here](https://github.com/apache/mynewt-nimble/pull/1027)!

It seems to have been accepted very quickly!  I still wonder why it doesn't seem to have come up as an issue elsewhere (I think it might also [affect scanning](https://github.com/apache/mynewt-nimble/blob/master/nimble/controller/src/ble_ll_scan.c#L1217)).  Perhaps there are not many users of NimBLE without the RF Management?

geekbozu:
> 
> 
> @JF002 Thanks for such a fantastic project -- I'm happy to help. This was quite a tricky issue to track down, so I was glad to finally spot the issue.
> 
> > I'm really curious about the feedback from nimble team [here](https://github.com/apache/mynewt-nimble/pull/1027)!
> 
> It seems to have been accepted very quickly! I still wonder why it doesn't seem to have come up as an issue elsewhere (I think it might also [affect scanning](https://github.com/apache/mynewt-nimble/blob/master/nimble/controller/src/ble_ll_scan.c#L1217)). Perhaps there are not many users of NimBLE without the RF Management?

This would be my guess. RF management is enabled by setting an xtal settle time..which defaults to 1500us for most platforms...we do not have a settle time it seems? Hence our lack of using it!

Riksu9000 (approved):
Able to connect at 23h. Nice!

hubmartin:
19 hours passed and PineTime and P8 still advertising. Awesome!

clemensvonmolo:
Sadly, this doesn't seem to be the entire story. I flashed the 1.4.99 build yesterday evening and right then Bluetooth worked like it always does after a reset, but after 18.5 hours of uptime Gadgetbridge refuses to reconnect (might be even earlier, didn't try for quite some time). My phone had Bluetooth disabled in the meantime if it makes a difference. 
I have already checked/tried:

- the commit hash on my watch matches this PR
- toggle Bluetooth, GPS, WiFi and Cellular on my phone
- reboot my phone
- unpair the watch in Gadgetbridge

I don't own a Bluetooth sniffer and I can't seem to get Bluetooth logs working on my phone without rooting it, so I'm unable to test if it is an issue other than advertising packets.

Avamander:
@clemensvonmolo 

This aims to fix advertising after the deadline, not reconnection issues after bonding. That's a separate PR, check it out.

clemensvonmolo:
I am not going to pretend that I know what I'm talking about, but after reading about BLE advertising it looks to me as if the watch still needs to advertise to allow connections, even if it knows the device already. I don't know if reconnect means something different in BLE land (like sending data after waking up from sleep?), but I meant pressing the connect button in Gadgetbridge after Bluetooth was disabled for a long time and there was no active or sleeping connection in that time. I also can't seem to find the PR you mentioned, only some GitHub issues.

PR #698, opened 2021-09-28
Alarm: Display the time until alarm, even if it is off

maksalees:
Previously, if the alarm was turned off, when you pressed the information button, a message was displayed that the alarm was not set. The amount of time until the alarm is now displayed as well as if it is on.
What for? Allows you to save two clicks on a button if you just need to see how much time is left before the alarm

## Before
Before

## After
After

### DIFF
--- a/src/components/alarm/AlarmController.cpp
+++ b/src/components/alarm/AlarmController.cpp
@@ -19,7 +19,6 @@
 #include "systemtask/SystemTask.h"
 #include "app_timer.h"
 #include "task.h"
-#include <chrono>
 
 using namespace Pinetime::Controllers;
 using namespace std::chrono_literals;
@@ -48,10 +47,7 @@ void AlarmController::SetAlarmTime(uint8_t alarmHr, uint8_t alarmMin) {
   minutes = alarmMin;
 }
 
-void AlarmController::ScheduleAlarm() {
-  // Determine the next time the alarm needs to go off and set the app_timer
-  app_timer_stop(alarmAppTimer);
-
+std::chrono::time_point<std::chrono::system_clock, std::chrono::nanoseconds> AlarmController::TimePoint() {
   auto now = dateTimeController.CurrentDateTime();
   alarmTime = now;
   time_t ttAlarmTime = std::chrono::system_clock::to_time_t(alarmTime);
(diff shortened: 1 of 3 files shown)


src/components/alarm/AlarmController.cpp
```diff
@@ -48,10 +47,7 @@ void AlarmController::SetAlarmTime(uint8_t alarmHr, uint8_t alarmMin) {
   minutes = alarmMin;
 }
 
-void AlarmController::ScheduleAlarm() {
-  // Determine the next time the alarm needs to go off and set the app_timer
-  app_timer_stop(alarmAppTimer);
-
+std::chrono::time_point<std::chrono::system_clock, std::chrono::nanoseconds> AlarmController::TimePoint() {
```
Riksu9000, line 50: This returns the timepoint of the alarm right? Maybe a name like `CalculateAlarmTimePoint()` would be more descriptive.

src/components/alarm/AlarmController.h
```diff
@@ -32,6 +33,7 @@ namespace Pinetime {
 
       void Init(System::SystemTask* systemTask);
       void SetAlarmTime(uint8_t alarmHr, uint8_t alarmMin);
+      std::chrono::time_point<std::chrono::system_clock, std::chrono::nanoseconds> TimePoint();
```
Riksu9000, line 36: Does this need to be public?

Riksu9000 (commented):
Looks like the alarmTime member might not be needed anymore and could be replaced with the new function. What do you think?

src/components/alarm/AlarmController.cpp
maksalees: Fixed

src/components/alarm/AlarmController.h
maksalees: Fixed

maksalees:
> Looks like the alarmTime member might not be needed anymore and could be replaced with the new function. What do you think?

I agree, I just forgot to fix it :slightly_smiling_face:

src/components/alarm/AlarmController.cpp
```diff
@@ -79,15 +74,25 @@ void AlarmController::ScheduleAlarm() {
...
   state = AlarmState::Set;
 }
 
 uint32_t AlarmController::SecondsToAlarm() {
-  return std::chrono::duration_cast<std::chrono::seconds>(alarmTime - dateTimeController.CurrentDateTime()).count();
+  if (state == AlarmState::Set) {
+    return std::chrono::duration_cast<std::chrono::seconds>(CalculateAlarmTimePoint() - dateTimeController.CurrentDateTime()).count();
+  } else {
+    return std::chrono::duration_cast<std::chrono::seconds>(CalculateAlarmTimePoint() - dateTimeController.CurrentDateTime()).count();
+  }
```
Riksu9000, line 95: This if statement isn't doing anything anymore.

maksalees: Fixed

mruss77:
Looks good to me!

Itai-Nelken:
I have been uaing this pr for some time, works fine and makes using the alarm for measuring time until something easier (don't have to turn alarm on every time).

maksalees:
Will this PR be merged?

FintasticMan:
Well at the moment it is based on quite an old revision of the `develop` branch, and it also has some conflicts. If you can rebase your branch onto the latest `develop`, it would be easier to see if it can be merged.

Riksu9000:
@maksalees I'm sorry it's been a long time. Probably the reason why this hasn't been merged is that we're unsure about what's the preferred behaviour. Indeed rebasing this will allow us to take another look at this.

Issue #703, opened 2021-09-29
[Feature Request] Reorder/hide apps


adocampo:
I think all apps are great, and I thank all devs for creating such a great piece of software.

But being realistic, for me there are apps more interesting than others, so I don't want to scroll down (now two pages) to configure an alarm while navigation, which I don't even know how to use it is on the first page. 

This will become even more frustrating as more apps will be on the watch, so I think a way to reorder or hide/show apps will be a nice to have feature.

calhix:
Having the ability to hide apps and move them around would be huge, out of the 12 I only find 6 useful in my case and they are all over the place.

dyamon:
I think one of the big todos atm is the ability to dinamically load apps into the watch without the need to recompile the firmware. The decoupling of firmware and apps will most likely solve the request of hiding apps, so I don't think it is wise to implement some patch code to simulate this now.

The reordering feature on the other hand is nice to have, and we would be able to reuse this code later on as well. That said, afaiu the app menu is really two different pages of apps. To simplify the task we would probably need to redesign the menu code.

In the meantime, if you are able to compile the firmware yourself (can help with that as well), a temporary solution would be to modify the code here:

https://github.com/JF002/InfiniTime/blob/e9bb0b3cdd0e1b95b7bfdcfb94db5a6a0966eed7/src/displayapp/screens/ApplicationList.cpp#L40-L66

The two lists of 6 elements determine the order of the apps, so moving a line around will result in a different order.

Changing one of the elements in the array to

```
{"", Apps::None},
```

will "disable" the app, resulting in an empty spot in the menu.

adocampo:
> The reordering feature on the other hand is nice to have, and we would be able to reuse this code later on as well. That said, afaiu the app menu is really two different pages of apps. To simplify the task we would probably need to redesign the menu code

What pebble did was let the phone companion app reorder (well, and install and uninstall) the app list, so it synced the desired order back to the watch. I think that's quite smart as it let the footprint of the firmware quite small,  if there is an "official" InfiniTime companion app, I doubt Gadgetbridgo or Siglo can have this in their roadmap. 
Without a proper companion app, it should be done via the watch itself, and as I don't know the internals at all, I can just suggest a semi-complex settings page where all the apps are listed and a number on the side indicating the position. (I don't even know if there is a drag and drop capability or if it would even be practical due to the screen size)

[...]

PR #708, opened 2021-09-30
Feature: Favorite app or last used app when swiped left on the watch face

coxtor:
This is a cleaner reworked version of #487 and adds a favorite option or opens the last used app when swiping left on the watch face.

https://user-images.githubusercontent.com/7712562/136454854-4242d885-1cc8-466e-9147-b04e1c783750.mp4

1. Hold an app to select/deselect it as a favorite app
2. Swipe left on the watch face to open the favorite app
3. If no favorite app is defined, the swipe left gesture opens the last used app


kieranc:
Thanks very much for recreating this - sadly after some discussion there might be a better approach (Issue #703) however I'd love to help you find out why it doesn't work right as a learning exercise :-) (I don't know either, we can both learn!)

src/displayapp/screens/settings/SettingFavoriteApp.h
```diff
@@ -0,0 +1,29 @@
...
+        SettingFavoriteApp(DisplayApp* app, Pinetime::Controllers::Settings& settingsController);
+        ~SettingFavoriteApp() override;
+        void UpdateSelected(lv_obj_t* object, lv_event_t event);
+
+      private:
+        Controllers::Settings& settingsController;
+        uint8_t optionsTotal;
+        lv_obj_t* cbOption[13];
+        Apps favoriteOptions[13] = {Apps::Music,Apps::Navigation,Apps::StopWatch,Apps::Timer,Apps::None,Apps::FlashLight,Apps::Paint,Apps::Paddle,Apps::Twos,Apps::HeartRate,Apps::Metronome,Apps::Motion,Apps::Steps};   
+        char favoriteNames[13][12] = { "Music","Navigation","Stop Watch","Timer","None","Flash Light","Paint","Paddle","Twos","Heart Rate","Metronome","Motion","Steps"};
```
tmilburn, line 25: Is this list in any particular order? I would expect "None" to be at the top or bottom.

src/displayapp/screens/settings/SettingFavoriteApp.h
```diff
@@ -0,0 +1,29 @@
...
+      class SettingFavoriteApp : public Screen {
+      public:
+        SettingFavoriteApp(DisplayApp* app, Pinetime::Controllers::Settings& settingsController);
+        ~SettingFavoriteApp() override;
+        void UpdateSelected(lv_obj_t* object, lv_event_t event);
+
+      private:
+        Controllers::Settings& settingsController;
+        uint8_t optionsTotal;
+        lv_obj_t* cbOption[13];
```
tmilburn, line 23: Should be a list to avoid lots of magic numbers.

src/displayapp/screens/Tile.h
```diff
@@ -49,6 +52,7 @@ namespace Pinetime {
 
         const char* btnmMap[8];
         Pinetime::Applications::Apps apps[6];
+        Pinetime::Applications::Apps currentFavoriteApp = Pinetime::Applications::Apps::None;
```
tmilburn, line 55: This isn't required since the value always fetched from the settings

src/displayapp/screens/Tile.cpp
```diff
@@ -123,9 +133,25 @@ void Tile::UpdateScreen() {
...
+void Tile::OnValueChangedEvent(lv_obj_t* obj, uint32_t buttonId, lv_event_t event) {
   if(obj != btnm1) return;
+  switch(event){
+    case LV_EVENT_VALUE_CHANGED:
+        app->StartApp(apps[buttonId], DisplayApp::FullRefreshDirections::Up);
+        running = false;    
+      break;
+    case LV_EVENT_LONG_PRESSED:
+      currentFavoriteApp = settingsController.GetFavoriteApp();
+      if(buttonId){
```
tmilburn, line 145: buttonId is the value from 0 to 5 of the current button on screen so this won't allow selection of the first button

coxtor:
> Thanks very much for recreating this - sadly after some discussion there might be a better approach (Issue #703) however I'd love to help you find out why it doesn't work right as a learning exercise :-) (I don't know either, we can both learn!)

I have been thinking about this too - however I do think that having a favorite quick to access app still is beneficial. I'll try to setup a proper debugging setup on the weekend, right now I am programming blindly 😅.

tmilburn (changes requested):
Overall structure of the code looks good but needs a few fixes.

tmilburn:
I like the idea but I feel that from a usability point of view it may be beneficial to use the swipe left to go the last opened app.

coxtor:
Thank you for looking at the code. Well if the majority feels that the swipe gesture should take you back to the last used app, there is no point on continuing work on this PR? Or would it make sense to make this feature configurable. However, that might clutter the settings long term.

src/displayapp/screens/Tile.cpp
coxtor: I realize that, however, all that is supposed to be done is to save the selected application as favorite - basically the same parameters are used to actually launch the application on a single tap. I don't understand how saving the application would be any different, as the types are ( or should be ) the same.

Riksu9000:
There's an issue inherent to swiping left. The transitions get confusing.

https://user-images.githubusercontent.com/37774658/135601438-ca022571-dbb8-4842-8295-341365e7ab82.mp4

If the screen moves right, it should move back left at some point to reach the clock. The same with going up. If the return gesture was a swipe right, then that would make return gestures inconsistent across the system. Maybe the button would be a better shortcut after all? #480

issue

Also replacing the app icon with an asterisk isn't the best way to display the favorite app. Instead the button color could be changed, or the star could be added to a corner, but without replacing the existing icon.

0x1a8510f2:
> Thank you for looking at the code. Well if the majority feels that the swipe gesture should take you back to the last used app, there is no point on continuing work on this PR? Or would it make sense to make this feature configurable. However, that might clutter the settings long term. 

I personally prefer this PR over it being the last app, but I'm also for making this configurable. That could be as simple as including a "last used app" at the top of the favourite app selector.

hubmartin:
@coxtor Tried it and it is awesome, thanks.
Regarding the swiping/screen moving logic. I don't mind if the swiping logic to close app (swipe down) and moving screen (to the left) is not in the same direction. But exiting the app should go to the watchface and not the menu. Could this be somehow improved? :+1:

coxtor:
Thanks for the feedback so far. 
@hubmartin I have a private seperate branch, where the function of the pushbutton can be configured implementing exactly what you have asked for. 
Right now I am not sure how much time I should continue to invest as the opinions are controversial on how or if this should be implemented.

hubmartin:
I feel that this PR is perfect and already usable. The only small issue I see is the exiting from the Favorite application which should go back to the watchface when you press the button or swipe down in the apps, that supports swipe to exit.

I played with it for a little bit and It seems that with using `ReturnApp` you can do it easily to go back to `Apps::Clock`.

```
case TouchEvents::SwipeLeft:
  favoriteApp = settingsController.GetFavoriteApp();
  if (favoriteApp == Apps::None)
    LoadApp(Apps::SettingFavoriteApp, DisplayApp::FullRefreshDirections::LeftAnim);
  else
    LoadApp(favoriteApp, DisplayApp::FullRefreshDirections::LeftAnim);
  ReturnApp(Apps::Clock, FullRefreshDirections::RightAnim, TouchEvents::SwipeDown);
  break;     
```

I would keep this PR simple. Adding long press in app menu to select favorite app or highlightning could be improved later in another PR.

coxtor:
> @coxtor Tried it and it is awesome, thanks. Regarding the swiping/screen moving logic. I don't mind if the swiping logic to close app (swipe down) and moving screen (to the left) is not in the same direction. But exiting the app should go to the watchface and not the menu. Could this be somehow improved? +1

@hubmartin checkout PR #717

Riksu9000:
> > @coxtor Tried it and it is awesome, thanks. Regarding the swiping/screen moving logic. I don't mind if the swiping logic to close app (swipe down) and moving screen (to the left) is not in the same direction. But exiting the app should go to the watchface and not the menu. Could this be somehow improved? +1
> 
> @hubmartin checkout PR #717 

The issue can be seen in the video I posted. When I swipe down, it goes to the app menu instead of the watch face. The issue isn't the button, but `ReturnApp()`.
https://github.com/InfiniTimeOrg/InfiniTime/pull/708#issuecomment-932104989

hubmartin:
I agree with @Riksu9000 that #717 does not solve it. It is not a configuration issue of the button but program issue which depends:
- if you run the app from the menu - then exiting app should go back to the menu
- if you run app by swipe left shortcu - then exiting the app should go back to the wachface (which is the only place you can do that swipe left gesture and run favorite app)

For the #717 I'll reply to that thread, since it really doesn't apply to this Favorite app thread.

coxtor:
I see, now I understand what the issue was. The current changes should fix the behavior making #717 obsolete. 
@Riksu9000 for now the favorite is indicated with icon + * . Unfortunately the button matrix does not allow the styling of individual buttons. An option would be to use individual buttons at the cost of more memory consumption, which is not the best option imho. 

Still to do:
1. remove the settings pane in favor of the "long press to select favorite" option. For this I need to understand why saving the favorite with the buttonid used to launch the app causes a crash - here advise would still be welcome
2.  if the setting page is removed  - a nice option to select "last used app" as favorite is needed using the long press selection method. Any inspiration here?

hubmartin:
I'm still not sure that favorite app should have any highlight in the menu. I'm thinking that #480 has also shortcuts like button doubleclick for starting app - would you think this also needs to be somehow highlighted in the menu? What about looong button press in the same PR - now we would have 3 highlighted apps in the menu.

But this brings me the idea - what if you long press on the item in the menu (or in the Settings) and "dialog" appears and will contain this:

Set current app as:
- Favorite
- Button Doubleclick 
- Button Loong press
- (Hide app from the menu for future? Or from the FLASH when we will be able to load them from external FLASH)

This might be really clean solution and we:
- get rid of separate settings selections UI
- we wouldn't duplicate code and use more FLASH
- it could be used for items in settings menu. Which could be awesome since I change Wakeup or Display timeouts quite often :)

What do you think?

Riksu9000:
>Unfortunately the button matrix does not allow the styling of individual buttons

It does allow styling by setting the state to checked at least.
https://docs.lvgl.io/latest/en/html/widgets/btnmatrix.html#simple-button-matrix

>if the setting page is removed - a nice option to select "last used app" as favorite is needed using the long press selection method. Any inspiration here?

When nothing is selected, it could go to the last used app.

I like @hubmartin's idea for a new menu on long pressing an app icon. The longer press button action should probably be saved for a power menu, but this makes sense for the double click, and maybe another shortcut for regular long press on the watch face?

hubmartin:
> saving the favorite with the buttonid used to launch the app causes a crash - here advise would still be welcome

Can you point me to the place in the code where I should start investigate that? Whether is it settings or somewhere in the app list? Not familiar with that areas yet.
Do you have a debugger?

coxtor:
> > saving the favorite with the buttonid used to launch the app causes a crash - here advise would still be welcome
> 
> Can you point me to the place in the code where I should start investigate that? Whether is it settings or somewhere in the app list? Not familiar with that areas yet. Do you have a debugger?

I have a stlink-v2 and the pinetime dev kit, however I haven't managed to set it up yet. 

It would be great to fix the crash to have an initial implementation before we start with the modal version with the new button handler. 

The crashing line is in src/displayapp/screens/Tile.cpp line 153. From what I can understand without a debugger I am using the exact same expression to store the favorite app as the one used to start the app (apps[buttonId]), the datatype is identical. If I use a static number as index like apps[2], it works without a problem. So something is up with buttonId. 


``` case LV_EVENT_VALUE_CHANGED:
        app->StartApp(apps[buttonId], DisplayApp::FullRefreshDirections::Up);
        running = false;    
      break;
    case LV_EVENT_LONG_PRESSED:
      currentFavoriteApp = settingsController.GetFavoriteApp();
      if(buttonId){
        if(currentFavoriteApp == apps[buttonId]){
          settingsController.SetFavoriteApp(Apps::None);
        }
        else{
          settingsController.SetFavoriteApp(apps[buttonId]); <---------------------------------- This line is the issue, where as apps[2] works fine.
        }   ```

coxtor:
@Riksu9000  thanks for pointing that out, it looks a lot better highlighting the favorite using the active select state, also now the last app used app is loaded, if no favorite is set.

Finally, if @hubmartin manages to understand what the crashing issue is it should be fine until we refine this to work with the updated button handler. As for that - is there already an implementation that can share data with other screens or would this need to be implemented too ( sending and retrieving information from the dialog window)

coxtor:
Okay, I have managed to setup my debugging environment and it appears that the pressed button is only omitted on the event "LV_EVENT_VALUE_CHANGED" , when this event is triggered, all values are as expected. 
However, when a different event is omitted, a random large int is omitted, thus not representing the correct index. Does anybody know how to go about this issue?

Riksu9000:
> Okay, I have managed to setup my debugging environment and it appears that the pressed button is only omitted on the event "LV_EVENT_VALUE_CHANGED" , when this event is triggered, all values are as expected. However, when a different event is omitted, a random large int is omitted, thus not representing the correct index. Does anybody know how to go about this issue?

Oh interesting. It must pass some other data as the event data for some reason. You should be able to use `lv_btnmatrix_get_active_btn()` instead to get the index.

hubmartin:
Tested, long press in app menu works fine. :+1: 

Please apply the `ResturnApp` suggestion from my previous post. Music app returns correctly but Stopwatch goes to the menu when I close it even I run it by swipe left from the watchface.
https://github.com/InfiniTimeOrg/InfiniTime/pull/708#issuecomment-933002880

Also is it really necessary to have the Favorite app highlighted in the menu? I don't see this information useful. If I forgot what the app is, I just swipe in the watchface to see it. It seems like it could confuse people. I understand that now second long press deactivates the app and choose the last opened. It would be great to try to think other solution to this.

hubmartin:
Can you please fix a bug?
 - after reboot when I do swipe left, then no last app was set and screeen stays on the Watchface, howvere when I swipe up for the app menu, then instead of slide-up animation the menu displays with left to right animation.
- (when selecting paint app, then the whole screen shifts a few pixels up. This seems like an error of Pain app and I'm sure I've seen it somewhere in issues)

coxtor:
@hubmartin thanks for the review! I have squashed your mentioned bugs.

- However, the offset on screen with paint and the exit animation inconsistenies of some apps shouldn't be part of this PR. 
- regarding the indicator of the current favorite app, I don't mind whether there is an indicator or not - @Riksu9000 proposed it and I liked the idea, I think it may be good to have some indication on what may be going on. 

Feel free to retest if you like .

hubmartin:
@coxtor Thanks for improvements. I've tested it a lot now and it seems absolutely perfect. :+1:

hubmartin:
(@coxtor not trying to make this PR more complicated, just a curious question how hard would it be to add support for settings items? Items like Watch face selection, Wake up, Display timeout, Battery would be also useful to choose for swipe left gesture.)

hubmartin:
Hmm, I looked deeper. First I thought that settings is separate, but if you use this enum then they seems to be together with apps. Hope I'm right.
https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/src/displayapp/Apps.h#L30

hubmartin:
By `where to put it` you mean exactly what? It will use the same swipe left gesture to run the favorite app/settings. You choose favorite app or favorite settings (so now it will be like favorite action, not just app). And you choose it by long press in the settings menu.

coxtor:
Ah now I understand, so you mean basically adding the hold to set favorite option in the settings menu too? That can be easily be done.

hubmartin:
@coxtor Absolutely awesome! Thanks for adding.
Noticed just small issue. When I choose one option in the settings, then in becomes green. Then when I choose item on the same screen, then the new selected settings becomes green, but the previous item stays green too. When I scroll to the next settings page and back then only the last item is correctly highlighted. Seems like some refresh is missing? In the App menu there is no such issue.
Thanks for these improvements!

hubmartin:
One small thing. There is no vibration when the item in Settings is selected. I might not be needed because items are big and green selection can be seem perfectly fine, but for consistency I would add it too.

I love how fast I can change faces! :)

https://user-images.githubusercontent.com/497212/136262859-e07fce3f-01f0-474c-87ed-3d954d91675e.mp4

coxtor:
@hubmartin thanks for highlighting those bugs! Should be ready to merge now ;) @JF002 ?

Riksu9000:
>> Which apps should be considered for previous app? I'd think only the apps in the app menu, but right now it could open many other screens.
>
>The initial implementation was exactly like this, but some users may want to be able to have the flexibility to use any app? My mind is not set on that topic, I guess the majority should decide? We could also decide on a whitelist of some sort?

Not favorite app, but previous app. It's was very unexpected for me when it went to the settings list and flashlight, because they don't seem like regular apps from a users perspective.

>> Then I think the issue I mentioned in #708 (comment) should still be considered, because weird transitions will be disorienting. I wonder if it would be bad if the return animation was in a different direction than the swipe. Nevertheless this will create some inconsistency, and using the button instead of a swipe left as the shortcut doesn't have these issues, but might not be quite as convinient? #480.
>
> I thought I already have considered that comment - the return animation takes you back from the previous swipe. Could you perhaps make more clear what you would like to see. I don't understand the issue here as I think it integrates well with the current integrations.

The issue is still similar to what happens in the video when swiping down, the app menu wasn't the main issue, and the drawing is still relevant https://github.com/InfiniTimeOrg/InfiniTime/pull/708#issuecomment-932104989. I've suggested using the button instead as the shortcut, which would require #480. The issue is inherent to the current design and can't easily be fixed otherwise.

I just noticed the previous app can also be the notification, which it really shouldn't be.

lman0:
any way to review this very good pull (@Avamander  or @geekbozu )?
because it feel like that the change requested by @Riksu9000 is more a personal preference than an actual blocker

Riksu9000:
The issue is real. Whether we choose to ignore it or not is the question. I don't think it would be very professional of us to simply ignore it.

I'm sure this feature can be useful and I hope we get to test these shortcuts with #782 button doubleclick action.

kieranc:
I tested this PR briefly and got confused at first because I didn't read how it was meant to work and it was just launching seemingly random stuff when I swiped left. We have to assume the users will also not read about how it works so I don't think this is ideal behaviour, maybe if no favourite is set, it could display a message explaining how it works? 
This is a nice feature, it works, but I can't help but feel maybe just the ability to reorder the app list would solve the problem in a cleaner way. I can't speak about the animation issue, I haven't used it enough, but I would like to get this into a test build and get some user feedback.

coxtor:
I'm not quite sure where to go from here. I'm glad to continue the discussion on this but I'll continue to work on this feature once we all are clear on what is the best solution.

KoalaV2:
Been running this for a while now and it runs great. Although it does feel like when going into the favorite app that swiping right should bring me back to the clock.

Riksu9000:
I'll close this as there are still unresolved issues and the usefulness of this feature is questionable. I've never felt that opening an app with two gestures is too much work. A way to remap double-clicking the button *could* be useful, since it might be even faster than a swipe.

leroivi:
Can we reconsider this kind of feature now that we have reached a 3rd page of application ?
A few gesture might not feel like much work when you are setting some alarm or any application that you will focus on, but when you just want to skip your current music I feel like it's a bit heavy. So I guess it depends on the target application and it is a good thing to let the user configure which app deserve a shortcut.

Ronkn:
Really hope this is reconsidered.

I'd almost prefer to have swipe left be the active or last used app (calulator, music player control, etc) for quick access. Right now, with music playing if I want to pause or play from the watch, I have to scroll down and select the music player. If I'm using it a frequently (it's consistently the last used app), would be nice to have that on swipe right menu.

I do like the idea of a screen to have favorite apps stored. I'd rather see the swipe right menu (flashlight, settings, brightness, notifications) customizable and to put favorite apps there. I don't ever use any of the icons. It's just a wasted swipe. Make the current menus on this screen appear in the "apps" screens. So if someone likes the way it is currently, they can make those their favorites again. 

That's my 2 cents anyway.

PR #711, opened 2021-10-02
Remove static declaration on batteryValue preventing read attribute from updating.

geekbozu:


xan-m:
Solved the problem for me! Battery read attribute is reporting correct values now

JF002:
Good catch! I wonder why this variable was declared as static!

Issue #693, opened 2021-09-27
Touch, button & power detection doesn't work when button is hold during boot


hubmartin:
I've noticed similar behavior also on P8 and now on PineTime, both with 1.4.0 Infinitime FW. It is hard to reproduce. Sometimes watch is frozen (cutting-edge PRs or something else maybe?), screen black and button does not react. I try to reboot by holding button. After reboot, the screen is on but the watch did not vibrate and button, touchscreen (touch or doubletap to wake up) or accelerometer (wake on wrist) does not work. After timeout the screen goes dark. Sometimes I need to make like 10 reboots. 
Other way to fix it is to short-circuit 3.3V rail for a while, or disconnect and reconnect the battery.

This happend few times to me. 

There was #566 which tried to fix it but the problem is that even after 8/16 SCL pulses the sensor still holds SDA low, which blocks the `Start condition` on the I2C bus. The sensor should give up after 1 byte transfer since it does not get ACK, but it seems is does not release the bus and needs more clocks.

You can read deeper info and scope signals when this happened on my P8 in the #566 . I thought that happened because of special accelerometer on P8, but now it happened also on PineTime.

Here I have video of like 13 resets which fixed that.
https://photos.app.goo.gl/j5HsZEDtwyau68A37

For now I have all my watches sealed, it would be great if someone could try this code devkit and test how it behaves if you ground SDA. I did some tests on my sealed watches but it seems like something more could be wrong with the code, maybe GPIO configuration. I wasn't able to get into `while (nrf_gpio_pin_read(Pinetime::PinMap::TwiSda) == 0)` loop.  Any ideas are welcome.

```
void declock_i2c()
{
    // I2C declocking issue
    nrf_gpio_cfg_input(Pinetime::PinMap::TwiSda, NRF_GPIO_PIN_PULLUP);

    volatile int i = 0;

    while (nrf_gpio_pin_read(Pinetime::PinMap::TwiSda) == 0)
    {
        nrf_gpio_cfg_output(Pinetime::PinMap::TwiScl);
        nrf_gpio_pin_clear(Pinetime::PinMap::TwiScl);
        nrf_delay_ms(2);
        nrf_gpio_cfg_input(Pinetime::PinMap::TwiScl, NRF_GPIO_PIN_PULLUP);
        nrf_delay_ms(2);

        if (++i > 100)
        {
            NVIC_SystemReset();
        }
    }

}
```

[...]

PR #715, opened 2021-10-03
Fix GPIOTE not calling IRQ when button is held during boot

hubmartin:
Should solve Issue #693 

Any feedback and testing is welcomed.

### DIFF
--- a/src/systemtask/SystemTask.cpp
+++ b/src/systemtask/SystemTask.cpp
@@ -158,19 +158,21 @@ void SystemTask::Work() {
   heartRateSensor.Disable();
   heartRateApp.Start();
 
-  nrf_gpio_cfg_sense_input(PinMap::Button, (nrf_gpio_pin_pull_t) GPIO_PIN_CNF_PULL_Pulldown, (nrf_gpio_pin_sense_t) GPIO_PIN_CNF_SENSE_High);
+  // Button
   nrf_gpio_cfg_output(15);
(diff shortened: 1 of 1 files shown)


Riksu9000:
I wonder why the issue happened in the first place. If the events don't go through while the button is pressed, then why didn't releasing the button fix it?

If the `nrf_gpio_cfg_sense_input()` functions aren't necessary, would it make sense to remove the last one as well, so all the GPIO initializations are more similar?

hubmartin:
>If the nrf_gpio_cfg_sense_input() functions aren't necessary, would it make sense to remove the last one as well, so all the GPIO initializations are more similar?

Do not fix it if ain't broken :) Maybe I could. But I guess there should be added a check for touch IRQ level in the GPIOTE ISR. So it triggers touch event reading only once and not on both IRQ gpio edges. What is your opinion?

JF002:
@Riksu9000 

> I wonder why the issue happened in the first place. If the events don't go through while the button is pressed, then why didn't releasing the button fix it?

There is a small note in the nrf52832 datasheet saying that the pin should not be in a state that would trigger the event when enabling it:
image

iirc I had this issue with the power detection too, and that's probably why I wrote https://github.com/InfiniTimeOrg/InfiniTime/commit/9ab298c09e273479822d10aad9f7bfe1d287ce75.

Your fix works fine for me, and it looks good to me, thanks ;)

PR #717, opened 2021-10-03
Add configurable push button function (back or watchface)

coxtor:
This PR adds the option to configure the action of the pushbutton (back or watchface
PXL_20211003_185333721
).

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -433,6 +433,7 @@ list(APPEND SOURCE_FILES
         displayapp/screens/settings/SettingDisplay.cpp
         displayapp/screens/settings/SettingSteps.cpp
         displayapp/screens/settings/SettingPineTimeStyle.cpp
+        displayapp/screens/settings/SettingPushButtonAction.cpp
 
         ## Watch faces
         displayapp/icons/bg_clock.c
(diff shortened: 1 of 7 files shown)


Riksu9000:
#480 already implements this without needing the option. Tap for back and long press 400ms for watchface.

coxtor:
The button handler is nice, however the user should still have the option to decide what happens when the button is actually pushed. In my opinion these prs implement different functions and do not exclude themselves, or am I missing something?

Riksu9000:
> The button handler is nice, however the user should still have the option to decide what happens when the button is actually pushed. In my opinion these prs implement different functions and do not exclude themselves, or am I missing something?

As I said, in #480 long pressing the button goes to the watch face, while a short press still goes back a single step. There's no reason to choose only one action, when you can have both at the same time, right?

hubmartin:
I agree with @Riksu9000. I use #480 for a long time and it has both options all solved by a single button and no config (plus doubleclick app and loooong press app). His PR also solves lot of other issues we have with debouncing, fast wake up etc.
When the #480 will be merged, then this configuration will not make a sense, otherwise short click will go to watchface and longer press will also go to watchface.

coxtor:
I see, under those circumstances we can consider this pr obsolete :) .

hubmartin:
@coxtor Sorry for that. I've been threre too. There's a lot of pull requests and lot of them are tangled together or "hidden" on the second page. However JF002 is working really hard to merge them together so it will be less cluttered. On the other side, more people are contributing which is also great :) Thanks for your contribution anyway. :+1:

PR #719, opened 2021-10-04
Improve battery percentage calculation and reporting

Riksu9000:
While charging, percentage should only go up, and while discharging, percentage should only go down.

This is simpler than a proper filter and achieves practically the same thing. Should reduce, or even eliminate the noise that's visible in Gadgetbridge battery graph.

Also notify the battery level over ble only when it has changed. This should make the graph smoother.

I would have liked to test this more and provide screenshots, but I want to work on other things as well and not wait for the battery to drain.


geekbozu:
I can test this later tonight with my devkit, I have an external Power supply that i can change the voltage on so no need to wait for a battery.

JF002:
I agree with @geekbozu, the battery level measurement and graph are getting better and better thanks to you, @Riksu9000, thanks!

PR #720, opened 2021-10-04
Improve gesture consistency across the firmware

### LINKED ISSUE #651: Inconsistent interface in InfiniTime 1.4.0
In most components of the watch, a simple slide down will bring the watch to the previous screen. However, the MUSIC and the TORCHLIGHT app does not behave that way.

Riksu9000:
Metronome can be closed by swiping down from the center. I could make it not close while dragging the arc, but I couldn't make it not close while dragging the dropdown menu, so this is the next best thing.

Music app shows volume buttons on swipe up and swiping down goes back to next/prev. now swiping down in the default state will close the app.

Flashlight can now be closed by swiping down. Also the transitions animation was changed to match other apps and returnapp is quicksettings instead of watchface.

The last apps that can't be closed through the touchscreen are Paint, Paddle and Twos, but they can't easily be made to use swipes for closing, so they would need some other solution.

Closes #651, unless we keep it open for these three.

https://user-images.githubusercontent.com/37774658/135816670-bd3f3c3a-bd4f-43f5-86ec-e5b1c018c050.mp4

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -380,7 +380,7 @@ void DisplayApp::LoadApp(Apps app, DisplayApp::FullRefreshDirections direction) 
       break;
     case Apps::FlashLight:
       currentScreen = std::make_unique<Screens::FlashLight>(this, *systemTask, brightnessController);
-      ReturnApp(Apps::Clock, FullRefreshDirections::Down, TouchEvents::None);
+      ReturnApp(Apps::QuickSettings, FullRefreshDirections::Down, TouchEvents::SwipeDown);
       break;
     case Apps::StopWatch:
       currentScreen = std::make_unique<Screens::StopWatch>(this, *systemTask);
--- a/src/displayapp/screens/Metronome.cpp
+++ b/src/displayapp/screens/Metronome.cpp
@@ -113,9 +113,15 @@ void Metronome::OnEvent(lv_obj_t* obj, lv_event_t event) {
           lv_label_set_text_fmt(bpmValue, "%03d", bpm);
         }
         tappedTime = xTaskGetTickCount();
+        allowExit = true;
       }
       break;
     }
+    case LV_EVENT_RELEASED:
+    case LV_EVENT_PRESS_LOST:
+      if (obj == bpmTap) {
+        allowExit = false;
+      }
     case LV_EVENT_CLICKED: {
       if (obj == playPause) {
         metronomeStarted = !metronomeStarted;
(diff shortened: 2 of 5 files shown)


Avamander:
I don't see a reason to even have a "swipe to close" gesture, that's what the button is for and it avoids all of the inconsistency and by now also avoid changes people have to get reaccustomed to.

Riksu9000:
> I don't see a reason to even have a "swipe to close" gesture, that's what the button is for and it avoids all of the inconsistency and by now also avoid changes people have to get reaccustomed to.

Pressing the button is inconvenient and I don't see how this creates any changes to get accustomed to.

Avamander:
>  Pressing the button is inconvenient

Matter of personal preference then.

> I don't see how this creates any changes to get accustomed to.

I have to be more precise in Music because I can accidentally exit it. I don't want that because the intention of gestures is that looking at the screen would not be necessary to control everything necessary.

Riksu9000:
>I have to be more precise in Music because I can accidentally exit it. I don't want that because the intention of gestures is that looking at the screen would not be necessary to control everything necessary.

Swiping left and right go to the previous and next song, but swiping up and down don't directly affect anything, they just change the buttons. You'd still need to look at the screen to press the buttons. I don't think swiping to change the buttons is that great in the first place, but that's another topic.

Avamander:
> You'd still need to look at the screen to press the buttons.

You don't actually because there's a gesture.

Riksu9000:
> > You'd still need to look at the screen to press the buttons.
> 
> You don't actually because there's a gesture.

I don't see a gesture for controlling the volume. Also are we still talking about the up/down gestures, which I also believe only affect the buttons?

hubmartin:
> Flashlight can now be closed by swiping down.

That bothered me many times, thanks. And that is mainly that flashlight starts by accident :) I never used it.

>> Pressing the button is inconvenient
> Matter of personal preference then.

I love swipe down to exit. When you have to press the button you have to use two fingers to support the watch from the right side to press button from the left side it correctly. I have quite a loose wachband.

Regarding the Music buttons, I also don't like switching between volume and prev/next arrows. We might get rid of rotating turntable animation and add buttons there, so they could be visible all the time. Or just keep next/prev for gestures.. That's fro another PR.

hubmartin:
Tested and works fine, thanks for this usage improvement!

lman0:
hi , maybe @hubmartin  @Riksu9000  you could put the music screen on the exclusion list ( like for 2048, paint ) for example?

by the way , since i use for the long time the music screen i have grown quite accustomed to switching to volume since with the gesture .
like @Avamander  it's a matter of preference 
but i think that the way to close all app should be  selectable with a menu 
because some people come from sailfishos , other from phosh (linux) , android or apple and 
each os have the way to close app (some with up gesture, other with down ...)

hubmartin:
@lman0 You can still swipe up in the Music app for Volume and swipe down for previous/next buttons. This stayed the same.
I also use Music app the most time, tested it now and it is a good solution.

lman0:
i'm thinking that should be reviewed  by @Avamander  or @geekbozu  since it might help give unity to the UI .

geekbozu (approved):
Code looks great!

JF002:
Looks good to me, thanks @Riksu9000 !

PR #723, opened 2021-10-05
Feature Swipe up to unlock screen 

coxtor:
This is a unfinished POC of a screen lock feature and I'm happy to discuss the features here like:

- What parameters should be configurable?
- When should the screen be locked?
- Should the screen lock, if the watch is activated using the button?
- Should a button press also unlock the watch?
- ...

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -268,9 +268,30 @@ void DisplayApp::Refresh() {
     nextApp = Apps::None;
   }
 
-  if (touchHandler.IsTouching()) {
-    currentScreen->OnTouchEvent(touchHandler.GetX(), touchHandler.GetY());
+  if (touchHandler.IsTouching()) {    
+      currentScreen->OnTouchEvent(touchHandler.GetX(), touchHandler.GetY());
   }
+}
+
+void DisplayApp::HideLockScreenDialog(){
+  if(!lockscreenVisible) return;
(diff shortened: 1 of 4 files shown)


hubmartin:
[thanks for video.. I was already asking for it, then you added it :) ]
Can you explain what led you to create this so we could better understand the need of it?

My watch sometimes turn on unintentionally only on wake-on-wrist. Never on pressing the button, maybe few times on doubleclick. So now I don't use wake on wrist and use doubleclick or the new ShakeWake™ #691 seems rock solid.

Based on the video it is really thoughtful that the screen is fully visible, only a small text displays asking for swipe. Let's see what others will say.

coxtor:
> [thanks for video.. I was already asking for it, then you added it :) ] Can you explain what led you to create this so we could better understand the need of it?
> 
> My watch sometimes turn on unintentionally only on wake-on-wrist. Never on pressing the button, maybe few times on doubleclick. So now I don't use wake on wrist and use doubleclick or the new ShakeWake™ #691 seems rock solid.
> 
> Based on the video it is really thoughtful that the screen is fully visible, only a small text displays asking for swipe. Let's see what others will say.

Essentially you summed it up why I feel that a unlock mechanism is necessary, the screen randomly wakes up from now and then. So in conjunction with certain wake options, like shake to wake or raise wrist this feature could be useful. 
I'm open to changes, designwise and in implementation.

PR #729, opened 2021-10-08
Removed the title prefix from issue template

Avamander:
We have labels for this, there's no need to prefix the title with one.


JF002:
> The template doesn't work anymore and can't be selected when opening an issue. The title can't be empty.

Indeed... @Avamander  could you have a look?

PR #731, opened 2021-10-08
indev_drv/fs_drv goes out of scope but reference is saved, can be overwritten and dereferenced.

Quantum-cross:
This variable goes out of scope after the init, but its reference is kept in an lv_indev object in the lv_indev_ll.

Store it as a member variable to keep it alive.

### DIFF
--- a/src/components/fs/FS.cpp
+++ b/src/components/fs/FS.cpp
@@ -179,8 +179,6 @@ namespace {
 }
 
 void FS::LVGLFileSystemInit() {
-
-  lv_fs_drv_t fs_drv;
   lv_fs_drv_init(&fs_drv);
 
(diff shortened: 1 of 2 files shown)


Quantum-cross:
well crap, looks like it was changed in lvgl at some point.

https://github.com/lvgl/lvgl/commit/7840f3577eb55a850a31f33a8c324d97802a988e#diff-af692396739a3a137ea3087ec0aa2034ab1d2fe1032479f841fd2560ba3db589L92-R93

I guess that makes sense why this became a problem only with lvgl 8...

JF002:
Ah ok, I didn't know you were experiencing with LVGL8 ;)
I think we can close this PR for now (as we are not running LVGL8) and try not to forget this when we update to lvgl8, then!

PR #735, opened 2021-10-08
Torch enhancements

coxtor:
Enhance functions of the torch app

https://user-images.githubusercontent.com/7712562/136627486-79fd644a-0538-4a07-bc3b-451e3d8fde2a.mp4

**Features:** 
- **Swipe up / swipe down:** cycle brightness 
- **Swipe left / swipe right:** change color
- **Long press:** go back to first color (white)
- **Double tap:** Toggle brightness on/medium
- Last used color is saved after the app is closed.

Features to possible add: 
- Shake to wake integration

Also:
The colors are now globally usable and can be included from all files easily.


hubmartin:
Thanks for improvement @coxtor.
Swipe down gesture should close the app, I'm already using #720 which adds swipe down gesture.

In 80% when my flashlight is started it is by accident and I use swipe down a lot to get back. It is very annoying to reach for the button every time.

What do you think? Would it possible to rething gestures in torch so the swipe down could be kept for closing? Thanks.

coxtor:
I agree - I'll have a look on how the music app handles the swipe down feature and the functionality within the app - it appears to have both is possible.

hubmartin:
Thanks for rethinking that. 
What about adding a buttons in the four quadrants in the screen? If the buttons will be transparent and have only small text, then it will not have practically any effect on the light output.

Dirty sketch from OO Calc :)
image

Each button could have bigger touch area like 1/4 of the screen.

Riksu9000:
Are these changes actually useful? I hope that InfiniTime doesn't get filled with features that people don't actually use.

I'm doubtful about the usefulness of colors and brightness control and I don't think the icon should be removed. Without the icon we can't tell what app is open, or if the watch is even working correctly without context. Also if you turn off the backlight, you can't tell if the watch is awake or sleeping.

There is #633, where a red flashlight has been discussed, but whether it ended up being usable or not hasn't been confirmed.

I would've instead just made the default state off, so the light doesn't blind you on launch, and used a lower brightness on flashlight off, so the icon isn't too bright. I'm willing to create this PR if you don't mind.

Riksu9000:
Since #740 was merged, can you check if there are still features that you would like to add and update this PR or create a new one?

PR #738, opened 2021-10-10
Added section / pics about InfiniTime User Interface in the doc

ncartron:
Hello,

this PR adds a section in the "README.md" with a bit more details on the InfiniTime User Interface.   
So far it's pretty basic, but I can expend if needs be - e.g. by adding more details on each of the applications with screenshots.

The idea is to provide new users of InfiniTime with a better onboarding.


Riksu9000:
The front page isn't the best place for this. On the front page, there's a section called [New to InfiniTime?](https://github.com/InfiniTimeOrg/InfiniTime/blob/b2829e57a81426a97770f4c40fe3070cc7f2f06e/README.md#new-to-infinitime) with links for new users. The getting started page has [this](https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/doc/gettingStarted/gettingStarted-1.0.md#infinitime-10-quick-user-guide) section, which already has a lot of this information, so maybe update that one instead?

Riksu9000:
I'm assuming this can be closed in favor of #838

Issue #702, opened 2021-09-29
[Feature Request] Several improvement ideas for torch watchapp


adocampo:
I use a lot the torchlight of my watch. I have a baby and when he's grumbling at nights, I always tiptoe in quietly and in the dark. I do not want to turn on the lamps if he's not really crying, or if I don't need to change his diaper, so this app is essential for me, and I was very glad when I saw it by default on InfiniTime!

Sometimes, though, I have to hold him with one arm while with the other I give him a drink, pick up a thermometer, or anything else.
At those times, I may want to turn off the torchlight, so the baby doesn't wake up or get distracted by the light and fall asleep quickly.
Interacting with the torch while holding a baby, trying to screw something on, or performing almost any task that requires the use of one hand (not to mention those who are one-handed), becomes a real challenge, besides it turns on or off accidentally quite often.

 - I would strongly suggest switching it off and on instead of tapping, by shaking the wrist (not to be confused with lifting the wrist, but to shake to one side to the other in rapid succession).
 - I would also remove the torchlight icon completely from the screen, especially when it is turned off as it lights quite a lot in complete darkness, and yet it can let the user know the app is running, it is quite easy to press the back button and exit the torchlight watchapp.
 - Finally, the swipe gestures could be used to increase or decrease brightness when turned on.

I said several times now, and I don't want to be so repetitive, I came from pebble, and I tested some torchlight watchapps, and the best one, IMHO, was one which you simply needed to shake the wrist to turn it off and on again.

JF002:
Thanks for your feedback based on actual use-cases! I was wondering how users were using the torchlight and I understand it can be really useful in some cases ;)

adocampo:
For me is an essential watchapp... in pebble you were able to configure all of its 4 physical buttons to start an app, I had configured one of them to start the torchlight. That's how much important it's to me. If we were being able to configure our only physical button on PT, I would probably set the torchlight as well, well, I would doubt between the torchlight or a possible 2FA (#709) 😛

Bugsbane:
I use the torch quite a lot, too. Mainly I'm using it at night when I have most of the lights off and I'm looking for something. I don't want to switch the main house lights on as too much light at night interrupts melatonin release and lowers sleep quality,  so I love having a fairly dim light always available on my wrist.

[...]

PR #740, opened 2021-10-13
Flashlight brightness control

Riksu9000:
At first I was planning on fixing #702 a different way https://github.com/InfiniTimeOrg/InfiniTime/pull/735#issuecomment-939448961, but then I found that brightness control can actually be useful in pitch black https://github.com/InfiniTimeOrg/InfiniTime/pull/735#issuecomment-940883632.

This PR adds adjustable brightness control to the flashlight app. The default brightness level will be the current backlight level.

By default the flashlight will be off, so it won't blind the user on launch.

The brightness control affects the backlight while the flashlight is disabled, which should make the icon dim enough that it isn't an issue.

Also added swipe down to exit and changed return app to QuickSettings.

If we find that a red flashlight color is actually usable and beneficial in some use case, that could be added behind a long press.

After recording this video, I changed the opening animation to a slide up, like the other apps.

Fixes half of #702. Shake algorithm should be added later.

https://user-images.githubusercontent.com/37774658/137112986-f5f60875-d6ad-4922-8062-dfe2b71ac82b.mp4

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -390,7 +390,7 @@ void DisplayApp::LoadApp(Apps app, DisplayApp::FullRefreshDirections direction) 
       break;
     case Apps::FlashLight:
       currentScreen = std::make_unique<Screens::FlashLight>(this, *systemTask, brightnessController);
-      ReturnApp(Apps::Clock, FullRefreshDirections::Down, TouchEvents::None);
+      ReturnApp(Apps::QuickSettings, FullRefreshDirections::Down, TouchEvents::SwipeDown);
       break;
     case Apps::StopWatch:
       currentScreen = std::make_unique<Screens::StopWatch>(this, *systemTask);
--- a/src/displayapp/screens/FlashLight.cpp
+++ b/src/displayapp/screens/FlashLight.cpp
@@ -5,30 +5,41 @@
 using namespace Pinetime::Applications::Screens;
 
 namespace {
-  static void event_handler(lv_obj_t* obj, lv_event_t event) {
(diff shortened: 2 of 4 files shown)


Itai-Nelken:
I have been using this pr for a few days, and it works great! the flashlight app is much more usable and pleasant to use.
A suggestion:
Usually when I open the flashlight app is because I want to use it, and the extra tap to turn it on is sometimes annoying.
I understand why it's good to have it open off, especially with the ghost touches issue. so maybe make it a setting (though the settings already feel cluttered on latest develop...)?

medeyko:
It's great that the torch is now ajustable, but personally for me it has became much less usable.

I often to routinely light a room. For this I need a torch that lights at maximum level with minimum gestures.

Now (1.7.1) I have to make FIVE gestures: 1. swipe left for menu; 2. click on the torch icon; 3. swipe left to increase light; 4. swipe right to increase light; 5. click to turn the torch on.

It's quite inconvient. I would like to either have an opportuninty to set the default light level to maximum and the torch state to on, or have some other short sequence (of no more than two gestures) that allows to turn the torch on maximum.

Also it's a pity that requests to support RED color were rejected. There are lot cases when low-energy photons are preferable, like photograhy or astronomy. Also, the red color is used to draw attention. Namely, I would use red torch when walking by roadside. It might be just a fourth "light level" option, a red one after three white ones.

For the latter usage, it would be also neat to have an option for a stroboscopic lighting (it draws more attention and keeps battery), but probably it's another topic for discussion.

(Please point where to post my comment if I chose a wrong place.)

Riksu9000:
@medeyko A new issue would be the best place to discuss this.

PR #751, opened 2021-10-16
Docs: Clarify Docker image needs to be built 1st, fix URL and typos.

carlosperate:
There was an incorrect DockerHub URL in the docs, so I've fixed it together with a couple of other minor things.

### DIFF
--- a/doc/buildWithDocker.md
+++ b/doc/buildWithDocker.md
@@ -11,7 +11,9 @@ Based on Ubuntu 18.04 with the following build dependencies:
 
 ## Run a container to build the project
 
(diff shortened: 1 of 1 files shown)


Avamander:
I'm getting the feeling more and more than these paragraphs should just be JavaDoc-style inside the source code. This would reduce documentation maintenance burden and keeps relevant things tightly tied together.

carlosperate:
Yeah, I run the build with that image, I didn't test it, but it did build without errors. It looks like the image was last updated 7 months ago, and the Docker file last updated 5, so is not too behind.

Issue #748, opened 2021-10-16
Evolution of the documentation


JF002:
The documentation (for end-users and for developers) is not well maintained for now. The doc is scattered across multiple pages and places, is probably out of date and not visible enough for the end-users.
I think it would be a good idea to try to update and improve the doc to improve the users and developers experience!

Now, the question is : do we keep the documentation (for developers and/or end-users) in the InfiniTime project or do we host it somewhere else?

- In another git repo (could be hosted in the InfiniTime organization, of course)
- In Github wiki/pages
- On InfiniTime' website (https://infinitime.io), which I unfortunately do not maintain quite well :/
- anything else?

Splitting the doc from the code would allow to maintain it more easily and to delegate the maintenance to some contributors. But it would have the disadvantage to de-correlate the documentation from versions of InfiniTime. Can we handle that?

So, I open this issue to talk about the documentation : how would you manage the documenation of the project?

ntindle:
GitHub pages works well for the arcane algorithm archive. It’s better than wiki imo because it supports more features and can be setup to use markdown as well. 

TL;DR; +1 GitHub pages

geekbozu:
I'm leaning towards GitHub pages as well. its a nice middle ground between all the options...oh and we do not have to maintain the back end...GitHub does that for us ;)

ncartron:
agreed with the GutHub, that makes maintaining the doc easier (not sure how this would be done with infinitime.io?) - and of course hosting is done by GH

ncartron:
@geekbozu agreed, docs.infinitime.io (or something similar) would be nice

MrMinimal:
@JF002 Regarding your GitHub pages question:

You can think of it as a directory. It is a seperate repository, but GitHub's servers only host the contents of the main branch - so it's like having a single directory to store static files in.

MrMinimal:
## Proposal
I want to make a case for **keeping the developer and user documentation in the same repository**.
The main reason for keeping developer documentation in the same place as code is to minimize them being out of sync.

But user documentation shouldn't be out of sync either - more crucially, the less technically versed users might even be more impacted by outdated documentation.

If your main motivation to split documentation and code into seperate repositories is maintainer workload, you could assign "documentation maintainers" that can only merge doc related PRs to offload the work from code maintainers.

I agree that user documentation is updated more frequently than the software is released, one could make the case for just keeping the docs in the same state as the "development" branch is.

## Hosting options
[...]

PR #753, opened 2021-10-17
Improve BLE documentation

Elara6331:
I have noticed, during my development of [`itd`](https://gitea.arsenm.dev/Arsen6331/itd) and [`infinitime`](https://gitea.arsenm.dev/Arsen6331/infinitime), that the BLE documentation for InfiniTime is lacking. I had to spend over 2 hours scouring through Nordic documentation and InfiniTime source code just to find a couple UUIDs and a message format table. I decided to make some documentation. I posted it in my repo: [docs.md](https://gitea.arsenm.dev/Arsen6331/infinitime/src/branch/master/docs.md). I wanted more people to be able to see it, so I merged it with the existing BLE docs and fixed some spelling and grammar mistakes.

### DIFF
--- a/doc/ble.md
+++ b/doc/ble.md
@@ -254,7 +254,7 @@ Enabling notifications on this characteristic gives you a single byte upon any e
 
 #### Status
 
-The status characteristic allows setting the playing status of music. Send `0x01` to the status characteristic for playing, and `0x01` for paused.
+The status characteristic allows setting the playing status of music. Send `0x01` to the status characteristic for playing, and `0x00` for paused.
 
 #### Artist, Track, and Album
 


JF002:
Thanks! That looks good to me!
As you may have seen, I create a PR that adds a new Motion service (https://github.com/InfiniTimeOrg/InfiniTime/pull/754) and included documentation about this service in an additional file. This way, the main page is not cluttered with many details about all the services. 

What do you think? Should we keep all the info in 1 page or create smaller pages about each service?

Avamander:
I'm getting the feeling more and more than these paragraphs should just be JavaDoc-style inside the source code. This would reduce documentation maintenance burden and keeps relevant things tightly tied together.

Arsen6331:
> I mean services are all very self contained and there's a few files to look at if you're interested in a single one. There's also search.
> 
> I don't think yours is a more common problem than documentation getting outdated when it's kept separate.

That is a good point, but in the absence of something like godoc, it is very inconvenient to look for docs in the source code. I think some parts of this can be put in the source code, but some, like the DFU section, are too long to place the entire thing into the code.

geekbozu:
I'm with avamander here, a large amount of this should be in the code and generated with some documentation generator that we can link to the planned #748
I love that its getting documented period however, and any an all is welcome. Since well none is worse :D

PR #754, opened 2021-10-17
Add motion service

JF002:
This PR adds a new Motion service, which exposes the "live" step count and raw motion values (X/Y/Z).

API definition:

 - Service : **00020000-78fc-48fe-8e23-433b3a1942d0**
   - Characteristics
     - Step count (1 * uint32_t) : ~~00020001-78fc-48fe-8e23-433b3a1942d0~~  **00030001-78fc-48fe-8e23-433b3a1942d0** (READ/NOTIFY)
     - Motion values (3 * int16_t) : ~~00020002-78fc-48fe-8e23-433b3a1942d0~~ **00030002-78fc-48fe-8e23-433b3a1942d0** (READ/NOTIFY)

I also improved notifications so that they are processed and send only if the companion app registered to them (and applied to change to the heart rate service).

motionService


Arsen6331:
The BLE characteristic UUID for step count is identical to the one used to get responses from call notifications. One of them needs to be changed.

JF002:
@Arsen6331  Good catch! I totally forgot about that custom "service"! And it's unfortunately not mentioned in the [doc page](https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/doc/ble.md)...
The Motion service will move to **0003xxxx-...**, then!

Issue #750, opened 2021-10-16
BLE FS Access


geekbozu:
Something looming on the horizon is to start getting assets and other files on the external memory and off the NRF52s internal memory is something we have wanted to accomplish for a while. 

This brings the question of the best way to handle the FS access over BLE.
A few ideas have been talked about so far
- Adding FS access to the DFU process
- Adding the [Adafruit simple (not fast) BLE FS ](https://github.com/adafruit/Adafruit_CircuitPython_BLE_File_Transfer) 
- Adding a fully custom implementation #164 As implemented in this old PR

Questions:
- Are there other standards out there?
- Which one seems like it will be the most easy to have companion applications implement?
- Which ones might implement breaking changes to existing software? 

I have looked into the DFU stuff done as including files as part of updates would be awesome...However doing so is likely to break compatibility with all current implementations of the DFU updater. This makes this not very well desired and FS updates will be much easier to handle through a separate channel. Which brings us to BLE FS in some form. 

No matter what we have littlefs as a back end which makes this more of an exercise in how to expose the FS instead of implement one thankfully!

### Updates
One of the major concerns here is handling updates.  Once assets are not in internal flash anymore we have todo something to ensure they are available to the application. Modifying the DFU process would be ideal here for just that. However as above that is not the most convenient thing to do for anyone... 
The current idea in my head is to make a new update package that just has the DFU zip internally, as well as the asset files. then gadgetbridge et al would just be updated to understand that package. Perform the DFU update then transfer the new files via BLE FS not ideal, but at least it would be transparent to the end users. 

### Lets discuss. 
Obviously this is something that will have quite a few moving parts and effect how things are done internally for a while to come on this project. Lets discuss to make sure we pick something that will actually work for our needs now, And later.

PR #756, opened 2021-10-17
BLE FS Using adafruits Simple (not fast) BLE FS  Api

geekbozu:
TODO:

FS Command implementation
- [x] Read
- [x] Read Pacing
- [x] Write
- [x] Write Pacing
- [x] Delete
- [x] Make Dir
- [x] List Dir
- [x] Move File/Folder

Companion/Test App
- [x] Feature Parity To Implementation

Documentation
- [x] Documentation in Infinitime For companion app developers implemented by #839 

The start of a PR to enabled access to the internal FS over a BLE characteristic. 

Discussion on merits of implementation in #750

Currently the marked features are working in this PR,
There is a deviation from the spec which states 

>Commands always start with a fixed header. The first entry is always the command number itself encoded in a single byte. The number of subsequent entries in the header will vary by command. The entire header must be sent as a unit so set the characteristic with the full header packet. You can combine multiple commands into a single write as long as the complete header is in the packet.

I expect one command per packet with no provisions for more, I could extend this to support more commands per packet...I just do not think the complexity is needed. 

I ignore and set all date time information to 0. DO not use date time for anything.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -487,6 +487,7 @@ list(APPEND SOURCE_FILES
         components/ble/NavigationService.cpp
         displayapp/fonts/lv_font_navi_80.c
         components/ble/BatteryInformationService.cpp
+        components/ble/FSService.cpp
         components/ble/ImmediateAlertService.cpp
         components/ble/ServiceDiscovery.cpp
         components/ble/HeartRateService.cpp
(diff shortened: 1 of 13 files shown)


Avamander:
+1 for the Adafruit protocol 😀 the even slightly bigger ecosystem will help.

Avamander:
@geekbozu is there a blocker to this PR?

geekbozu:
It heavily depends on my IRL schedule. There is quite a bit Todo I hope to have a task lost up today, and to continue chipping away at it! 
There are no major blockers as of the moment implementation wise. Just time to work on it.

geekbozu:
Added a task list for this, It is slowly nearing completion!
I added some comments in the OP about deviations from the OG spec as well as asked questions about it above. 
Feedback is requested!

geekbozu:
Move works! At this point its feature complete. It needs some better status code returns as well as some documentation. 
Have we decided where that documentation is going to live yet?

PR #764, opened 2021-10-20
Added Trip meter to Steps app

### LINKED ISSUE #694: Manual step counter reset
Is there a way to reset the step counter without having to reboot?

stephanie-eng:
Fixes #694 

- Added a lap button (similar to the lap button in Stopwatch) to reset the step counter.
- Added a total step count in that does not get reset in the bottom right corner.
- Moved/shrank existing Steps UI to fit the new button and total step count
- Added a variable to MotionController to save the steps since the last lap.

Note that using the lap button does not reset the progress to the daily step goal. The progress bar uses the total step count.

The image below demos the new button after fresh boot with (1) 0 steps, (2) the watch after taking some steps (step count and total updated and equal), (3) pressing the lap button (step count reset, no change to total), and (4) after walking again (step count and total updated, but not equal).

image

I also checked that large step counts do not cause display issues.

image

### DIFF
--- a/src/components/motion/MotionController.cpp
+++ b/src/components/motion/MotionController.cpp
@@ -8,7 +8,7 @@ void MotionController::Update(int16_t x, int16_t y, int16_t z, uint32_t nbSteps)
   this->z = z;
   int32_t deltaSteps = nbSteps - this->nbSteps;
   this->nbSteps = nbSteps;
-  if(deltaSteps > 0){
+  if (deltaSteps > 0) {
     currentTripSteps += deltaSteps;
   }
 }
@@ -39,9 +39,15 @@ void MotionController::IsSensorOk(bool isOk) {
   isSensorOk = isOk;
 }
 void MotionController::Init(Pinetime::Drivers::Bma421::DeviceTypes types) {
-  switch(types){
-    case Drivers::Bma421::DeviceTypes::BMA421: this->deviceType = DeviceTypes::BMA421; break;
-    case Drivers::Bma421::DeviceTypes::BMA425: this->deviceType = DeviceTypes::BMA425; break;
-    default: this->deviceType = DeviceTypes::Unknown; break;
+  switch (types) {
+    case Drivers::Bma421::DeviceTypes::BMA421:
+      this->deviceType = DeviceTypes::BMA421;
+      break;
+    case Drivers::Bma421::DeviceTypes::BMA425:
+      this->deviceType = DeviceTypes::BMA425;
+      break;
+    default:
+      this->deviceType = DeviceTypes::Unknown;
+      break;
   }
 }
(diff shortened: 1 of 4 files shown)


stephanie-eng:
Thanks for your input. I've changed the lap button to a trip reset button and made the UI changes. The below image shows the new behaviour:

1. The watch after fresh boot, no steps taken. Steps and trip meter are both 0.
2. The watch after taking 36 steps, trip meter matches step count
3. Pressing the reset button resets the trip meter to 0, without changing the step count. Progress bar for step goal also remains unchanged.
4. Setting the time to before midnight to reset the daily step count
5. Step count in the UI is shown to be 0
6. Step count after midnight in the step app: step count reset to 0, but trip meter remains unchanged from 36. Progress bar for step goal is reset to 0.
7. Taking an additional 20 steps increments step count to 20, and trip meter to 56 (36 steps from yesterday + 20 from the new day)
8. Trip meter can be set back to 0 with the reset button

image

stephanie-eng:
Good point - thanks for the catch. I've changed the trip meter to update with the steps rather than from the app. I also found a bug in my logic for detecting new days and fixed that as well.

src/components/motion/MotionController.cpp
```diff
@@ -6,7 +6,11 @@ void MotionController::Update(int16_t x, int16_t y, int16_t z, uint32_t nbSteps)
   this->x = x;
   this->y = y;
   this->z = z;
+  int32_t deltaSteps = nbSteps - this->nbSteps;
   this->nbSteps = nbSteps;
+  if(deltaSteps > 0){
```
Avamander, line 11: Requires formatting with clang-format.

src/components/motion/MotionController.h
```diff
@@ -27,6 +27,13 @@ namespace Pinetime {
       uint32_t NbSteps() const {
         return nbSteps;
       }
+    
+      void ResetTrip() {
+        currentTripSteps = 0;
+      }
+      uint32_t GetTripSteps() const {
```
Avamander, line 34: Requires formatting with clang-format.

Avamander (changes requested):

Avamander (approved):
Looks good style and format-wise. Needs testing on a physical device.

stephanie-eng:
To produce the above screenshots, I did test the code with a physical device. If anyone else would like to try it as another point of data, that could be good.

As an aside, clang-format didn't initially work with VSCode in the container out of the box for me - I had to tweak some settings to get it to use the right executable and format on save. I can create a separate issue to document this.

JF002:
@stephanie-eng Thank you for this nice first PR! It's very well documented with test cases and pictures! I would be happy if we received more PR like this one as it makes the review much easier!

PR #765, opened 2021-10-20
Add stopwatch setting

lectrician1:
This adds a setting page to control whether the stopwatch screen sleeps or not while the stopwatch is running. 

I am a runner and often use the stopwatch to time workouts. I didn't like how the screen remained on while the stopwatch was running as I do not want to drain the battery, especially on longer runs.

The settings page allows the user to enable or disable the watch from sleeping. It also offers a brief explanation of the setting.

**Demo video** 

https://user-images.githubusercontent.com/22477298/138088529-ce71b03f-0206-4463-b6fd-1269e265df8b.mp4

The instruction positioning on the settings page has been moved up a bit than what is seen on the video. Feel free to change the UI positioning if further tuning is required.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -436,6 +436,7 @@ list(APPEND SOURCE_FILES
         displayapp/screens/settings/SettingPineTimeStyle.cpp
         displayapp/screens/settings/SettingSetDate.cpp
         displayapp/screens/settings/SettingSetTime.cpp
+        displayapp/screens/settings/SettingStopWatch.cpp
 
         ## Watch faces
         displayapp/icons/bg_clock.c
--- a/src/components/settings/Settings.h
+++ b/src/components/settings/Settings.h
@@ -20,6 +20,7 @@ namespace Pinetime {
       enum class Colors : uint8_t {
         White, Silver, Gray, Black, Red, Maroon, Yellow, Olive, Lime, Green, Cyan, Teal, Blue, Navy, Magenta, Purple, Orange
       };
+      enum class StopWatchSleepMode : uint8_t { Enable, Disable };
       struct PineTimeStyle {
         Colors ColorTime = Colors::Teal;
         Colors ColorBar = Colors::Teal;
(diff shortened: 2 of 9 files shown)


geekbozu:
> 
> 
> > What about making the stopwatch sleep by default, but by long pressing on the screen it vibrates and "locks" the screen on, disabling sleep, and maybe display a lock icon somewhere on the screen?
> 
> This is a good idea. This should probably be a global function for any app. You would need to make sure that no app utilizes a touch-and-hold function though.
> 
> Some users might want apps to not sleep by default and would get annoyed if every app's sleep was disabled by default and they had to use the "long press function" every time they used the app. Therefore I think a settings page per-app that specifies the default sleep behavior like this PR is necessary.
> 
> Obviously, creating a settings page for every app and it's sleep mode would make the settings list very long, so that is why I think we need to shift app-specific settings to a settings button present on every app.
> 
> > It has also been discussed that sleep shouldn't even be disabled in apps, but the screen timeout should be longer instead.
> 
> I would just leave this to the user and maybe make the screen wake time more customizable. We can increase the default if we need to.

This is something @kieranc is looking into for PineTimeStyle as well with his new color picker interface.

kieranc:
> > What about making the stopwatch sleep by default, but by long pressing on the screen it vibrates and "locks" the screen on, disabling sleep, and maybe display a lock icon somewhere on the screen?
> 
> This is a good idea. This should probably be a global function for any app. You would need to make sure that no app utilizes a touch-and-hold function though.
> 
> Some users might want apps to not sleep by default and would get annoyed if every app's sleep was disabled by default and they had to use the "long press function" every time they used the app. Therefore I think a settings page per-app that specifies the default sleep behavior like this PR is necessary.
> 
> Obviously, creating a settings page for every app and it's sleep mode would make the settings list very long, so that is why I think we need to shift app-specific settings to a settings button present on every app.
> 
> > It has also been discussed that sleep shouldn't even be disabled in apps, but the screen timeout should be longer instead.
> 
> I would just leave this to the user and maybe make the screen wake time more customizable. We can increase the default if we need to.

I quite like the idea of being able to lock the screen on for stopwatch or similar but solid wrist raise wake or shake wake mostly eliminates this need.
Per-app settings is certainly a goal of my recent PR #718 it adds a settings button on long press and can hopefully be applied to all watchfaces, step count, whatever else needs settings but doesn't want to clutter the settings menu. I can't guarantee the current solution is perfect but I'm open to input!

lectrician1:
If someone wants to work on this in the future or if I ever get back to this, I was planning on moving the settings page from the Settings page to a page accessed on through pressing a settings icon at the top right corner of the opened stopwatch app. This is how Fitbit does it and it works well.

I hope on making this the norm for application settings/options rather than putting them on the Settings page so that they are easily accessible and the Settings page does not have too many options. 

I'd also prefer this per-app settings button rather than a long-press @kieranc so that it's more intuitive and we don't take up a potential form of user input for an app.

Also, this PR should probably be merged with #783.

Riksu9000:
I'll close this as no app should disable sleep forever and there should instead be a longer screen timeout when using apps than when just checking the time, so adding this setting isn't necessary.

PR #767, opened 2021-10-20
Fixed the feature request template

Avamander:
According to https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms#top-level-syntax `title` is optional but can't be empty. So let's try that.


JF002:
@Avamander  Looks good! 
I guess we can do the same change for the bug reports?

PR #776, opened 2021-10-23
WatchFaceAnalog : move notification icon

SteveAmor:
Standardise notification icon to display in top left.


Riksu9000 (approved):
This feels like a better placement to me.

PR #785, opened 2021-10-26
About : Remove misleading info

SteveAmor:
Removed Steps which is hard coded to display 0.
Could be misleading and is using up memory.


SteveAmor:
Of course, my view will be biased but this is hard coded to zero so I think it should be removed to avoid being misleading. If it correctly showed the step count for the BMA421 but not for the latest accelerometer (BMA425?) I would concur that a bug fix is required.

Riksu9000 (approved):
About is a place for verbose information that might not exist elsewhere. The step count can be seen in the steps app, so it's not necessary here.

PR #786, opened 2021-10-26
Move Status Bar code to DisplayApp and support 12-hour format

mruss77:
The "status bar" that appears above the quick settings & app menu was directly coded into those screens. Developers wanting to add a status bar to the top of their apps would need to copy the code and get a reference to the battery controller and datetime controller.

This PR moves the code to draw the status bar into DisplayApp. Apps that want a status bar just need to declare a struct (DisplayApp::StatusBar) and call the methods to draw the status bar and to update it periodically.

Another benefit of this approach is that the status bar can be updated globally in one place. I updated it to show 12-hour or 24-hour time based on the time format setting.

I updated the app menu code in ApplicationList and Tile and the QuickSettings code to use this new functionality to draw the status bar.

Not many apps use the status bar currently, but I think it would be useful to see the current time when setting the alarm.

I used the StatusBar struct to hold the pointers for the LVGL objects in the status bar, so that the screen actually displaying them would own them and they wouldn't go out of scope... not sure if this is the best approach, but it works.


Riksu9000:
We might need more time to figure out what to do about the status bars, so could you create another PR with just the time format changes, so we can get that merged quicker?

mruss77:
@Riksu9000 sure, will do that today

PR #791, opened 2021-10-29
Fix recovery firmware build

evergreen22:
Fix the "recovery" firmware build targets

Error introduced in #746

### DIFF
--- a/src/displayapp/DisplayAppRecovery.h
+++ b/src/displayapp/DisplayAppRecovery.h
@@ -10,7 +10,7 @@
 #include <date/date.h>
 #include <drivers/Watchdog.h>
 #include <components/motor/MotorController.h>
-#include <BootErrors.h>
+#include "BootErrors.h"
 #include "TouchEvents.h"
 #include "Apps.h"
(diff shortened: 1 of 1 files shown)


JF002:
@evergreen22 Why did you reopen this PR? The [last build](https://github.com/InfiniTimeOrg/InfiniTime/actions/runs/1402602746) on develop succeeded.

evergreen22:
I reopened because building recovery targets didn't work for me. The compiler looked in the systems folders for the BootErrors.h file instead of locally. The CI build works, but I don't think it is building any of the recovery targets.

evergreen22:
I can see that the recovery target is building on CI (wasn't logged in to GH when I replied before). All I can say is that the recovery targets will not build on my local machine unless the change in this PR is made.

JF002:
Ok, I got confused by the new colors in Github. I thought you re-open PR that was already merged...
I had a look at the changes, and they make sense, I'll merge it ;)
I'm not sure why it works on github/docker/my setup and not on your side, though.

PR #792, opened 2021-10-29
Memory savings by replacing strings

Riksu9000:
Saves 996 bytes!


JF002:
@Riksu9000 Impressive! Does it free RAM or FLASH memory?

Riksu9000:
> @Riksu9000 Impressive! Does it free RAM or FLASH memory?

These changes only free FLASH memory.

JF002:
How come those std::to_string were using so much space in flash?

geekbozu:
there are still a few more instances of `std:to_string` left in the repo it seems. In the date time library. I wonder if removing those will yield some more space? I'm confused as well as to why this frees up so much room. Is it just because we no longer link to_string in the infinitime blob? Or is that just that hungry of a routine...in flash?

PR #793, opened 2021-10-30
Exclude some clang-tidy checks

Riksu9000:
The first check shows a warning on almost every include. Doesn't seem very helpful.
The second one says to use dynamic_cast, but we can't because we use `-fno-rtti`

I don't see these warnings when running clang-tidy from CLI, but they're shown in my editor. Not sure why that is, but disabling them should be fine right?

https://clang.llvm.org/extra/clang-tidy/checks/llvmlibc-restrict-system-libc-headers.html
https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines-pro-type-static-cast-downcast.html


Avamander:
> Not sure why that is,

Because you have a lower version in CLI than your editor.

Avamander (approved):

Issue #707, opened 2021-09-30
Bluetooth connection still dropping out


Bugsbane:
So, I updated to version 1.6, but I still have the issue where after a number of hours that the bluetooth connection drops out, and Gadgetbridge can't reconnect until the watch is restarted. After restarting, pairing works fine.

Steps to reproduce:
1. Pair watch to phone
2. Wait 12-24 hours.
3. Watch bluetooth connection resets.

Expected behavior:
Connection remains active.

Specs:
Infinitime 1.6 with 1.0 bootloader flashed by Gadgetbridge 0.59.3 on Android 10 on a Samsung Galaxy A51. Is there any other info that would be useful?

omartijn:
I am having a similar issue. For me, it drops connection _all_ the time, but _that_ at least seems to be due to Sailfish power-management, since it doesn't happen when the screen remains on (and no, setting the suspend policy didn't help).

However, after a day of constant connection dropping and reconnects, the watch stopped accepting connections. Every time I tried to connect from the phone side, it would briefly show connection and then give an error that the passcode was incorrect. The only way to fix it was rebooting the watch.

kieranc:
It is looking like the heart rate monitor is causing Bluetooth issues, are either of you using the HRM, or not?

hallebubbsan:
> It is looking like the heart rate monitor is causing Bluetooth issues, are either of you using the HRM, or not?

I started the hrm at the same time I got the connection issue

omartijn:
I didn't notice the link at the time, but I do remember at some point in the evening turning it on for a little bit and then switching it off. Later I did experience the connectivity issues. I can't be sure of the exact time, but it sounds plausible that it's the culprit.

akutuzov:
> It is looking like the heart rate monitor is causing Bluetooth issues, are either of you using the HRM, or not?

I notice exactly the same. Without HRM, the watch can stay connected for whole days (with occasional re-connects). As soon as HRM is turned on, the Bluetooth connection dies in a matter of an hour if not less. Only resetting the watch helps then.

InfiniTime 1.6.0, Sailfish OS 4.2 on Sony Xperia 10

RichardNewb1e:
I'm having same issue on Android 10 of XiaoMi 9 without using HRM. The first 12 hours or so worked fine, then Bluetooth was dead and no longer discoverable.

Avamander:
@RichardNewb1e 

[...]

Issue #628, opened 2021-08-30
Security: Auth Request should be Pinetime-first


pdolinic:
Not everyone should be able to connect to the Pinetime to execute code. This is security risk, even though the watch has to execute updates. 

Solution: Enable Requests prior to code upload or connections.

JF002:
Indeed, there's not security feature in InfiniTime right now. NimBLE provides many security features, but we need to enable/intergrate them in the project.

evergreen22:
I've been working on encryption+authentication and bonding for a while. I have successfully enabled encryption and authentication with a pass key, but not bonding. Once I get bonding sorted out I'll make a PR.

Note that when the watch is connected, no one else can upload firmware because InfiniTime only supports one ble connection at a time.

I don't know what your threat model is, but there is not an easy way to prevent an evil maid attack. If you leave the watch unattended and unconnected anyone can get to it to modify the settings or firmware even if encryption+authentication is enabled for connections.

GIJack:
Threat models. Someone gets close with a handheld device and executes one of the following attacks

1. MITM  - can read messages, see who's calling, intercept bio-data like heartrate
2. Exploit against the phone - use address/call log permissions to attempt and read this data off the phone using gadgetbridge
3. exploit against the watch - impersonate the phone to upload and run code.
4. 4 de-auth and connect. jam/block/deauth the connection, or wait for one of the numerous disconnects the pinetime has and reconnect before your device has.

Most of these hinge on a handheld device that is a modified cellphone, or something of similar size. Devices like this are relatively cheap, and the physical methodology of such "bump" attacks with small concealed computers with wireless devices has been demonstrated against NFC devices.

If you as much make the complexity of the attack being you have to physically take possession of the device, and then scroll through the settings and remove the current pairing to add a new one, you've done a lot as its pretty feasible to keep the thing on your wrist most of the time. Also, if you pair it, and need to go into settings to remove the pairing, and can't add back an old pairing, it will be tamper evident.

Doing as much as bonding + encrypt + authentication that needs to be manually removed in settings before adding a new one, would go a long way. You'd increase the complexity needed greatly

pdolinic:
I entirely agree with @GIJack regardless if only 1 connection is allowed or not, pairing the watch with a phone is a giant security risk currently until the appropriate authentication is given, and this should be communicated publicly imo.

I personally would not grand that many rights over my phone as of now.

Avamander:
[...]

PR #796, opened 2021-10-30
Connect and bond with a passkey

### LINKED ISSUE #224: Bluetooth / BLE connection watch face symbol unreliable.
The symbol on the watch face that indicates a Bluetooth connection loses state or fails to update.
Reproduce:
Press the button on PineTime to wake it up.
Connect with nRF Connect 2.4.8 on iOS.
Note symbol appears.
Press disconnect / connect a few times, it should turn on and off.
With nRF Connect in the connected state (and symbol on the watch face), scroll the watch face to the first menu.
Touch the watch app icon to go back to the watch face
Test connectivity / symbol by disconnecting/connecting in nRF Connect, sometimes it is stuck.
If it still works, while connected scroll watch face and go to the "list" icon (information app)
Use the button to return to the first menu
Touch the watch app icon to go back to the watch face
Test connectivity / symbol by disconnecting/connecting in nRF Connect, this is always stuck in my testing.

Going to different apps and back to the watch face will generally reset this and the symbol state will start tracking again.

### LINKED ISSUE #302: Watch does not automatically reconnect to GadgetBridge after turning airplane mode on and off

### LINKED ISSUE #502: Authentication/Authorization Support BLE Bonding
EDIT:
Add native authorization using NimBLE

The experts have informed me that the BLE stack they are using supports proper Bluetooth authentication, so this post was edited to reflect that.

evergreen22:
Add Feature - Connect and bond with a passkey

- [X] Passkey pairing - passkey is displayed on watch and entered on phone
- [X] Swipe down to clear passkey screen
- [X] Connection encryption
- [X] Connection bonding
- [X] Automatic reconnects to bonded peripheral\*
- [X] Trusted device on Android
- [X] Persist bond between reboots

\* ~~**Rebooting the watch will cause reconnect failures. You must delete the bond from the phone to reconnect/pair.** Persisting the bond between reboots will resolve this.~~

Closes #224 
Closes #302 
Closes #502 

There are several tasks listed below that I have listed as future work in order to make this PR easier to review and test.

Future work includes the following tasks:
* Investigate encrypting other characteristics
* Explore if a delete bond option in settings is desirable/needed
* Enable RPA
* Automatically clear the passkey screen on pairing completion
* Enable the whitelist to conserve power and prevent non-bonded devices from pairing


src/components/ble/NimbleController.cpp
```diff
@@ -94,8 +101,22 @@ void NimbleController::Init() {
   Pinetime::Controllers::Ble::BleAddress address;
   rc = ble_hs_id_copy_addr(addrType, address.data(), nullptr);
   ASSERT(rc == 0);
-  bleController.AddressType((addrType == 0) ? Ble::AddressTypes::Public : Ble::AddressTypes::Random);
+
   bleController.Address(std::move(address));
+  switch (addrType) {
+    case BLE_OWN_ADDR_PUBLIC :
```
Avamander, line 107: This does not seem formatted with `clang-format`

src/components/ble/NimbleController.cpp
```diff
@@ -105,17 +126,11 @@ void NimbleController::Init() {
...
-  int rc;
-
-  /* set adv parameters */
   struct ble_gap_adv_params adv_params;
   struct ble_hs_adv_fields fields;
-  /* advertising payload is split into advertising data and advertising
-     response, because all data cannot fit into single packet; name of device
-     is sent as response to scan request */
   struct ble_hs_adv_fields rsp_fields;
+  int rc;
```
Avamander, line 132: I know you didn't add this, but what does this variable do? Seems very nondescript.

Also maybe an fixed-size integer would be better here?

src/components/ble/NimbleController.cpp
```diff
@@ -138,6 +153,12 @@ void NimbleController::StartAdvertising() {
   fields.uuids128_is_complete = 1;
   fields.tx_pwr_lvl = BLE_HS_ADV_TX_PWR_LVL_AUTO;
 
+  /* The advertising payload is split into advertising data fields
+   * and advertising response fields because all of the data cannot
+   * fit into single packet. Therefore the name of device is sent as
+   * a response to the scan request.
+   */
+
   rsp_fields.name = (uint8_t*) deviceName;
```
Avamander, line 162: C-style cast, could get rid of.

src/components/ble/NimbleController.cpp
```diff
@@ -138,6 +153,12 @@ void NimbleController::StartAdvertising() {
   fields.uuids128_is_complete = 1;
   fields.tx_pwr_lvl = BLE_HS_ADV_TX_PWR_LVL_AUTO;
 
+  /* The advertising payload is split into advertising data fields
+   * and advertising response fields because all of the data cannot
+   * fit into single packet. Therefore the name of device is sent as
+   * a response to the scan request.
+   */
+
```
Avamander, line 161: Should be `/**` and without the ` ` after the `*/` to become a Javadoc-style comment that tooling can pick up.

Avamander (changes requested):

lman0:
hi 
@evergreen22  i've tested this pull and it work great!

yet there is some point that would make better:
- the passkey screen should not be tied to the timeout screen because if the screen timeout is 5 s , for example , then there is  not enough time to read the passkey properly.
- could you add a button , in settings (for example), in order to initiate again a new bond (with the passkey screen) with the phone?
because the only way , when someone don't manage to put the key (either by having the screen time out , touching the screen) , there is no way to restart a bonding unless infinitime is restarted...
(resetting infinitime mean losing notification , timer ....) thatwould be the case if you add a button to do the trick a initiate again a bonding.

src/components/ble/NimbleController.cpp
evergreen22: Fixed

src/components/ble/NimbleController.cpp
evergreen22: I relocated the variable to where it is used to make the purpose clearer. Note that this variable will be optimized out by the compiler unless debugging is enabled.

This code is essentially extracted from nimble and I'm following their convention of naming the return code "rc"  (I prefer rv for "return value"). The called nimble functions all return a full int.

src/components/ble/NimbleController.cpp
evergreen22: Fixed

src/components/ble/NimbleController.cpp
evergreen22: The comment language was copied from the nimble example code. I made a couple of editorial changes and relocated it closer to the code it is describing in this PR. The advertising payload and scan response is clearly documented in the BLE spec so I think the comment is unnecessary (it is still in the nimble example code).

lman0:
hi 
@evergreen22  i've tested this pull and it work great! 
(i tested with gadgetbridge and by stop and starting  the Bluetooth of the phone)

yet there is some point that would make it better:
- the passkey screen should not be tied to the timeout screen because if the screen timeout is 5 s , for example , then there is  not enough time to read the passkey properly.
- could you add a button , in settings (for example), in order to initiate again a new bond (with the passkey screen) with the phone?
because the only way , when someone don't manage to read the key , either  because of the screen time out or touching the screen by mistake, is to restart infinitime in order to restart the  bonding  ...
(resetting infinitime mean losing notification , timer ....) that would not  be the case anymore  if you add a button to do the  initiate again a bonding.

evergreen22:
Thanks for testing @lman0 and your feedback.
All of the the items you suggested are good ideas and listed in the "Future work" section.

lman0:
is these will be on another pr?
because the first time , i had to restart infinitme because the screen passkey was dimmed....to fast

evergreen22:
@lman0 - you can tap on the screen to wake it up and see the passkey again - see the "Wake Up" setting on your phone to set how fast your screen turns off.

If you miss the passkey and can't get it back by tapping, just wait. Your phone will restart pairing automatically if you don't enter the passkey.

If that doesn't work, press cancel on your phone and then wait. Your phone will restart pairing automatically if you cancel the pairing attempt.

I am currently planning on the "Future work" items to be a new PR.

lman0:
i don't know why @evergreen22  , but after 3 or so restart i can't no longer initiate the screen passkey at launch....
  have many time i need to wait for the pairing restart? (i'm on android)

lman0:
there is a screen corruption @evergreen22  if the phone request more than once the passkey:
if you don't validate the passkey and you don't cancel on phone , the next time the passkey shwo up it will corrupt the screen
and only a restart is able  it restore to normalcy

evergreen22:
Thanks for the feedback, I'll see if I can reproduce the issue.

I'd recommend not running it yet on a sealed watch. I'm only running it on my dev-kit right now because it is not finished. Once it is merged then it is ready to install on a sealed watch.

lman0:
it's seem be the case when the passkey is still shown when the new passkey is generated that the corruption begin @evergreen22

src/displayapp/screens/PassKey.cpp
```diff
@@ -0,0 +1,23 @@
...
+  lv_label_set_text_fmt(lpasskey, "%06u", key);
+  lv_obj_align(lpasskey, nullptr, LV_ALIGN_CENTER, 0, -20);
+}
+
+PassKey::~PassKey() {
+  lv_obj_clean(lv_scr_act());
+}
+
+bool PassKey::OnTouchEvent(Pinetime::Applications::TouchEvents event) {
+  if (event == Pinetime::Applications::TouchEvents::SwipeDown)
```
Avamander, line 19: Lack of braces is not consistent with the rest of the code

Avamander (changes requested):

src/displayapp/screens/PassKey.cpp
evergreen22: That bit of code was bothering me from a readability/clarity standpoint and I was contemplating whether an "else" would help. You are correct, the idiomatic solution is braces - fixed.

geekbozu:
How many bonds do we want to store? 1 or more? I can attempt soon to get it to store to littleFS since I'm...well versed in that how that works right now :)

evergreen22:
I am just storing 1 in my offline test code. Nimble is set to persist
3, but I don't see any reason to keep more than 1 in the fs.

I wrote two functions; StoreBond and RestoreBond. Every time I access
the fs, the bus freezes and then the watchdog fires. Your expertise
would be greatly appreciated.

Avamander (approved):
Formatting wise seems to be fine. 

Needs testing on physical device.

evergreen22:
I am testing a proposed fix for the problem reported by @Iman0 and will
push later today if it resolves the issue.

geekbozu:
So I almost have an PR for this PR to support storing bonds to flash... Seems the device wants to store 2 SEC key pairs and 2 CCCD entries per device. I'm only allowing one CCCD entry... hopefully that will make it persist.

evergreen22:
Awesome!

I am trying to store one key set for us, one set for the peer, and 2
CCCD entries since I think that is what it will take to restore the
connection.

If the peer subscribes to everything we offer that number can grow. GB
does not subscribe to the additional items. I can make nrfconnect
subscribe and that causes nimble to store the additional info.

geekbozu:
I'll just expand on the CCCD store on my local copy to hold up to the config entries then. Its not a huge deal either way....I was taking the lazy way out and just using files called
`BLE_STORE_PEER_SEC` , `BLE_STORE_OUR_SEC`, and `BLE_STORE_PEER_CCCD`....aka one key for each. Ill have to expand CCCD to use a struct of the proper type and persist that...not trivial but not hard persay either! Just going to take more time!

geekbozu:
Do you have it working with the BLE_RAM_STORE in the host folder? If so I can easily modify that to persist...currently all of what i'm doing is pretty much just implementing that in our NimbleController class...which feels really bad.

lman0:
> I am testing a proposed fix for the problem reported by @iman0 and will push later today if it resolves the issue. Tested and pushed - [13a6fbf](https://github.com/InfiniTimeOrg/InfiniTime/commit/13a6fbf39936544416cb580db74b525bceacd13e)

Hi 
first and foremost i"m @lman0  and not @iman0 , L not i at the beginning , i remind you.

i have tested the pr with the fix and it work great  with no bug of corruption anymore !

I think it could could be reviewed /merged by @Avamander , @geekbozu  in order to be merged by @JF002 . 

Since it seem the pr for persisting between reboot seem to be done by @geekbozu  
if this this pull were to be merged in the develop, it would make @geekbozu pr to be created more easier

lman0:
in this form ,( even without the persist between infintime reboot) ,
 it would be a great inclusion in 1.7.0  @JF002 
 since most  reboot of infinitime done by the user **are** in order to make a new bond with gadget.
 the persist between infinitime would less need , since the purpose for reboot is resolved by this pull for most case

trman:
@evergreen22 , @geekbozu  any news on the pr?

i'am also quite interested by this pr and have this pr on my pinetime for quite some time 
 since it  resolve these issue 
 https://github.com/InfiniTimeOrg/InfiniTime/issues/707 
https://github.com/InfiniTimeOrg/InfiniTime/issues/302

and 2/3  this one https://github.com/InfiniTimeOrg/InfiniTime/issues/628 , because we can still choose which smartphone unlike peeble for example.

do the task 7 :Persist bond between reboots ,still need to be necessary for this pr @evergreen22 ?

for what i have read @geekbozu  will do pr  that will resolve this task so it might be not necessary to let this part be blocker of this pr completion .

for instance , i use only this pr for 1 week already and i didn't need to restart infinitime since gadgetbridge pickup infinitime instantly.
that a great improvement on Bluetooth reliability   on par , or even above ,with @danielgjackson PR !

maybe @JF002  will even make a release only for this pr on this completion since the improvement on Bluetooth are this great!


@evergreen22    keep the good work on this pr !
@geekbozu  good luck on the persist part as it will be the needed base for be the creation of a screen where user will be able to select which smartphone will be removed or remembered !

src/displayapp/screens/PassKey.cpp
```diff
@@ -0,0 +1,31 @@
...
+  lv_obj_clean(lv_scr_act());
+}
+
+bool PassKey::OnTouchEvent(Pinetime::Applications::TouchEvents event) {
+  if (event == Pinetime::Applications::TouchEvents::SwipeDown) {
+    return false;
+  }
+
+  return true;
+}
```
Riksu9000, line 31: Looks like this override might not be necessary. Returning false allows displayapp to check if the gesture should close the app and by default this always returns false.

src/displayapp/DisplayApp.cpp
```diff
@@ -331,6 +335,11 @@ void DisplayApp::LoadApp(Apps app, DisplayApp::FullRefreshDirections direction)
       ReturnApp(Apps::Clock, FullRefreshDirections::Down, TouchEvents::None);
       break;
 
+    case Apps::PassKey:
+      currentScreen = std::make_unique<Screens::PassKey>(this, bleController.GetPairingKey());
+      ReturnApp(Apps::Clock, FullRefreshDirections::Up, TouchEvents::SwipeDown);
```
Riksu9000, line 340: The FullRefreshDirection should be down

evergreen22:
@Riksu9000 thanks for the feedback.

src/displayapp/screens/PassKey.cpp
```diff
@@ -0,0 +1,31 @@
+#include "PassKey.h"
+#include "../DisplayApp.h"
```
Avamander, line 2: Maybe we can avoid relative imports in new code?

Avamander (changes requested):

geekbozu:
> 
> 
> @evergreen22 , @geekbozu any news on the pr?
> 
> i'am also quite interested by this pr and have this pr on my pinetime for quite some time since it resolve these issue #707 #302
> 
> and 2/3 this one #628 , because we can't still choose which smartphone unlike peeble for example.
> 
> do the task 7 :Persist bond between reboots ,still need to be necessary for this pr @evergreen22 ?
> 
> for what i have read @geekbozu will do a pr that will resolve this task so it might be not necessary to let this part be blocker of this pr completion .
> 
> for instance , i use only this pr for 1 week already and i didn't need to restart infinitime since gadgetbridge pickup infinitime instantly. that a great improvement on Bluetooth reliability on par , or even above ,with @danielgjackson PR !
> 
> maybe @JF002 will even make a release only for this pr on this completion since the improvement on Bluetooth are this great!
> 
> @evergreen22 keep the good work on this pr ! @geekbozu good luck on the persist part as it will be the needed base for be the creation of a screen where user will be able to select which smartphone will be removed or remembered !
> 
> @Avamander , do you thinks that could that the pr could be merged as is?

No update on my end. I passed the hand back to @evergreen22 for now. After forwarding the issues regarding File System Access they were running into. I am waiting to hear more as well! I do however think this is a requirement in this PR to have persist working. :)

trman:
> No update on my end. I passed the hand back to @evergreen22 for now. After forwarding the issues regarding File System Access they were running into. I am waiting to hear more as well! I do however think this is a requirement in this PR to have persist working. :)


@geekbozu ,   you have written in a previous comment that you had a pr nearly done for this pr  , yet you say that you have passed the hand back to @evergreen22 .

does it mean that you have given to @evergreen22  , the nearly finished code in order for him to include it in his pr?
i don't understand otherwise , your statement n this part.
(the persist for me is not necessary per see , for this pr at least,, 
but more  for the later pr , because we will need to have a screen to remove persisted device  , and it seem that it will be  for a latter pr as per @evergreen22

geekbozu:
> 
> 
> > No update on my end. I passed the hand back to @evergreen22 for now. After forwarding the issues regarding File System Access they were running into. I am waiting to hear more as well! I do however think this is a requirement in this PR to have persist working. :)
> 
> @geekbozu , you have written in a previous comment that you had a pr nearly done for this pr , yet you say that you have passed the hand back to @evergreen22 .
> 
> does it mean that you have given to @evergreen22 , the nearly finished code in order for him to include it in his pr? i don't understand otherwise , your statement n this part. (the persist for me is not necessary per see , for this pr at least,, but more for the later pr , because we will need to have a screen to remove persisted device , and it seem that it will be for a latter pr as per @evergreen22

I had something nearly working, Which I then realized was missing a very large chunk of the implementation, and on the chat forums handed the information related to the Filesystem implementation to evergreen so they could attempt it on their end. He has not received any code from me. Due mostly to him having an implementation on his end already in the works as well. 
That said we are working on this on our free time as we feel like it. I know evergreen and my self have both been very busy with work. This will get completed in good time. :) 

Removing a persisted device is a non issue, as the way the store is currently getting setup is to just only store one devices worth of security info and pairing a new device will overwrite that.

trman:
> Removing a persisted device is a non issue, as the way the store is currently getting setup is to just only store one devices worth of security info and pairing a new device will overwrite that.

it's a issue  @geekbozu because , if the persist is done   , without any screen to remove device then it will be 
allways be the same device that will be persisted  : only a firwmare flash in a previous version would a allow to add again another device in place of the previous one.

( the passket screen show up at the begin **because** there is no device persisted ( in memory or flash) , it don't show afterward once you have persited , in memory a device.
if the persist is in flash , then it will not show anymore.... unless the previous device is removed from flash (it would have not point  otherwise to persist , if  it not for this case)

geekbozu:
> 
> 
> > Removing a persisted device is a non issue, as the way the store is currently getting setup is to just only store one devices worth of security info and pairing a new device will overwrite that.
> 
> it's a issue @geekbozu because , if the persist is done , without any screen to remove device then it will be allways be the same device that will be persisted : only a firwmare flash in a previous version would a allow to add again another device in place of the previous one.
> 
> ( the passket screen show up at the begin **because** there is no device persisted ( in memory or flash) , it don't show afterward once you have persited , in memory a device. if the persist is in flash , then it will not show anymore.... unless the previous device is removed from flash (it would have not point otherwise to persist , if it not for this case)

Ill let evergreen confirm, but I'm under the belief trying to pair a new device it will replace the previously persisted device. No need to clear the persist it will automatically get overwritten when a new unit is paired.

trman:
and how to pair a new device? by restarting infinitme?
the purpose of the persist seem **is** to store the information of the previous bonded  device **across** restart of infinitme.
then it will not even show a pass screen , since it's in flash already bonded

evergreen22:
Thanks to @Riksu9000 and @Avamander for the helpful comments and thanks to @geekbozu for sharing info on the FS peculiarities.

Writing and reading the bond info into the FS is working, but reconnecting has more work to do before it is ready for review.

kieranc:
I've included this in a test build today and it works as expected when pairing with my phone but I can't get it connected to my desktop, I don't get a prompt for the PIN, the watch doesn't display a PIN, and while it claims to be connected I can't transfer any data.
Using Siglo, when I try to pair it gives me a system dialog asking if i want to pair, pair and trust or deny, I chose pair and trust and the BLE icon is displayed but I can't communicate. 
When I try to pair via the bluetooth device wizard again I get the BLE icon displayed but the pairing fails.

evergreen22:
The encryption/passkey process is triggered when the central reads the
battery state. You will get a "just works" pairing before you read the
battery state (which won't work as expected). In my testing, I am
looking into whether using a different characteristic is better.
However, passkey bonding does work as expected with GB because GB
always queries the battery state when connecting.

geekbozu:
I wonder if we should be choosing to encrypt most endpoints and have it trigger on any encrypted endpoint?

JF002:
@evergreen22  Thank you very much for this PR! As you may have noticed, I added it to the next milestone as I would like to get this feature integrated soon!

What's the current state of this PR? Do you still want to add something? Are there any issues that needs to be addressed? Does it need more test or review?

I'll review it and will probably come back with more questions :)

evergreen22:
I am in the final testing phase for the 'persist' task. I expect to push those changes in the next week along with some changes suggested by other reviewers.

Due to the significant changes in the code base since I started this branch, I plan to rebase and then push. @JF002 if you do _NOT_ want me to rebase before pushing the 'persist bond between reboots' addition, please let me know.

JF002:
@evergreen22 Thanks for the update! I'm looking forward to next commits, then! And I'll do my best to provide feedback asap!
I guess that's ok if you rebase your branch on the current develop, as we commit/merged quite a few changes!

JF002:
@evergreen22 In the mean time, I have a few question regarding this PR.
First, I think it would be nice to list what exactly this new feature will bring. I guess it'll provide security (encryption) and companion apps will be able to reconnect more easily to the watch (without needing to scan and wait for an advertisement message. Am I right? Is there anything else?

Also, If I understand correctly, bonding is triggered when the companion apps requests to read the battery level. Is it definitive or temporary as a test? What is the "state of the art" regarding bonding? What should trigger the start of a bond procedure?

evergreen22:
TL;DR secure connection and seamless reconnects

The BLE link is encrypted with MITM protection after passkey pairing with this PR. The pairing process will require the user to read the randomly generated six digit code displayed on the watch and then enter it on the phone. This will mitigate the risk of third parties connecting to the watch and exfiltrating information.

Furthermore, the watch will become a "trusted device" in Android. This enables the capability to automatically unlock the phone when the watch is present.

After the phone and watch are separated and disconnect (out of range) and then return in range they will automatically securely reconnect without requiring pairing again. This is a combination of this PR and my previous BLE refactoring PR which was previously merged.

Finally, when the watch is rebooted, the bond is restored. The bonding information will automatically be restored and the watch and phone will automatically securely reconnect without requiring pairing again.

> Also, If I understand correctly, bonding is triggered when the companion apps requests to read the battery level. Is it definitive or temporary as a test? What is the "state of the art" regarding bonding? What should trigger the start of a bond procedure?
> 

The pairing encryption process is triggered when either the central or peripheral attempts to read a characteristic and the insufficient authorization message is returned. Therefore, we need at least one characteristic on the watch that requires authentication and encryption.  This PR changes the battery level characteristic to require encryption and authentication. I chose battery level because GB always reads the battery level upon connection. 

It would be nice to know if other companion apps also automatically read the battery level. @kieranc noted that the siglo desktop app doesn't behave as expected when pairing from the desktop (PC). It would be helpful if someone could capture the connection sequence and share the data. When I manually connect from nRFConnect the pairing doesn't bond until I read the battery characteristic. This is expected.

FWIW, I did try using CTS and device info in prior testing but neither worked. I'll revisit those.

src/displayapp/screens/PassKey.cpp
evergreen22: Agreed

src/displayapp/DisplayApp.cpp
evergreen22: Agreed

src/displayapp/screens/PassKey.cpp
evergreen22: Agreed

geekbozu:
Personally if there is minimal penalty, I think ALL characteristics should be encrypted. Which would solve that issue.

evergreen22:
I didn't consider execution time as my criteria. I tried different characteristics to trigger the auth+enc process and battery was the only one that worked with GB and nRFConnect.

Neither CTS or ANS worked in my testing.

I agree, in principle, it would be best to require auth+enc on all characteristics. I'll add that to my list of future work to investigate.

src/components/ble/NimbleController.cpp
```diff
@@ -141,10 +154,11 @@ void NimbleController::StartAdvertising() {
   fields.uuids128_is_complete = 1;
   fields.tx_pwr_lvl = BLE_HS_ADV_TX_PWR_LVL_AUTO;
 
-  rsp_fields.name = (uint8_t*) deviceName;
+  rsp_fields.name = reinterpret_cast<const uint8_t*>(deviceName);
```
Avamander, line 157: Why is it `reinterpret_cast`?

src/components/ble/NimbleController.cpp
```diff
@@ -141,10 +154,11 @@ void NimbleController::StartAdvertising() {
...
   rsp_fields.name_len = strlen(deviceName);
   rsp_fields.name_is_complete = 1;
 
+  int rc;
   rc = ble_gap_adv_set_fields(&fields);
```
Avamander, line 162: Initialize on the same line as definition?

src/components/ble/NimbleController.cpp
```diff
@@ -45,16 +48,18 @@ NimbleController::NimbleController(Pinetime::System::SystemTask& systemTask,
...
 
 void nimble_on_sync(void) {
-    int rc;
+  int rc;
 
-    rc = ble_hs_util_ensure_addr(0);
-    ASSERT(rc == 0);
+  NRF_LOG_INFO("Nimble is synced");
+
+  rc = ble_hs_util_ensure_addr(0);
```
Avamander, line 59: I think you could initialize and define on the same line here as well.

src/components/ble/NimbleController.cpp
```diff
@@ -199,18 +211,45 @@ int NimbleController::OnGAPEvent(ble_gap_event* event) {
...
+      NRF_LOG_INFO("encryption change event; status=%0X ", event->enc_change.status);
+      break;
+
+    case BLE_GAP_EVENT_PASSKEY_ACTION:
+      /* Authentication has been requested for this connection.
+       * Standards insist that the rand() PRNG be deterministic.
+       * Use the nimble TRNG since rand() is predictable.
+       */
+      NRF_LOG_INFO("Security event : BLE_GAP_EVENT_PASSKEY_ACTION");
+      if (event->passkey.params.action == BLE_SM_IOACT_DISP) {
```
Avamander, line 241: Is there only a single action that can happen here?

lman0:
Wow , the persist commit disappeared! (I received el commit notification in box mail).
Is it github that acting up?

Avamander (changes requested):

Avamander:
> I'll add that to my list of future work to investigate.

I think this is a good idea for one more reason, requiring bonding might be a reason for a major version increment as it's a very strong API breakage.

src/components/ble/NimbleController.cpp
evergreen22: you request I lose the C-style cast ni a previous review :-)

src/components/ble/NimbleController.cpp
evergreen22: I consider that bad style and we don't have a rule for it here

src/components/ble/NimbleController.cpp
evergreen22: I consider that bad style and we don't have a rule for it here

src/components/ble/NimbleController.cpp
evergreen22: No, but this is the only one I care about.

Avamander: The others are safe to not handle? Please leave a comment

src/components/ble/NimbleController.cpp
Avamander: Yes, I guess I should've been more clear, does `static_cast` not work?

evergreen22: I have vague memories of trying that first and it did not work. I'll try it (again?) and report back.

src/components/ble/NimbleController.cpp
evergreen22: Great suggestion! Will do.

src/components/ble/NimbleController.cpp
evergreen22: I did try a static_cast first and here is what I found.

Using a static_cast won't allow casting from a signed to unsigned type so I need to reinterpret. For some inexplicable reason the device name is a char* everywhere in nimble except the response fields where it is a const uint8_t*

JF002:
> The pairing encryption process is triggered when either the central or peripheral attempts to read a characteristic and the insufficient authorization message is returned. Therefore, we need at least one characteristic on the watch that requires authentication and encryption. 

Once the pairing encryption process is successful, will all following exchanges be encrypted? Or will it be enabled only on characteristics that are explicitly secured (`BLE_GATT_CHR_F_READ_ENC | BLE_GATT_CHR_F_READ_AUTHEN`)

I guess there is no way for the central or peripheral to request the pairing encryption process at connection time, even before a characteristic is accessed?

> I agree, in principle, it would be best to require auth+enc on all characteristics.

Do BLE/GATT allow this? Can we enable security feature on all services/characteristics and still be "BLE compliant"? Or do some of them still need to be available without those features?

> I think this is a good idea for one more reason, requiring bonding might be a reason for a major version increment as it's a very strong API breakage.

Yep, I was going to ask if companion apps needed to do anything to support these new security features and checked with **Amazfish** : it does not work properly anymore. I guess it fails when trying to read the battery level. I'm not sure if there's a generic way to implement bonding on Linux apps or if the companion app need to implement it from scratch, but I guess that means that enabling security is a *major* API break.

So now I'm wondering : should we
 - enable auth+sec on a characteristic every companion app will read to ensure that the secure bonding process will be triggered ?
 - enable them on all services/characteristics so we are 100% secure
 - enable 2 modes:
   - legacy/unsecure : infinitime will work exactly as it is working now which would require no work from companion app
   - secure : auth+enc is enabled, and additional characteristics are available
 - anything else?

JF002:
FYI I just did a size check : this PR uses 2156KB in Flash. I have to assume I was afraid it would be much more than that! :+1:

Avamander:
I would try avoiding complexity and not make this toggle-able once implemented. This saves both companion complexity and doesn't increase this PR's scope.

trman:
hi,

@evergreen22  first thanks for adding the persist , and add way to remove bonded state !

but sadly there is a big bug when i tested the pr  .
 here the step to reproduce it :
1-first bond the smartphone with infinitime ( with the 6 digit of this pr)

 2 a -  once complete bonded  (gadget bridge show 'connected' ) reset infinitime (hold around 5 sec the button)
2- b - during reset stop the smartphone bluetooth
3-  when infinitme is fully restarted ,  wait 3 sec (or don't wait?) and start the smartphone bluetooth
--->  smartphone (and gadgetbridge) show that infinitme is connected
4) crash of infinitme (the bug) and reset
5)  infinitme lose the bond and want again to be bonded with 6 digit passkey

evergreen22:
> Once the pairing encryption process is successful, will all following exchanges be encrypted? Or will it be enabled only on characteristics that are explicitly secured (`BLE_GATT_CHR_F_READ_ENC | BLE_GATT_CHR_F_READ_AUTHEN`)
> 

I don't have any way to find out since I don't own a BLE sniffer. It would be useful if someone who does own a sniffer could do a capture and analyze it to determine if all traffic is encrypted after bonding. FWIW nimble tells me that we are encrypted, authenticated, and bonded after the passkey pairing is complete, but nimble has lied to us before, so I don't really know.

> I guess there is no way for the central or peripheral to request the pairing encryption process at connection time, even before a characteristic is accessed?
> 

Based on my understanding of comments in nimble and the ble spec, only the central should initiate a secure connection. Of course that would require a change in every companion app as well as changes in InfiniTime.

The approach I've taken is to return an auth+enc required message to the central to force it to initiate security. This is how the spec says to do it (as well as several online guides). This, obviously, does not require a change in the companion as long it already handles the auth+enc required response.

> Do BLE/GATT allow this? Can we enable security feature on all services/characteristics and still be "BLE compliant"? Or do some of them still need to be available without those features?
> 

Yes, again based on my understanding of the spec, this would be allowed. As @Avamander and @kieranc pointed out previously it has the potential to cause a lot of breakage.

> Yep, I was going to ask if companion apps needed to do anything to support these new security features and checked with **Amazfish** : it does not work properly anymore. I guess it fails when trying to read the battery level. I'm not sure if there's a generic way to implement bonding on Linux apps or if the companion app need to implement it from scratch, but I guess that means that enabling security is a _major_ API break.
> 

Try re-anabling "Legacy pairing" and see if that fixes Amazfish.

I disabled it to force all messages to be encrypted. Since I don't own a ble sniffer I can't prove, or disprove, that it does force an encrypted link for all messages. But, based on my understanding of the ble spec it does. I invite someone with a ble sniffer to confirm or refute my claim.

> So now I'm wondering : should we
> 
>     * enable auth+sec on a characteristic every companion app will read to ensure that the secure bonding process will be triggered ?
> 

That is exactly my reasoning for requiring auth+enc on battery level. I use GB and it always reads the battery level on connect.

As stated before, I also tried the CTS and device information characteristics but they did not work for me. I concede that was in early days so I'll probably revisit them (future work).

>     * enable them on all services/characteristics so we are 100% secure
> 

I added this to future tasks to determine feasibility.

IMO ble cannot ever be 100% secure because the architecture is fundamentally insecure.

>     * enable 2 modes:
>

My gut says no - the complexity would lead to a frustrating user experience for those who failed to grasp the pairing process and limitations (for example - people who tried this PR but didn't read the commit messages to understand what exactly it does and does not do). Also, the additional complexity would likely mean the code would be hard(er) to understand, debug, and maintain.

>       
>       * legacy/unsecure : infinitime will work exactly as it is working now which would require no work from companion app
>

Abandoning this PR would mean the watch would not work as a trusted device on Android (unlock your phone). Otherwise, I don't see any real disadvantage to the project in skipping this PR. BLE connections are reliable and easy now without bonding (thanks to everyone who spent hours and hours on the ble issues).

>       * secure : auth+enc is enabled, and additional characteristics are available
> 
>     * anything else?

In general, I've been trying to improve the reliability and usability of the BLE side of the watch. My pebble seamlessly bonded, reconnected, and unlocked my phone and those are all features I personally wanted in InfiniTime.

I think this PR accomplishes my goals with only one minor nit. The bond is not available to be stored until there is a disconnection or a completed successful encryption change. Maybe upgrading nimble to a newer version (or sub-module) would resolve this. However, that is a problem for another day/PR.

trman:
@evergreen22  did you have checked the bug  i've , found ? 
because i have updated the step, (i forgot a step) , so you should check it again

JF002:
Thanks for all these info, @evergreen22 !

The plan is not to drop this PR, quite the contrary! Making the connection easier will certainly improve the user experience, and secured (encrypted) connection is a must in our times!

I can setup a BLE sniffer using a laptop and a NRF52-DK board, I'll try to do it soon to confirm encrypted communication !

Regarding the bonding, indeed, it looks like the central can request it : NRFConnect has a "bond" button. So I guess companion app will be able to trigger the bonding process themselves, without needing to read a specific characteristic.

Now, do we want to keep the CTS (or any other characteristic like the one for the notification, which requires more privacy) to force the bonding process? Or should we create a new characteristic specific to infinitime for that purpose?

Sooner today, I did a few tests, but couldn't read the battery level from Gadgetbridge and NRFConnect. Gadgetbridge would just connect but did not display the battery level. NRFConnect would connect too, bond when I tried to read the battery level, but never displayed the battery level. According to the log, it was bonded but it never requested the PIN code (and infinitime never displayed it). I'll do more test to give more info about this!

trman:
@JF002  , @geekbozu   did you have tested the bug (or it's a feature?) i 've found , 
i don't know if it's only related to my infinitme or it's truly a bug on this pr.
'btw , with the step that have been updated , it crash always at 4 step

darkdragon-001:
What is the current approach to multiple devices (e.g. Phone+Laptop)? Is only one connection supported at the moment and re-pair required to change the device? Or are multiple connected devices (maybe also at the same time) supported? If multiple devices are supported, I think there is need for a forget device UI. If pairing a new device effectively deletes the old key, I think such an UI is not needed.

evergreen22:
InfiniTime only supports a single connection at a time.

You'd have to disconnect and repair to a second device. This PR handles the possibility that you might do this.

picomatic:
I'm not a bluetooth expert, but merged this pr into my build, and it works very well!

trman:
Hi @evergreen22 ,

I've tested the latest build and I don't have anymore the crash so far ,so all good for me!!.

I 'be seen that you have explained on comments how to reset the persist but it should be good as well to add an explanation on github MD wiki because user would not know how to do it.

JF002:
This PR have already received some good feedback! Thanks to everyone who participated in this work, and especially to @evergreen22 !

JF002:
I did a quick sniffing test:

InfiniTime 1.7.1 : everything is visible : characteristic handles, BLE commands and data ("Hello world" notification in this case :
unencrypted

Using changes from this PR:
encrypted

I'm not a security expert but... it looks good to me :+1:

heinrich-ulbricht:
Just want to leave POSITIVE feedback: before this new passkey bonding I could hardly use the watch. But ever since the latest update and the passkey bonding I never ever encountered a single Bluetooth connectivity issue again, no matter how often I go in and out of flight mode or out of range and back.

Thank you thank you thank you!

JF002:
@heinrich-ulbricht Thank you very much for this nice feedback! I'm glad this feature is working as well for you as it has for me and many other contributors and users!

evergreen22:
@heinrich-ulbricht thanks for the kind words. We are all volunteers and is always nice to know our work is appreciated.

Issue #674, opened 2021-09-18
Use Docker in GitHub Actions?


JF002:
So, currently, the [Github Actions script](https://github.com/JF002/InfiniTime/blob/develop/.github/workflows/main.yml) runs the builds in a "github" virtual machine based on Ubuntu. 
As I have already said, I don't know Github Actions that much and barely use them, but they are really useful for other developers, testers and allow to check that the branches build before merging them (and will also allow to add more tests in the future).

My only complain about Github Actions is that they are described in this specific `yml` file and run into a VM from Github, which means that the build chain is specific to github and cannot be reproduced on developers dev stations or moved to another CI.

Also, this `yml` file is more or less a duplicate of the [Docker container](https://github.com/JF002/InfiniTime/tree/develop/docker) we've also added in the project. I really like this container as it allows anyone to build the whole project on their computer without needing to setup a complex dev environment. 
When I first created this docker container, my plan was to use exactly the same container on dev computers and on the CI. Unfortunately, installing, managing and maintaining a CI build server takes time, and I prefer to spend my time working on InfiniTime instead of a CI infrastructure.

Now, I've recently discovered it should be possible to use [the Docker container into the Github Actions workflow](https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action). I think it would be great to use this feature, so that we ensure that the builds we do on our computers and the ones built on Github are the same (same versions, sames options, same tools).
I think this would also allow to improve the build chain (with more features, tests,...) without having to do it twice (once in Docker, once in Github Actions).

Is anyone up to the task to move our Actions to Docker  ?

JF002:
@ntindle Exactly! We already have the build [docker container](https://github.com/InfiniTimeOrg/InfiniTime/tree/develop/docker). It builds the entire firmware and all the variations (standalone firmware, firmware for MCUBoot, MCUBoot image file, DFU, recovery firmware). I personally use it exclusively to generate binaries for the [releases](https://github.com/InfiniTimeOrg/InfiniTime/releases).

The Github Action is nice and allows to easily build binaries without a dev environmnent on the computer, but it's currently not complete (only builds the dfu file) and it's basically the same as the build docker written into another form.

So the idea is to use the docker container into the Github action so that we avoid the code duplication and ensure it's working fine and it's maintained. If I understand the doc correctly, this should be possible, but someone has to do it ;)

carlosperate:
[...]

PR #799, opened 2021-10-31
Add the name of the current GIT branch in the filenames

### LINKED ISSUE #795: Automatically rename workflow artifacts after the PR
### 

### Pitch us your idea!

Rename the artifact so I can tell which one it is after downloading.

### Description

It would be nice if the dfu artifact was automatically renamed to something like `infinitime-pr-1234`. That way I can tell what they are after downloading, especially if I have multiple downloaded.

JF002:
Add the name of the current GIT branch in the name of the files that are generated during the build (bin, hex, zip,...).
If the branch name is master, the filenames do not contain the name of the branch.

Fixes #795


JF002:
This PR works perfectly fine on my setup and on docker, but not on Github Action :/

This is one additional reason to try to implement https://github.com/InfiniTimeOrg/InfiniTime/issues/674...

trman:
I see , that you didn't have modified the github action itself.
It's most likely here that need to do the modification?
Like taking the final file and rename it in github action

Here a link a found by searching how to use branch name on github action:
https://stackoverflow.com/questions/58033366/how-to-get-current-branch-within-github-actions

JF002:
My changes are made in CMake files. As CMake must be executed in every build environments (local, docker, devcontainer, github action), all these environments should benefit from this single modification.

And it works (at least on native builds and docker builds). The only change I expected on the git workflow was to rename a filename or 2.

But that's not the case. Here, the workflow fails at getting the git hash and branch

```
PROJECT_GIT_COMMIT_HASH_SUCCESS? 0
fatal: ref HEAD is not a symbolic ref
CMake Error at CMakeLists.txt:77 (string):
  string sub-command STRIP requires two arguments.
```

Why would CMake and Git not work correctly in github action? It makes no sense to me.

If we follow the instructions from the link you posted, it will probably work in github action, but not in any other build env like local or docker.

That's why I think github action should run our docker container (which I personally maintain and use to build released versions of InfiniTime) to build the project instead of using an alternative (duplicated) description of the build process in the .yml file.

trman:
For the few ,I managed to understand @JF002  docker on github is a black box that output nothing but the result file.
Unlike the github action where all steps are swing by people like me...
So if docker no one but Dev could help because they can use docker.
Where user could point breakage on github action steps...

trman:
By the way ,even if you add the docker to github action , @JF002  it seem that you still need to say to github action which file (name) to take so you still need to modifiy github action at the end ....

JF002:
Yeah sorry, It's not easy for me to explain my thoughts :) And my view on the ideal CI is probably in conflict with the state of the art of the CI/CD platforms hosted in the cloud.

Basically, I do not know github actions and barely use them. Also, I don't want to the project to be dependent of Github (or any other platform). We should be able to build and maintain the project independently of the hosting platform. That's why I emphasize on the possibility for anyone to build the whole project on their own computer, either 'natively' or via a Docker image. This way, even if github break down, go offline, decide to stop supporting open source, or whatever, we'll be able to switch platform easily.

The way github actions work forces us to write the build steps (environment configuration, cmake generation, build, and files generation) 2 times : once for local build and once for github builds. That makes 2 build chains to support, maintain and debug, which is, in my opinion, a waste of time and energy. And we have no guarantee that those 2 chains will produce the same results.

Being able to run the exact same build process on my pc and on github action looks really comfortable. The facts that github does not display the output of a docker build and that we have to edit the workflow each time the output filename changes  look like a major issue to me.

Any idea on how other FOSS project use github action for their workflows?

trman:
> Any idea on how other FOSS project use github action for their workflows?

i thinks you mean't how other project manage to not use github action , right?
because it's seem you don't like at all the github action , and if this the case you should seek external ci to do the job
like for instancehttps://dev.to/coteh/6-continuous-integration-tools-you-can-use-for-an-open-source-project-313m

most opensource project use github action for cloud and docker for local (some have on their git the gitlab or drone ci  file as well for selfhost)


but regardless of what you do , you still need to do a lot of work to make it functional...

JF002:
> because it's seem you don't like very much the github action

I didn't mean to complain about github actions. I guess I'm trying to use them the way I want to instead of the way they are intended to be used.

I just a bit surprised that this simple modification in cmake files make the project non-functional on github actions.

I think github actions do a great job for a lot of people and projects, so that's nice! I only I could use them just by pointing to my Dockerfile, but that's not how it works :)

trman:
i think that you need to only change the name value on this part of the main.yaml:
      - name: Upload DFU package
        uses: actions/upload-artifact@v2
        with:
          name: pinetime-mcuboot-app-dfu
          path: build/src/pinetime-mcuboot-app-dfu/*
          
     Since it create a zip with pinetime-mcuboot-app-dfu , with the value from the links i provided would resolve this pr without 
     any change to docker or anything....
     
     I you thinks it's useful for dev to have a generated name from branch  then you should make it in docker result as well 
     in the docker file as well (not only on cmake)

trman:
i think that  @JF002 you need to only change the name value on this part of the main.yaml:
  ```
    - name: Upload DFU package
        uses: actions/upload-artifact@v2
        with:
          name: pinetime-mcuboot-app-dfu
          path: build/src/pinetime-mcuboot-app-dfu/*
          
```

Since it create a zip with 'pinetime-mcuboot-app-dfu' as name , you could replace the name with an variable  with the value from the links i provided would resolve this pr  (for the github actions)
 
And you do ,for docker , the same things  that is  renaming at the very end in another file aka the post_build.sh 
the final name with the github branch .

so working with github action is not that different

trman:
by the way @JF002  on the way i provided aka 
https://stackoverflow.com/questions/58033366/how-to-get-current-branch-within-github-actions

amongs the response there is one that use docker in github
for info

Riksu9000:
@JF002 Will you be working on this?

JF002:
> @JF002 Will you be working on this?

IIRC, the main blocker was that it was not working on Github Actions, but this should not be an issue anymore since it's now based on the docker container. I also think it has been implemented another way in this PR (https://github.com/InfiniTimeOrg/InfiniTime/pull/1115) so I guess we can close this one.

PR #801, opened 2021-10-31
Remove CMake targets specific to the debugging environment

JF002:
Remove CMake targets that allowed to flash the firmware and erase the flash memory of the CPU.

These targets were enabled/disabled by CLI parameters `USE_JLINK`, `USE_GDB_CLIENT` and `USE_OPENOCD`. 

These targets were not easily portable on every developers setups, were mostly confusing (you had to enable the openocd mode if you didn't want to use any of the other possibility), and I think very few people actually used them.

I think the CMake project files should only focus on building the project. Other things like accessing the hardware, configure the debugger, flash the firmware, erase the memory,.. should not be part of the project as it'll most probably differ from one developer to another.


Avamander (approved):
Great. I thought about doing the same at one point, but people used it extensively.

evergreen22:
Agreed because as stated "These targets were not easily portable...". It should be fairly easy to build the project to encourage participation. Folks using hardware debuggers to flash and test can (and should) look at the documentation on these topics.

kieranc:
Hmm. Well I used them and found them useful.

JF002:
> Hmm. Well I used them and found them useful. I guess I can figure out a different approach.

Interesting! What config do you use (jlink, openocd, gbd)? I can add instructions in the doc if they are missing!

Riksu9000:
While I only use DFU, I think these targets could be convenient and lower the barrier to entry. I would probably use them if I had a devkit. I think instead of `USE_JLINK`, `USE_GDB_CLIENT` and `USE_OPENOCD`, all the targets could always be enabled, but with different names like `jlink_flash`, `jlink_erase` etc.

geekbozu:
I would like to see these removed from the cmake as well. Let the cmake focus on getting a build. Leave the flashing to the other tools/user setup available. 
@kieranc we can totally work on getting documentation setup so you can continue to work without them and understand the tools instead of this black box!

JF002:
> Hmm. Well I used them and found them useful. I guess I can figure out a different approach.

@kieranc Have you already found a new approach to build the project? If not, what would you need to be able to work without those targets? Documentation? Examples? Scripts? IDE support?

kieranc:
Don't worry about me, if the consensus is that this is a good move then I don't want to hold you back, I will certainly figure out a solution once I do wired programming again, and document it!

JF002:
You're probably not the only one to use those targets. If I can do anything to help them too, that would be nice :)

geekbozu (approved):

StarGate01:
In response to https://github.com/InfiniTimeOrg/InfiniTime/pull/1051#issuecomment-1159796455 :

Personally, I do not use any of these targets. I invoke `nrfjprog` manually or via scripts (see https://github.com/StarGate01/p8b-infinitime/tree/master/scripts ). I can see how the targets might be appealing to newcomers, but I felt rather restricted by the somewhat cumbersome CMake configuration.

Riksu9000:
Since merging #1236 I'm fine with removing these, as I believe it's now easy enough to find the instructions.

PR #803, opened 2021-10-31
InfiniPaint vibrate on colorchange, fix color rotation

clemensvonmolo:
This is a very simple pull requests that changes the Paint app in order to vibrate on color change (because It's hard to know if the touch controller actually registered a long press or a small swipe). I also fixed the logic for the color rotation. Before, the color magenta was unreachable and instead black was chosen a second time. The `color`  variable was changed to initialize to 3 instead of 2, so you don't have to press twice on the first rotation.

- [x] the changes were successfully tested on hardware and showed no problems.

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -410,7 +410,7 @@ void DisplayApp::LoadApp(Apps app, DisplayApp::FullRefreshDirections direction) 
       currentScreen = std::make_unique<Screens::Twos>(this);
       break;
     case Apps::Paint:
-      currentScreen = std::make_unique<Screens::InfiniPaint>(this, lvgl,motorController);
+      currentScreen = std::make_unique<Screens::InfiniPaint>(this, lvgl, motorController);
       break;
     case Apps::Paddle:
       currentScreen = std::make_unique<Screens::Paddle>(this, lvgl);
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -410,7 +410,7 @@ void DisplayApp::LoadApp(Apps app, DisplayApp::FullRefreshDirections direction) 
       currentScreen = std::make_unique<Screens::Twos>(this);
       break;
     case Apps::Paint:
-      currentScreen = std::make_unique<Screens::InfiniPaint>(this, lvgl);
+      currentScreen = std::make_unique<Screens::InfiniPaint>(this, lvgl,motorController);
       break;
     case Apps::Paddle:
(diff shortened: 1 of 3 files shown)


0x1a8510f2:
Whoa, didn't even realise the colours could be changed!

geekbozu:
Not super related to this PR but I would love to see a way to just set a system event for the motor instead of having to pass this around so much....Nice pr tho didn't know you could change the color.

src/displayapp/screens/InfiniPaint.cpp
```diff
@@ -4,7 +4,7 @@
 
 using namespace Pinetime::Applications::Screens;
 
-InfiniPaint::InfiniPaint(Pinetime::Applications::DisplayApp* app, Pinetime::Components::LittleVgl& lvgl) : Screen(app), lvgl {lvgl} {
+InfiniPaint::InfiniPaint(Pinetime::Applications::DisplayApp* app, Pinetime::Components::LittleVgl& lvgl, Pinetime::Controllers::MotorController& motor) : Screen(app), lvgl {lvgl}, motor{motor} {
```
Riksu9000, line 7: This line is too long now. Use clang-format to break it up.

src/displayapp/DisplayApp.cpp
```diff
@@ -410,7 +410,7 @@ void DisplayApp::LoadApp(Apps app, DisplayApp::FullRefreshDirections direction)
       currentScreen = std::make_unique<Screens::Twos>(this);
       break;
     case Apps::Paint:
-      currentScreen = std::make_unique<Screens::InfiniPaint>(this, lvgl);
+      currentScreen = std::make_unique<Screens::InfiniPaint>(this, lvgl,motorController);
```
Riksu9000, line 413: Reformat with clang-format.

Riksu9000 (changes requested):
A proper fix for the color would be to increment the color before the switch and modulo with 8. That way we use 0-7 instead of 1-8. Currently the color variable points to the next color instead of the current one which is strange.

The vibration is a good idea.

Riksu9000 (commented):
Maybe the vibration doesn't have to be so strong. 35 duration is usually used elsewhere.

Other than that this looks good.

clemensvonmolo:
The run duration seems a bit all over the place. (Currently 50, 90, 30, 35 and 15 are used.) Maybe a design guideline with times for strong/medium/weak vibration would help. For now I will change it to 35 since that really is the most common one.

Riksu9000 (approved):

geekbozu (approved):

PR #743, opened 2021-10-13
Lvgl simulation run PineTime screens on your PC

NeroBurner:
Add simulator to have pinetime Screens on PC for development.

The first Screen can be run locally, using your mouse to play Paddle.h!

2021-10-13_lv_sim_paddle

based on lvgl 8 port #734 

lv_sim based on: https://github.com/lvgl/lv_sim_eclipse_sdl

---
2021-11-03 updated screens with PineTime fonts

Screenshot_20211103_214140 Screenshot_20211103_214323 Screenshot_20211031_100600 

---
2021-11-07 new Screen Metronome and the first simulated external: MotorController (shown by red dot)

Metronome.h

---
2021-11-15 new Screen `BatteryInfo.h` with a very low Battery, according to the `BatteryController.h` stub

BatteryInfo.h

---
2021-11-28 rebased on `develop` branch getting the new `v1.7.1` version of `FlashLight.h`

flashlight_light flashlight_dark

---
2021-11-30 All three PineTime Watch Faces are working now: `WatchFaceDigital.h`, `WatchFaceAnalog.h` and `PineTimeStyle.h`

WatchFaceDigital WatchFaceAnalog PineTimeStyle

---
2021-12-02 Can trigger a Notification, got the system time, and the first Settings-Screen `QuickSettings.h`

20211202_WatchFaceDigital 20211202_WatchFaceAnalog 20211202_PineTimeStyle 2021-12-02_Quicksettings

---
2021-12-03 Fix "score" alignment in `Twos.h` by upgrading lvgl submodule to the current master (`v8.1.0-64-g628299676`) 

2021-12-03 Twos fixed align and with lvgl-8.1-master
---

2022-01-16 Ported the simulator to use lvgl7 like current PineTime firmware! much less porting and maintenance burden! :tada: And `Music.h` has shown up!

WatchfaceDigital WatchfaceAnalog WatchfacePineStyle

Paddle Twos Metronome

Meter Battery

Torch Black Torch White QuickSettings

Music InfiniPaint

---

2022-01-21: `SystemInfo.h`'s first Screen, others available, but no swipe yet

SystemInfo Screen 1

---

2022-01-25: Swiping now works and most Screens are implemented

2022-01-25_lv_sim_swiping

---

2022-01-27: add `SettingSetDate.h` and `SettingSetTime.h`

SettingsSetDateSettingSetTime

---

2022-01-29: add `Screen is OFF` overlay when brightness is set to `Off`. Also add `Notifications.h`

[...]

PR #805, opened 2021-11-01
Non-relative includes, restructure, add missing includes

NeroBurner:
This is a two-part PR, first use relative includes, and secondly restructure all includes.

The non-relative includes are needed for me in #743 to inject my header-stubs for the simulator and is a more robust coding style for header files

---

First commit message:

Don't use relative imports like `../foo.h` as those depend on the
relative position of both files. Rather than that use imports relative
to the `src` directory, which explicitly is part of the include
directories.

---

Second commit message:

Update the include structure:
- use `#include "foo.h"` for headers from the InfiniTime project (except
  bundled libraries)
- use `#include <foo.h>` for everything else (external libraries or standard
  library)

Change the include order from the most project specific down to the most
generic include. This helps to enforce, that our headers include what
they use. For example if `foo.h` uses `std::array` but forgets the
include. This won't be cought if `main.cpp` includes `<array>` before
`foo.h`.

For example `DeviceInformationService.h` did miss `<stdint>` include for
`uint16_t` and `BootloaderVersion.h` missed `<stddef>` for `size_t`.
As last example `InfiniPaint.cpp` missed include `<algorithm>` for `std::fill`.

The following include order is used:
- project headers
- nimble library
- littlefs library
- date library
- lvgl library
- FreeRTOS library
- nrf5 sdk
- standard libraries

### DIFF
--- a/src/components/alarm/AlarmController.cpp
+++ b/src/components/alarm/AlarmController.cpp
@@ -15,7 +15,7 @@
     You should have received a copy of the GNU General Public License
     along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
-#include "AlarmController.h"
+#include "components/alarm/AlarmController.h"
 #include "systemtask/SystemTask.h"
 #include "app_timer.h"
 #include "task.h"
--- a/doc/contribute.md
+++ b/doc/contribute.md
@@ -94,6 +94,7 @@ If there are no preconfigured rules for your IDE, you can use one of the existin
  - **Includes** :
     - files from the project : `#include "relative/path/to/the/file.h"`
     - external files and std : `#include <file.h>`
+    - use includes relative to included directories like `src`, not relative to the current file. Don't do: `#include "../file.h"`
  - Only use [primary spellings for operators and tokens](https://en.cppreference.com/w/cpp/language/operator_alternative)
  - Use auto sparingly. Don't use auto for [fundamental/built-in types](https://en.cppreference.com/w/cpp/language/types) and [fixed width integer types](https://en.cppreference.com/w/cpp/types/integer), except when initializing with a cast to avoid duplicating the type name.
  - Examples:
--- a/src/BootloaderVersion.h
+++ b/src/BootloaderVersion.h
@@ -1,5 +1,8 @@
 #pragma once
 
+#include <cstdint>
+#include <cstddef>
+
 namespace Pinetime {
   class BootloaderVersion {
   public:
--- a/src/Version.h.in
+++ b/src/Version.h.in
@@ -2,6 +2,8 @@
 
 @VERSION_EDIT_WARNING@
 
+#include <cstdint>
+
 namespace Pinetime {
   class Version {
     public:
--- a/src/components/battery/BatteryController.cpp
+++ b/src/components/battery/BatteryController.cpp
@@ -1,4 +1,4 @@
-#include "BatteryController.h"
+#include "components/battery/BatteryController.h"
 #include "drivers/PinMap.h"
 #include <hal/nrf_gpio.h>
 #include <nrfx_saadc.h>
--- a/src/components/ble/AlertNotificationClient.cpp
+++ b/src/components/ble/AlertNotificationClient.cpp
@@ -1,6 +1,6 @@
-#include "AlertNotificationClient.h"
+#include "components/ble/AlertNotificationClient.h"
 #include <algorithm>
-#include "NotificationManager.h"
+#include "components/ble/NotificationManager.h"
 #include "systemtask/SystemTask.h"
 
 using namespace Pinetime::Controllers;
--- a/src/components/ble/AlertNotificationClient.h
+++ b/src/components/ble/AlertNotificationClient.h
@@ -7,7 +7,7 @@
 #include <host/ble_gap.h>
 #undef max
 #undef min
-#include "BleClient.h"
+#include "components/ble/BleClient.h"
 
 namespace Pinetime {
 
--- a/src/components/ble/AlertNotificationService.cpp
+++ b/src/components/ble/AlertNotificationService.cpp
@@ -1,8 +1,8 @@
-#include "AlertNotificationService.h"
+#include "components/ble/AlertNotificationService.h"
 #include <hal/nrf_rtc.h>
 #include <cstring>
 #include <algorithm>
-#include "NotificationManager.h"
+#include "components/ble/NotificationManager.h"
 #include "systemtask/SystemTask.h"
 
 using namespace Pinetime::Controllers;
--- a/src/components/ble/BatteryInformationService.cpp
+++ b/src/components/ble/BatteryInformationService.cpp
@@ -1,5 +1,5 @@
-#include <nrf_log.h>
 #include "components/ble/BatteryInformationService.h"
+#include <nrf_log.h>
 #include "components/battery/BatteryController.h"
 
 using namespace Pinetime::Controllers;
--- a/src/components/ble/BatteryInformationService.cpp
+++ b/src/components/ble/BatteryInformationService.cpp
@@ -1,5 +1,5 @@
 #include <nrf_log.h>
-#include "BatteryInformationService.h"
+#include "components/ble/BatteryInformationService.h"
 #include "components/battery/BatteryController.h"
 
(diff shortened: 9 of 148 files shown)


src/components/alarm/AlarmController.cpp
```diff
@@ -15,10 +15,13 @@
...
-#include "AlarmController.h"
+#include "components/alarm/AlarmController.h"
 #include "systemtask/SystemTask.h"
-#include "app_timer.h"
-#include "task.h"
+// nrf5 sdk
+#include <app_timer.h>
+// FreeRTOS library
+#include <task.h>
+// standard libraries
```
Avamander, line 24: I don't think the comments are necessary.

NeroBurner: They add a bit of structure for future contributions to keep the include order. And it would have been helpful for my first exploration of the source code. For example I had no idea, that `<task.h>` is part of FreeRTOS. With the comment this is explicitly listet

Avamander:
Includes were sorted once a while ago, I don't see a reason to redo that part with the potential downside that InfiniTime's code conflicts with any of the libraries'.

Original sort order
```
Sort includes:
* source files include their header files first
* library includes next
* own code includes last
* alphabetical
```

src/components/alarm/AlarmController.cpp
Avamander: Sounds like a dev documentation addition instead of each file.

NeroBurner:
One advantage of the new ordering is, that I was able to find some missing includes, that were hidden by the previous ordering.

I don't understand the comment about conflicting library code, all the targets still compile. Would you please be so kind to explain this point to me so I can understand as well?

Avamander:
> I don't understand the comment about conflicting library code, all the targets still compile. Would you please be so kind to explain this point to me so I can understand as well?

Yes, at the moment. There'd be a potential for conflicts or nasty bugs when we define something and it ends up being used in a library, because ours comes first instead of being last.

NeroBurner:
I think we won't (and shouldn't) define so short and generic macros, that they get picked up in libraries. Those changes would be bad code for me (personally).

Even if that happens we could move those specific library includes up and still get the "include-what-you-use"-checkings

Another way to argue is, that library defines could bleed into our headers and subtly break stuff. I think the argument works in both directions

---

sorry for being pedantic about this. I just think it is "more right" and more resilient with the new ordering. In the end it's your project and your guidelines and I'll adhere to them (after a proper discussion :grin: )

Avamander:
> Another way to argue is, that library defines could bleed into our headers and subtly break stuff. I think the argument works in both directions

Yes, but from the direction of libraries to us it's less error prone.

Avamander:
> I think we won't (and shouldn't) define so short and generic macros, that they get picked up in libraries. Those changes would be bad code for me (personally).

Yes absolutely.

Basically the thing is that I see a very minor upside to not changing the order (less changes + our code affects only our own), but I think e.g. @geekbozu or @Riksu9000 should break the tie here.

Riksu9000:
I would agree with @Avamander here. I don't see enough reason to change it.

src/components/alarm/AlarmController.cpp
```diff
@@ -15,10 +15,13 @@
     You should have received a copy of the GNU General Public License
     along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
-#include "AlarmController.h"
+#include "components/alarm/AlarmController.h"
```
Riksu9000, line 18: Do these cpp file specific headers also need to be non relative? Normally it's pretty safe to assume they're in the same directory.

NeroBurner: not strictly necessary, just consistent

NeroBurner:
reverted ordering commit, readd the found missing standard includes

src/components/alarm/AlarmController.cpp
Riksu9000: I would personally rather not type the entire path for these cases as it seems unnecessary. For everything else it's fine.

Riksu9000:
Since we're already restructuring, maybe the different sections should be separated by an empty line, which would make the sorting order more obvious and then clang-tidy wouldn't suggest sorting them in a different way.

src/components/alarm/AlarmController.cpp
NeroBurner: @geekbozu @Riksu9000 what's your stance on this?

- :+1: for consistently using `#include "components/alarm/AlarmController.h"`
- :-1: for using the shorter `#include "AlarmController.h"` (IMHO looks inconsistent as soon as a include from the same component is directly after the main header include)

Riksu9000: I have been trying to think about which would be better, but honestly I'm fine with either. If no one has anything to add, then we can just leave it as it is.

Riksu9000:
If every include must be non relative, it should be mentioned here https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/doc/contribute.md#coding-style

src/components/alarm/AlarmController.cpp
Avamander: I don't think we'll end up with name conflicts, unless it's something generic. But I guess it wouldn't hurt to use the full path

geekbozu: I'm a fan of verbosity, while it is a pain to have them long...its only a few more keystrokes for explicitness, And explicit saves headache later.

NeroBurner:
moving up the file header include to the top, one example from `WatchFaceAnalog.cpp`:

```cpp
#include "displayapp/screens/WatchFaceAnalog.h"
#include <lvgl/lvgl.h>
#include "displayapp/screens/BatteryIcon.h"
#include "displayapp/screens/BleIcon.h"
#include "displayapp/screens/NotificationIcon.h"
```

Is this OK? Or should there be a space before and after the `lvgl/lvgl.h` include?

NeroBurner:
> If every include must be non relative, it should be mentioned here https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/doc/contribute.md#coding-style

@Riksu9000 is this Ok? https://github.com/InfiniTimeOrg/InfiniTime/pull/805/commits/0d745a0ad10378d71196af908f031108252b1214

Avamander (approved):
Seems fine now. Someone should take a look as well.

Avamander:
Oh no, a conflict, if you fix that and we can merge, we can avoid breaking this PR with each and every other PR.

NeroBurner:
rebased and conflict resolved

Riksu9000 (approved):

NeroBurner:
rebased on current develop (after v1.7.0 release)

NeroBurner:
rebased on `develop` after v1.7.1 release

NeroBurner:
@Avamander Is something still missing, did I forget something? Or is it just limited free time for this awesome hobby? :)

Avamander:
Everything seems to be in order, I think it's in the queue waiting for a merge.

JF002:
@NeroBurner 

> Is something still missing, did I forget something? Or is it just limited free time for this awesome hobby? :)

I'm sorry for the time it took for me to have a look at your PR! I am currently the only one that is allowed to merge PRs to develop (that might change in the future) and I couldn't allocate much time to InfiniTime recently!

However, I think you did a great job cleaning the includes of the project! Thank you for that! 
I've just fixed the last conflicts that were introduce by PRs I've just merged and I merged this one!

NeroBurner:
@JF002 Thanks :tada: Happy it's merged

InfiniTime is a hobby-project. If there is no personal time left to work on it, then it is ok as well ;) It's true, that waiting for feedback for weeks was bad for motivation, but now I'm happy my contribution got merged.

I'm VERY grateful to you, that you created InfiniTime. Better take some personal time, than burning out on the project

In short: thank you @JF002  :)

JF002:
@NeroBurner Thank you so much for your kind words! Do not forget I'm not alone in this adventure! There are so many contributors and enthousiasts and users that supports the project! Thank to all of them!
I understand that waiting several weeks for a feedback is not ideal, and that's exactly for that reason that I'm creating teams of contributors that help me to triage the issues, PRs, review the code, help with specific topics,...
Thanks again, and have fun contributing again to the project :-)

PR #811, opened 2021-11-06
Add data validity check and retries in CST816S driver

### LINKED ISSUE #763: Touch controller error warning on PineTime
### 

### What happened?

Touch controller error warning appears on PineTime on boot

Touch controller error warning should not appear on PineTime

### Reproduction steps

- Install the DFU artifact from #754
- Reboot PineTime

### More details?

This did not happen before installing the artifact from the mentioned PR.
After tapping Proceed, touch works as expected.

### Version

add-motion-service

### Companion app

_No response_

JF002:
I noticed that I couldn't get 100% reliable readings from the CST816S (touch controller) (see https://github.com/InfiniTimeOrg/InfiniTime/issues/763#issuecomment-962436976). I added a few checks in `Cst816S::GetTouchInfo` to ensure it only returns is data are valid. I also added a retry functionality for the Check ID procedure. It'll hopefully reduce the number of false positive when checking the device IDs.

This PR might fix #763, but I think we still have other issues with I²C (see https://github.com/InfiniTimeOrg/InfiniTime/issues/763#issuecomment-962489299).

Anyway, it would be nice if some people could test that PR and ensure it does not show the warning about an incompatible touch controller. @Arsen6331 @geekbozu @kieranc I would be really interested to know if this PR fixes that issue.

### DIFF
--- a/src/drivers/Cst816s.cpp
+++ b/src/drivers/Cst816s.cpp
@@ -33,14 +33,14 @@ bool Cst816S::Init() {
   vTaskDelay(5);
 
   static constexpr uint8_t maxRetries = 3;
-  bool isDeviceOk = false;
+  bool isDeviceOk;
   uint8_t retries = 0;
   do {
     isDeviceOk = CheckDeviceIds();
     retries++;
-  } while(!isDeviceOk && retries < maxRetries);
+  } while (!isDeviceOk && retries < maxRetries);
 
-  if(!isDeviceOk) {
+  if (!isDeviceOk) {
     return false;
   }
 
--- a/src/drivers/Cst816s.cpp
+++ b/src/drivers/Cst816s.cpp
@@ -32,6 +32,18 @@ bool Cst816S::Init() {
   twiMaster.Read(twiAddress, 0xa7, &dummy, 1);
   vTaskDelay(5);
 
+  static constexpr uint8_t maxRetries = 3;
+  bool isDeviceOk = false;
+  uint8_t retries = 0;
+  do {
+    isDeviceOk = CheckDeviceIds();
+    retries++;
+  } while(!isDeviceOk && retries < maxRetries);
+
+  if(!isDeviceOk) {
+    return false;
+  }
+
   /*
   [2] EnConLR - Continuous operation can slide around
   [1] EnConUD - Slide up and down to enable continuous operation
@@ -50,21 +62,7 @@ bool Cst816S::Init() {
   static constexpr uint8_t irqCtl = 0b01110000;
   twiMaster.Write(twiAddress, 0xFA, &irqCtl, 1);
 
-  // There's mixed information about which register contains which information
-  if (twiMaster.Read(twiAddress, 0xA7, &chipId, 1) == TwiMaster::ErrorCodes::TransactionFailed) {
-    chipId = 0xFF;
-    return false;
-  }
-  if (twiMaster.Read(twiAddress, 0xA8, &vendorId, 1) == TwiMaster::ErrorCodes::TransactionFailed) {
-    vendorId = 0xFF;
-    return false;
-  }
-  if (twiMaster.Read(twiAddress, 0xA9, &fwVersion, 1) == TwiMaster::ErrorCodes::TransactionFailed) {
-    fwVersion = 0xFF;
-    return false;
-  }
-
-  return chipId == 0xb4 && vendorId == 0 && fwVersion == 1;
+  return true;
 }
 
 Cst816S::TouchInfos Cst816S::GetTouchInfo() {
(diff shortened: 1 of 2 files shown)


Arsen6331:
I just installed this PR. I don't seem to be getting the warning anymore. I have restarted my watch multiple times and haven't gotten it once. The warning seems to be fixed so far.

src/drivers/Cst816s.cpp
```diff
@@ -32,6 +32,18 @@ bool Cst816S::Init() {
   twiMaster.Read(twiAddress, 0xa7, &dummy, 1);
   vTaskDelay(5);
 
+  static constexpr uint8_t maxRetries = 3;
+  bool isDeviceOk = false;
+  uint8_t retries = 0;
+  do {
+    isDeviceOk = CheckDeviceIds();
+    retries++;
+  } while(!isDeviceOk && retries < maxRetries);
```
Avamander, line 41: It's not really a huge issue, but is this style the one we want to use?

src/drivers/Cst816s.cpp
```diff
@@ -32,6 +32,18 @@ bool Cst816S::Init() {
...
+
+  if(!isDeviceOk) {
```
Avamander, line 43: Hasn't been formatted :)

Avamander (changes requested):

src/drivers/Cst816s.cpp
JF002: Do you mean `!` vs `not` and `&&` vs `and` ?

src/drivers/Cst816s.cpp
JF002: I'm not sure what's wrong here?

src/drivers/Cst816s.cpp
Avamander: `do {} while()` instead of `while () {`

src/drivers/Cst816s.cpp
Avamander: It would be `if (!isDeviceOk) {` after it has been formatted

kieranc:
I've just done 10+ test cycles with this PR - reboot watch, play a full game of twos, return to menu, reboot, repeat.
No touch controller errors on boot and no crashes. The first 5 cycles I ran with the watch with a back on, taped to the charger, after that I removed the tape and back, put it in a non-conductive stand and only touched the screen. There was one point when it was on the stand when the touch screen seemed to stop responding until I touched the metal bezel, then it went back to working normally, but I couldn't repeat this.

src/drivers/Cst816s.cpp
JF002: To me, the `do while` loop makes sense as I want to ensure the check is done at least once. But I agree the initialization of `isDeviceOk` is probably redundant in this case...

src/drivers/Cst816s.cpp
JF002: Oh I forgot we enforced the space after the `if`. Should be ok now :)

JF002:
Thanks for the test @kieranc ! I'm wondering if the presence of the SWD debugger could make the i²c communication less reliable?

kieranc:
> 
> 
> Thanks for the test @kieranc ! I'm wondering if the presence of the SWD debugger could make the i²c communication less reliable?
> 
> EDIT : I've just tested one devkit without SWD connection : no crash anymore 0_o

It looks that way but I can't figure out the how/why.. 
My dev watch (RIP) was intermittently unreliable after 1.4/new touch handler, at one point reseating the touch cable made it work for a few days but then it reverted to previous behaviour (I could swipe but not tap), at other times holding the device or touching the screen with a metallic object seemed to make it work temporarily. It was definitely after I soldered the SWD cables that it started breaking but I kind of assumed I broke something during soldering.
I tried removing the programmer from the soldered cables which had no impact but didn't try removing the cables themselves. It feels to me like some sort of grounding/charge buildup thing but it's difficult to replicate reliably so I don't have a lot more to go on than feelings.
All the references I've seen to this touch controller make reference to it responding via I2C only after a touch event, does our driver implement this? I'm not sure I'm understanding it fully, either it's a case of boot -> init -> no i2c -> touch event -> normal i2c, or, i2c only works correctly immediately after a touch event, and will not work right during a touch event (which is kind of what we're seeing? and kind of aligns with swipe but not tap working?)
Apologies for the wall of text, I'm fairly sure I've been running into this bug for a while, maybe my rambling will spark something useful :p

refs:
https://doc.riot-os.org/group__drivers__cst816s.html
https://wiki.pine64.org/wiki/PineTime#Touch_panel
https://github.com/fbiego/CST816S

Avamander (approved):

PR #814, opened 2021-11-07
clang-format fix

SteveAmor:
General clang-format fix so not as to get in the way of any other feature changes that are made to the file.


SteveAmor:
Partial fix in #782

PR #815, opened 2021-11-07
SystemTask applied clang-format

SteveAmor:


geekbozu:
With this line in particular would it make more sense to remove the casts in some form? Formatting is fine just feels like a very long line.

Riksu9000:
This has been fixed.

PR #816, opened 2021-11-09
Fix Alarm app crashing on buttonpress

### LINKED ISSUE #770: Crash when exiting alarm app with button after setting it
### 

- [x] I searched for similar bug reports and found none was relevant.

### What happened?

Exiting alarm app with button after seting it crashes InfiniTime sometimes. 

InfiniTime shouldn't crash. 

### Reproduction steps

Open alarm app, set it, exit it with the button, watch will lock up and InfiniTime will reboot.
I haven't been able to reproduce this reliably though. Sometimes it happens and sometimes it doesn't. 

### More details?

_No response_

### Version

Develop 1d76d91 (with pr's #746, #740, #720, #698, #679, #747, #648, #652, #586, buy none of them change the alarm other than 698) 

### Companion app

_No response_

Riksu9000:
Fixes #770 

This should be fixed before releasing 1.7.0

### DIFF
--- a/src/displayapp/screens/Alarm.h
+++ b/src/displayapp/screens/Alarm.h
@@ -40,7 +40,9 @@ namespace Pinetime {
         Controllers::AlarmController& alarmController;
 
         lv_obj_t *time, *btnEnable, *txtEnable, *btnMinutesUp, *btnMinutesDown, *btnHoursUp, *btnHoursDown, *txtMinUp, *txtMinDown,
-          *txtHrUp, *txtHrDown, *btnRecur, *txtRecur, *btnMessage, *txtMessage, *btnInfo, *txtInfo;
+          *txtHrUp, *txtHrDown, *btnRecur, *txtRecur, *btnInfo, *txtInfo;
+        lv_obj_t* txtMessage = nullptr;
+        lv_obj_t* btnMessage = nullptr;
 
         enum class EnableButtonState { On, Off, Alerting };
         void SetEnableButtonState();


mruss77:
Thanks, I came in here to open this same PR this morning!

I've been testing the same fix for several days and the alarm is no longer crashing.

Avamander:
Why not initialize them all to `nullptr`?

Riksu9000:
> Why not initialize them all to `nullptr`?

Everything else gets initialized in the constructor and we don't initialize them to nullptr in other screens either.

Avamander:
> Everything else gets initialized in the constructor and we don't initialize them to nullptr in other screens either.

Do you have any ideas how such uninitialized variables could be better avoided? Any warnings?

Avamander (approved):

Riksu9000:
> > Everything else gets initialized in the constructor and we don't initialize them to nullptr in other screens either.
> 
> Do you have any ideas how such uninitialized variables could be better avoided? Any warnings?

There was a warning that the constructor doesn't initialize those variables.

Avamander:
> There was a warning that the constructor doesn't initialize those variables.

What was the name? Maybe it should be turned into an error with `-Werror=<name>`.

Riksu9000:
> > There was a warning that the constructor doesn't initialize those variables.
> 
> What was the name? Maybe it should be turned into an error with `-Werror=<name>`.

Oh not a compilation warning but a clang-tidy one. This one https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines-pro-type-member-init.html

geekbozu (approved):

ncartron:
Hi, I had a pre 1.7 DFU without this PR and the Alarm app was constantly crashing.
I confirm that with a DFU which includes that PR, the Alarm app works as expected, no more crash ! <3

petterhs:
PR is fixing the issue :+1:

petterhs (approved):

JF002:
Thanks! Just in time for 1.7! (Initially, I was going to release it sooner, but I was too busy these last few days...).

PR #820, opened 2021-11-10
Fix animation when long pressing on screens adjacent to watch face

### LINKED ISSUE #817: Long Press of power button always plays the same animation.
### 

### What happened?

Long press of home button always plays coming down animation of watchface even if you are on the notifications menu or settings menu resulting in a weird looking animation.

If on for example the notifications menu i expect it to look like the watchface animation is coming up instead of down. If on the quick access menu i expect the animation to look like it's coming from the right.

### Reproduction steps

1. Be on the latest commit.
2. Go to either notifications menu or quick access menu.
3. Long press power button until it returns to watchface.

### More details?

https://user-images.githubusercontent.com/27083886/140960257-770c10dd-a668-4bb2-a94e-e02ae7a57620.mp4

### Version

Latest git branch.

### Companion app

_No response_

Riksu9000:
Add extra handling when in adjacent screens to use the most fitting animation.

Fixes #817

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -262,7 +262,13 @@ void DisplayApp::Refresh() {
         break;
       case Messages::ButtonLongPressed:
         if (currentApp != Apps::Clock) {
-          LoadApp(Apps::Clock, DisplayApp::FullRefreshDirections::Down);
+          if (currentApp == Apps::Notifications) {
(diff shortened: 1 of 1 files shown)


geekbozu (approved):

KoalaV2:
Just tried this fix. Works perfectly.

petterhs (approved):

PR #821, opened 2021-11-11
Status bar displays time in 12 or 24 hour format based on settings

mruss77:
This makes the status bar display time in the correct format per the "Time format" setting.

New function is added to DateTime called FormattedTime returns a string with the correct time format, so the caller doesn't need to check the settings directly.

I had to remove one unused include from Settings.h to avoid circular include problems.

This is a scaled-back version of #786 (per @Riksu9000 's suggestion) to fix the time display while thinking about how to handle other changes to the status bar.


Riksu9000 (approved):

Itai-Nelken:
Works nicely. This is a great improvement.

PR #826, opened 2021-11-12
Improve Raise to Wake algorithm

### LINKED ISSUE #548: Wake on wrist raise too sensitive
I think the wake-up on wrist raise is too sensitive. 

While it is nice to not have to twist your arm quickly by 90 degrees to activate the screen (like on my MiBand 4), it is pointles to have it activated almost constantly during normal day activities. 

My PineTime lights up even when I'm driving on a highway and barely turn the steering wheel. 

I think that battery life suffer greatly from that. Maybe tuning it down a little bit might be a good idea.

FintasticMan:
Depends on #1696 and #1726.

Fixes #548.

With this implementation, your hand needs to be raised at a certain speed to register.

The values for the speed and threshold are values that feel about right to me, but I'm up for changing them.

### DIFF
--- a/src/components/motion/MotionController.cpp
+++ b/src/components/motion/MotionController.cpp
@@ -9,7 +9,7 @@ void MotionController::Update(int16_t x, int16_t y, int16_t z, uint32_t nbSteps)
     service->OnNewStepCountValue(nbSteps);
   }
 
-  if (service != nullptr && (this->x != x || this->y != y || this->z != z)) {
+  if (service != nullptr && (this->x != x || yHistory[0] != y || zHistory[0] != z)) {
     service->OnNewMotionValues(x, y, z);
   }
 
@@ -18,10 +18,10 @@ void MotionController::Update(int16_t x, int16_t y, int16_t z, uint32_t nbSteps)
 
   lastX = this->x;
   this->x = x;
-  lastY = this->y;
-  this->y = y;
-  lastZ = this->z;
-  this->z = z;
+  yHistory++;
+  yHistory[0] = y;
+  zHistory++;
+  zHistory[0] = z;
 
   int32_t deltaSteps = nbSteps - this->nbSteps;
   if (deltaSteps > 0) {
@@ -31,21 +31,21 @@ void MotionController::Update(int16_t x, int16_t y, int16_t z, uint32_t nbSteps)
 }
 
 bool MotionController::ShouldRaiseWake(bool isSleeping) {
-  if ((x + 335) <= 670 && z < 0) {
+  if ((x + 335) <= 670 && zHistory[0] < 0) {
     if (!isSleeping) {
-      if (y <= 0) {
+      if (yHistory[0] <= 0) {
         return false;
       }
       lastYForRaiseWake = 0;
       return false;
     }
 
-    if (y >= 0) {
+    if (yHistory[0] >= 0) {
       lastYForRaiseWake = 0;
       return false;
     }
-    if (y + 230 < lastYForRaiseWake) {
-      lastYForRaiseWake = y;
+    if (yHistory[0] + 230 < lastYForRaiseWake) {
+      lastYForRaiseWake = yHistory[0];
       return true;
     }
   }
@@ -54,7 +54,8 @@ bool MotionController::ShouldRaiseWake(bool isSleeping) {
 
 bool MotionController::ShouldShakeWake(uint16_t thresh) {
   /* Currently Polling at 10hz, If this ever goes faster scalar and EMA might need adjusting */
-  int32_t speed = std::abs(z - lastZ + (y - lastY) / 2 + (x - lastX) / 4) * 100 / (time - lastTime);
+  int32_t speed =
+    std::abs(zHistory[0] - zHistory[histSize - 1] + (yHistory[0] - yHistory[histSize - 1]) / 2 + (x - lastX) / 4) * 100 / (time - lastTime);
   // (.2 * speed) + ((1 - .2) * accumulatedSpeed);
   accumulatedSpeed = speed / 5 + accumulatedSpeed * 4 / 5;
 
--- a/src/components/motion/MotionController.cpp
+++ b/src/components/motion/MotionController.cpp
@@ -87,26 +87,22 @@ MotionController::AccelStats MotionController::GetAccelStats() const {
   return stats;
 }
 
-bool MotionController::ShouldRaiseWake(bool isSleeping) {
-  if ((x + 335) <= 670 && zHistory[0] < 0) {
-    if (!isSleeping) {
-      if (yHistory[0] <= 0) {
-        return false;
-      }
-      lastYForRaiseWake = 0;
-      return false;
-    }
+bool MotionController::ShouldRaiseWake() const {
+  constexpr uint32_t varianceThresh = 56 * 56;
+  constexpr int16_t xThresh = 384;
+  constexpr int16_t yThresh = -64;
+  constexpr int16_t rollDegreesThresh = -45;
 
-    if (yHistory[0] >= 0) {
-      lastYForRaiseWake = 0;
-      return false;
-    }
-    if (yHistory[0] + 230 < lastYForRaiseWake) {
-      lastYForRaiseWake = yHistory[0];
-      return true;
-    }
+  if (x < -xThresh || x > xThresh) {
+    return false;
   }
-  return false;
+
+  // if the variance is below the threshold, the accelerometer values can be considered to be from acceleration due to gravity
+  if (stats.yVariance > varianceThresh || (stats.yMean < -724 && stats.zVariance > varianceThresh) || stats.yMean > yThresh) {
+    return false;
+  }
+
+  return DegreesRolled(stats.yMean, stats.zMean, stats.prevYMean, stats.prevZMean) < rollDegreesThresh;
 }
 
 bool MotionController::ShouldShakeWake(uint16_t thresh) {
--- a/src/components/motion/MotionController.cpp
+++ b/src/components/motion/MotionController.cpp
@@ -54,10 +54,9 @@ bool MotionController::ShouldRaiseWake(bool isSleeping) {
 
 bool MotionController::ShouldShakeWake(uint16_t thresh) {
   /* Currently Polling at 10hz, If this ever goes faster scalar and EMA might need adjusting */
-  int32_t speed = std::abs(z - lastZ + (y / 2) - (lastY / 2) + (x / 4) - (lastX / 4)) / (time - lastTime) * 100;
-  //(.2 * speed) + ((1 - .2) * accumulatedSpeed);
-  // implemented without floats as .25Alpha
-  accumulatedSpeed = (speed / 5) + ((accumulatedSpeed / 5) * 4);
+  int32_t speed = std::abs(z - lastZ + (y - lastY) / 2 + (x - lastX) / 4) * 100 / (time - lastTime);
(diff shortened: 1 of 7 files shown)


SteveAmor:
Seems to be working well for me.

FintasticMan:
After running it for a couple of days, I think the speed and threshold could be lowered a bit.

kieranc:
I've been testing this a bit too, it seems like an improvement but I am still getting some false positives, usually when I'm lying down/on the sofa when my arm is close to horizontal but angled slightly up, a small movement will wake the display. It's difficult to pin down as it only seems to happen accidentally, if I try to replicate it I can't! I'll keep testing and try get more info.

FintasticMan:
Rebased onto latest develop.

Riksu9000 (commented):
There are two things that I'm noticing.

While the watch is already facing slightly towards me, it will wake up on tiny movements. In this case I think it's too sensitive. If the watch is positioned towards me and it's off, I think it should require a flick to turn on.

Previously the accepted range of values for X wasn't limited in the negatives. By requiring X to be larger than 384, I think it's harder to turn on when lying down. I think this also depends on handedness.

FintasticMan:
I've just updated it to use either the y or z values depending on which is more accurate. This means that it also works when the facing down and towards you. The threshold values and speed values are now also variables so it's easier to adjust them, and I've increased them.

Riksu9000:
I don't think using Z in place of Y is correct, because the axis don't receive comparable acceleration. Y is basically detecting wrist rolling, which is what we want. Z detects moving the arm up, which can be useful data, but it's not the same as Y.

I think motion algorithm changes are difficult to test, because it only depends on feeling. What should we look for when testing this? Please explain your thoughts behind the changes in detail and for each future revision, so we can get an idea of how this should work. People with two PineTimes can help compare the behaviour.

FintasticMan:
I believe that using Z is fine, because Y and Z are both the same axis for some reason, just offset by 90 degrees. This means that when Y is at (-)1024, Z is at 0, and vice versa. I've noticed that the larger the values are, the less they change with the same motion (A.K.A they are less accurate). For example, if the watch is facing up, Y is at 0 and Z is at -1024, and a slight movement towards you will have more of an effect on Y than on Z.

In the current state of this PR, you should need around the same movement to get it to wake up all the time. It should also work when the watch is facing down and towards you, so that should help when you're lying down.

Riksu9000:
> because Y and Z are both the same axis, just offset by 90 degrees

I don't think I understand what you mean. Z is from the front of the screen to the back of the device. Y is from the top of the device to the bottom. When you tilt the device, the acceleration due to gravity will affect one of the axis more than the other, but the tilting of the wrist only affects the Y acceleration, since there's no movement along Z.

FintasticMan:
A visual might help, but my drawing skills are not up to par, especially not with a mouse ;-). It's quite difficult to explain using words, but maybe you can try it on your own device using the motion app.

If the screen is facing up, the Y value is 0 and the Z value will be shown as -64. Then, if you slowly roll the screen towards you, the Y value will drop below zero, and the Z value will increase. Once the screen is completely facing you, the Y value should show as -64 and the Z value should be 0.

I personally haven't looked at the spec sheet, so I don't know what the values are supposed to represent, I just figured it out through a lot of testing. I can't see anything that the Z value could be other than roll offset by 90 degrees, but maybe I missed something.

FintasticMan:
image
Here is a very crude visual I just made.

Riksu9000:
If there was no gravity, these would all show 0 by default. What you're showing is acceleration due to gravity. An accelerometer shows device movement and gravity combined. Please note the difference between a gyroscope and an accelerometer.

FintasticMan:
Ah, I see now. I did make a mistake, I misunderstood what Z does. In the code, I do use the values of Z in a way that doesn't align with what it means, but I think that because of the X threshold it can't cause incorrect behaviour.

FintasticMan:
This change should be the final change (bar maybe some tweaking of the values if people think it's necessary). I think the is good and readable enough to be merged. I have made it so that the required speed to wake the device is slightly dependent on the current y value, because the bigger it is, the less sensitive it is.

I would like some testing on it, because at the moment the values are right for me, and I would like to know if that is also the case for other people.

src/components/motion/MotionController.cpp
```diff
@@ -21,27 +23,26 @@ void MotionController::Update(int16_t x, int16_t y, int16_t z, uint32_t nbSteps)
...
-      lastYForWakeUp = y;
-      return true;
-    }
+  if (x < -xThresh || x > xThresh || y > yThresh) {
+    return false;
   }
-  return false;
+
+  // Use z here because it's not possible to accurately detect movement between around y == -768 and y == -1024
+  // y can't be used in this range because it needs more movement for the values to change the same amount
```
Riksu9000, line 37: I'm not sure this is true. I set the Motion app range to +-2000 and couldn't see a difference in the amplitude of the detected motion whether the watch was placed horizontally or vertically. The BMA425 datasheet claims a typical non linearity of 0.5%FS, which I think means across the entire range.

Riksu9000: Again, this is gravity and not motion. This behaviour can be explained with simple trigonometry.
sin0 = 0
sin15 = 0.26
sin75 = 0.97
sin90 = 1

FintasticMan: Ah ok, then I shouldn't have put that in the comment then. The second point I make does still stand though I think, because if the watch moves from facing up with a y value of -800 for example to facing down with a y value of -800, you can't tell whether the wrist was moved or not.

Riksu9000: No, I think you're still thinking about gravity. If you have the watch in any orientation and quickly jerk it along Y, it will show a positive and a negative peak in acceleration, assuming it updated fast enough and the acceleration is within the operating range of the accelerometer, and then it will return to the same old value which is acceleration due to gravity.

FintasticMan: I think you're right, but that the MotionController doesn't update fast enough for that to be recorded in the deltaY variable. I also think that the flick to wake the device up is usually not going to be greater than the acceleration due to gravity. Because of this I think it makes most sense to assume that the acceleration values are all acceleration due to gravity.

FintasticMan: Thanks helping me get clarification on how the accelerometer works! I appreciate it.

I think that changing the polling rate won't affect this, as long as the SystemTask still runs at the same speed. I will go through my logic and make sure that it still holds up with my new knowledge.

Riksu9000:
The issue with this PR is that these changes are hard to understand. Especially considering the confusion along the way. So how exactly does this work? What is it measuring and how? How would this work when the motion values are updated more frequently?

FintasticMan:
I've finally had some time to work on this again, other than just keeping it updated. I decided to put pen to paper and make sure that this covers all edge cases. I discovered that my previous algorithm wasn't exactly the most mathematically sound, so I've made some changes that both make it more correct and improve readability.

The algorithm assumes that the acceleration data is from the acceleration due to gravity. It works by checking if the watch is facing the user (between around -8 degrees and -135 degrees, where the screen facing upwards is 0 degrees), then checking if the amount the watch "rolled" since the last update is above a threshold. This threshold is dependent on the time since the last update, meaning that it doesn't matter at what speed the SystemTask is running.

There are three cases that need to be handled in the AmountRolled function, to prevent the need for costly sine functions. Between -PI/4 and PI/4, the sine function is largely linear, so that can be handled without actually using it. The use of the Z axis is not a problem, because it is bounded to only work between quite strict bounds in the X axis.

I've also done some cleaning up of the MotionController class as a whole, so now there's less code duplication. There were a number of `last*ForShake` variables, which I removed and replaced with their more general counterparts. The Shake Wake function has been simplified, but the functionality hasn't been changed at all.

Riksu9000:
There are still a lot of changes here and reading the commits individually isn't proving any simpler. Not sure why shakewake was changed in the same PR and also the same commit where a lot of other stuff was done. LVGL has cheap sine functions that are already used in the analog watchface if that would be useful. `_lv_trigo_sin()`.

FintasticMan:
I think this is ready to be reviewed now. I have made some changes that mean that this has as few false positives as possible, while making sure it does wake when the user wants it to. It checks that the accelerometer values can reasonably be assumed to be from gravity, then checks that the wrist was rolled more than 45 degrees within 6 SystemTask cycles (around 600ms).

To make the recovery firmware not fail when linking, I've added an ifdef to the arcsin function so it doesn't use lvgl. This is the best solution I could think of, other than creating a separate SystemTask for the recovery firmware.

There are a lot of separate cases in the DegreesRolled function, but I'm quite sure that's the only way to make it work in every situation.

src/utility/Math.cpp
```diff
@@ -0,0 +1,47 @@
+#include "utility/Math.h"
+
+#ifndef PINETIME_IS_RECOVERY
+
+  #include <lvgl/src/lv_misc/lv_math.h>
+
+int16_t Asin(int16_t arg) {
```
minacode, line 7: Does it work, if you add
```cpp
if (PINETIME_IS_RECOVERY) {
  return 0;
}
```
here instead and let the compiler figure out that the include is not needed?

src/components/motion/MotionController.cpp
```diff
@@ -29,31 +62,52 @@ void MotionController::Update(int16_t x, int16_t y, int16_t z, uint32_t nbSteps)
...
+  }
+
+  return DegreesRolled(stats.yMean, stats.zMean, stats.prevYMean, stats.prevZMean) > rollDegreesThresh;
 }
 
 bool MotionController::ShouldShakeWake(uint16_t thresh) {
   /* Currently Polling at 10hz, If this ever goes faster scalar and EMA might need adjusting */
-  int32_t speed = std::abs(z + (y / 2) + (x / 4) - lastY / 2 - lastZ) / (time - lastTime) * 100;
+  int32_t speed =
+    std::abs(zHistory[0] + (yHistory[0] / 2) + (x / 4) - yHistory[histSize - 1] / 2 - zHistory[histSize - 1]) / (time - lastTime) * 100;
```
minacode, line 110: Maybe it's s good idea for documentation to split this into intermediate results with meaningful names.

src/utility/Math.cpp
FintasticMan: No, because `PINETIME_IS_RECOVERY` isn't defined at all if it's not building the recovery, so that'll just give an error.

src/components/motion/MotionController.cpp
FintasticMan: That's a good idea, but I think that that should be done in another PR. I also think that #1691 should be merged before that.

JF002:
Hey @FintasticMan ! Thanks for this PR and for taking the time to refine it over and over again :)

I've just tested this PR and compared it with the current implementation. At first sight, this new implementation reject more false positives that the current one. It's also a bit slower, but the reaction time is still reasonable, in my opinion.

About the code, I'm very bad at signal processing, and I can't really judge if this new algorithm "makes sense" or is "mathematically correct". It does however seem to work fine :+1!

Did you have the opportunity to get feedback from other users? 

All in all, if this new implementation is still comfortable enough bug avoid waking up the watch when it's not necessary, it'll probably help with unwanted display and touch activation, and will also improve the battery life!

JF002 (approved):
@FintasticMan told me he uses this implementation for a few months, and also received positive feedback from a few users. 

So I suggest we merge this branch, and refine the implementation later on if necessary.

PR #828, opened 2021-11-12
Adjust systemtask to respect doNotGoToSleep.

geekbozu:
System task getting a GoToSleep event will not respect doNotGoToSleep and attempt to sleep anyway.

### DIFF
--- a/src/systemtask/SystemTask.cpp
+++ b/src/systemtask/SystemTask.cpp
@@ -279,7 +279,7 @@ void SystemTask::Work() {
         } break;
         case Messages::GoToSleep:
           if (doNotGoToSleep) {
-            return;
+            break;
           }
           isGoingToSleep = true;
           NRF_LOG_INFO("[systemtask] Going to sleep");
--- a/src/systemtask/SystemTask.cpp
+++ b/src/systemtask/SystemTask.cpp
@@ -278,6 +278,9 @@ void SystemTask::Work() {
           }
         } break;
         case Messages::GoToSleep:
+          if (doNotGoToSleep) {
+            return;
+          }
           isGoingToSleep = true;
           NRF_LOG_INFO("[systemtask] Going to sleep");
           xTimerStop(idleTimer, 0);
@@ -497,7 +500,7 @@ void SystemTask::OnTouchEvent() {
 }
 
 void SystemTask::PushMessage(System::Messages msg) {
-  if (msg == Messages::GoToSleep) {
+  if (msg == Messages::GoToSleep && !doNotGoToSleep) {
     isGoingToSleep = true;
   }
 


FintasticMan:
This is definitely an important fix. While testing my lower to sleep pr, I accidentally lowered my wrist while applying a DFU, which caused some issues.

src/systemtask/SystemTask.cpp
```diff
@@ -497,7 +500,7 @@ void SystemTask::OnTouchEvent() {
 }
 
 void SystemTask::PushMessage(System::Messages msg) {
-  if (msg == Messages::GoToSleep) {
+  if (msg == Messages::GoToSleep && !doNotGoToSleep) {
     isGoingToSleep = true;
   }
```
Riksu9000, line 505: I'm thinking this part shouldn't even be here. This function is for pushing messages, but it changes the state of system task too and `isGoingToSleep` is set to true in the message handling too so it's set twice.

geekbozu: My research said it was don't there for some race condition fixing. I couldn't completely logic it away so I chose to leave it.

geekbozu:
> This is definitely an important fix. While testing my lower to sleep pr, I accidentally lowered my wrist while applying a DFU, which caused some issues.

This is exactly why I did this ;) I have a similar bug in the filesystem code I'm working on and this would have caused a crash as well for me!

src/systemtask/SystemTask.cpp
geekbozu: So looked again, I think the whole push routine needs some adjustment since it is allowing tasks to fire before the event is processed, and im not sure that is the correct behavior, 
See the `portYIELD_FROM_ISR` call in that routine.

SteveAmor: Was that the race condition fixed in commit 20f5b0f

geekbozu: Thats what the git blame says yes, I was on mobile when I typed this earlier and never got around to adding it thanks!

Riksu9000: The variable was added to fix some race condition, but it was added to both places at the same time, which looks like a mistake to me. It wasn't added to `PushMessage()` specifically to fix the issue.

geekbozu: I disagree, it was added in PushMessage since after that the display task or similar can preempt the GoingToSleep event queuing up Display task stuff after its to late for it to short circuit. Really this whole SPI/i2c side of things needs an overhaul to not rely completely on the System task for their control In my opinion. Which is out of the scope of this PR. For now fixing a potential crash is the goal :)

FintasticMan:
If I send a `goToSleep` message while `doNotGoToSleep` is set, InfiniTime crashes.  
I tested this with latest develop + this pr + lower to sleep pr. Turning on the torch and lowering your wrist crashes InfiniTime.

FintasticMan:
Found the issue. You don't want to be returning if `doNotGoToSleep` is set, just breaking the switch.

geekbozu:
> 
> 
> Found the issue. You don't want to be returning if `doNotGoToSleep` is set, just breaking the switch.

...soon as you said it I face palmed. Thanks!

src/systemtask/SystemTask.cpp
JF002: I agree with @Riksu9000, it looks like `isGoingToSleep` is set to `true` in `SystemTask::PushMessage()` when the message is `Messages::GoToSleep` and also when the message `Messages::GoToSleep` is handled, which looks redundant. 

We would need to cross-check, but we could probably remote the assignation in the message handler.

> I'm thinking this part shouldn't even be here.

I also agree with this. The state machine of SystemTask needs a bit of refactoring and rationalization. There are currently multiple boolean variable that represents the state, and it's getting tricky to understand and maintain it.
But as @geekbozu said, this is not the scope of this PR. 

If everyone agrees with this, could you close this conversation so that I can merge it?

geekbozu: I'm of the thought that we really should not have the assignation in Push Message. However this is something to be fixed with a State machine cleanup. So for now lets shelf this and get this patch in and we can clean it up when we actually fix our state machine.

PR #833, opened 2021-11-15
BatteryInfo: remove unused FreeRTOS.h and timer.h includes

NeroBurner:
Those two includes are not used. The project still compiles without them.

Not tested on PineTime, just compiled


NeroBurner:
rebased on `develop` after v1.7.1 release

PR #834, opened 2021-11-15
DateTimeControler: remove unused include

NeroBurner:
Compiles fine without it. Not tested on PineTime, just compiled


NeroBurner:
it compiles fine, but `DateTimeController.cpp` uses `NRF_LOG_INFO()`, so include is indeed needed

PR #835, opened 2021-11-17
Notification age (x minutes ago) on notifcation screen

clemensvonmolo:
This PR adds "x time ago" to the bottom of the notification screen. Every notification item now stores the time at which it was received as a std::time_t (saw no reason to use std::chrono::time_point as it takes more memory with no advantages for storage). The time arrived is set in the notification manager so existing apps don't have to change anything.

I'm not really sure about the location of the text (image below) and if it would be better to use C-style string operations for memory usage and allocation optimisation.
IMG_20211117_150831
IMG_20211117_151348
..e indicator does not interfere with the message


SteveAmor:
Great feature.

Itai-Nelken:
Works nicely. Only problem is that the 'no notifications' notification has it (and it doesn't seem to update).

src/displayapp/screens/Notifications.cpp
```diff
@@ -166,6 +181,21 @@ Notifications::NotificationItem::NotificationItem(const char* title,
...
+  std::chrono::minutes age = std::chrono::duration_cast<std::chrono::minutes>(diff);
+  std::string ageString;
+  ageString.reserve(10);
+  if ((age.count() / (60 * 24)) >= 1) {
+    ageString = std::to_string(static_cast<uint16_t>(age.count() / (60 * 24))) + "d ago";
+  } else if ((age.count() / 60) >= 1) {
+    ageString = std::to_string(static_cast<uint8_t>(age.count() / 60)) + "h ago";
+  } else {
+    ageString = std::to_string(static_cast<uint8_t>(age.count())) + "m ago";
+  }
```
Riksu9000, line 194: `to_string` has been found to use a lot of memory #792. Using `sprintf` and char arrays might save some memory here too.

src/displayapp/screens/Notifications.cpp
```diff
@@ -166,6 +181,21 @@ Notifications::NotificationItem::NotificationItem(const char* title,
...
+  lv_obj_t* alert_age = lv_label_create(lv_scr_act(), nullptr);
+  lv_label_set_text(alert_age, ageString.c_str());
+  lv_obj_align(alert_age, container1, LV_ALIGN_IN_BOTTOM_RIGHT, 0, 0);
```
Riksu9000, line 197: I think this label should have similar "padding" as the message.

src/displayapp/screens/Notifications.cpp
clemensvonmolo: Or maybe use lv_label_set_text_fmt for the int to string conversion?

Tiggilyboo:
Not sure if this is still active, but this would be great to get merged. What is left to do / blocking this from getting in?

clemensvonmolo:
The problem seems to have been me forgetting about it. Now I think this has to be adjusted to newer changes that happened in the meantime. I don't have time today but I can try doing that tomorrow.

devnoname120:
> The problem seems to have been me forgetting about it. Now I think this has to be adjusted to newer changes that happened in the meantime. I don't have time today but I can try doing that tomorrow.

Friendly bump 🙏

JF002:
@clemensvonmolo  @devnoname120 Sorry for not providing any feedback about this PR for so long. Is there any reason for it to be still a draft?

clemensvonmolo:
I can set it to a normal PR for now, but Github says there will be conflicts and I can't do anything about that in the next three days while I'm on a class trip. It might also be a good idea to discuss the positioning of the text again. Anyway thanks for the positive feedback and patience everyone!

clemensvonmolo:
Okay this should work now. I've also created InfiniTimeOrg/InfiniSim#46 to fix the build issue it is currently facing with this PR.

src/displayapp/screens/Notifications.cpp
```diff
@@ -270,6 +284,26 @@ Notifications::NotificationItem::NotificationItem(const char* title,
...
+    if ((ageInt / (60 * 24)) >= 1) {
+      ageInt /= (60 * 24);
+      timeUnit = 'd';
+    } else if ((ageInt / 60) >= 1) {
+      ageInt /= 60;
+      timeUnit = 'h';
+    } else {
+      timeUnit = 'm';
+    }
+    lv_obj_t* alert_age = lv_label_create(container, nullptr);
```
Riksu9000, line 302: Variables names should be like this. Don't commit this.
```suggestion
    lv_obj_t* alertAge = lv_label_create(container, nullptr);
```

src/displayapp/screens/Notifications.cpp
```diff
@@ -270,6 +284,26 @@ Notifications::NotificationItem::NotificationItem(const char* title,
...
+    lv_label_set_text_fmt(alert_age, "%d%c ago", ageInt, timeUnit);
+    // same format as alert_count
+    lv_obj_align(alert_age, container, LV_ALIGN_IN_BOTTOM_RIGHT, 0, -16);
```
Riksu9000, line 305: This will make the padding similar to the text.
```suggestion
    lv_obj_align(alert_age, container, LV_ALIGN_IN_BOTTOM_RIGHT, -10, -10);
```

src/displayapp/screens/Notifications.cpp
```diff
@@ -270,6 +284,26 @@ Notifications::NotificationItem::NotificationItem(const char* title,
...
+  // almost impossible to receive a real notification at time 0, so skip because it is the "no notifications" notification
+  if (timeNow != 0) {
+    auto diff = std::chrono::system_clock::from_time_t(timeNow) - std::chrono::system_clock::from_time_t(timeArrived);
+    std::chrono::minutes age = std::chrono::duration_cast<std::chrono::minutes>(diff);
+    uint32_t ageInt = static_cast<uint32_t>(age.count());
+    char timeUnit;
+    if ((ageInt / (60 * 24)) >= 1) {
+      ageInt /= (60 * 24);
+      timeUnit = 'd';
+    } else if ((ageInt / 60) >= 1) {
```
Riksu9000, line 296: These can be simplified.
```suggestion
    if (ageInt >= 60 * 24) {
      ageInt /= (60 * 24);
      timeUnit = 'd';
    } else if (ageInt >= 60) {
```

Riksu9000 (changes requested):
In a follow up PR there could be special messages like "Just now" or "Yesterday".

src/components/ble/NotificationManager.h
```diff
@@ -57,6 +62,7 @@ namespace Pinetime {
       size_t NbNotifications() const;
 
     private:
+      Controllers::DateTime& dateTimeController;
```
NeroBurner, line 65: we can even store a `const` reference, to make it explicit, that a notification can't change the date-time, just get the current date and time

```suggestion
      const Controllers::DateTime& dateTimeController;
```

src/components/ble/NotificationManager.h
```diff
@@ -30,13 +32,16 @@ namespace Pinetime {
...
         uint8_t size;
+        std::time_t timeArrived;
         std::array<char, MessageSize + 1> message;
         Categories category = Categories::Unknown;
 
         const char* Message() const;
         const char* Title() const;
       };
 
+      NotificationManager(Controllers::DateTime& dateTimeController) : dateTimeController {dateTimeController} {
```
NeroBurner, line 43: we can even store a `const` reference, to make it explicit, that a notification can't change the date-time, just get the current date and time
```suggestion
      NotificationManager(const Controllers::DateTime& dateTimeController) : dateTimeController {dateTimeController} {
```

NeroBurner (changes requested):

yusufmte:
Any plans to rebase/eventually merge this sometime? Seems it would be a nice improvement.

RageGamerBoi:
This would go very well with #1697 if the grey notification box was made slightly smaller vertically to allow a small bottom bar for this and the notification icons.

Boteium:
lv_obj_t* alert_age should be the last object to be created. Otherwise, objects such as alert_caller will be on top of it and cause the alert_age to be invisible in IncomingCall message.

This also needs a rebase.

vchigrin:
Yet another friendly bump  :)

mark9064:
It looks like the original author is not working on this anymore, so if anyone would like to rebase this + apply the feedback onto a new PR you'd be totally welcome to do that

src/components/ble/NotificationManager.h
vchigrin: At present it is not possible since DateTime controller updates it member variables during time retrieving.
https://github.com/InfiniTimeOrg/InfiniTime/blob/d69cfcfb136e45486437ad773d362da7870042e3/src/components/datetime/DateTimeController.cpp#L72-L102

It can be overcomed with `mutable` but I am not sure whether this is a good idea.

vchigrin:
Rebased on the latest `main` branch in PR https://github.com/InfiniTimeOrg/InfiniTime/pull/2198 .
@clemensvonmolo  - if you prefer continue work on your own - I can step back. I am interested only in final result.

clemensvonmolo:
> Rebased on the latest `main` branch in PR #2198 . @clemensvonmolo - if you prefer continue work on your own - I can step back. I am interested only in final result.

Feel free to continue. I wouldn't have had time until after Christmas anyways and on top of that I'm not currently up to date with the codebase of InfiniTime which would only mean more work for me.

PR #839, opened 2021-11-22
Add documentation for BLE FS

Arsen6331:
This PR adds documentation for the BLE side of the BLE FS protocol, as implemented in #756, and will be updated with the mentioned PR.

### DIFF
--- a/doc/BLEFS.md
+++ b/doc/BLEFS.md
@@ -1,0 +1,167 @@
+# BLE FS
+---
+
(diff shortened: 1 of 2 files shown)


JF002:
I think it would be nice to merge this PR at (more or less) the same time than #756.
@Arsen6331 Does it still need some work or only reviews?

Arsen6331:
> I think it would be nice to merge this PR at (more or less) the same time than #756. @Arsen6331 Does it still need some work or only reviews?

BLE FS is probably going to have some changes, and I need to document those changes as they occur. The current version is ready for the current version of BLE FS.

Arsen6331:
It should be ready for merge at the same time as BLE FS, since I will document the changes as they happen.

geekbozu (approved):
As of right now I'm quite happy with this. The holiday season has kept me busy from shaping up the actual implementation of the BLE FS but @Arsen6331 has done a great service getting the documentation done!

I do not for see any major changes just additions, However this can all be crossed as we get there. For now getting a functional base so we can start to get companion app integration and move assets off to external flash is the main goal!

PR #845, opened 2021-11-28
Settings: more specific read and write mode

NeroBurner:
For each filesystem interaction be more specific if we want to read from
the file or write to it.

Doing a non-creating read on the loading of the settings file, otherwise
an empty file could be created, and when reading that empty file for the
initial settings I would expect an error (or random data) when reading.

I found this, trying to create a simulator for the file-access. Maybe I'm not understanding how `littlefs` works, but for me it seems like this should be an issue for the first read of the `settings.dat` file

---

Not tested on an actual device

### DIFF
--- a/src/components/settings/Settings.cpp
+++ b/src/components/settings/Settings.cpp
@@ -26,7 +26,7 @@ void Settings::LoadSettingsFromFile() {
   SettingsData bufferSettings;
   lfs_file_t settingsFile;
 
-  if ( fs.FileOpen(&settingsFile, "/settings.dat", LFS_O_RDWR | LFS_O_CREAT) != LFS_ERR_OK) {
+  if ( fs.FileOpen(&settingsFile, "/settings.dat", LFS_O_RDONLY) != LFS_ERR_OK) {
     return;
   }
   fs.FileRead(&settingsFile, reinterpret_cast<uint8_t*>(&bufferSettings), sizeof(settings));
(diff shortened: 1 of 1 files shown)


NeroBurner:
@geekbozu a friendly reminder/plea to review my PR :)

JF002:
We are currently looking for some bad memory corruption bugs. We refrain from merging new PRs until we understand a bit more about them. Sorry for the delay ;-)

geekbozu:
> 

I missed this the first time it came around, Apologies thanks for the nag!

So this looks properish. We need to triple check that when the settings file is actually loaded/dealt with else where that it will create the file if it does not exist as well as make sure that its loading sane defaults when the file does not exist either. 
On a technicality it does this already as the immediate next step is to check the version. Which 1/65565 times will not match on a random new file ;)

Needless to say I'll triple check how the file is saved but looks proper and a good catch from over here!

NeroBurner:
as far as I understand the following is done if there is no file present (maybe slightly different with littlefs, can't say. My understanding is from using `FILE` or `std::ifstream`)

- try to read settings file and failing because it is not available.
- We get the default values from the Settings.h default constructor
- modify some settings
- write settings and creating the settings file in the process

Without the current PR the following happened for me (on the PC):
- try to read non-existing settings file, creating a 0-length settings-file and succeeding
- throwing exception on `fs.FileRead(&settingsFile, reinterpret_cast<uint8_t*>(&bufferSettings), sizeof(settings));` because of the 0-length settings file

NeroBurner:
@geekbozu gentle ping to triple check these changes :bowing_man:

geekbozu (approved):
This looks good under the assumption that failing to load a file means that there are already sane defaults loaded into memory.
Thanks @NeroBurner

trman:
let's pray that   @JF002  will add it to 1.9.0 @NeroBurner  !

PR #848, opened 2021-11-29
Switch CI/CD to build using ninja vs. make

jharmer95:
By switching the CI/CD workflow to use CMake's `ninja` backend, the build time goes down considerably compared to GNU Make:

| | ninja | make |
| ----- | ----- | ---- |
| non-cached | ~5 min. | ~8 min. |
| cached | ~2 min. | ~4.5 min. |

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -113,7 +113,7 @@ jobs:
         run:  |
           mkdir -p build
           cd build
-          cmake -DARM_NONE_EABI_TOOLCHAIN_PATH=${{ runner.temp }}/arm-none-eabi -DNRF5_SDK_PATH=${{ runner.temp }}/nrf5_sdk -DUSE_OPENOCD=1 -DBUILD_DFU=1 ../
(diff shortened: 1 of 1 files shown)


geekbozu:
I see no issue with this, Specially since its only github specific.

geekbozu (approved):

.github/workflows/main.yml
```diff
@@ -126,7 +126,7 @@ jobs:
       - name: Make pinetime-mcuboot-app
         run:  |
           cd build
-          make pinetime-mcuboot-app
+          ninja pinetime-mcuboot-app
```
NeroBurner, line 129:
```suggestion
          cmake --build build --target pinetime-mcuboot-app
```

when already touching this part, you could switch away from manually calling the build command (`make` vs `ninja`), and needing to `cd build` beforehand. The two commands could be replaced by using just `cmake --build build --target pinetime-mcuboot-app` called from the source directory

jharmer95: That is a good point, we should try and follow modern CMake best practices where possible and make things universal

jharmer95: Made the change, thanks!

NeroBurner (approved):
LGTM :+1:

PR #856, opened 2021-12-02
Clock: add missing Settings include

NeroBurner:
Add missing include in `Clock.cpp` for `Settings.h`. The Settings class
is forward declared in the header file, but it needs to be included in
the cpp file.


NeroBurner:
@JF002 You're on a roll! Nice :)

PR #858, opened 2021-12-02
PineTimeStyle: move down notification icon away from bluetooth icon

NeroBurner:
The notification icon and the bluetooth icon overlapped. Move the Notification icon down a bit for better alignment.

up until now the PineTimeStyle looks like this:

20211202_PineTimeStyle

With the new notification placement it looks as follows (edit: now pixel perfect)
- Battery - 7px - bluetooth - 7px - notification - 8px - FRI
- Charger - 4px - bluetooth - 7px - notification - 8px - FRI

2021-12-03 battery 7px 7px 8px 2021-12-03 charge 4px 7px 8px

edit old pictures of first try:

20211202_PineTimeStyle with fixed notification charging 20211202_PineTimeStyle with fixed notification on battery

### DIFF
--- a/src/displayapp/screens/PineTimeStyle.cpp
+++ b/src/displayapp/screens/PineTimeStyle.cpp
@@ -108,7 +108,7 @@ PineTimeStyle::PineTimeStyle(DisplayApp* app,
 
   notificationIcon = lv_label_create(lv_scr_act(), nullptr);
   lv_obj_set_style_local_text_color(notificationIcon, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(0x000000));
-  lv_obj_align(notificationIcon, sidebar, LV_ALIGN_IN_TOP_MID, 0, 40);
+  lv_obj_align(notificationIcon, sidebar, LV_ALIGN_IN_TOP_MID, 0, 51);
 
   // Calendar icon
   calendarOuter = lv_obj_create(lv_scr_act(), nullptr);


NeroBurner:
I'll move the `i` a few more pixel further down, to have the alignment in the battery-case matching better. The daily-driver is on battery, not on charging, so I'll make it fit the battery usecase as perfect as possible

NeroBurner:
Now pixel perfect placement

- Battery - 7px - bluetooth - 7px - notification - 8px - FRI
- Charger - 4px - bluetooth - 7px - notification - 8px - FRI

2021-12-03 battery 7px 7px 8px 2021-12-03 charge 4px 7px 8px

src/displayapp/screens/PineTimeStyle.cpp
```diff
@@ -108,7 +108,7 @@ PineTimeStyle::PineTimeStyle(DisplayApp* app,
 
   notificationIcon = lv_label_create(lv_scr_act(), nullptr);
   lv_obj_set_style_local_text_color(notificationIcon, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(0x000000));
-  lv_obj_align(notificationIcon, sidebar, LV_ALIGN_IN_TOP_MID, 0, 40);
+  lv_obj_align(notificationIcon, sidebar, LV_ALIGN_IN_TOP_MID, 0, 51);
```
Riksu9000, line 111: Wouldn't it make more sense to make the positions 0, 25 and 50? Even if it wouldn't be pixel perfect, it would be independent of the font. In #718 this actually gets set to 50 as well. Also there will be a gap when the bluetooth icon isn't shown, so maybe it would be better to automatically realign the icons like in the digital watchface.

NeroBurner: I just wanted to fix a small alignment bug by changing as little as possible. Furthermore I assumed the non-realigning icons are a design choice (so I didn't want to touch it)

I don't get the argument with different fonts, we got the fonts fixed here. It's not like you upload different fonts to the pinetime?

The PR #718 looks very promising. Much better interaction to get to the color settings

Riksu9000: I meant to say it's basically a magic number that depends on the exact font and characters we are using, instead of a constant distance of 25.

NeroBurner: Yes, totally a magic number and font dependent :sweat_smile: 

Using the 0, 25, 50 proposal the following distances of the icons are seen:
- 6 px - battery - 9px bluetooth - 6px - notification - 9px - FRI
- 2 px - charger - 6px bluetooth - 6px - notification - 9px - FRI

battery 0 25 50 charger 0 25 50

If we can get consensus, that the 0, 25, 50 strategy is preferable to the magic 2, 25, 51 strategy I'll gladly change the PR. Everything is better than the overlapping icons ;)

Riksu9000: Oh actually I didn't realize that it started from 2 instead of 0. Let's just leave it like it is now then and leave the rest of the tweaking to if/when the icons are made dynamically aligning.

kieranc:
This fix is included in #718 but I have a better solution which places the icons side by side below the battery icon, and realigns them depending on how many icons are visible. I'll push it shortly...

NeroBurner:
closed in favor of #869

PR #860, opened 2021-12-04
Fix assertion failure in ButtonHandler

evergreen22:
FreeRTOS says zero is not a valid value for xTimerPeriodInTicks.
Zero value fires an assertion on line 361 in timers.h

### DIFF
--- a/src/buttonhandler/ButtonHandler.cpp
+++ b/src/buttonhandler/ButtonHandler.cpp
@@ -8,7 +8,7 @@ void ButtonTimerCallback(TimerHandle_t xTimer) {
 }
 
 void ButtonHandler::Init(Pinetime::System::SystemTask* systemTask) {
-  buttonTimer = xTimerCreate("buttonTimer", 0, pdFALSE, systemTask, ButtonTimerCallback);
+  buttonTimer = xTimerCreate("buttonTimer", pdMS_TO_TICKS(200), pdFALSE, systemTask, ButtonTimerCallback);
 }
 
 ButtonActions ButtonHandler::HandleEvent(Events event) {


src/buttonhandler/ButtonHandler.cpp
```diff
@@ -8,7 +8,7 @@ void ButtonTimerCallback(TimerHandle_t xTimer) {
 }
 
 void ButtonHandler::Init(Pinetime::System::SystemTask* systemTask) {
-  buttonTimer = xTimerCreate("buttonTimer", 0, pdFALSE, systemTask, ButtonTimerCallback);
+  buttonTimer = xTimerCreate("buttonTimer", pdMS_TO_TICKS(200), pdFALSE, systemTask, ButtonTimerCallback);
```
Riksu9000, line 11: I couldn't find the assert, but this change is fine as the default value is never used. Using `pdMS_TO_TICKS(200)` might be misleading though so setting it to something like 1 might imply a placeholder value better.

evergreen22:
I tested with 1 and 200ms. Both resolved the issue.

I selected 200ms because that is the same as a doubleClickTime in the
function below.

src/buttonhandler/ButtonHandler.cpp
telent: I hit this too. The assert is in nRF5_SDK_15.3.0/external/freertos/source/timers.c:361 (don't know if that's online in a linkable form anywhere) which says `configASSERT( ( xTimerPeriodInTicks > 0 ) );`

https://www.freertos.org/FreeRTOS-timers-xTimerCreate.html doc says "The timer period must be greater than 0. "

PR #862, opened 2021-12-04
Enable some extra checks and warnings

Avamander:
This PR enables some extra warnings, but also explicitly enables `-ftree-vrp` because `-Warray-bounds=2` won't work otherwise.


src/CMakeLists.txt
```diff
@@ -782,7 +782,7 @@ link_directories(
 )
 
 
-set(COMMON_FLAGS -MP -MD -mthumb -mabi=aapcs -Wall -Wno-unknown-pragmas -g3 -ffunction-sections -fdata-sections -fno-strict-aliasing -fno-builtin --short-enums -mcpu=cortex-m4 -mfloat-abi=hard -mfpu=fpv4-sp-d16 -Wreturn-type -Werror=return-type -fstack-usage -fno-exceptions -fno-non-call-exceptions)
+set(COMMON_FLAGS -MP -MD -mthumb -mabi=aapcs -Wall -Wextra -Warray-bounds=2 -Wformat=2 -Wformat-overflow=2 -Wformat-truncation=2 -Wformat-nonliteral -ftree-vrp -Wno-unused-parameter -Wno-missing-field-initializers -Wno-unknown-pragmas -Wno-expansion-to-defined -Wno-unknown-pragmas -g3 -ffunction-sections -fdata-sections -fno-strict-aliasing -fno-builtin --short-enums -mcpu=cortex-m4 -mfloat-abi=hard -mfpu=fpv4-sp-d16 -Wreturn-type -Werror=return-type -fstack-usage -fno-exceptions -fno-non-call-exceptions)
```
NeroBurner, line 785: `-Wno-unknown-pragmas` is listed two times now

Avamander: Good catch.

Avamander:
Fixed.

NeroBurner (approved):

Riksu9000:
This already found a bug in Displayapp where AlarmTriggered handling falls through to TouchEvent handling, which is nice. Just wondering if we should enable these just for our code because now nimble generates many warnings?

Avamander:
> This already found a bug in Displayapp where AlarmTriggered handling falls through to TouchEvent handling, which is nice. Just wondering if we should enable these just for our code because now nimble generates many warnings?

If there's a specific error in NimBLE that keeps spamming, I'd disable it for just that single target.

I've fixed a few warnings upstream already, so that's an option as well.

PR #863, opened 2021-12-04
Enabled stronger optimizations on nrf-sdk and disabled a warning

Avamander:
I think that the nRF SDK is fine to optimize as much as possible because it's IMO the most speed-sensitive (ISR, logging etc.) part of InfiniTime besides NimBLE.

I've also disabled a warning we can't fix, this significantly cuts down clutter coming from that module.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -804,10 +804,10 @@ add_library(nrf-sdk STATIC ${SDK_SOURCE_FILES})
 target_include_directories(nrf-sdk SYSTEM PUBLIC . ../)
 target_include_directories(nrf-sdk SYSTEM PUBLIC ${INCLUDES_FROM_LIBS})
 target_compile_options(nrf-sdk PRIVATE
-        $<$<AND:$<COMPILE_LANGUAGE:C>,$<CONFIG:DEBUG>>: ${COMMON_FLAGS} -Og -g3>
-        $<$<AND:$<COMPILE_LANGUAGE:C>,$<CONFIG:RELEASE>>: ${COMMON_FLAGS} -Os>
-        $<$<AND:$<COMPILE_LANGUAGE:CXX>,$<CONFIG:DEBUG>>: ${COMMON_FLAGS} -Og -fno-rtti>
-        $<$<AND:$<COMPILE_LANGUAGE:CXX>,$<CONFIG:RELEASE>>: ${COMMON_FLAGS} -Os -fno-rtti>
+        $<$<AND:$<COMPILE_LANGUAGE:C>,$<CONFIG:DEBUG>>: ${COMMON_FLAGS} -Wno-expansion-to-defined -Og -g3>
+        $<$<AND:$<COMPILE_LANGUAGE:C>,$<CONFIG:RELEASE>>: ${COMMON_FLAGS} -Wno-expansion-to-defined -O3>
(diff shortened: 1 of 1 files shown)


NeroBurner:
for reference size of target pinetime-app with current develop branch (commit 85a25302bfac215f9ec7b993f6a2c21a20ee223b Clock: add missing Settings include)

```sh
$ cmake --build build --target pinetime-app  --clean-first
[100%] Linking CXX executable pinetime-app-1.7.1.out
           FLASH:      364752 B       480 KB     74.21%
             RAM:       53672 B        64 KB     81.90%
post build steps for pinetime-app-1.7.1
   text    data     bss     dec     hex filename
 364752     940   52724  418416   66270 pinetime-app-1.7.1.out
[100%] Built target pinetime-app
```

size with optimizations (this PR)

```sh
$ cmake --build build --target pinetime-app  --clean-first
[100%] Linking CXX executable pinetime-app-1.7.1.out
           FLASH:      369776 B       480 KB     75.23%
             RAM:       53672 B        64 KB     81.90%
post build steps for pinetime-app-1.7.1
   text    data     bss     dec     hex filename
 369776     940   52724  423440   67610 pinetime-app-1.7.1.out
[100%] Built target pinetime-app
```

Riksu9000:
I think there would need to be a noticable improvement in some aspect to justify increasing the memory usage by 5k. I'm thinking that most user interaction "speed" is practically limited by the screen refresh rate, so does this improve the user experience in some way, or does the speedup improve reliability or have some other effects?

NeroBurner:
I think I used a Debug build (I didn't change the type)

```
[cmake] BUILD CONFIGURATION
[cmake] -------------------
[cmake]     * Version : 1.7.1
[cmake]     * Toolchain : /home/nero/repos/pinetime/gcc-arm-none-eabi-9-2020-q2-update
[cmake]     * GitRef(S) : 85a2530
[cmake]     * NRF52 SDK : /home/nero/repos/pinetime/nRF5_SDK_15.3.0_59ac345
[cmake]     * Target device : PineTime
[cmake]     * Programmer/debugger : OpenOCD Client
[cmake]     * Debug pins : Disabled
[cmake]     * Build DFU (using adafruit-nrfutil) : Disabled
```

Which config should I choose to get comparable (to the release) results?

Avamander:
>  I'm thinking that most user interaction "speed" is practically limited by the screen refresh rate, so does this improve the user experience in some way, or does the speedup improve reliability or have some other effects?

I was disassembling the firmware and given how the contents of nrf-sdk is very often used, speeding up things like ISR's, task switching, and bunch of other components that all depend on it, is probably worth it. But honestly, I don't have the tooling to empirically verify "what and how exactly".

Before:
```
           FLASH:      364660 B       480 KB     74.19%
             RAM:       53616 B        64 KB     81.81%
```

After:
```
           FLASH:      369428 B       480 KB     75.16%
             RAM:       53624 B        64 KB     81.82%
```

I would say the increase is quite minor

geekbozu (approved):
Personally I'm for this, It is something we can revert later if we need the space back. The gains might seem minimal but even if execution time is a little faster that means we get to sleep sooner which means we get more battery life as well. So until we are really scrambling for space. This seems logical to me.

JF002:
Well, I agree with @Riksu9000 regarding the increase of the binary size vs. the visible effect for the user, and also with @Avamander  and @geekbozu regarding the performance gains... 
@geekbozu 's arguments is good, let's merge and see what happens :)

PR #866, opened 2021-12-05
TouchHandler: remove unused FreeRTOS.h and task.h includes from header

NeroBurner:
FreeRTOS and task.h aren't used in the Header file. Furthermore the
SystemTask forward declaration isn't needed as it isn't used in the
header.


Avamander (approved):
Seem fine to remove from here.

Issue #831, opened 2021-11-15
Update 1.6.0 to 1.7.0 fails with "Error"


InFerYes:
### What Happened?

The update starts, but fails with "Error".

### What should happen instead?

The device should update from 1.6.0 to 1.7.0

### Reproduction Steps

Using GadgetBridge:

1. Download pinetime-mcuboot-app-dfu-1.7.0.zip
2. Open the file with GadgetBridge
3. The update process in GadgetBridge starts, GadgetBridge appears to be sending data but waits for the PineTime to receive
4. The PineTime shows `Firmware update 0%`
5. After a short while the PineTime shows `Firmware update error!`

Using Siglo:

1. Open Siglo and select Automatic under "Firmware update"
2. Select tag "1.7.0"
3. Select asset "pinetime-mcuboot-app-dfu-1.7.0.zip"
4. The update process in Siglo starts, Siglo appears to be sending data but waits for the PineTime to receive
5. The PineTime shows `Firmware update 0%`
5. After a short while the PineTime shows `Firmware update error!`

![https://i.imgur.com/SijiYW9.png](https://i.imgur.com/SijiYW9.png)

### More Details

I have tested the update process on both Android 11 and PinePhone Arch Linux.

### Version

1.6.0
### Companion App

Siglo 0.9.5-1, GadgetBridge 0.59.3, GadgetBridge 0.62.0

Avamander:
Did you restart both sides in the equation?

InFerYes:
The PineTime has been restarted many times, and GadgetBridge and Siglo run on different devices/platforms.

Avamander:
Yeah but do try restarting the other side as well, this switching between the two might've left something lingering.

InFerYes:
I have both started the Android phone and PinePhone cold and rebooted, and I have restarted the PineTime inbetween attempts.

TechnologyClassroom:
I could not reproduce.  Updating with GadgetBridge worked for me on the first try.

InFerYes:
I have always done it with GadgetBridge, but I just can't get it to work. I have restarted both devices a tens of times, the update process just shows the red `Error!`. Downgrading gives the same error.

I tried Siglo because it GadgetBridge wasn't working for the update, but to me it rules out the problem lies with GadgetBridge (or Siglo). 

The PineTime still receives time updates and notifications via GadgetBridge. 

What can I try?

TechnologyClassroom:
[...]

PR #867, opened 2021-12-05
Enabled LTO (and whitespace fixes)

Avamander:
Basically, the compiler we're using has a bug that would require us to patch `nrf-sdk` if we'd want to bypass it.

It's more reasonable to simply upgrade.

The RAM usage is yet unexplained, I need to double check the `asm` if it isn't some kind of padding or caused by mismatch of optimization flags.


evergreen22:
Does this PR close or obsolete #495 ?

Avamander:
I don't think so.

Avamander:
I see a lot of potential to further improve the set of optimization flags and the CMakefile (as done in https://github.com/InfiniTimeOrg/InfiniTime/pull/495). 

However, I don't want to do it with this PR, it is already large as it is and pr-rot will soon get it unless merged.

.github/workflows/main.yml
```diff
@@ -32,24 +32,24 @@ jobs:
...
-        uses: fiam/arm-none-eabi-gcc@v1.0.4
+      #- name: Check cache for Embedded Arm Toolchain arm-none-eabi-gcc
+      #  id: cache-toolchain
+      #  uses: actions/cache@v2
+      #  env:
+      #    cache-name: cache-toolchain-10.3-2021.10
+      #  with:
+      #    path: ${{ runner.temp }}/arm-none-eabi
+      #    key: ${{ runner.os }}-build-${{ env.cache-name }}
+      #    restore-keys: ${{ runner.os }}-build-${{ env.cache-name }}
```
Avamander, line 43: Couldn't get it working properly and it does not seem required (the toolchain download is cached), I think it's fine to disable?

Avamander:
@geekbozu If you could take a look if everything seems fine, that'd be nice.

.github/workflows/main.yml
geekbozu: I'm fine with this being disabled for now, we can re-enable it later for the actions speed up later.

Avamander:
@FintasticMan on Discord reported they're getting freezes. Investigating.

JF002:
It appears that FreeRTOS heap and the stack of the main task (systemtask) are probably overflown.
I increased the size of the heap (18*1024), the size of the stack of SystemTask and disabled the HeartRate task (to free a bit of RAM) in this DFU : 
[pinetime-mcuboot-app-dfu-1.7.98.zip](https://github.com/InfiniTimeOrg/InfiniTime/files/7781495/pinetime-mcuboot-app-dfu-1.7.98.zip)
This DFU also has RTT logging enabled, btw.

I'm running it right now on my sealed device, no issue to report right now.

ManuelBoe14:
> Patch : [Test_LTO.txt](https://github.com/InfiniTimeOrg/InfiniTime/files/7781499/Test_LTO.txt)
> 
> I'm running it right now on my sealed device, no issue to report right now.

I have also installed version 1.7.98 on my Pinetime.
So far everything seems to be working.

trman:
i installed 1.78 and the hr data no longer work (i need it for fitotrack)

do you plan to put back before merging in 1.8.0 @JF002 ?

if anything , and if it's truly needed to remove a task , it might be better to remove something else than a core functionality of the pinetime
 (  the core functionality a the functionality that have a physical counterpart : steps (step counter hw ) / hr (green led HW)).
 
 i see that the 1.78 is far bigger than usual ....

 do that mean that enabling lto will forbid to add new functionality (like the near completed calculator of @Raupinger )?
 
 for al i see  , and if this the case this pr should not been merged until it doesn't need to remove functionality and doesn't take more space for no visible /perceptible improvement .
 
 if the pr were to be merge as is by removing functionality and making impossible to add new one with removing existing app 
 it would be more a big regression than an improvement!!

ManuelBoe14:
> i installed 1.78 and the hr data no longer work

@trman version 1.7.98 is for testing purposes only.
JF002 has noted in his post above that he has deactivated the hr.

trman:
> > i installed 1.78 and the hr data no longer work
> 
> @trman version 1.7.98 is for testing purposes only. JF002 has noted in his post above that he has deactivated the hr.

i do understand this , but he removed it becaause otherwise the dfu would not work .....
that's why iw wish that he don't use this 'trick' to validate this pr .... because i would not be able to use fitotrack

Avamander:
@trman As written above, HR task was disabled. 

If you need HR, don't flash random test builds meant to demonstrate something wholly different. It is not meant for regular usage and wasn't labeled as such. Any qualms about a random test build are really not relevant here and such test builds shouldn't be taken as something that is going to be instantly released. 

Changes made are not all also immediately perceptible to the user, or really should be, there are other things that matter here as well.

Any issues you have with DFU are not relevant to this PR. You should really be expecting rough corners on something unreleased, it is unreleased for a reason. If you actually discover a bug that hasn't been reported, neither under that feature's PR or in an issue, then open a new one for that single bug.

JF002:
@trman
I'll give a bit of context about that 1.7.98 version. This is really a test build I built to check a single assumption : does increasing the heap of FreeRTOS and the stack size of SystemTask solve our memory corruption issue. The only purpose of this test build was to check the stability of InfiniTime with LTO enabled and the increased RAM allocation. Nothing else.

You've probably noticed : I didn't create a Pull-Request with those changes, because I don't want them to be merged. 

But, as I had to free a bit of RAM to increase the heap/stack, I had to disable the HR task so I can use the RAM memory allocated to this task for other purposes. Again, I did this only for the purpose of this test and I won't merge that in develop/master

> do that mean that enabling lto will forbid to add new functionality (like the near completed calculator of @Raupinger )?

The FLASH and RAM memories are already almost full, and that's why I cannot merge as many PR as I would like : I want to keep the memory usage under control, and some of these PRs have a lot of impact on the memory usage. The goal of LTO is to reduce the binary size (FLASH usage) so we can include some more functionalities. But, for some reason, using LTO seems to increase the RAM usage, and we still have to understand why and see what we can do about this.

> by the way , i need to go in recovery in order to flash dfu because , in secured bluetooth , i can't no longer do a dfu update ....
is it related of this pr ? if not it shoiuld be resolved as well before 1.8.0
because , i don't think user will like no be able to update dfu ....

This is not related to this PR, please see [this](https://github.com/InfiniTimeOrg/InfiniTime/issues/831#issuecomment-1003000691)

JF002:
Here's a TODO list so we know what must still be done in this PR:

- [ ] Measure the gain in FLASH space
- [ ] Analyze RAM memory usage, adapt FreeRTOS heap and tasks stack size so they don't overflow. Try to provide an explanation why the RAM usage increased when enabling LTO
- [ ] Fix DEBUG build
- [ ] Fix Docker build
- [ ] Check Github Actions build
- [ ] Check the build warnings returned by the compiler and fix them if relevant

JF002:
@Avamander The test build ran for more than 6 days on my sealed unit (until I need to reboot to test other features). So I think it's quite stable once the stacks/heaps are correctly set :+1:

trman:
any news on the remaining task?

src/main.cpp
```diff
@@ -299,7 +304,14 @@ void nimble_port_ll_task_func(void* args) {
 }
 }
 
-int main(void) {
+/**
+ * This is for fooling LTO to not remove a function referenced in inline asm
+ */
+static __attribute__((used)) void TaskSwitchDummy() {
+  vTaskSwitchContext();
+}
```
AlexXZero, line 312: This solution looks like a dirty hack, I would suggest to do the following more clear hack instead:
```diff
port.c:
void xPortPendSVHandler( void )
{
...
-    "   bl vTaskSwitchContext               \n"
+    "   bl _vTaskSwitchContext              \n"
...
}

+void __attribute__((used)) _vTaskSwitchContext(void)
+{
+    vTaskSwitchContext();
+}
```
Note: I found that this solution shouldn't add any overhead as `vTaskSwitchContext` will be inlined into `_vTaskSwitchContext`.

src/main.cpp
```diff
@@ -341,7 +352,11 @@ int main(void) {
 
   vTaskStartScheduler();
 
-  for (;;) {
+  // Yes, this should never be called, never is here, but it MUST be kept.
+  TaskSwitchDummy();
```
AlexXZero, line 356: In fact, this function will be called as it is placed before infinity loop!

src/FreeRTOS/port.c
```diff
@@ -73,26 +73,25 @@ void vPortStartFirstTask( void )
 #ifdef SOFTDEVICE_PRESENT
                     ::"i"(configKERNEL_INTERRUPT_PRIORITY  << (8 - configPRIO_BITS))
 #endif
-                );
+      " .ltorg                \n");
```
AlexXZero, line 76: Should this line be added before `#ifdef SOFTDEVICE_PRESENT`?

PR #868, opened 2021-12-05
Style improvements

Avamander:
This PR contains primarily style improvements, meaning no functional impact

* Improved format specifiers.
* Removed a few C-style casts.
* Removed useless trailing whitespace
* Removed Tiles shadowing (when LTOing, it threw a warning)


src/systemtask/SystemTask.cpp
```diff
@@ -417,18 +422,20 @@ void SystemTask::Work() {
     uint32_t systick_counter = nrf_rtc_counter_get(portNRF_RTC_REG);
     dateTimeController.UpdateTime(systick_counter);
     NoInit_BackUpTime = dateTimeController.CurrentDateTime();
-    if (!nrf_gpio_pin_read(PinMap::Button))
+    if (!nrf_gpio_pin_read(PinMap::Button)) {
       watchdog.Kick();
+    }
   }
-// Clear diagnostic suppression
-#pragma clang diagnostic pop
```
Riksu9000, line 424: Why was this removed? If this were to be removed, then shouldn't the corresponding push be removed too?

Avamander: I wanted to remove the comment, but accidentally removed both.

It's really difficult to separate all the changes into neat commits if you stumble upon them while investigating other things. Also the reason I try avoid larger PRs.

PR #872, opened 2021-12-10
Wrap PineTimeStyle step counter

aereaux:
This changes the PineTimeStyle step counter gauge to continue counting with wraparound after you achieve your step goal.

Let me know if this change isn't wanted or if there's anything I can improve.  I like it better this way, but I mostly did it to play around with developing for the Pinetime.

### DIFF
--- a/src/displayapp/screens/WatchFacePineTimeStyle.cpp
+++ b/src/displayapp/screens/WatchFacePineTimeStyle.cpp
@@ -431,7 +431,7 @@ void WatchFacePineTimeStyle::Refresh() {
   stepCount = motionController.NbSteps();
   motionSensorOk = motionController.IsSensorOk();
   if (stepCount.IsUpdated() || motionSensorOk.IsUpdated()) {
-    lv_gauge_set_value(stepGauge, 0, (stepCount.Get() / (settingsController.GetStepsGoal() / 100)));
+    lv_gauge_set_value(stepGauge, 0, (stepCount.Get() / (settingsController.GetStepsGoal() / 100)) % 100);
     lv_obj_realign(stepGauge);
     if (stepCount.Get() > settingsController.GetStepsGoal()) {
       lv_obj_set_style_local_line_color(stepGauge, LV_GAUGE_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_WHITE);


kieranc:
I have seen this but not got round to testing it properly yet, sorry. Looks like a useful addition, I'll up date here as soon as I've had a proper look

kieranc:
I've used this in a couple of test builds and it works great, @JF002 I think this could happily be merged?

Riksu9000:
Why should the counter wrap around? You wouldn't be able to know how many times it has wrapped around.

I think the gauge can already be a bit confusing to read, and this can make it even more so. I would rather have an arc, so the start and end points are clear, and maybe put a step icon in the middle which turns into a checkmark when the goal has been reached.

aereaux:
I feel that a circle is a lot easier to read a value from than an arc.  One can at least tell if it has wrapped around at all by the color of the dots, although it's true that it isn't too intuitive the first time.  I find myself not going over 20000 steps that often so this wraparound change would let me easily tell how many steps I've gone without going into the steps app.

One possible improvement would be to make the top dot of the gauge larger so that it's clear that it begins and ends there.

Riksu9000:
> I feel that a circle is a lot easier to read a value from than an arc.

Maybe estimating that for example a quarter of a 20k goal is 5k is easier with a circle, but the idea of the gauge is to visualize the progress towards the goal and not necessarily indicate the number of steps. Otherwise why not show the number of steps right there? The original Pebble TimeStyle actually shows the number of steps, which makes me wonder if we should do the same.

kieranc:
I would like to optionally display the number of steps (or HRM data, or weather) but this involves adding another page of settings etc, so this is a long term goal unrelated to this PR.

As it stands we have a gauge which goes round until you hit your goal then stops, and we can make it continue going round with a simple fix. Currently if my goal is 10k steps, once i hit that and look at my watch, I can't tell if I've done 10001, 15001 etc steps. With this change I can see if the gauge is white and it's half way round, that I've done 15000 steps roughly. It might be that I've done 25000, but if this is the case I'd probably know about it.

I think this PR is a useful improvement to the current step gauge - other improvements will happen but not in this PR.

Riksu9000:
It's not a fix, but a change in behaviour. It shouldn't be changed if it is likely to be changed again soon. I think it's worth considering replacing the gauge altogether, so we should discuss that before changing the current behaviour.

kieranc:
> It's not a fix, but a change in behaviour. 
Yes, but I (and the PR's author) consider it an improvement, I probably would have done it like this originally if I'd though of it (and known how).

> It shouldn't be changed if it is likely to be changed again soon. 
'Soon' could be 6 months away, or never, I'd have to get feedback on the current gauge and potential alternatives, I'd be surprised if there was a convincing consensus on which option is best. I did try several alternatives while working on this one, including an incomplete and half circle, but this was the one which looked and worked best.

> I think it's worth considering replacing the gauge altogether, so we should discuss that before changing the current behaviour.
Sure, I can test some alternatives while doing the weather integration but this being merged in the mean time doesn't impact that, or anything much else. We have quite a lot of PRs stagnating because they're not perfect, and people who have made the effort to contribute being discouraged as a result. Perhaps not the most technically sound reason for merging stuff, but I feel it's important, particularly in this case where there is no question about code quality, it's clean, simple and does exactly what it says on the tin.

JF002:
I have no firm opinion about this one, UI and UX are not my domain of expertise :)
I think PTS will have a lot of additions and improvements in the (near) future as Kieran have a lot of ideas (configurable display, weather widgets,...), and the layout of the watchface will probably need a few fine-tuning to fit all of them.

So, should the gauge wrap around? I don't know... I guess it's a matter of personal preference. In this case, as @kieranc is the author of PTS and as he maintains it quite well, I think we can let him have the final say on how it should evolve (after ~~listenning~~ reading to our opinions, of course :) )

kieranc:
I would prefer this to be merged, I feel it's an improvement, and a tidy PR. The step gauge should continue to do _something_ once the goal is reached, and continuing to show progress seems logical. Even if it's not the final/perfect form of the step gauge, it will not cause any issues for future changes. Once I have weather integration functional I'll work on adding the extra settings so the user can choose step count or gauge etc etc

aereaux:
The other idea I had was to continue changing colors as you went above 20k, 30k....  That seems like it would get pretty confusing, though

aereaux:
Looks like there's been a lot of good improvements in the new release!  It's been a while on this PR, is there anything else that would help it get merged?

Riksu9000:
I still feel that making it wrap would make it more confusing and I would look for other solutions.

kieranc:
If it's confusing of not is difficult to quantify, the behavior is documented and people occasionally ask what the gauge is/how it works but it doesn't take much explaining. I think we have 2 options, one of which is that when you reach your goal the gauge becomes useless as it just stops, you don't see how far beyond your goal you are, alternatively the needle continues to move with the divisions a different color so you know you've passed your goal, and by roughly how much. I am firmly in favor of the 2nd option.

Riksu9000:
I could keep arguing, but I don't think it is worth continuing this discussion. I'll close this as the change is insignificant and the readability of the gauge should be improved instead.

aereaux:
What sort of readability changes would be accepted?  I'd be willing to write more code for this if there's something that would be merged.

kieranc:
> What sort of readability changes would be accepted? I'd be willing to write more code for this if there's something that would be merged.

I'm also interested in an answer to this question

Riksu9000:
I previously wrote an arc or a numbered step count as examples. If it should show the progress towards the goal, whatever the indicator, it should have clear start and end points and it should get filled to indicate progress. These changes will make it immediately understandable.

kieranc:
Numerical representation is a possibility once the sidebar widgets are configurable (if I can ever make weather work...) but for the time being I much prefer a visual representation to a numerical one, so while there's only one option, it is visual.
I did try an arc before choosing the gauge but it was aesthetically difficult and significantly smaller/less easily readable than the gauge since it's limited by the width of the sidebar. In theory an arc could be a large portion of the circle, 300+ degrees but I don't think it's better, and iirc there were technical issues making it work with our version of lvgl and the divisions.
I don't think that an arc is significantly or inherently more understandable than the gauge, especially given it would be smaller, and an arc doesn't have the ability to continue providing useful information after the step count has been reached which the gauge does.
In any case, without wanting to be an ass, JF indicated that this should be my decision and what I would like is for this to be merged while other alternatives/improvements are worked on.

aereaux:
After some further testing apparently you can do this sort of gradient:
20220714_17h23m05s_grim

Not sure if that would help with some of the start/end point concerns

PR #873, opened 2021-12-10
bug: turn screen on in recovery

telent:
# symptom

When I run the pinetime-recovery image I can't see anything on the
screen, although the device appears to be operating otherwise (shows up
in nRF Connect, transmits RTT logs)

# hypothesis

Brightness is set to  Off

# resolution

Set it to something else

### DIFF
--- a/src/displayapp/DisplayAppRecovery.cpp
+++ b/src/displayapp/DisplayAppRecovery.cpp
@@ -48,6 +48,8 @@ void DisplayApp::Process(void* instance) {
 }
 
 void DisplayApp::InitHw() {
+  brightnessController.Init();
+  brightnessController.Set(Controllers::BrightnessController::Levels::High); // Controllers::BrightnessController::Levels.
   DisplayLogo(colorWhite);
 }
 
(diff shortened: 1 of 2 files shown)


petterhs:
The recovery firmware is intended to enable BLE OTA firware updates and be as small as minimal as possible thus not needing the display

escoand:
Any background color is not better than any other IMO. You'll never know is it frozen or defect. Only more useful option than a plain color is some kind of description what and how to do an update.

telent:
> The recovery firmware is intended to enable BLE OTA firware updates and be as minimal as possible thus not needing the display

There is existing code in the  recovery firmware to show the Infinitime logo, to change its color when a bluetooth peer connects, and to show progress of an OTA update - which IMO is a vastly better UX than a blank black screen with no indication of what is happening or what to do. Or it would be if any of it were visible, but on my watch it *is* a blank black screen because the brightness defaults to `Off`

JF002:
The recovery firmware should display the InfiniTime logo colored in white when no BLE connection is established and in blue when it's connected to a BLE device. It also shows a progress bar when a DFU file is being transferred. So I think @telent  is right : the recovery firmware should display something.

Due to a lack of time, we don't test and release the recovery firmware each time we release a new version of InfiniTime. With the recent addition made to the BLE implementation (secure pairing, mostly), we'll probably need to apply some changes to the recovery firmware to ensure it's still working fine and reliably (and probably without the secure pairing feature enabled).

PR #874, opened 2021-12-12
FIX: AlarmTriggered fall through to ShowPairingKey

evergreen22:
Add missing break statement - AlarmTriggered case falls through to ShowPairingKey case


Riksu9000 (approved):

Avamander (approved):

Avamander:
@JF002 this would be a nice quick fix to merge.

JF002:
Yep, good catch! Thanks!

PR #875, opened 2021-12-12
Remove unused bits in Hrs3300::ReadAls, ReadHrs

yehoshuapw:
First, thanks a lot! great work making this and continuing to make it.

(commit message:)
Since both functions `Hrs3300::ReadHrs`, `Hrs3300::ReadAls` return
`uint16_t`:
* removed extra 2 bits (at position 16,17) in HRS, which will be thrown
  out
* removed unneeded checks of garbage bits in ALS, since they too will be
  thrown out.

I have a bit more doubts about the second change (not doing ` & 0x3f `), since while it still does happen, it does not clearly say that the MSB of the H register is trash. (since for the uint16_t all the 3 MSB are trash)


yehoshuapw:
So after a discussion, it seems that the low level driver should not cut off the extra 2 bits, and that can (and perhaps should for memory issues) be done at a higher level.

Therefore, I am closing this PR.

PR #876, opened 2021-12-13
Hrs driver changes

yehoshuapw:
changed low level driver for HRS3300 to return uint32_t values (since it gives 18 bit values) 
and changed hidden uint16_t to float convert, to a static_cast.

supersedes #875

thanks for all the great work!


Riksu9000 (approved):
The resolution is limited to 16 bits, which is why this hasn't been an issue, but it should be fixed so there won't be any issues in the future.

JF002:
@yehoshuapw  Thanks, and congratulations for your first contribution to the project!

PR #878, opened 2021-12-13
Add airplane mode feature

### LINKED ISSUE #632: Disable Bluetooth
Hi - is there a way to disable bluetooth from the watch?  If not, could it be added as a feature request?

evergreen22:
~~# Add airplane mode feature~~

Refactor the code for style in preparation for adding an airplane mode feature. More design work needs to be done.

### DIFF
--- a/src/components/ble/NimbleController.cpp
+++ b/src/components/ble/NimbleController.cpp
@@ -23,14 +23,14 @@
 using namespace Pinetime::Controllers;
 
 NimbleController::NimbleController(Pinetime::System::SystemTask& systemTask,
-                                   Pinetime::Controllers::Ble& bleController,
+                                   Ble& bleController,
                                    DateTime& dateTimeController,
-                                   Pinetime::Controllers::NotificationManager& notificationManager,
-                                   Controllers::Battery& batteryController,
+                                   NotificationManager& notificationManager,
+                                   Battery& batteryController,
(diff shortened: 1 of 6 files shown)


yehoshuapw:
Maybe recovery should ignore this, otherwise if there is some error, while this is on - you cannot update the firmware over Bluetooth.

ItzSwirlz:
I personally think we just need better options with the Bluetooth (choose connections to connect to). It would make things easier

Itai-Nelken:
The 'firmware' and 'about' settings sbould be at the bottom as they are the least used setting in my opinion.

lman0:
@evergreen22   can you add a plane icon  when in "airplane mode" that replace the bluetooth icon?
in the the digital watchface first (and maybe other watch if you want/can ).

thanks


the reason :
because , when i on a plane i deactivate the bluetooth of my phone , then  the bluetooth icon disappear on infinitime yet it's still advertise.
having a icon for airplane mode would be helpfull and less stressfull (i would not have to check every time if i'm on airplane mode , in the seting)

uis246:
> I personally think we just need better options with the Bluetooth (choose connections to connect to). It would make things easier

From my comment under issue:
> I think there will several modes: "noisy"(with adv), normal(without adv), quiet(if possible to use rx only) and disabled.

Issue #880, opened 2021-12-17
Breaking change - Secure/BLE pairing


evergreen22:
This applies to a 1.7.1 develop build that includes secure pairing PR #796.

While chatting about whether or not to persist the airplane mode setting, @kieranc  asked why not. After explaining the potential issue with recovery, the question was asked about how secure pairing might impact recovery. It appeared on my screen at the same moment I realized it is also a potential issue. Thanks @kieranc !

Good news - when using the recovery firmware from a NOT bonded companion everything works (tested with GB and nRFConnect).

However, when using the recovery firmware from a bonded companion you will be unable to establish a connection until the bond is deleted from the companion.

This is because the recovery firmware installed on watches is too old to understand secure connections.

This could be addressed at least three ways:

1. Clearly document that the bond must be deleted from the companion to use the recovery firmware.

2. Update the recovery firmware to handle secure and non-secure connections. This only solves the problem for people who update their recovery firmware.

3. Do not persist bonds on the watch. Recovery firmware works for everyone, but you'll have to re-pair your watch on every boot.

In a related note, this means that saving the airplane mode setting doesn't impact recovery today. It might in the future though if people install a recovery firmware that enforces airplane mode.

Finally, should we consider separating the recovery firmware and making it a standalone repo similar to mcuboot?

evergreen22:
I'm migrating this conversation to an issue in order to organize the discussion as suggested by @geekbozu .

FWIW, the consensus to date is option (1)

geekbozu:
I really think we should just have a check for recovery to bluntly ignore all loading from flash. Incase it gets damaged or similar. Recovery should be as standalone as possible. Even if that makes it an inconvenience. IT working reliably is More important then all.

evergreen22:
That was one of my initial thoughts as well. However, that will mean updating the recovery firmware and getting everyone to install the updated version.

I think we are stuck with some period of time where folks will have the current 'basic/dumb' recovery firmware and a new recovery-aware version.

evergreen22:
Also, your argument @geekbozu  clearly articulates why I think the recovery firmware should be separated from the feature firmware.

geekbozu:
> That was one of my initial thoughts as well. However, that will mean updating the recovery firmware and getting everyone to install the updated version.
> 
> I think we are stuck with some period of time where folks will have the current 'basic/dumb' recovery firmware and a new recovery-aware version.

[...]

PR #886, opened 2021-12-21
Optional secure pairing with a passkey

evergreen22:
Support passkey pairing when requested by the central.
Also, fixes a comment and a reorder warning introduced by a faulty conflict resolution.

### DIFF
--- a/src/components/ble/NimbleController.cpp
+++ b/src/components/ble/NimbleController.cpp
@@ -134,9 +134,7 @@ void NimbleController::Init() {
 
   RestoreBond();
 
-  if (!ble_gap_adv_active() && !bleController.IsConnected()) {
-    StartAdvertising();
-  }
+  StartAdvertising();
 }
 
 void NimbleController::StartAdvertising() {
@@ -274,7 +272,7 @@ int NimbleController::OnGAPEvent(ble_gap_event* event) {
        * display capability only so we only handle the "display" action here.
        *
        * Standards insist that the rand() PRNG be deterministic.
-       * Use the nimble TRNG here since rand() is predictable.
+       * Use the tinycrypt prng here since rand() is predictable.
        */
       NRF_LOG_INFO("Security event : BLE_GAP_EVENT_PASSKEY_ACTION");
       if (event->passkey.params.action == BLE_SM_IOACT_DISP) {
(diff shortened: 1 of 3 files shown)


geekbozu (approved):
Looks good, 
Thanks for staying on top of this!

lman0:
work properly .
moreover unlike before , the passkey is shown only if gadgetbridge request a bonding unlike before , that is good
(basically , the passkey was automatically shown once bluetooth was activated , now it show itself only when we add infinitme as bond from gadgetbridge by the add device button).

thanks

src/components/ble/NimbleController.cpp
```diff
@@ -274,7 +272,7 @@ int NimbleController::OnGAPEvent(ble_gap_event* event) {
        * display capability only so we only handle the "display" action here.
        *
        * Standards insist that the rand() PRNG be deterministic.
-       * Use the nimble TRNG here since rand() is predictable.
+       * Use the tinycrypt prng here since rand() is predictable.
```
Avamander, line 275: This is intentional?

Avamander (approved):

evergreen22:
Closes #880

JF002:
Thank you very much, @evergreen22! Bluetooth pairing now works exactly as expected:

- Gadgetbridge : when connecting to the PinetTime, Gadgetbridge asks if the user wants to pair with the device. If he chose to pair, the PIN code is displayed on the PinetTime and can be entered in Gadgetbridge. If the user does not want to pair, vanilla connection is established.
- On Linux (Manjaro, KDE plasma), Siglo and Amazfish work exactly like before : they connect and communicate with the watch. Sometimes, KDE asks me if I want to "authorize and trust" my PineTime. I don't know what it does, but it doesn't ask for the PIN code and siglo/amazfish cannot connect anymore.
- On Linux, I can manually pair using `bluetoothctl` (type the command `pair xx:yy:zz:11:22:33` and enter the PIN code displayed on the PineTime). Once it's done, Siglo/Amazfish can connect to the PineTime again.
- ITD supports secure pairing out of the box, and it works!

Great job everyone!

If I had only 1 suggestion : add a visual indicator showing if the connectivity is secured or not (ble icon with another color? another icon? something else?).

JF002:
Update : according to my observations, OTA is much more reliable on **not bonded** connections! Using my new phone, it's nearly impossible to start an OTA (it fails at 0% after a few seconds) when bonded. WIthout the bond, OTA works as expected (and a lot faster with my new phone : 12kbps).

For some reason, it still works (bonded or not) on my older phone (but it's so slow... 2-3 kbps).

evergreen22:
>If I had only 1 suggestion : add a visual indicator showing if the connectivity is secured or not (ble icon with another color? another icon? something else?).
>

A padlock icon to the left of the ble icon would be nice when the 
connection is encrypted.

PR #888, opened 2021-12-24
Airplane mode feature (redesigned)

### LINKED ISSUE #632: Disable Bluetooth

evergreen22:
# Add airplane mode feature (redesigned)

- [X] Implement 'Airplane mode' to disable and enable bluetooth
- [X] Add a setting menu for switching airplane mode on and off
- [X] Add an airplane icon that is displayed when airplane mode is enabled

You turn on airplane mode by swiping and selecting settings, selecting *Airplane mode*, then choosing **Yes**. This will disconnect the watch if it is currently connected or cancel advertising if it is not connected. When airplane mode is on you cannot connect to the watch (i.e. bluetooth communications are disabled).

To turn airplane mode off, swipe and select settings, select *Airplane mode*, and choose **No**. If your companion app is set to automatically reconnect then your watch will automatically connect.

Note that in both cases, turning airplane mode on and off, the change happens when you **close** the *Airplane mode* setting menu (swipe down).

The airplane mode setting is **not** saved. Rebooting the watch will turn off airplane mode.

In addition, the following style and organizational changes are included in PR:
* Alphabetize the settings menu options
* Style cleanups

Resolves #632


src/displayapp/screens/settings/Settings.cpp
```diff
@@ -36,10 +36,10 @@ bool Settings::OnTouchEvent(Pinetime::Applications::TouchEvents event) {
 std::unique_ptr<Screen> Settings::CreateScreen1() {
 
   std::array<Screens::List::Applications, 4> applications {{
```
mattchrist, line 38: Reordering the settings list seems unrelated to airplane mode to me.

trman:
hi ,

i have tested the pr and it work great !

the setting order is pretty disturbing at first but become intuitive after two day or so .

great redesign @evergreen22  ! 

i wish that it would be merged by @JF002  for 1.8.0 too !

geekbozu:
Code looks clean, 
Do you have the size increases associated with this?

trman:
> Code looks clean, Do you have the size increases associated with this?

for what i see in the action @geekbozu  , i'ts around 1k more (maybe related to the new icon )

src/displayapp/screens/settings/SettingAirplaneMode.h
```diff
@@ -0,0 +1,30 @@
...
+
+      class SettingAirplaneMode : public Screen {
+      public:
+        SettingAirplaneMode(DisplayApp* app, Pinetime::Controllers::Settings& settingsController);
+        ~SettingAirplaneMode() override;
+
+        void UpdateSelected(lv_obj_t* object, lv_event_t event);
+
+      private:
+        static constexpr std::array<const char*, 2> options = {" No", " Yes"};
```
Riksu9000, line 23: I think the "Yes" option should be above "No"

src/components/ble/NimbleController.cpp
```diff
@@ -376,6 +380,22 @@ void NimbleController::NotifyBatteryLevel(uint8_t level) {
...
+      bleController.SetConnectState(Ble::ConnectStates::Airplane);
+      ble_gap_adv_stop();
+    }
+  } else {
+    bleController.SetConnectState(Ble::ConnectStates::Disconnected);
+    fastAdvCount = 0;
+    StartAdvertising();
+  }
+}
+
```
Riksu9000, line 398: I would rather have this be two functions without a boolean parameter.

src/displayapp/screens/settings/Settings.cpp
```diff
@@ -36,10 +36,10 @@ bool Settings::OnTouchEvent(Pinetime::Applications::TouchEvents event) {
...
-    {Symbols::sun, "Display", Apps::SettingDisplay},
-    {Symbols::eye, "Wake Up", Apps::SettingWakeUp},
-    {Symbols::clock, "Time format", Apps::SettingTimeFormat},
-    {Symbols::home, "Watch face", Apps::SettingWatchFace},
+    {Symbols::list, "About", Apps::SysInfo},
+    {Symbols::airplane, "Airplane mode", Apps::SettingAirplaneMode},
+    {Symbols::batteryHalf, "Battery", Apps::BatteryInfo},
+    {Symbols::sun, "Display", Apps::SettingDisplay}
```
Riksu9000, line 42: The order shouldn't be alphabetized and shouldn't be changed in this PR.

Riksu9000 (commented):
Code looks good and seems to work as expected. I can't test if the watch leaks *any* signals, but it disconnects and can't be found when scanning.

It would be much better to have this option in the quick settings, though that's obviously not possible yet. We should look into reworking that so it can hold more functionalities as they come.

src/displayapp/screens/settings/Settings.cpp
JF002: I agree with @Riksu9000, I'm not sure the alphabetized order is the best one, and I would rather like to see this change in a dedicated PR.

src/displayapp/screens/settings/SettingAirplaneMode.h
JF002: +1

src/components/ble/BleController.cpp
```diff
@@ -2,12 +2,12 @@
 
 using namespace Pinetime::Controllers;
 
-void Ble::Connect() {
-  isConnected = true;
+void Ble::SetConnectState(Ble::ConnectStates newState) {
```
JF002, line 5: I'm not sure we should mix "connection state" (connected and disconnected are implemented, and other future states could be "error", "failed", "refused",...) with an "operating mode" (normal and airplane mode).
I think they are 2 distinct concept that should be kept separate : 
 - ConnectionMode : What is the status of the BLE connectivity : are we connected or not ?
 - OperatingMode (or ConnectionAllowed, or something similar?) : Do we accept new BLE connections ?

What do you think?

src/components/ble/NimbleController.h
```diff
@@ -42,18 +42,19 @@ namespace Pinetime {
...
-                       Controllers::MotionController& motionController,
-                       Pinetime::Controllers::FS& fs);
+                       HeartRateController& heartRateController,
+                       MotionController& motionController,
+                       FS& fs);
       void Init();
       void StartAdvertising();
       int OnGAPEvent(ble_gap_event* event);
 
+      /* these are not implemented yet
```
JF002, line 57: You can safely remove those declarations of the methods are not implemented in NimbleController.cpp.

JF002 (changes requested):
Thanks for this nice and complete PR (new feature + settings + icon, perfect!).

Could you have a look at the @Riksu9000's and mine reviews?

JF002:
@evergreen22 I have one question regarding this PR : If I understand correctly the changes, the radio is not "physically" switched off. Do you know if there's any chance the radio is still emitting signal when the airplane mode is enabled? 
I think this PR is still valid even if it does, but I think this would be good to know!

Also, what prevents a bonded device from re-connecting to the watch, as a bonded device does not need an advertising packet to be able to connect to the device ?

src/displayapp/screens/settings/SettingAirplaneMode.h
evergreen22: I disagree.

I did a HCI/UX paper prototype study to test the following options before selecting this option (n=16):
* Yes/No 56% of respondents performed correctly
* On/Off 25% of respondents performed correctly
* Enable/Disable 19% of respondents performed correctly

Regarding the order, the existing UX displays the default choice first and "No airplane mode" is the default mode. Therefore the order should be "No" then "Yes" for consistency.

src/components/ble/NimbleController.cpp
evergreen22: I disagree. I see no reason to make the code larger.

src/displayapp/screens/settings/Settings.cpp
evergreen22: I disagree. While I concede this may not be an ordering that makes 100% of people happy (does that even exist?), I feel strongly that the aribtrary order is wrong from a UX standpoint because it creates an unecessary cognitive burden on the user.

If we order the options alphabetically now I expect it will kick off a serious, evidence based discussion on how this is best resolved. Folks posting about which option they use the most (FWIW I use "About" the most) is not moving the ball forward and just adding new options in random spots is procrastination.

At the very least, alphabetical ordering is a common, accepted UX pattern for options that is widely understood.

src/components/ble/NimbleController.h
evergreen22: I didn't remove them because I wasn't sure if they had any value from a documentation standpoint.

src/displayapp/screens/settings/Settings.cpp
Riksu9000: I agree that alphabetical ordering has its benefits. However when changing languages should the order change according to the used language? Alphabetical ordering isn't common for settings on other platforms either. At the very least this will not be changed in a PR about airplane mode.

src/displayapp/screens/settings/SettingAirplaneMode.h
Riksu9000: I don't really understand what you mean. Do you mean that if a person goes to enable the airplane mode, only 19% would be able to press the correct button to enable it? I highly doubt that is the case. "Airplane mode. Yes/No" doesn't really make sense in a literal sense either.

Having the default option first sounds reasonable, but somehow having "No" above "Yes" doesn't feel right. Unfortunately I'm not able to explain why.

This could also just be a switch widget.

src/components/ble/NimbleController.h
JF002: I think I forgot to remove them a long time ago. They are not needed since these functionalities are implemented is services.

JF002:
I applied the changes I asked for and merged the PR. Thanks @evergreen22  and everyone who contributed to this PR!

WhyNotHugo:
Why is this called "airplane mode"? The watch only has bluetooth, which is nowadays allowed in flights and also actively used. The watch has no cellular antenna either, which _does_ need to be disabled in planes.

I relate deactivating bluetooth more with privacy and battery saving, TBH.

Anyway, I wanted to bring up https://github.com/InfiniTimeOrg/InfiniTime/issues/1060, I think there's an overlap in functionality between this PR and that feature request.

Mostly, I'd like bluetooth to be turned on after the device boots from an out-of-battery situation. The use case is: I've let it run out of battery and put it in the charging case. When it wakes up, I want it to auto-sync the time with my laptop ASAP, and then turn bluetooth off again. I think this PR is a `on`/`off` switch, and I'm mostly looking for an `auto` setting.

Riksu9000:
@WhyNotHugo Nobody had thought of that. Renaming it might be a good idea.

PR #894, opened 2021-12-30
Implement showermode

gnome71:
This pull request adds a "showermode". It can be enabled by a long tap on the Analog- or DigitalWatchFace. It basically just saves the current WakeUpMode and then disables it. The watch will then not be woken up by contact with water - hence "showermode". To disable the showermode, wake up the watch by ButtonPress and LongTap again on Analog- or DigitalWatchFace. If the showermode is active, is shown by an icon on these WatchFaces.

What's not implemented yet, is to send the watch to sleep 5 seconds after activating the showermode to not have to wait for whatever ScreenTimeout is set.


geekbozu:
I like the Idea of this. However I'm not sure we want to actually recommend the PT go in the shower....depending on the glue work I have seen a few units fail due to water ingress...they really are only splash proof. 

Good work either way!

Itai-Nelken:
This probably conflicts with the PTS settings that alsonuses the long press since #718

gnome71:
Thanks for your reply. At least I learned a lot about InfiniTime and the Git workflow. Keep up your good work!!

trman:
feature need to be reworked  @gnome71 , and name of functionality changed  to no longer talk about putting pinetime inside water like requirement of @Riksu9000   in  #961

Riksu9000:
There's a simpler way to do this #961. I'm also assuming this isn't something that we'll be adding.

shvechkov:
> There's a simpler way to do this #961. I'm also assuming this isn't something that we'll be adding.

I would still recommend adding this feature - when exercising long (sweaty) sleeves  touch the screen and mess all the settings on my PT ... Locking screen is MUST HAVE feature for folks who wear PT in the gym ...

PR #904, opened 2022-01-01
Added 12 hour format to set time screen

eliwss0:
Changed the SettingSetTime screen to automatically use a 12 hour interface when format is set to 12 hour. Only changed interface; logic still uses 24 hour time.

Still cleaning up to avoid if else chains.


eliwss0:
Merging from wrong branch. Opening new PR

PR #905, opened 2022-01-02
Fix indev_drv and fs_drv going out of scope, store as member variable

NeroBurner:
The variables fs_drv (FS.h) and indev_drv (LittleVgl.h) are going out of
scope after the init.

Store them as member variables to keep them alive.

Originally found by @Quantum-cross in lvgl8 porting branch. Cleaned up and extracted by me to make the lvgl8 branch smaller. I don't know how critical this change is, but fs and input device driver objects running out of scope seems non-intentional to me.

Tested on pintime:
- connects to gadgetbridge
- apps like twos and others start
- switching Watchfaces works

### DIFF
--- a/src/components/fs/FS.cpp
+++ b/src/components/fs/FS.cpp
@@ -181,8 +181,6 @@ namespace {
 }
 
 void FS::LVGLFileSystemInit() {
-
-  lv_fs_drv_t fs_drv;
   lv_fs_drv_init(&fs_drv);
 
   fs_drv.file_size = sizeof(lfs_file_t);
--- a/src/components/fs/FS.h
+++ b/src/components/fs/FS.h
@@ -3,6 +3,7 @@
 #include <cstdint>
 #include "drivers/SpiNorFlash.h"
 #include <littlefs/lfs.h>
+#include <lvgl/lvgl.h>
 
 namespace Pinetime {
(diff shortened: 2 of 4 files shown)


JF002:
I totally agree with you, this code seems buggy BUT... in this case, I don't think this is a bug, that's how LVGL is designed.

Let's have a look at [lv_fs_drv_register()](https://github.com/lvgl/lvgl/blob/v7.11.0/src/lv_misc/lv_fs.c#L503) : it copies (`lv_memcpy()`) the content of the struct pointed by drv_p (which is `fs_drv`) into a variable maintained by LVGL. See the comment above the function : `drv_p pointer to an lv_fs_drv_t structure which is inited with the corresponding function pointers. The data will be copied so the variable can be local.`

And it's the same for `indev_drv` : in [lv_hal.c](https://github.com/lvgl/lvgl/blob/v7.11.0/src/lv_hal/lv_hal_indev.c#L62), the content of `indev_drv` is copied into a struct maintained by LVGL, and the function has the same comment  `driver pointer to an initialized 'lv_indev_drv_t' variable (can be local variable)`.

NeroBurner:
Thanks for taking the time and explaining it so thoroughly!

closing and removing the commit from the lvgl8 branch

PR #907, opened 2022-01-02
Added 12 hour set time interface

eliwss0:
Added appropriate set time interface for 12 hour setting. Time is still kept and set as 24 hours, but interface changes to reflect 12 hour time if user chooses 12 hour format.


Izaic:
Does this also fix the alarm app time setting?

eliwss0:
No it does not.

eliwss0:
Opening new pull request because branch was deleted

eliwss0:
Reopening pull request

eliwss0:
This PR only changes the set time interface. I will look into the alarm app too.

Riksu9000 (changes requested):
Please clang-format SettingSetTime.cpp

This is what it looks like now. I'm wondering what would be the best place for the ampm label, though the position might slightly change soon.
current

eliwss0:
Thanks for the review! I will get started on addressing the issues.

src/displayapp/screens/settings/SettingSetTime.cpp
```diff
@@ -16,23 +18,31 @@ namespace {
...
     screen->HandleButtonPress(obj, event);
   }
 }
 
-SettingSetTime::SettingSetTime(Pinetime::Applications::DisplayApp *app, Pinetime::Controllers::DateTime &dateTimeController) :
-  Screen(app),
-  dateTimeController {dateTimeController} {
-  lv_obj_t * title = lv_label_create(lv_scr_act(), nullptr);
+struct Time12H {
+  std::string ampm;
```
Itai-Nelken, line 28: As only string literals are ever assigned to this member, wouldn't it be more memory efficient to use `const char *`?

src/displayapp/screens/settings/SettingSetTime.cpp
```diff
@@ -105,38 +121,90 @@ SettingSetTime::SettingSetTime(Pinetime::Applications::DisplayApp *app, Pinetime
...
+  is24H = (settingsController.GetClockType() == Controllers::Settings::ClockType::H24);
+
   if (object == btnHoursPlus) {
     hoursValue++;
-    if (hoursValue > 23)
+    if (hoursValue > 23) {
       hoursValue = 0;
-    lv_label_set_text_fmt(lblHours, "%02d", hoursValue);
+    }
+    struct Time12H convertedTime = timeConvert(hoursValue);
```
Riksu9000, line 175: I don't think returning a struct was an improvement.

src/displayapp/screens/settings/SettingSetTime.cpp
```diff
@@ -105,38 +121,90 @@ SettingSetTime::SettingSetTime(Pinetime::Applications::DisplayApp *app, Pinetime
...
+    if (!is24H) {
+      lv_label_set_text_static(lblampm, convertedTime.ampm.c_str());
+      lv_label_set_text_fmt(lblHours, "%02d", convertedTime.hours);
+    } else {
+      lv_label_set_text_fmt(lblHours, "%02d", hoursValue);
+    }
```
Riksu9000, line 181: This block of code is still duplicated above and below. +125..+133 does pretty much the same thing as this block of code.

src/displayapp/screens/settings/SettingSetTime.cpp
```diff
@@ -105,38 +121,90 @@ SettingSetTime::SettingSetTime(Pinetime::Applications::DisplayApp *app, Pinetime
...
+  }
+  return time12H;
+}
+
+void SettingSetTime::HandleButtonPress(lv_obj_t* object, lv_event_t event) {
+  bool is24H;
   if (event != LV_EVENT_CLICKED)
     return;
 
+  is24H = (settingsController.GetClockType() == Controllers::Settings::ClockType::H24);
```
Riksu9000, line 168: This variable is still redundant. Do you have it so the code looks neater?

eliwss0: Would it be better practice to check `settingsController.GetClockType()` in place of is24H? Or is there a variable that contains this info already that I am missing?

src/displayapp/screens/settings/SettingSetTime.cpp
eliwss0: I had originally replaced lines 125-133 with`lv_label_set_text_static(lblampm, convertedTime.ampm.c_str());`, however with this, the ampm label was not populated until an hour button was pressed. The only resolution I found was to use an if statement instead. I will continue working on a better fix.

src/displayapp/screens/settings/SettingSetTime.cpp
eliwss0: I moved towards using a struct so I could create one function that returned both the hours value, and whether it was AM or PM in one function instead of nested if statements. Otherwise I would be creating two functions for relatively simple purposes. I am open to suggestions on how to handle this if I am going about it the wrong way.

src/displayapp/screens/settings/SettingSetTime.cpp
eliwss0: Issue is resolved. The if else is replaced by `lv_label_set_text_static(lblampm, convertedTime.ampm)` and will be included in the next commit.

src/displayapp/screens/settings/SettingSetTime.cpp
Riksu9000: I would prefer that for now.

src/displayapp/screens/settings/SettingSetTime.cpp
Riksu9000: I meant to suggest moving the LVGL functions into the function as well. This way you don't need to return anything and the blocks of code can be replaced by a single function. The function would be called something like `SetHourLabels()`.

src/displayapp/screens/settings/SettingSetTime.cpp
eliwss0: I see. I have made that change for the next commit.

src/displayapp/screens/settings/SettingSetTime.cpp
eliwss0: That is a much more elegant solution. I will implement that for the next commit.

Riksu9000 (approved):
I made a couple of comments, but it might not require changes.

src/displayapp/screens/settings/SettingSetTime.cpp
```diff
@@ -105,38 +115,75 @@ SettingSetTime::SettingSetTime(Pinetime::Applications::DisplayApp *app, Pinetime
...
-    if (hoursValue > 23)
+    if (hoursValue > 23) {
       hoursValue = 0;
-    lv_label_set_text_fmt(lblHours, "%02d", hoursValue);
+    }
+    if (settingsController.GetClockType() == Controllers::Settings::ClockType::H12) {
+      set12HourLabels(hoursValue, lblampm, lblHours);
+    } else {
+      lv_label_set_text_fmt(lblHours, "%02d", hoursValue);
+    }
```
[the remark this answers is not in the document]
eliwss0: I originally intended to, but would need to pass more parameters. I will explore it as a possibility making the function a private class member though.

src/displayapp/screens/settings/SettingSetTime.cpp
```diff
@@ -105,38 +115,75 @@ SettingSetTime::SettingSetTime(Pinetime::Applications::DisplayApp *app, Pinetime
...
 
 SettingSetTime::~SettingSetTime() {
   lv_obj_clean(lv_scr_act());
 }
 
-void SettingSetTime::HandleButtonPress(lv_obj_t *object, lv_event_t event) {
+void set12HourLabels(int time24H, lv_obj_t* lblampm, lv_obj_t* lblHours) {
+  switch (time24H) {
+    case 0:
+      lv_label_set_text_fmt(lblHours, "%02d", 12);
```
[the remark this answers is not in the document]
eliwss0: Changed in the new commit

src/displayapp/screens/settings/SettingSetTime.cpp
```diff
@@ -16,23 +17,26 @@ namespace {
...
     screen->HandleButtonPress(obj, event);
   }
 }
 
-SettingSetTime::SettingSetTime(Pinetime::Applications::DisplayApp *app, Pinetime::Controllers::DateTime &dateTimeController) :
-  Screen(app),
-  dateTimeController {dateTimeController} {
-  lv_obj_t * title = lv_label_create(lv_scr_act(), nullptr);
+void set12HourLabels(int time24H, lv_obj_t* lblampm, lv_obj_t* lblHours);
+
```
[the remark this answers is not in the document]
eliwss0: That's a good point. I've gone ahead and done that in the new commit.

src/displayapp/screens/settings/SettingSetTime.h
```diff
@@ -32,6 +32,31 @@ namespace Pinetime {
...
+                break;
+              case 13 ... 23:
+                lv_label_set_text_fmt(lblHours, "%02d", time24H - 12);
+                lv_label_set_text_static(lblampm, "PM");
+                break;
+            } else {
+              lv_label_set_text_fmt(lblHours, "%02d", hoursValue);
+            }
+          }
+        }
```
Riksu9000, line 59: Code shouldn't exist in the header file. Only put the declaration `void SetHourLabels();` here, and the following in the cpp file
```
void SettingSetTime::SetHourLabels() {
  // code here
}
```
The first letter is generally uppercase for member functions, and I think the time24H parameter is unnecessary too as it's always just `hoursValue`.

Riksu9000 (changes requested):

src/displayapp/screens/settings/SettingSetTime.h
eliwss0: Corrected

Riksu9000 (approved):

JF002:
Thanks for the PR and for the review, @eliwss0 & @Riksu9000 !

PR #911, opened 2022-01-04
Reduce RAM overhead and free ~9kB of memory

JF002:
Using `SystemMonitor`, I checked the memory usage of the global heap, the FreeRTOS heap, the LVGL heap and the stack of the task.

Here are the results on develop (e0013e7):
```
           FLASH:      412136 B       480 KB     83.85%
             RAM:       57520 B        64 KB     87.77%
```

```
Free heap : 624
<info> app: Task [IDL] - 43
<info> app: Task [MAI] - 94
<info> app: Task [dis] - 308
<info> app: Task [LOG] - 71
<info> app: Task [Tmr] - 209
<info> app: Task [Hea] - 411
<info> app: Task [ble] - 491
<info> app: Task [ll] - 246
<info> app: #LVGL Memory#
 used# 6504 (46%) / free 7832
 max used# 6820
 frag# 1%
 free# 7820
<info> app: heap : 2740
```

 * **Free heap** (first line) is the memory available on the FreeRTOS heap
 * **Task[xxx]** shows the minimum space available in the stack of the task (watermacked value).
 * **LVGL max used** is also a watermacked value of the maximum memory used by LVGL
 * **heap** (last line) is the space available in the global heap (malloc).

This PR reduces the following buffers:
 * Global heap : from 4kB to 3kB (-1024B)
 * FreeRTOS heap : from 17kB to 14kB (-3072B)
 * LVGL heap : from 14kB to 8kB (-6144B)

And the following stacks:
 * Timer task stack : from 1200B to 1000B (200B)
 * Display task stack : from 3200B to 2400B (-800B)
 * BLE ll task stack : from 1280B to 560B (-720B)
 * BLE host task stack : from 2880B to 1320B (-1560B)
 * System task stack : from 1400B to 1200B (-200B)
 
In total, this PR frees 10507B of RAM memory, which is quite nice! This memory has always been available, but it was over-allocated in those buffers. Reducing the size of the buffers according to the needs of the current code allows us to have a better overview of the memory currently available.

```
           FLASH:      412056 B       480 KB     83.83%
             RAM:       48304 B        64 KB     73.71%
```

```
Free heap : 1032
<info> app: Task [MAI] - 44
<info> app: Task [IDL] - 43
<info> app: Task [Hea] - 425
<info> app: Task [LOG] - 71
<info> app: Task [Tmr] - 159
<info> app: Task [dis] - 111
<info> app: Task [ll] - 69
<info> app: Task [ble] - 103
<info> app: #LVGL Memory#
 used# 6484 (80%) / free 1708
 max used# 6800
 frag# 1%
 free# 1696
<info> app: heap : 2740
```

## How to test this PR ?
You can build this branch with RTT logging enabled to check the logs from the `System Monitor`.

Most of those info are also available in the SystemMonitor app in InfiniTime.
What should you check?
 - Free heap should not decrease under 0
 - Task stack should not decrease under 0
 - LVGL max used should no overflow 8192
 - Global heap should not overflow 3072

Example on my sealed PineTime running for ~10hours:

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -809,7 +809,7 @@ add_definitions(-DOS_CPUTIME_FREQ)
 add_definitions(-DNRF52 -DNRF52832 -DNRF52832_XXAA -DNRF52_PAN_74 -DNRF52_PAN_64 -DNRF52_PAN_12 -DNRF52_PAN_58 -DNRF52_PAN_54 -DNRF52_PAN_31 -DNRF52_PAN_51 -DNRF52_PAN_36 -DNRF52_PAN_15 -DNRF52_PAN_20 -DNRF52_PAN_55 -DBOARD_PCA10040)
 add_definitions(-DFREERTOS)
 add_definitions(-D__STACK_SIZE=1024)
-add_definitions(-D__HEAP_SIZE=4096)
+add_definitions(-D__HEAP_SIZE=3072)
 
 # NOTE : Add the following defines to enable debug mode of the NRF SDK:
 #add_definitions(-DDEBUG)
--- a/src/FreeRTOSConfig.h
+++ b/src/FreeRTOSConfig.h
@@ -62,7 +62,7 @@
 #define configTICK_RATE_HZ                      1024
 #define configMAX_PRIORITIES                    (3)
 #define configMINIMAL_STACK_SIZE                (120)
-#define configTOTAL_HEAP_SIZE                   (1024 * 17)
+#define configTOTAL_HEAP_SIZE                   (1024 * 14)
 #define configMAX_TASK_NAME_LEN                 (4)
 #define configUSE_16_BIT_TICKS                  0
 #define configIDLE_SHOULD_YIELD                 1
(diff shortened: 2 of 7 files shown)


JF002:
This version has just crashed on my PineTime just after secure pairing with my phone. Maybe a buffer is a bit too tight? I'll debug this asap!

lman0:
@JF002 i don't see the heap for nimble, does it need as well to be setted?

evergreen22:
Don't forget that I already adjusted the BLE and LL stack sizes in #796 (1.8.0).

kieranc:
> This version has just crashed on my PineTime just after secure pairing with my phone. Maybe a buffer is a bit too tight? I'll debug this asap!

I just found the same thing.

kieranc:
I just tried building this with the BLE/LL stack sizes as before (600/200) and the watch resets during boot before getting to the watchface.
I had a second try setting the BLE/LL back to 210/20 and FreeRTOS heap to 15 rather than 14, the watch boots correctly but secure pairing still causes a reboot.
Third try with FreeRTOS 15, BLE/LL 600/200 still doesn't boot.
As ever I don't really know what I'm doing but maybe this info will be useful :shrug:

JF002:
Thanks for those tests :) I haven't had the opportunity to analyze those issues deeper, but I have probably overlooked something... As @evergreen22 said, the stack of BLE tasks were increased for secure pairing, and I reduce them in this PR, which is probably wrong...

kieranc:
I figured the same, but putting them back to their previous values makes it not boot. I assumed some other 'container' in which they reside needed to be increased to make room, I figured it'd be the FreeRTOS heap but apparently not? I'll keep fiddling!
Also interesting is it's only new secure pairings which cause a reboot, it will reconnect to an existing bond with this PR no problem.

trman:
any news @JF002  , is this still crashing ?

JF002:
I haven't taken the time to check out these changes yet. I think I was a little too aggressive with the stack size reduction.

PR #915, opened 2022-01-06
PTS: add seconds display

ialokim:
After receiving a PineTime for Christmas, I started trying to code a bit for it. The first thing I noticed was that the very nice PineTimeStyle watch face does not show the seconds, so I changed that. The design is adapted from the [original Pebble TimeStyle](https://apps.rebble.io/en_US/application/55a5c024f4510f794c000071), where the seconds are also displayed in the bottom right corner. I therefore had to move the step counter up and to avoid coming to near to the calendar, I also changed the step counter to be a half circle instead of a full circle.

I am using the very same changes (but applied to master) since more than a week now on my PineTime and it works perfectly fine, but since I only have a sealed version, I didn't dare to flash the develop branch with the changes. However, I now also simulated it using the awesome work of @NeroBurner in https://github.com/NeroBurner/InfiniTime/tree/lv_simulation_lvgl7 on the current develop branch and everything seems to work here as well.

This is how it looks like:

https://user-images.githubusercontent.com/18088991/148376673-56090395-fa88-4022-b68f-e6dc4a500719.mp4

Also mentioning @kieranc since he is the original author of the PineTimeStyle watchface. Would be really nice to have this included in the official InfiniTime builds!

### DIFF
--- a/src/displayapp/screens/WatchFacePineTimeStyle.cpp
+++ b/src/displayapp/screens/WatchFacePineTimeStyle.cpp
@@ -192,7 +192,7 @@ WatchFacePineTimeStyle::WatchFacePineTimeStyle(DisplayApp* app,
 
   // Display seconds
   timeDD3 = lv_label_create(lv_scr_act(), nullptr);
-  lv_obj_set_style_local_text_color(timeDD3, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(0x000000));
+  lv_obj_set_style_local_text_color(timeDD3, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_BLACK);
   lv_label_set_text_static(timeDD3, ":00");
   lv_obj_align(timeDD3, sidebar, LV_ALIGN_IN_BOTTOM_MID, 0, 0);
 
@@ -423,7 +423,6 @@ void WatchFacePineTimeStyle::Refresh() {
 
     if (displayedSecond != second) {
       displayedSecond = second;
-
       lv_label_set_text_fmt(timeDD3, ":%02d", second);
     }
 
--- a/src/displayapp/screens/WatchFacePineTimeStyle.cpp
+++ b/src/displayapp/screens/WatchFacePineTimeStyle.cpp
@@ -172,11 +172,10 @@ WatchFacePineTimeStyle::WatchFacePineTimeStyle(DisplayApp* app,
   }
   stepGauge = lv_gauge_create(lv_scr_act(), nullptr);
   lv_gauge_set_needle_count(stepGauge, 1, needle_colors);
-  lv_obj_set_size(stepGauge, 40, 40);
-  lv_obj_align(stepGauge, sidebar, LV_ALIGN_IN_BOTTOM_MID, 0, 0);
-  lv_gauge_set_scale(stepGauge, 360, 11, 0);
-  lv_gauge_set_angle_offset(stepGauge, 180);
-  lv_gauge_set_critical_value(stepGauge, 100);
+  lv_obj_set_size(stepGauge, 37, 37);
+  lv_obj_align(stepGauge, sidebar, LV_ALIGN_IN_BOTTOM_MID, 0, -10);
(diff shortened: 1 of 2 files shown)


kieranc:
Nice work, it looks good and the half circle gauge is pretty nifty.... However, this needs to be a configurable option, I don't think it should be the default for everyone. I'm working on getting some weather data into the watchface currently but at some point I will need to add additional settings for which items are displayed in the sidebar, a seconds display can certainly be one of them!

NeroBurner:
my simulator branch is useful :tada:

ialokim:
@kieranc Excited to hear about weather info being integrated into the watch face! In case you would like some help on that or on the settings screen, I'd be happy to provide it!

Riksu9000:
I quite like the look of this and it makes the gauge easier to understand. Can you fix the conflicts, so we can consider merging this @ialokim?

ialokim:
I struggled a bit with the docker build environment: building the docker file locally resulted in InfiniTime build failures until I used the latest version of the Dockerfile from `develop`. Afterwards, I was not able to build InfiniTime on current `develop`, as it fails with several `fatal error: date/date.h: No such file or directory` errors.

I did however apply the same changes to the last released version 1.10.0 where I was able to build InfiniTime. That version is running smoothly on my PineTime now.

NeroBurner:
missing `date/date.h` could be because of the introduction of the `date` library as a git submodule and may be fixed by running `git submodule update --init --recursive`

https://github.com/InfiniTimeOrg/InfiniTime/pull/1183

src/displayapp/screens/WatchFacePineTimeStyle.cpp
```diff
@@ -191,6 +190,12 @@ WatchFacePineTimeStyle::WatchFacePineTimeStyle(DisplayApp* app,
   lv_obj_set_style_local_line_width(stepGauge, LV_GAUGE_PART_NEEDLE, LV_STATE_DEFAULT, 3);
   lv_obj_set_style_local_pad_inner(stepGauge, LV_GAUGE_PART_NEEDLE, LV_STATE_DEFAULT, 4);
 
+  // Display seconds
+  timeDD3 = lv_label_create(lv_scr_act(), nullptr);
+  lv_obj_set_style_local_text_color(timeDD3, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(0x000000));
```
NeroBurner, line 195: suggest explicitly naming the used color instead of using the hex value
```suggestion
  lv_obj_set_style_local_text_color(timeDD3, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_BLACK);
```

src/displayapp/screens/WatchFacePineTimeStyle.cpp
```diff
@@ -415,6 +421,12 @@ void WatchFacePineTimeStyle::Refresh() {
       }
     }
 
+    if (displayedSecond != second) {
+      displayedSecond = second;
+
```
NeroBurner, line 426: optional, I don't think this newline is needed
```suggestion
```

ialokim:
Yes, that was it, obviously! Now I can build this branch as well locally. Thanks @NeroBurner!

NeroBurner (changes requested):

NeroBurner (approved):
LGTM

Riksu9000 (approved):

kieranc:
I'm still not keen on this being implemented as a default, I personally don't want it, I have no problem with it being implemented with a selector to choose between this and the current default (which I'm working on) but as a default replacement for the current gauge, I'm firmly opposed.

JF002:
This is probably a matter of taste and an artistic choice, but, as @kieranc is the original author of this watchface and continues to maintain it, I would not like to go against his opinion. 

@kieranc Keep us up to date when you have something to show us regarding the selector :)

kieranc:
InfiniSim_2022-08-16_212831

I've got this far, adding an options settings button in addition to the color settings, which so far allows toggling between full step gauge, half gauge and seconds or numerical step count. Todo, implement an additional persistent setting to save the selected option and pretty it up a bit. I'll keep going when I can!

Issue #779, opened 2021-10-24
Pokétch Apps


moriel5:
During my later childhood, I remember having played Pokémon Diamond version, and well, the in-game gadget there was the "Pokétch" (short for Pokémon Watch) that was what caused me to want a smartwatch (back then, even the term "smartphone" didn't exist, much less "smartwatch").

Now, obviously the Pokétch's interface is very bad for a smartwatch (one button to switch between apps (watchfaces, for those who prefer the term) in a linear fashion, with the only way to get to apps that are earlier in the list is to press the button enough times to get to the end, and then more times to start over, something that gets tiring when there are, I believe it got to at least 16), however the different apps you could get in the game (I would expect them to be detailed over at Bulbapedia) could give some inspirations for any developers wanting to build apps for InfiniTime (such as the calculator or calendar apps).

TechnologyClassroom:
- [X] Digital Watch
- [ ] Calculator
- [ ] Memo Pad
- [X] Pedometer
- [ ] Pokémon List
- [ ] Friendship Checker
- [ ] Dowsing Machine
- [ ] Berry Searcher
- [ ] Day-Care Checker
- [ ] Pokémon History
- [ ] Counter
- [X] Analog Watch
- [ ] Marking Map
- [ ] Link Searcher
- [ ] Coin Toss
- [ ] Move Tester
- [ ] Calendar
- [ ] Dot Artist
- [ ] Roulette
- [ ] Trainer Counter
- [X] Kitchen Timer
- [X] Color Changer
- [ ] Matchup Checker
- [X] Stopwatch
- [X] Alarm Clock

https://bulbapedia.bulbagarden.net/wiki/Poketch

dyamon:
I'm weirdly intrigued by a friendship checker!
*checks watch* Cool, I'm still friends with my buddy Joe!

PR #919, opened 2022-01-07
Create a coin flip app

ItzSwirlz:
Hello, I'm new to the project, and I want to get involved. I was going to make a Snake game, but as a start I decided to start creating a coin flip app. (Idea in #779)

This is just a base app and doesn't do anything yet.Since I'm new feel free to critique me and leave comments!

### DIFF
--- a/src/displayapp/screens/CoinFlip.cpp
+++ b/src/displayapp/screens/CoinFlip.cpp
@@ -1,0 +1,44 @@
+#include "displayapp/screens/CoinFlip.h"
+#include "displayapp/DisplayApp.h"
+
+#include <cstdlib> // for rand()
+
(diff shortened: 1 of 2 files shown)


medeyko:
Thanks! However, I think it would be better to create an app supporting arbitrary sized dices. It would be useful for a lot of games.

ItzSwirlz:
maybe eventually have a section for all the randomized stuff/apps later down the road?

Riksu9000:
There's a dice rolling app PR which also has d2 #565

ItzSwirlz:
If there is a dice app it should just become a general Randomizer app and have pages (swipe right to a different section, like on the main screen left to the settings)

PR #926, opened 2022-01-10
Added repeat option to timer

### LINKED ISSUE #918: Repeat Option for Timer
### Verification

### Pitch us your idea!

My cross country/track team uses watches that vibrate automatically every like 17-30s (depending on your pace group) to keep you on pace per 100m. This would be very useful for me and likely other runners.

### Description

Basically, set the timer to what you want the repeat interval to be, then, with "repeat mode" enabled, whenever the timer hits 0, the watch immediately vibrates and starts the timer again at the same number. Another button next to start/stop button in the normal timer app to toggle "repeat mode" would be an easy way to implement this. However, for my use case, the touch button to start/stop the button is very inconvenient. If a "tap" of the hardware button was able to start/stop it that would be very nice (exiting would then just be a hold) but that should probably be part of a separate issue.

clay53:
Closes #918 
I moved the play/pause button to the left and put a "Repeat" button next to it which just toggles `repeating` when it's pressed. When repeating is disabled, it follows the default theme for buttons and the timer behaves exactly as it did before but when repeating is enabled, it turns green and when the timer ends, the timer is started again at whatever duration it was started at last. I thought about modifying the TimerController to get repeating working by changing the timer mode to `APP_TIMER_MODE_REPEATED` but I has trouble getting it working. So currently, the timer app itself restarts the timer when `repeating` is enabled. I made it so that repeating can be toggled even when the timer is enabled since having the button not work but still visible as if it did when the timer was running seemed weird to me. I also didn't want to make it not visible because then it would seem the play/pause button was off to the side for no reason when the timer was running and I didn't want the play/pause button to move when the timer started. I also didn't have another good place to put the repeat button.

I tested this on a sealed device so I don't have any debugging info for it and I haven't done C++ in years but as far as I can tell, everything is well so this can be merged once reviewed.

https://user-images.githubusercontent.com/16981283/148719309-8b7bca0b-d34f-4b71-9bec-4ade939901fc.mp4

### DIFF
--- a/src/displayapp/screens/Timer.cpp
+++ b/src/displayapp/screens/Timer.cpp
@@ -81,7 +81,7 @@ Timer::Timer(DisplayApp* app, Controllers::TimerController& timerController)
   lv_obj_set_height(btnRepeat, 40);
   txtRepeat = lv_label_create(btnRepeat, nullptr);
   lv_label_set_text(txtRepeat, "Repeat");
-  // Always starts off not repeating so color is already what it should be.
+  lv_btn_set_checkable(btnRepeat, true);
 
   taskRefresh = lv_task_create(RefreshTaskCallback, LV_DISP_DEF_REFR_PERIOD, LV_TASK_PRIO_MID, this);
 }
@@ -122,13 +122,6 @@ void Timer::OnButtonEvent(lv_obj_t* obj, lv_event_t event) {
         lv_obj_del(btnMinutesUp);
         btnMinutesUp = nullptr;
       }
-    } else if (obj == btnRepeat) {
-        repeating = !repeating;
-        if (repeating) {
-          lv_obj_set_style_local_bg_color(btnRepeat, LV_BTNMATRIX_PART_BG, LV_STATE_DEFAULT, LV_COLOR_GREEN);
-        } else {
-          lv_theme_apply(btnRepeat, LV_THEME_BTN);
-        }
     } else {
       if (!timerController.IsRunning()) {
         if (obj == btnMinutesUp) {
@@ -169,7 +162,7 @@ void Timer::OnButtonEvent(lv_obj_t* obj, lv_event_t event) {
 }
 
 void Timer::setDone() {
-  if (repeating) {
+  if (lv_obj_get_state(btnRepeat, LV_BTN_PART_MAIN) & LV_STATE_CHECKED) {
     timerController.StartTimer((secondsToSet + minutesToSet * 60) * 1000);
   } else {
     lv_label_set_text(time, "00:00");
--- a/src/displayapp/screens/Timer.cpp
+++ b/src/displayapp/screens/Timer.cpp
@@ -64,7 +64,7 @@ Timer::Timer(DisplayApp* app, Controllers::TimerController& timerController)
   btnPlayPause = lv_btn_create(lv_scr_act(), nullptr);
   btnPlayPause->user_data = this;
   lv_obj_set_event_cb(btnPlayPause, btnEventHandler);
-  lv_obj_align(btnPlayPause, lv_scr_act(), LV_ALIGN_IN_BOTTOM_MID, 0, -10);
+  lv_obj_align(btnPlayPause, lv_scr_act(), LV_ALIGN_IN_BOTTOM_LEFT, 0, -10);
   lv_obj_set_height(btnPlayPause, 40);
   txtPlayPause = lv_label_create(btnPlayPause, nullptr);
   if (timerController.IsRunning()) {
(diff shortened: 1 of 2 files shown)


src/displayapp/screens/Timer.cpp
```diff
@@ -74,6 +74,15 @@ Timer::Timer(DisplayApp* app, Controllers::TimerController& timerController)
...
 
+  btnRepeat = lv_btn_create(lv_scr_act(), nullptr);
+  btnRepeat->user_data = this;
+  lv_obj_set_event_cb(btnRepeat, btnEventHandler);
+  lv_obj_align(btnRepeat, lv_scr_act(), LV_ALIGN_IN_BOTTOM_RIGHT, 0, -10);
+  lv_obj_set_height(btnRepeat, 40);
+  txtRepeat = lv_label_create(btnRepeat, nullptr);
+  lv_label_set_text(txtRepeat, "Repeat");
+  // Always starts off not repeating so color is already what it should be.
+
```
Riksu9000, line 85: It would make sense to make this button toggleable/checkable in LVGL with `lv_btn_set_checkable(btnRepeat, true);`. You can take a look at QuickSettings for reference. Then you won't need the `repeating` variable.

https://github.com/InfiniTimeOrg/InfiniTime/blob/26ae828e393d28390c5ea718bec9d4c1a96435ac/src/displayapp/screens/settings/QuickSettings.cpp#L82

src/displayapp/screens/Timer.cpp
```diff
@@ -113,6 +122,13 @@ void Timer::OnButtonEvent(lv_obj_t* obj, lv_event_t event) {
         lv_obj_del(btnMinutesUp);
         btnMinutesUp = nullptr;
       }
+    } else if (obj == btnRepeat) {
+        repeating = !repeating;
+        if (repeating) {
+          lv_obj_set_style_local_bg_color(btnRepeat, LV_BTNMATRIX_PART_BG, LV_STATE_DEFAULT, LV_COLOR_GREEN);
+        } else {
+          lv_theme_apply(btnRepeat, LV_THEME_BTN);
+        }
```
Riksu9000, line 131: All of this will be handled automatically when the button is checkable.

Riksu9000:
@clay53 Hi. Can you take a look at my review?

clay53:
> @clay53 Hi. Can you take a look at my review?

Sorry, I've been very busy. I'll try for this Saturday.

clay53:
@Riksu9000 Did what you said, just wanted to note that (for both iterations) I noticed that the repeat option would be lost if the app was exited. Do you think it should be persistent (like is that worth it or even desired)?

Riksu9000:
Because the settings doesn't persist, if I exit the app while a repeating timer is running, it will notify once and won't repeat anymore. I also think that for the use case you suggested in #918, the watch shouldn't wake up when repeating is on. Thus I think this option should be added in TimerController, so in SystemTask or DisplayApp it can be made to not wake up when repeating is enabled. https://github.com/InfiniTimeOrg/InfiniTime/blob/4f649a85448fc79b55202d9db0e8f4c8828975e3/src/systemtask/SystemTask.cpp#L313

We also need to figure out if the repeating timer should keep running after the app is closed. If it doesn't, someone might accidentally close the timer they meant to set, but if it does keep running, should the timer app open each time the timer ends or not?

A repeating timer seems like a completely separate use case from a normal timer, so maybe it could be a separate app? Maybe the Metronome app could be used for this actually.

PetersOtto:
That would be a great feature for interval training. How is the status here, is it possible that this feature is part of the next versions of PineTime?

clay53:
> That would be a great feature for interval training. How is the status here, is it possible that this feature is part of the next versions of PineTime?

The current version works without any trouble on my fork. You can clone it from there and figure out how to build it. Only thing I had trouble with after following the instructions on the GitHub was finding the actual build script. I created my own script to overlay everything that had to be done for it (you'll need Linux, pyenv, make, and adafruit-nrfutil from pip for my script - other deps in instructions and might be duplicated here):
```bash
source .venv/bin/activate
cd build
make -j pinetime-mcuboot-app
cd ..
python ./tools/mcuboot/imgtool.py create --align 4 --version 1.8.0 --header-size 32 --slot-size 475136 --pad-header ./build/src/pinetime-mcuboot-app-1.8.0.bin image.bin
adafruit-nrfutil dfu genpkg --dev-type 0x0052 --application image.bin ~/Downloads/DELETE/dfu.zip
```
change the `~/Downloads/DELETE/dfu.zip` if you want it to put the package somewhere else. You can then take that package and flash your watch with it. I just copy it over to my Android phone and flash it with Gadgetbridge. Probably a better system but that's what I've been doing.

I imagine it'll be included in future versions.

Riksu9000:
I do think the repeating timer should stop when closing the app, but my main concern is that someone makes a 10 min timer, closes the app and forgets about it, but they accidentally left repeat mode on, so it would be stopped. If use cases of an independently repeating timer only require at most 60 seconds for example, we could disable the minutes when repeating is enabled. Then the user would have to disable the repeating to set the timer to more than a minute. What do you think?

@clay53 Did you enable `-DBUILD_DFU=1` when generating the project with cmake? Github also builds these PRs as checks and the DFUs can be grabbed from there.

PetersOtto:
Yesterday and today I test the repeat-function at the training. It is a great feature. There is no need to have an eye on the clock and it is possible to here music during the training because the time use the vibration-alarm.

clay53:
I just want to make it clear that at this point, I am not working on this feature so someome else is clear to work on it. I haven't been keeping up with the latest updates to InfiniTime (iirc there was a timer rework) since I've switched back to my Pebble Time for more reasons than just this feature so maybe something similar's already been added idk.

PR #938, opened 2022-01-16
Added 12 hour alarm interface

eliwss0:
Added 12 hour time interface to the alarm app, similar to my set time 12 hour [PR](https://github.com/InfiniTimeOrg/InfiniTime/pull/907).


src/displayapp/screens/Alarm.cpp
```diff
@@ -180,8 +189,34 @@ bool Alarm::OnButtonPushed() {
...
+        lv_label_set_text_fmt(time, "%02d:%02d", alarmHours - 12, alarmMinutes);
+        alarmController.SetAlarmTime(alarmHours, alarmMinutes);
+        break;
+    }
+  } else {
+    lv_label_set_text_static(lblampm, "  ");
+    lv_label_set_text_fmt(time, "%02d:%02d", alarmHours, alarmMinutes);
+    alarmController.SetAlarmTime(alarmHours, alarmMinutes);
+  }
 }
```
Riksu9000, line 220: `alarmController.SetAlarmTime(alarmHours, alarmMinutes);` can always be called with the same parameters, so you only need to put this once at the end of this function

src/displayapp/screens/Alarm.cpp
```diff
@@ -180,8 +189,31 @@ bool Alarm::OnButtonPushed() {
...
+        lv_label_set_text_fmt(time, "%02d:%02d", alarmHours - 12, alarmMinutes);
+        break;
+    }
+    alarmController.SetAlarmTime(alarmHours, alarmMinutes);
+  } else {
+    lv_label_set_text_static(lblampm, "  ");
+    lv_label_set_text_fmt(time, "%02d:%02d", alarmHours, alarmMinutes);
+    alarmController.SetAlarmTime(alarmHours, alarmMinutes);
+  }
 }
```
Riksu9000, line 217: You can still move the `alarmController.SetAlarmTime` to right after the if-else, so it's only needed once

src/displayapp/screens/Alarm.cpp
```diff
@@ -180,8 +189,31 @@ bool Alarm::OnButtonPushed() {
...
+        lv_label_set_text_fmt(time, "%02d:%02d", 12, alarmMinutes);
+        break;
+      case 13 ... 23:
+        lv_label_set_text_static(lblampm, "PM");
+        lv_label_set_text_fmt(time, "%02d:%02d", alarmHours - 12, alarmMinutes);
+        break;
+    }
+    alarmController.SetAlarmTime(alarmHours, alarmMinutes);
+  } else {
+    lv_label_set_text_static(lblampm, "  ");
```
Riksu9000, line 213: This label doesn't need to be set when in 24H mode as it's always empty.

Riksu9000 (approved):

PR #941, opened 2022-01-16
Let TouchHandler return TouchEvents instead of driver specific enum

NeroBurner:
Let the TouchHandler::GestureGet() function return a TouchEvent instead
of the touchpanel-driver specific enum.

This helps to move the driver specific helper function `ConvertGesture`
from `DisplayApp` into `TouchHandler`.

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -60,28 +60,6 @@ namespace {
   static inline bool in_isr(void) {
     return (SCB->ICSR & SCB_ICSR_VECTACTIVE_Msk) != 0;
   }
-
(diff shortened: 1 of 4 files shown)


NeroBurner:
@Avamander @geekbozu could you review this PR. I've tested it on my dev-kit and everything works as expected (double tap to wake up, swipes, normal taps)

geekbozu (approved):
Ill trust your testing of it, software wise I'm happy with it. Seems like a better place for all of this.

Riksu9000 (approved):
The only thing I'm wondering about is if the conversion should be done in Cst816.cpp instead, but it's hard to predict how TouchHandler will evolve and what kinds of new drivers we'll need, so for now this change is fine as a small step in the right direction.

PR #946, opened 2022-01-18
Use Git Timestamp for build date and time

NeroBurner:
For a clean repository use the timestamp of the current commit as build
time instead of the current time. This helps with reproducible builds
(not actually checked if the build is reproducible).

Create a `BuildTime.cpp` file in the build directory. It's a cpp file for
the compiler to recompile just one file to get the most up to date
timestamp after a cmake reconfigure. Just the updated h file
doesn't trigger a recompile for `SystemInfo.cpp`.

For compatibility the `Pinetime::Version` class is changed to a namespace
so that the `BuildTime.h` can add additional functions like `GitCommitDate()`
and `GitCommitTime()`.

SystemInfo.h


NeroBurner:
With this PR we get much more information about the git repository. Should we do something with it? Maybe adding `-dev` to the version string for all non-tag builds? Add the current branch name? Some indication if the repo wasn't clean?

Riksu9000:
There seems to be some interesting info that can be extracted, but I'm not sure the timestamp is one of them. It's not the same as build time and it shouldn't be referred to as such at least.

I would like the branch name to be displayed. I would prefer that over the shortref, or just show both. I also like the idea of showing `-dev` for non tag builds.

NeroBurner:
My rationale of replacing the build time with the time of commit: it isn't really important when the firmware was built, but rather what is in the firmware. When was the change done for the firmware running on my device.

I'm having a hard time finding something descriptive to replace "build date". Here are some alternatives

- commit date
- latest change
- date
- nothing, just write the date and time
- from
- modification date

Riksu9000:
Well the build time is useful for checking that the last build booted and didn't revert during bootup. The timestamp makes this more complicated, although you do use the real time for non clean builds.

In what use case is the timestamp beneficial?. When I'm working on something, I want the real build time, and If I review something, the timestamp doesn't really tell me anything. The branch name would say a lot more about what's in the firmware build. For the release builds none of this really matters.

NeroBurner:
The timestamp tells you how old the state of your firmware is.

Riksu9000:
That doesn't really tell me anything about what changes are in the firmware though.

Riksu9000:
We could remove the build time and replace it with the branch name. Are you interested in updating this?

PR #951, opened 2022-01-22
Move steps goal setting to steps app

kieranc:
I'm not sure if this is the way we want to go, so I've done a basic prototype and am submitting this as a draft to get some feedback.
The idea is to reduce the number of settings items, and duplicate paths, so instead of the step goal being a settings item, it is accessed within the steps app via a button which enables other buttons to increase/decrease the step goal.

Potential improvements:
- plus/minus buttons can be made bigger for easier tapping
- I'd like the reset button to say 'reset trip' ideally
- make the physical button close the settings, as with PTS

image
image


Riksu9000:
Maybe just open the existing step setting screen when pressing the set goal button. Unlike PTS, there's no need to have everything on screen at once.

kieranc:
Leaving the settings entry in place, or removing it? Seem to me that if we're removing it, the functionality should be integrated with the Steps app, otherwise there's no point. If the concern is the appearance, I can totally hide the steps app while the goal setting is open.

Riksu9000:
Yeah, the settings entry can be removed. I just thought that we could make use of that screen as it already exists. The screens don't necessarily need to be combined into one screen/file. I think that apps can contain multiple screens.

PR #954, opened 2022-01-23
Music: initialize totalLength to have non-random data from the start

NeroBurner:
In the first `Music::Refresh` call the lastLength and totalLength are
not initialized. The `lastLength` value is read first from the
musicService. Naturally in the first itereation that is changed and
`UpdateLength()` is called. There the variable `totalLength` is used as
well, but that variable isn't initialed to a sensible value yet. This
leads to sometimes the `Music.h` screen segfaulting (depending on the
random uninitialized data in the `lastLength` variable)

### DIFF
--- a/src/displayapp/screens/Music.h
+++ b/src/displayapp/screens/Music.h
@@ -69,7 +69,7 @@ namespace Pinetime {
         std::string track;
 
         /** Total length in seconds */
-        int totalLength;
+        int totalLength = 0;
         /** Current length in seconds */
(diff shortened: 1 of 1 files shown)


Riksu9000 (approved):
I guess this happens if totalLength is a negative value. With the extra minus, it doesn't fit in the char array right?

NeroBurner:
yes, in my case it was a huge negative number, but that is random memory and can be any number in the integer range

PR #960, opened 2022-01-26
Add contact_links to issues

Riksu9000:
Add links to the PineTime community and PineTIme developer chatrooms to show up when trying to open a new issue. This will encourage users to first try and get help there.

We can also disable blank issues and force people to use the issue forms. People don't always use them and thus they may not provide all the necessary information. Should we disable them?

Screenshot from 2022-01-26 15-21-42

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,8 @@
+blank_issues_enabled: true
+contact_links:
+  - name: PineTime community chat (Matrix)
+    url: https://app.element.io/#/room/#pinetime:matrix.org
+    about: Please ask questions about PineTime here.
+  - name: PineTime developers chat (Matrix)
+    url: https://app.element.io/#/room/#pinetime-dev:matrix.org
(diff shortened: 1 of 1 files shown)


JF002:
I think it's a good idea to add those links. I didn't know it was possible to do that with Github!

Regarding the blank issue, I know some users use it to create issues that would fit into one of the templates, and we should encourage them to use the template next time. However, I don't think those templates will apply to all kind of message, questions and inquiries that users might have, so, in my opinion, I think we should keep it as well.

Maybe we need to add a few templates like "Developer questions" for example?

Riksu9000:
It's true that there are situations that don't fit in a bug report or a feature request. We could add a generic template that replaces the blank issue. Not sure if that would help in any way though. It's hard to predict what kinds of templates would be required, so a blank/generic issue should probably be kept. Anyway, this PR keeps the blank issues and only adds the links, so wouldn't this be worth merging by itself?

JF002:
Yes! I probably kept this PR open while waiting for feedback about additional template, but I think it's worth merging as it is right now. Thanks!

Issue #697, opened 2021-09-28
Unlock watch mechanism


ajack2001my:
I need everyone feedback on this. I constantly find that when I look at the watch, it is in one menu or another that was randomly registered while I was wearing the watch.

Can we have it such (suggestion only) where when we take or raise the wrist, it shows the watchface and any other touch input (while watchface is displayed before going to sleep) requires a "slide to unlock" before we can access the menus.

I am also open to any other suggestion. That said, it's only a discussion to gather ideas and hopefully, for a dev to take up as I don't code in "modern" languages... :)

aemkai:
I would support this request. I have also the problem, that sometimes there is another screen, often also the alarm is disabled or brightness has changed by unwanted inputs.
Instead of slide to unlock I could imagine a double tap maybe better suited, because a unwanted sliding gesture  seems to occur often.

The issue appeared with the changed touch behaviour in 1.4.0, with Version 1.3.0 and prior there were rarely unwanted inputs.

Riksu9000:
Duplicate of #395 and #618

the-voidl:
This in combination with #395 would be a great improvement for the every day use of the watch!
I often find the clock in some unwanted state (like already mentioned by others). Since I also wear my clock in the shower and while doing sports, water and capacitive touchscreens are not the best friends - so I would also support some lock/unlock mechanism.

Riksu9000:
Let's track this issue in #618

Issue #395, opened 2021-05-23
Raise Wrist to wake, but Double Tap to unlock


Soundtoxin:
It is useful to raise your wrist to see the time, but currently the setting is too sensitive and I keep doing random actions on my watch by accident throughout the day. If I could check the time at a glance, but still have to tap to unlock (allow actions), this would feel like a good combo.

I could see a couple ways of implementing this.
1) Raise Wrist displays the currently-open app whether it's the clock or not
2) Raise Wrist displays the time specifically, like a screensaver or Android lockscreen

After thinking about it a bit, I'm leaning toward the second option.

MaWalla:
I'm definitely agreeing on this (specifically option 2). My PineTime does lots of random things throughout the day, and when raising the wrist, I'm sometimes in some menu all of sudden.

This could be a great combination with #377 since long-press on the watchface could be used to unlock/manually lock the screen for example (which could further be indicated by a lock icon in the top left corner since we have plenty of space there, at least for the digital watchface), though I could also see the use of the physical button here in some way.

Furthermore I could also see settings for seperate brightness/screen-on time for the lockscreen, since there is a fair amount of "accidental" wakeups, especially with the raise writst to wakeup option. Being able to lower brightness and duration here would help with battery life a lot and is often sufficient for a quick glance.

Riksu9000:
Let's track this issue in #618.

PR #961, opened 2022-01-26
Rain mode feature

FintasticMan:
Adds a rain mode toggle to the quick settings, which disables the touch
panel.
Currently the symbols for rain mode and flashlight don't show up, and
the touch panel only updates after it is touched once.


Riksu9000:
Do we really need a rain mode? What inspired you to create this feature?

FintasticMan:
I've noticed that if I'm walking in the rain, and I check my watch, the raindrops make a bunch of touch events that take me off the clock screen. By disabling the touch panel I can check the time in the rain.

hatmajster:
I was once in the mountains and soaked completely after few hours. The watch was completely useless. This feature would be awesome if it would work nice

Hactarus:
+1. Usage:
Under shower to control music without water drops messing in all menu.
At pool to see time (can't wear corrective glasses/lens)

escoand:
Maybe duplicate of #894

FintasticMan:
I think that this is different from 894, because that one just disables waking the watch entirely, which means you can't use it to tell the time anymore. This PR just disables the touch panel, so you can still check the time.

Riksu9000:
We don't want to encourage people putting their PineTimes in water, as they're not actually all that waterproof. It's unlikely for these PRs to be merged.

FintasticMan:
The way I understood it, PineTimes are waterproof enough for rain, but not really for submerging or showering, because of the heat and soap. I'd be happy to accept if it's not even waterproof enough for rain, and then I'll just close this.

Riksu9000:
Maybe it's fine for a bit of rain, but if we provide a feature like this, people would misuse it anyway ☝

trman:
@Riksu9000  this feature is good not only for rain/shower but as well  when putting the pinetime in the your jeans' pocket !

i explain : when it's rain a lot or when i need to put the pinetime inside the jean's pocket 
, that allow me to have the steps tracked,
but 
in the jeans pocket there is also , other things like money / key / cards ..., that activate the screen unknow to me ,and drain the pinetime !
even more when i retreive the pinetime from the pocket and use it , most time than not it's already showing the screen !
if the shake to wake is activated then it will remain activated by every thing here at evey movement and evey touch...
(sometime , it even activated the music app , on bus...)



that's why i truly need this pr because regardless of rain or not , it will allow me to put the pinetime in the pocket and it will not be drained where the shake to wake i activated  or not , because the  screen will remain disabled until i push the button and activated the screen again!

what do you thinks of this purpose  and this pr @JF002  ?

Hactarus:
Some points I'll you known too.

# Dry feature
It's winter here, so long coats and sweaters, somehow the cloth activate menu and often my watch play pong by itself. 

# Water
First time I used watch underwater to test the  "PineTime comes with a IP67 rating, meaning that it can handle up to one meter of water and is fully dust proof", the menu come crazy and to watch time I need to push a lot the button (that break the seal and allow water to come in), and the watch break, put it in rice and dry it, work fine since.

Was a bad move but the website is unclear, maybe we could work on it instead of restrict users feature.

My point is to keep the seal and waterproof you need to keep the button unpushed underwater, so avoiding user to use it could help keeping it sealed, even on misuse.

Now I still use it underwater, and the feature will help but without that don't prevent not use it.

In shower I don't keep it on wrist, keep as dry as possible but still some droplets on it, like rain.

I wasn't clear on my last comment, but allowing disable touch feature could help for other usage than just try to prevent user to misuse they watch.

Thanks for reading have a great day.

hellojaccc:
Or maybe say just show a warning that pinetime is not really made for use in heavy rain, and this mode is meant to block unintended touches.

levipinkard:
It's useful regardless of rain IMO, there's been many times where I'm doing heavy activity, I go to check the time, and it's either on a different screen or it's changed a setting somehow (I've accidentally turned off vibration before, causing me to miss notifications).

If you view it as a lock/hold mode, it makes a lot of sense. Similar to how old MP3 players had a hold switch for when it was rumbling around in your pocket. It'd be useful too for the times that I have to throw my pinetime in a bag.

Would love to see this feature added.

Riksu9000:
The issue here is that this is a rain mode, not a locking mode. They may work very similarily, but the other references water, which is not good. People are suggesting use cases that don't have to do with rain, but in the end this feature is still presented as a rain mode. Check #395 and #697 for locking related issues. Some issues mentioned here probably relate to raise wrist being too sensitive #548.

trman:
> The issue here is that this is a rain mode, not a locking mode. They may work very similarily, but the other references water, which is not good. People are suggesting use cases that don't have to do with rain, but in the end this feature is still presented as a rain mode. Check #395 and #697 for locking related issues. Some issues mentioned here probably relate to raise wrist being too sensitive #548.

if i understand right @Riksu9000  , as long as the title is changed from "[Rain mode feature]" to for example "[lock screen mode feature] , you are willing to approve the feature ?
if it's only a title , maybe @FintasticMan  could change it (the pr title and if needed the title setting as well).

for the rain or water part , that's the "lock mode" could help with , if you have resentment about it send it to the pine64 team not on this pr , it's fruitless and misplaced :
because  , as long as  it's written that the product page of the pinetime that it can withstand rain or shower , then people will use on these case whether you like it or not ,because it's written that they could do it!

FintasticMan:
This pr isn't finished, I haven't had time to work on it.

I do understand that calling it rain mode is not a good idea, and I'd be more than happy to change it. I don't think that screen lock is a good term, as while that is what is doing, I think people understand screen lock to mean that it would need something like a PIN code to unlock. Maybe something as simple as just saying it disables touchscreen is good enough?

I do also think that this might not be an optimal solution for the problem. If someone has a better one, please mention it!

JF002:
There's a lot going on here, let me try to re-focus the discussion.

Regarding the IP67 rating and the product page : Pine64 sells the **hardware** and not the software. The hardware **is** certified with IP67 rating, and there is nothing to change on the product page.
I am totally not an expert in IP ratings, but I'm not sure that "protected against short periods of immersion in water while under pressure under 15cm and 1m" means that you can use and operate the watch while under 1m of water (or that it will work at all), or under jets of hot and soapy water from the shower, or under the pressure of moving water when you're swimming with your watch on your wrist.
If you have concerns or questions about the IP rating of the PineTime, I would suggest you ask your question directly to Pine64, as the software has nothing to do with that.

By reading the comments in this discussions, I notice that you all have various and specific use cases : use the watch under heavy rain, while walking in the mountains, control the music under the shower, put the watch in your jean pocket with a lot of other pieces and objects in the pocket, use the watch while doing heavy activities. All these use-cases are probably valid but... I'm not sure we'll be able to implement the specific functionalities to support all of them. We are talking about a 25$ watch that is running on very constrained resources and not about a high-end device specifically built to resist to extreme environments and usages. 

I do very little sport, I rarely have the opportunity to walk in the mountains under heavy rain and I always remove my watch when going to the pool and in the shower, so I might not fully understand your requests. But to me, it seems that the "lock screen" or  "rain mode" or "shower mode" feature is trying to work around other issues : the touch panel is too sensitive (and is triggered by a small drop of water) and the rise/shake wake algorithm are too sensitive and wake the watch too often.

So I think we should try to fix those issues before trying to implement various complex algorithms to work around them.

trman:
> the touch panel is too sensitive (and is triggered by a small drop of water)

>So I think we should try to fix those issues before trying to implement various complex algorithms to work around them.


I thinking that this pr resolve this issue **without**  complex algorithm  , so it's good for the use case of sensitivity issue
 with the simple way : disable the touchscreen

any other way would need to create complex algorithms to work around  the sensitivity ....
that you don't want @JF002 .

(example : @hubmartin  created a pr for screen level of brightness but it was never merged , even thought it was final, because it seem it was too complex)

FintasticMan:
I believe that while the touchscreen in general might need some tuning to avoid ghost touches, I don't think it's likely or even possible to tune it in such a way that it doesn't register rain as touch inputs. I made this PR as it's an issue I face, and this solution is one I've seen used in other smartwatches ([see this article from Samsung](https://www.samsung.com/us/support/answer/ANS00078058/)).

hellojaccc:
https://support.apple.com/en-us/HT210931 Same feature is also avaiable in apple watch, i agree with comment above.

trman:
@hellojaccc and @FintasticMan   yours example are not good because , the pinetime is not able to go inside 50 m of water
 unlike apple and samsung watch

trman:
by the way , @Riksu9000  , i have the banglej2 , that have the same ip and form factor than the pinetime
and there is a auto lock /disable touch screen functionality  (that is disabled with long press of button)

shvechkov:
@Riksu9000 .. When exercising long (sweaty) sleeves touch the screen and mess all the settings on my PT ... Locking screen is MUST HAVE feature for folks who wear PT in the gym ...

Riksu9000:
There are multiple ways to tackle an issue. The most obvious ways may not be the best ways. It has already been mentioned here that we would rather tune the touchscreen first to reduce this issue. Wouldn't it be even better if you didn't even need to lock the device, and it just worked? #1011

trman:
> There are multiple ways to tackle an issue. The most obvious ways may not be the best ways. It has already been mentioned here that we would rather tune the touchscreen first to reduce this issue. Wouldn't it be even better if you didn't even need to lock the device, and it just worked? #1011
> 
> (We're not saying deal with it, we're saying we may be able to do even better.)

@Riksu9000 sometime a good intention aka in your case tuning the touch for all case (https://github.com/InfiniTimeOrg/InfiniTime/issues/1011) is not the proper way because 
instead a simply disabling a the touch screen  (but not the screen !) that by it's very nature is causing problem when touched , regardless of type of touch (a sleeve , a a key  or whatever **can ** do **long/little** touch that you can't predict!)

you want to create an all purpose functionality ,  a labyrinth,  the fine tuning task that would 'magicaly' adresse all case !
**but** 
-first it will never be able to resolve the original problem (because it's  unpredictable)
-second it will create by it very nature , unlimited of problem that would make it unable to be usable
- trying to figure all case , all the time (so even when screen off!) will drain the pinetime way faster, because it's not a smartphone,  making the purpose moot !!!

i would want to have @JF002 , or @Avamander  or other to give a way because i think it's not resolvable but with a feature to disable the touchscreen

shvechkov:
disabling touchscreen would be an overkill - instead just disable navigating from main screen to various submenus/settings and display lock icon in the center (optional) ...    This is typical behavior that I observed in other products that I owned  in the past.

(In my case when I accidentally touch my PT multiple times w/o noticing I end up navigating into various menus and changing settings ... )

trman:
> disabling touchscreen would be an overkill - instead just disable navigating from main screen to various submenus/settings and display lock icon in the center (optional) ... This is typical behavior that I observed in other products that I owned in the past.
> 
> (In my case when I accidentally touch my PT multiple times w/o noticing I end up navigating into various menus and changing settings ... )

@shvechkov  which product ?

shvechkov:
@trman  garmin watches (e.g. vivoactive 3) , MSFT band (rip) ... 
Here is video: https://www.youtube.com/watch?v=Dpk2vM7KtZA&ab_channel=MrPerfect

trman:
@shvechkov  i think  you understand , but  the  first  product is  watch with far more processing power than pinetime 
and more battery and because of these functionality they last at best 3 day ....

i repeat , a functionality that  drain the pinetime to make it last 3 day  , because of its processing power needed make the purpose moot...

the second is not a screen with lock wthat would not resolve drain  the problem , because , the pinetime would show the pin screen every time  , and thus draining the battery as well....

JF002:
My position about this issue is still the same : I do agree that false positive and ghost inputs from the touch sensor are annoying and that we should find a solution to that. I still think that the first step to solve this issue is to try to filter the inputs as much as possible at the source (ie at the touch panel level). We will then be able to decide if further changes are needed based on the results of this first step.

@FintasticMan Thank you very much for sharing your code with us! You noticed an issue in InfiniTime and tried to find a fix for it. This fix is perfectly fine and functional which is great. However, as I said, I would like to explore other ways to fix the original issue before applying your changes.

@trman In case you haven't noticed yet, @Riksu9000 is part of the InfiniTime core-developers team. I invited them to join the team because they do really good contributions to the projects and because I trust them to take the right decisions for the project. Please do not @mention me each time someone writes something you do not agree with. Everyone is free to comment and to express their opinion, **and I'll react if I want to and when I want to.** 
Now, I think you have expressed your opinion quite clearly here : you know exactly what is the best solution for this issue. I think your next move should be to create the pull-request, document it to prove that this is the best solution and wait for someone to review it. In the meantime, please **stop** harassing other contributors that try to express their ideas and to provide another opinion or solution!

4bitFox:
Hi,

I recently bought the PineTime and had issues with wrong inputs aswell. I'm in the Military and move a lot during the whole day. Here is my take on this topic:

I think there really should be an option to disable the touch screen when woken up by eg. shake or raise.
I have raise to wake on an it false triggers ALOT (which would be fine by itself because of tge 5s time)! I then accidentally enable the heart rate monitor, change the watch face, enable the flashlight, ect... This can use a lot of power and is inconvenient. A feature to only enable touch after raise or shake to wake by pressing the button would be a possible solution. You have to bring up your other hand to touch the screen anyways. You don't need touch if you just want to see the time by raising / shaking. (In this case, raise to wake would need to always show the watchface, to make it actually convenient / useable.)

minacode:
I actually read through all this and we all seem to have weird use cases, BUT:

_"Show the time without reacting to touch until I press the button"_

sounds like a solid and reasonable feature to me.

Totally independent of problems with touch input.

FintasticMan:
I've just updated this to the latest develop, and fixed the icon issues. There are still some issues where even though I'm almost certain the touch panel is asleep, it sometimes does still react to touch events. Because of this I'll probably switch to just not reacting to touch events even though the panel does recognise them.

The user interface is pretty terrible at the moment, it's very unintuitive. It's not obvious how to turn the panel back on or even that the panel is disabled. I'm going to have a look at how other watches I have handle it for inspiration.

As for whether this is the correct solution to the problem, I think it's a pretty reasonable one. Like @JF002 said, there should be some kind of filtering of inputs that don't make sense at a lower level. However I think that things like raindrops can be almost indistinguishable from fingers, and so I think that this is quite a good solution to that.

While we definitely shouldn't be picking features just because it is done that way by others, I think the fact that other watches have similar features to disable touch input is quite a good indicator that it is a reasonable fix.

I think that the "Show the time without reacting to touch until I press the button" idea is quite a good one as well, but it seems like quite a big difference in default experience. A good way to combine retaining the old experience and implementing that would be to only do it after the screen has been off for a certain amount of time, 5 minutes for example.

lman0:
@FintasticMan  this feature is on by default on bangle js

escoand:
I'd rather have a more generic lock icon and also an appropriate name that isn't explicitly associated with "water".

FintasticMan:
I can change the icon to whatever, I just copied my other watch because then I don't have to think about the specifics lol. I'm more interested right now in getting the functionality working properly, because for some reason my touch panel is reacting to some inputs even when it's asleep.

FintasticMan:
I've just finished some commits that fix the issue I was having where the touch panel would still respond to touch inputs if there was a tap first. I've also moved the logic to the TouchHandler because then it won't persist across reboots like it would in Settings. This does mean that I had to pass TouchHandler to every class that needs to access that data, which might not be ideal. If anyone has any better suggestions, I'm very open to them.

I've also renamed it internally to ScreenLock, and changed the text in the confirmation dialogue to not mention water, as that doesn't seem right. Again, if anyone has any better suggestions, I'm very open to them.

image  
image

src/components/settings/Settings.h
```diff
@@ -212,7 +212,7 @@ namespace Pinetime {
     private:
       Pinetime::Controllers::FS& fs;
 
-      static constexpr uint32_t settingsVersion = 0x0003;
+      static constexpr uint32_t settingsVersion = 0x0004;
```
FintasticMan, line 215: Oops, I accidentally forgot to undo the bump in settings version. I'll fix that in the next commit.

WhyNotHugo:
Vaguely related: https://github.com/InfiniTimeOrg/InfiniTime/issues/1839

PR #964, opened 2022-01-27
Added command line method for generating jetbrains_mono_42.c to README.md

yehoshuapw:
Added `lv_font_conv` based command to `src/displayapp/fonts/README.md`.


Riksu9000:
I think we should only support one method to keep the instructions as simple as possible. The fonts may some day be automatically generated with `lv_font_conv`, but for now I would keep the instructions as they are.

yehoshuapw:
In that case, everything in the file should probably be changed to one method, maybe with a reference at the end that it can also be done using command line.

Also, about auto-generating: I think that some config file (a file json for example) should include all the generated fonts,
and a build_font script be created. if this is an accepted method, I can do it. any better ideas? (parsing the relevant info from the README is doable, but untrustworthy)

Riksu9000:
Yeah, I agree. The font readme was recently updated, but there's a mistake where it says the offline converter is required, when it actually isn't and I've been meaning to fix it.

Auto generating is definitely something we would like to have. I can't really comment on what the best method would be, or if it should integrated with CMake somehow. No need to parse the readme at least. The readme won't be needed after auto generation has been implemented.

PR #932, opened 2022-01-12
add terminal watchface

13werwolf13:
I tried to add watchface found here https://zephyrlabs.github.io/Watchfaces/Terminal/, I asked the developer and he doesn't mind. but now I am not able to check its work.

PR #965, opened 2022-01-27
Added Binary watch face (based on digital)

yehoshuapw:
Created a (rather pointless) watchface.
it's mostly a copy of WatchFaceDigital, with only the time part changed.

in case someone wants to try it out:
[pinetime-mcuboot-app-dfu-1.8.0.zip](https://github.com/InfiniTimeOrg/InfiniTime/files/7952675/pinetime-mcuboot-app-dfu-1.8.0.zip)

and a quick recording (using the simulator. thanks @NeroBurner )
record


yehoshuapw:
Right now, this is using face number 3, if #932 gets merged, I will update to 4.

yehoshuapw:
Note: this is not "real" binary for the time, or even for hours minutes and seconds. this is binary for each digit separately.
(since that is the way I have seen this done before.)

NeroBurner:
I like the idea! I've had a binary watch myself

binary_watch

The background of the watch is black (much darker than on the photo) and the LEDs emit a blue light for a short time when pushing the top button

We could use 4 bits for 12-hour mode, and 5 bits for 24 hour mode

(Nice! My simulator is used before I'm even finished with it! I believe this means the simulator fills a need :blush: )

Riksu9000:
Binary coded decimal is a valid binary format and it's easier to read. Looks nice. Btw GitHub creates DFUs of PRs for us, so you don't need to provide them.

yehoshuapw:
Added 12 hour style (using the setting), using the MSB (which isn't used, since that digit is only 0,1 or 2).
painted it a bit different to differentiate.
record

Any thoughts about better colors are welcome.

0x1a8510f2:
First time flashing a non-release firmware on my sealed PineTime. Terrifying but worth it :P

Some suggestions:

- Perhaps the date could also be in binary but a different colour (blue?)
- It might be good to increase the size of the binary display a bit to make it a little more readable
- The up/down padding for each bit is smaller than the left/right padding which looks a little weird. May be worth increasing u/d a little and possibly decreasing l/r

But overall, would love to see this get merged since I think it's really cool.

yehoshuapw:
>     * Perhaps the date could also be in binary but a different colour (blue?)

that is doable. (but i'm not sure if that will make it too much of a mess. maybe just removing the date?)

>     * It might be good to increase the size of the binary display a bit to make it a little more readable

this too is actually planed, I used the regular font (since LVGL's LED did not work) to make the squares, and there are many other ways to do that too - including just using a larger font - or drawing manually. (which might waste to much memory.)

>     * The up/down padding for each bit is smaller than the left/right padding which looks a little weird. May be worth increasing u/d a little and possibly decreasing l/r

I actually put the spacing there on purpose, to differentiate between the different digits. anyone else have thoughts one way or another?

0x1a8510f2:
>but i'm not sure if that will make it too much of a mess. maybe just removing the date?

I definitely wouldn't remove the date. The way I imagined it is if the current display was duplicated and used to represent DD/MM/YY. The day of the week could also be included in a different colour yet as one set of 4 bits (orange?). Perhaps something like the time at the top, date below, and day of the week to the right but larger so the 4 bits take the same height as the 8 of time+date.

I'll see if I can make a little mockup to see whether that would be overcomplicated as you say.

yehoshuapw:
the real issue here, is actually not the code, but memory. each bit adds a whole LVGL object.

So in some point, If I manage to convert all those things into more efficient representations - then such things can be done.
(maybe by using the "canvas" for each digit. I have no idea right now how much memory each thing actually costs.)

also, what there is right now can be reduced in code-complexity by making a "widget" for bit-displayed digit - which If this is used for more then 6 digits, is definitely worth it. (while it won't be quite zero-cost abstraction, it can be done pretty low cost, by having a pointer back, and doing the minimum there)
[This actually can be extended further, to everything on the watchfaces, which will allow changing them with much less hassle. and after adding a "loader" from some storable format, it will even allow (almost) complete freedom in making watchfaces. but that will wait until I'm less busy]

0x1a8510f2:
That makes sense. I'm not used to embedded dev so I tend to forget memory constraints.

In which case, I'll leave you to it. Hopefully eventually we could get a 100% binary watchface (Bluetooth status, notification status etc. as bits too) because I think that would look awesome, but that'll no doubt take a lot more time.

yehoshuapw:
Actually, changing the other symbols to be just squares in the right color and position is very easy, but... will be unreadable.
(battery, hr and step counter are the only other things on the screen which are not just a symbol. (and time and date of course))

yehoshuapw:
So, by creating a font with only the square, in different sizes:
(using a real picture for a better feel)

24:


26:


28:


30:


32:


personally, I like the 24 (and the 22 - the current), and think that so long as there is more things on the face, the others don't look nice. (and 32 is just too much)

thoughts?

Riksu9000:
The bigger the better. It's just about laying it out nicely and not cramming too much stuff in.

watchsmall

yehoshuapw:
> The bigger the better. It's just about laying it out nicely and not cramming too much stuff in.

however, just as you are saying here - you need to take out stuff.
with the rest of the info as it is, it does not work.

so either:
A) leave things as is and maybe make bigger but not too much
B) have *only* time (or create a "simplistic" setting for the face which does that), and then the layout can be changed.
C) have mostly time, and in a less intrusive method have more info (and not all of it). for example a tiny dot for Bluetooth, (and maybe a tiny binary date display.?)

long term - where this really takes us however, is that watchfaces up to a certain level should be non-code.

0x1a8510f2:
>so either:
A) leave things as is and maybe make bigger but not too much
B) have only time (or create a "simplistic" setting for the face which does that), and then the layout can be changed.
C) have mostly time, and in a less intrusive method have more info (and not all of it). for example a tiny dot for Bluetooth, (and maybe a tiny binary date display.?)

What about something like:

image

yehoshuapw:
> What about something like:

you're adding even _more_ information?
and things like DoW should be smaller then time. (size ~ notability which should also approximate importance)

0x1a8510f2:
>you're adding even *more* information?

I am, and I think it looks cool too :P I understand if you're not a fan, but I figured it wouldn't hurt to suggest it.

>and things like DoW should be smaller then time. (size ~ notability which should also approximate importance)

That's fair and I have thought about it, it's just difficult to make it line up. The sizes and layouts can always be tweaked around, I'm no designer. I primarily was aiming for how things could be represented.

yehoshuapw:
> I am, and I think it looks cool too :P I understand if you're not a fan, but I figured it wouldn't hurt to suggest it.
cool-ness I agree. but consider also usability? what's better, a very cool watch you'd use only rarely to show off, or a cool (but less so) which is usable enough that you can use it as the main watch?


> That's fair and I have thought about it, it's just difficult to make it line up. The sizes and layouts can always be tweaked around, I'm no designer. I primarily was aiming for how things could be represented.
that was what I guessed, but it still needed to be said. (and I think even having empty space is better then completely filling up if it causes that type of thing.)

0x1a8510f2:
>but consider also usability? what's better, a very cool watch you'd use only rarely to show off, or a cool (but less so) which is usable enough that you can use it as the main watch?

Depends on what you bought it for :P but yeah, overall, definitely the latter. However, I do think my design, or something close to it, is still practical enough given that everything is clearly separated. The user need only remember where a piece of information is located or what colour it is, which should come quickly (I've tried to make locations and colours somewhat related to function where I could). Otherwise, if the user is well-versed in binary, it should take almost no time to read. Bear in mind that this watch face still represents only 7 individual pieces of information, which isn't all that much.

>and I think even having empty space is better then completely filling up if it causes that type of thing

That's also a fair point. I guess the lack of empty space could make the watchface feel cluttered. Not sure how to solve that in my design other than maybe increasing gaps between things.

Riksu9000:
The digital watchface is the fully-featured watchface. If you want all the information, use that. Other watchfaces should exist primarily for different looks, and the only requirement is to show the time. A binary watchface is hardly practical anyway.

yehoshuapw:
> The digital watchface is the fully-featured watchface. If you want all the information, use that. Other watchfaces should exist primarily for different looks, and the only requirement is to show the time. A binary watchface is hardly practical anyway.

while the purpose is not to be practical, so some unpracticalness is allowed by definition, it does not mean it needs to be unpractical. (I didn't manage to make that have more "practicals" then your sentence had "watchface". oh well)


@0x1a8510f2  how about if we put the time in the middle, slight smaller under it the date (i think YY-MM-DD, which is closer to MSB-LSB, and helps avoid the MM/DD or DD/MM fight. but this actually can be easily configurable),
battery and notifications as small dots on the top as a single line, and steps and hr on the buttom?
that leaves bluetooth and charging, which can be a thin line on each side

0x1a8510f2:
>how about if we put the time in the middle, slight smaller under it the date (i think YY-MM-DD, which is closer to MSB-LSB, and helps avoid the MM/DD or DD/MM fight. but this actually can be easily configurable),
battery and notifications as small dots on the top as a single line, and steps and hr on the buttom?
that leaves bluetooth and charging, which can be a thin line on each side

I like that idea. Only thing is, instead of dots I'd stick to squares for battery and notifications (but smaller naturally). Otherwise, I think that would be pretty awesome :D

yehoshuapw:
redone some stuff, mainly for personal use (since I wanted it, but didn't have time to really create a binary watch face as discussed)

PR #966, opened 2022-01-27
PineTimeStyle: SetClockFace(2)

yehoshuapw:
I notices all the other watch faces do this (for the relevant number),
and PineTimeStyle does not - so here's a tiny change to make it call
`SetClockFace` too.

(alternatively, maybe all the watchfaces should _not_ do this.)


kieranc:
I think that this used to be necessary to return to the correct watchface after leaving the menu, but is not needed any more. Can anyone confirm?

Riksu9000:
We should rather remove this from the other watchfaces. I don't think it does anything currently.

PR #967, opened 2022-01-27
SystemMonitor: implement FreeRtosMonitor only if trace facility is set

NeroBurner:
No need to implement `SystemMonitor<FreeRtosMonitor>` if the variable
`configUSE_TRACE_FACILITY` is set to `1`. Otherwise implement the
`DummyMonitor` class.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -516,6 +516,7 @@ list(APPEND SOURCE_FILES
         displayapp/lv_pinetime_theme.c
 
         systemtask/SystemTask.cpp
+        systemtask/SystemMonitor.cpp
         drivers/TwiMaster.cpp
 
         heartratetask/HeartRateTask.cpp
@@ -577,6 +578,7 @@ list(APPEND RECOVERY_SOURCE_FILES
         FreeRTOS/port_cmsis.c
 
         systemtask/SystemTask.cpp
+        systemtask/SystemMonitor.cpp
         drivers/TwiMaster.cpp
         components/gfx/Gfx.cpp
         components/rle/RleDecoder.cpp
--- a/src/components/ble/AlertNotificationClient.cpp
+++ b/src/components/ble/AlertNotificationClient.cpp
@@ -2,6 +2,7 @@
 #include <algorithm>
 #include "components/ble/NotificationManager.h"
 #include "systemtask/SystemTask.h"
+#include <nrf_log.h>
 
 using namespace Pinetime::Controllers;
 constexpr ble_uuid16_t AlertNotificationClient::ansServiceUuid;
--- a/src/components/ble/DfuService.cpp
+++ b/src/components/ble/DfuService.cpp
@@ -3,6 +3,7 @@
 #include "components/ble/BleController.h"
 #include "drivers/SpiNorFlash.h"
 #include "systemtask/SystemTask.h"
+#include <nrf_log.h>
 
 using namespace Pinetime::Controllers;
 
--- a/src/components/ble/HeartRateService.cpp
+++ b/src/components/ble/HeartRateService.cpp
@@ -1,6 +1,7 @@
 #include "components/ble/HeartRateService.h"
 #include "components/heartrate/HeartRateController.h"
 #include "systemtask/SystemTask.h"
+#include <nrf_log.h>
 
 using namespace Pinetime::Controllers;
 
--- a/src/components/ble/MotionService.cpp
+++ b/src/components/ble/MotionService.cpp
@@ -1,6 +1,7 @@
 #include "components/ble/MotionService.h"
 #include "components/motion/MotionController.h"
 #include "systemtask/SystemTask.h"
+#include <nrf_log.h>
 
 using namespace Pinetime::Controllers;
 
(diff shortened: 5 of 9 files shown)


JF002:
The definition of the template `SystemMonitor<FreeRtosMonitor>` should not cause any issue if it's not instantiated by the code that is being built. Did you get any error while building the simulator?

NeroBurner:
yes, without the commit I'd have to implement/stub `xPortGetFreeHeapSize()` for a logger, that isn't used in the simulator

```
[build] /usr/bin/c++ -DLV_CONF_INCLUDE_SIMPLE -I/home/nero/repos/pinetime/InfiniSim/sim -I/home/nero/repos/pinetime/InfiniSim/sim/libraries/log -I/home/nero/repos/pinetime/InfiniSim/sim/libraries/timer -I/home/nero/repos/pinetime/InfiniSim/sim/nrfx -I/home/nero/repos/pinetime/InfiniSim/sim/nrfx/hal -I/home/nero/repos/pinetime/InfiniSim -I/home/nero/repos/pinetime/InfiniSim/InfiniTime/src/libs -I/home/nero/repos/pinetime/InfiniSim/lv_drivers -I/home/nero/repos/pinetime/InfiniSim/InfiniTime/src -I/home/nero/repos/pinetime/InfiniSim/build -isystem /home/nero/repos/pinetime/InfiniSim/InfiniTime/src/libs/date/includes -isystem /usr/include/SDL2 -g -std=c++17 -MD -MT CMakeFiles/infinisim.dir/sim/components/battery/BatteryController.cpp.o -MF CMakeFiles/infinisim.dir/sim/components/battery/BatteryController.cpp.o.d -o CMakeFiles/infinisim.dir/sim/components/battery/BatteryController.cpp.o -c /home/nero/repos/pinetime/InfiniSim/sim/components/battery/BatteryController.cpp
[build] In file included from /home/nero/repos/pinetime/InfiniSim/InfiniTime/src/systemtask/SystemMonitor.h:4,
[build]                  from /home/nero/repos/pinetime/InfiniSim/InfiniTime/src/systemtask/SystemTask.h:15,
[build]                  from /home/nero/repos/pinetime/InfiniSim/sim/components/battery/BatteryController.h:3,
[build]                  from /home/nero/repos/pinetime/InfiniSim/sim/components/battery/BatteryController.cpp:1:
[build] /home/nero/repos/pinetime/InfiniSim/InfiniTime/src/systemtask/SystemMonitor.h: In member function ‘void Pinetime::System::SystemMonitor<Pinetime::System::FreeRtosMonitor>::Process() const’:
[build] /home/nero/repos/pinetime/InfiniSim/InfiniTime/src/systemtask/SystemMonitor.h:26:83: error: ‘xPortGetFreeHeapSize’ was not declared in this scope
[build]    26 |           NRF_LOG_INFO("---------------------------------------\nFree heap : %d", xPortGetFreeHeapSize());
[build]       |                                                                                   ^~~~~~~~~~~~~~~~~~~~
[build] /home/nero/repos/pinetime/InfiniSim/sim/libraries/log/nrf_log.h:56:66: note: in definition of macro ‘NRF_LOG_INFO’
[build]    56 | #define NRF_LOG_INFO(...)         do {printf("info:  "); printf( __VA_ARGS__ ); printf("\n"); } while(false);
[build]       |                                                                  ^~~~~~~~~~~
```

The only place where this logger is instantiated is in `SystemTask.h` with the following statement:
```cpp
#if configUSE_TRACE_FACILITY == 1
      SystemMonitor<FreeRtosMonitor> monitor;
#else
      SystemMonitor<DummyMonitor> monitor;
#endif
```

One thing I just noticed is that we could get rid of the templated System-Monitor class and use just a `SystemMonitor monitor;` in `SystemTask.h` if we use the `configUSE_TRACE_FACILITY` macro to decide the implementation of the monitor.

If this restructure isn't wanted I'll have to stub the `xPortGetFreeHeapSize()` in `InfiniSim`. Your choice, we got all the freedom to choose :)

JF002:
Ok, I understand :) The goal with the template class was to avoid using `#ifdef` : both versions exist in the code but only 1 will be linked in the final binary. Well I failed at that goal as the `#ifdef` is still needed in `systemtask.h` any way...

On the long term, I would like to find a better design to handle multiple platforms but for now, I think your first solution will be the best : 
> One thing I just noticed is that we could get rid of the templated System-Monitor class and use just a SystemMonitor monitor; in SystemTask.h if we use the configUSE_TRACE_FACILITY macro to decide the implementation of the monitor.

NeroBurner:
With the restructure of `SystemMonitor.h` some missing `nrf_log.h` includes were detected which I've fixed in [Add missing nrf_log.h includes shadowed by SystemMonitor.h](https://github.com/InfiniTimeOrg/InfiniTime/pull/967/commits/eb7fc2df19d528904f13d327ddd5b9ffff3faeaf)

> Some components were missing a `nrf_log.h` include. This missing
> include was accidentally provided by the SystemMonitor.h header, which
> was included by Systemtask.h

And I've split the SystemMonitor into a h and cpp pair, as the implementation of a non-template function shouldn't be in the header. As this could cause errors with the one-definition rule (not here, but generally)

Furthermore I've changed the `Process()` function to be a non-const function, as it modifies a member-variable, and thus shouldn't be marked as `const`

NeroBurner:
rebased on current `develop` branch 0e2b27d7926035f63790851db5c00af431b2eb07

PR #971, opened 2022-01-29
Adjust screen gamma and fix UI colors

### LINKED ISSUE #745: Orange displaying as a goldish yellow
https://github.com/InfiniTimeOrg/InfiniTime/blob/e9bb0b3cdd0e1b95b7bfdcfb94db5a6a0966eed7/src/displayapp/Colors.cpp#L24

When playing around with the watch face colors I noticed that I couldn't find an orange color, found it in the src but it's displaying as a darker yellow for me at least on my unit. May need to adjust the color values to get it right if this is widespread.

Riksu9000:
The PineTime screen has poor gamma by default. This is why the colors look very different when simulated on a computer with more accurate colors #743.

Here I've adjusted the gamma of the screen and tweaked the colors to match what they previously looked like by eye. It's not perfect, but hopefully close enough.

Fixes #745

### DIFF
--- a/src/displayapp/screens/Navigation.cpp
+++ b/src/displayapp/screens/Navigation.cpp
@@ -146,7 +146,7 @@ Navigation::Navigation(Pinetime::Applications::DisplayApp* app, Pinetime::Contro
 
   txtManDist = lv_label_create(lv_scr_act(), nullptr);
   lv_label_set_long_mode(txtManDist, LV_LABEL_LONG_BREAK);
-  lv_obj_set_style_local_text_color(txtManDist, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_GREEN);
+  lv_obj_set_style_local_text_color(txtManDist, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_MAKE(0x00, 0xc0, 0x00));
   lv_obj_set_width(txtManDist, LV_HOR_RES);
   lv_label_set_text(txtManDist, "--M");
   lv_label_set_align(txtManDist, LV_LABEL_ALIGN_CENTER);
--- a/src/displayapp/lv_pinetime_theme.c
+++ b/src/displayapp/lv_pinetime_theme.c
@@ -38,6 +38,7 @@ static lv_style_t style_title;
 static lv_style_t style_label_white;
 static lv_style_t style_back;
 static lv_style_t style_icon;
+static lv_style_t style_bar;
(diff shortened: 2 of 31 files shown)


NeroBurner:
I'd split the PR into two and do here just the gamma correction (to be more color accurate), and in another one the color changes (color is very subjective and I think discussion about color taste may block the gamma change part of this PR)

Riksu9000:
@NeroBurner Fair. The color changes will require more work. I've now tried to match the colors closer, apart from a few changes which I consider fixes. They are also in #978. Swapping arc colors, brightening steps label and creating a default background for a bar. Since I had to touch those colors anyway, I decided to fix them too.

trman:
> Since I had to touch those colors anyway, I decided to fix them too.

swapping colors , is not fix colors , is changing the theming of infinitme to suit  your taste
and if this accepted  , every people will folow suit and do say the same :
-> create pull to change the colors scheme to something that suit their taste and call it a fix !

Riksu9000:
> swapping colors , is not fix colors , is changing the theming of infinitme to suit your taste

I think this concern has been resolved in the thread at #978.

src/displayapp/screens/Navigation.cpp
```diff
@@ -146,7 +146,7 @@ Navigation::Navigation(Pinetime::Applications::DisplayApp* app, Pinetime::Contro
 
   txtManDist = lv_label_create(lv_scr_act(), nullptr);
   lv_label_set_long_mode(txtManDist, LV_LABEL_LONG_BREAK);
-  lv_obj_set_style_local_text_color(txtManDist, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_GREEN);
+  lv_obj_set_style_local_text_color(txtManDist, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_MAKE(0x00, 0xc0, 0x00));
```
NeroBurner, line 149: why define LV_PINTEIMTE_COLOR_GREEN and then use `LV_COLOR_MAKE` in the code? I don't understand

Riksu9000: None of the other theme defined colors are used in code either. I don't think `lv_pinetime_theme.h` is the best place to keep our colors, so I didn't want to use that. Ideally we would have the colors in some cpp format, but then they can't be included in the C theme file. I thought this issue should be resolved separately.

Riksu9000:
Closed in favor of #1103

PR #973, opened 2022-01-29
New checkable button design

Riksu9000:
Alarm is the first app to get a checkable button, the on/off button. A switch probably wouldn't work in this app, so the button needs to be made distinguishable somehow. This is what I came up with. What do you think?

off
 
on


JF002:
I'm far from being a reference in UI design, but I think this is nice. The border around the button "announces" that it's a special kind of button!

Riksu9000:
I decided to use a switch after all as it is a design people are already familiar with, instead of creating something new. #982

PR #974, opened 2022-01-29
SystemTask: fix static cast missing brackets syntax

NeroBurner:
The static_cast syntax requires brackets around the input variable. The
implementation worked because the used input are defines, which add the
missing brackets like the following:
```cpp
  #define GPIO_PIN_CNF_SENSE_Low (3UL)
```


NeroBurner:
The change was done in https://github.com/InfiniTimeOrg/InfiniTime/commit/6a442b90a1d310d082e2626128d0f2bf8e932851#diff-e56aa4b7322b0d628dc46d9cd07cea2accd612ca7f4eb8f7ec271e3596325237R206

@Avamander was this intentional? please review

Riksu9000 (approved):

JF002:
The change from @Avamander was probably not intentional. And I'm surprised that it compiled without issue/warning!

PR #976, opened 2022-01-29
Notifications: replace newlines in label-copy instead of const char* title directly

NeroBurner:
The variable `title` is defined as `const char*`, which means, that
`strchr()` returns a `const char*` as well according to
https://www.cplusplus.com/reference/cstring/strchr/

But in the same line the return value is assigned to a non-const
`char*`, which shouldn't be allowed (error with `-pedantic`).

To prevent manual memory management just use `std::string` to copy the
provided title and replace newlines in the copied string.

### DIFF
--- a/src/displayapp/screens/Notifications.cpp
+++ b/src/displayapp/screens/Notifications.cpp
@@ -198,15 +198,18 @@ Notifications::NotificationItem::NotificationItem(const char* title,
 
   lv_obj_t* alert_type = lv_label_create(lv_scr_act(), nullptr);
   lv_obj_set_style_local_text_color(alert_type, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(0x888888));
-  if (title == nullptr)
-    title = "Notification";
-  char* pchar;
-  pchar = strchr(title, '\n');
-  while (pchar != nullptr) {
-    *pchar = ' ';
-    pchar = strchr(pchar + 1, '\n');
+  if(title == nullptr) {
+    lv_label_set_text_static(alert_type, "Notification");
+  } else {
+    // copy title to label and replace newlines with spaces
+    lv_label_set_text(alert_type, title);
+    char *pchar = strchr(lv_label_get_text(alert_type), '\n');
+    while (pchar != nullptr) {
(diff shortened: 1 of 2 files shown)


JF002:
In this case, it's not clear if we use the strchr() function from the C library (`char * strchr ( const char *, int ); `) or the const override from the C++ library (`const char * strchr ( const char * str, int character );`).

According to CLion, strchr() comes from string.h, which is the C version that does not define a `const` return value.

**However**, even if we are using the C version, we still modify the content of the string `title` which beaks the `const` declaration of the variable, so your concern to totally valid!

I guess the intent of the code was to avoid the copy and the memory overhead of std::string (with potential dynamic allocation).

I'm wondering : is there any other solution that would respect the constness that does not need a std::string temporary variable?

NeroBurner:
Found a solution: `lv_label_set_text` creates a copy of the string itself. Access and modify that `char*` instead of the `const char* title`

Also removed an unused struct in `Notifications.h`

JF002:
Did you check that it actually works? The text buffer is edited, but LVGL is not aware of that change. According to the code of `lv_label_set_text()` you'll probably need to call it again with `NULL` as parameter, or call `lv_label_refr_text()` directly.

NeroBurner:
It worked in the simulator, but I'll add the `lv_label_refr_text()` statement just to be sure

JF002:
Did you check with notification title that contains a \n character?

NeroBurner:
Yes, but only on the Simulator, as it is easier for me to get a notification with a title with newlines there

The shown notification has as input the following title-message-pair:

```cpp
"category\nUnknown",
"Lorem ipsum\ndolor sit amet,\nconsectetur adipiscing elit,\n",
```

Unknown

Riksu9000:
If the newline is replced, dolor should be on the first line right?

test

NeroBurner:
The newline is only replaced in the title (the `category Unknown` part in my example). The newlines in the message are kept as they are (the message is not modified)

NeroBurner:
rebased on current `develop` branch 0e2b27d7926035f63790851db5c00af431b2eb07

Issue #953, opened 2022-01-22
Drop down list: scrollbar goes over the edge


PhilDevProg:
### Verification

- [X] I searched for similar bug reports and found none was relevant.

### What happened?

The scrollbar goes over the edge in the drop down list of the metronome app 

### What should happen instead?

The scrollbar shouldn't go over the edge.

### Reproduction steps

I open the metronom app.
I press the button with the text "4 bpb".
I scroll down the list to the bottom.
The scrollbar goes over the edge here:
![IMG_20220122_225015_1](https://user-images.githubusercontent.com/91820316/150656672-7506a9e1-519e-4aa0-91e6-6efdbc1d6069.jpg)

### Version

1.8.0

JF002:
I guess this bug is related to LVGL (the UI lib used in InfiniTime) and not directly to InfiniTime. 

Unless someone is willing to write a patch for LVGL, this bug might be fixed in a future/more recent version of LVGL, but I don't know when we'll update LVGL in InfiniTime.

Riksu9000:
This issue is at least partially caused by incorrect theme configuration in InfiniTime.

JF002:
> This issue is at least partially caused by incorrect theme configuration in InfiniTime.

Oh really? I'll re-open it, then!

PR #978, opened 2022-01-30
Theme fixes and cleanup

Riksu9000:
A disabled state was added to the default button theme. Now Stopwatch, SetTime and SetDate match.
Arrow icons were added. These are used for dropdowns.
Arc colors were swapped.
Steps text was made lighter as it was too dark.
The default bar theme now has a background. This is visible on the firmware update screen and navigation.
Unused styles were removed and other small improvements in `lv_pinetime_theme.c`

This touches #953, but the dropdown still has some issues.

date
metronome
navigation
steps
update


trman:
why did you swap the arc color ? 
to force your color taste?
the rest of the modification could related to button and drop down were good but 
but tying these with color change... is tasteless...

Riksu9000:
It's common for the arc to fill clockwise and for the foreground to be brighter than the background.

trman:
> It's common for an arc to fill clockwise and for the foreground to be brighter than the background.

no , it wholly depend from themes it not "common" : 
if theme is black  for foreground
or 
if theme is white for  foreground

ilke night and white theme on android....

trman:
by the way , in my case , i like the actual theme rather that your's

Riksu9000:
Let me rephrase that. Foreground should have more contrast. It should "pop" more. Also the InfiniTime theme is a dark theme, where background is dark and foreground is bright.

Screenshot from 2022-01-30 19-59-05

trman:
thanks ! 
this explanation is more understandable !
now i understand this fix , because without it it felt like a color taste 'fix' !

PR #979, opened 2022-01-30
Provide reference to BrightnessController in DisplayApp

NeroBurner:
For the simulator I need a way to get to the brightnessController object
and handle the set brightness-levels accordingly.

This is done by the constructor expecting a brightnessController object
instead of initializing one itself.

If this change is too intrusive adding a function like `getBrightnessLevel()` to just the main `DisplayApp.h` forwarding the brightness level of the internally stored `BrightnessController` would be another way forward for me.

Or return a reference with `getBrightnessController()`


JF002:
I agree with the changes. Could you fix the conflict so we can merge?

NeroBurner:
rebased :)

PR #980, opened 2022-02-01
feat(watchfacedigital): change color of the step counter to show when the step goal has been reached

FrazerClews:
Open to different color options when the goal has been reached.

Used LV_COLOR_YELLOW just for testing to mimic gold and it looked good in my opinion.


NeroBurner (approved):
I like the idea! Change seems minimal and clean

a few more steps step goal reached

kieranc:
Looks good to me too, I'd prefer to see the colors both defined as LV_COLOR_SOMETHING rather than a mix of hex and standard colors if possible.

PR #982, opened 2022-02-02
Alarm: Change checkable button for a switch

Riksu9000:
I tried to create a checkable button design in #973, but wasn't totally happy with it, so I changed it to a switch here anyway. This is better, because it uses a design people are already familiar with instead of creating something new.

alarm

EDIT: The alert layout doesn't need to be changed, so I reverted that change. The adjustment buttons don't need to be visible during alerting, but that's for another PR.

### DIFF
--- a/doc/ui_guidelines.md
+++ b/doc/ui_guidelines.md
@@ -9,5 +9,6 @@
 - Top bar takes at least 20px + padding
 	- Top bar right icons move 8px to the left when using a page indicator
 - A black background helps to hide the screen border, allowing the UI to look less cramped when utilizing the entire display area.
+- A switch should be twice as wide as it is tall.
 
 ![example layouts](./ui/example.png)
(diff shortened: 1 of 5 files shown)


escoand:
For me there is nothing that needs to be fixed at the 1.8.0 design. IMHO it's absolutely understandable.

trman:
it woulf be better if the switch contain letter 'on ' inside when on
and off when off
it would make more consitant

Riksu9000:
@escoand True, the current design is understandable, but I think it's important to use the appropriate widgets where possible. I'm hoping this also makes it easier to use colors for other buttons.

@trman LVGL Switches don't support labels by default. I hope it's understandable without it. We can worry about color change when the time comes to add that feature. This is what it looks like disabled. 

disabled

medeyko:
Normally the button captions contains a description of the state AFTER the button is pressed, NOT the CURRENT state.

So current design is a bit misleading. At least, when I saw it first time, I was unsure of the exact meaning.

I guess that "on"/"off" state might be indicated by separate text, for example between the "-" buttons.

Issue #988, opened 2022-02-07
Zero character


medeyko:
### Verification

- [X] I searched for similar bug reports and found none was relevant.

### What happened?

The dot inside the zero character is glued to the 0 itself. That resembles 8 and makes reading more difficult.

### What should happen instead?

The dot inside 0 should not glue to the boundary.

### Reproduction steps

See for example 0 in 2022 on the digital watch face.

### More details?

Here's a sample image:

![изображение](https://user-images.githubusercontent.com/7012236/152727763-1ab3c5c4-daca-486a-855a-d58dfe1d822d.png)

Zero in 20 is current 0 character. Zero standing alone is how I propose to change the 0 character.

To make this, the jetbrains_mono_bold_20.c should be edited:

    /* U+0030 "0" */
    0x3f, 0x1f, 0xef, 0x3f, 0x87, 0xe1, 0xf8, 0x7f,
    0xdf, 0xf7, 0xe1, 0xf8, 0x7e, 0x1f, 0xcf, 0x7f,
    0x8f, 0xc0,

should be replaced with

    /* U+0030 "0" */
    0x3f, 0x1f, 0xef, 0x3f, 0x87, 0xe1, 0xf8, 0x7e,
    0xdf, 0xb7, 0xe1, 0xf8, 0x7e, 0x1f, 0xcf, 0x7f,
    0x8f, 0xc0,

(there are two changes: 7f -> 7e and f7 -> b7.)

I understand that the process of the font generation is automated, but I think that this manual change worth doing.

### Version

v1.8.0

### Companion app

-

escoand:
This is maybe fixed with #955.

medeyko:
@escoand 
I don't think that the Jetbrains mono font should be removed. Anyway, #955 issue will be discussed some time, I think. The fix I proposed is quite simple and may be introduced in 1.9.0 already.

Riksu9000:
I think the reason this hasn't been fixed is that the fix may easily get lost when someone updates the font. But I think this may be worth fixing anyway, as long as it is well documented and we make sure the fix doesn't get lost with future updates. Since you have the solution to the issue, would you like to create the PR for it @medeyko ? Make sure to update https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/src/displayapp/fonts/README.md

medeyko:
@Riksu9000 
Thank you! I made #989 PR, hopefully it's fine...

JF002:
[...]

PR #989, opened 2022-02-07
JetBrains Mono number zero character improvement

medeyko:
This should fix 0 (zero) symbol.
For more details, see #988


src/displayapp/fonts/README.md
```diff
@@ -31,6 +31,23 @@ Add new symbols:
...
+
+with
+
+    /* U+0030 "0" */
+    0x3f, 0x1f, 0xef, 0x3f, 0x87, 0xe1, 0xf8, 0x7e,
+    0xdf, 0xb7, 0xe1, 0xf8, 0x7e, 0x1f, 0xcf, 0x7f,
+    0x8f, 0xc0,
+
+(there are two changes: 7f -> 7e and f7 -> b7)
+
```
Riksu9000, line 50: I think these instructions should be moved to the "Generate the fonts" section, maybe below the line where it says to download the file. Also the instructions should be worded in a way that makes this a requirement rather than a decision. This is to make sure this fix is included each time the font is updated in InfiniTime. It could be something like "To fix an error that happens during the font conversion...".

medeyko: But it already was in the "Generate the fonts" section!
Anyway, I removed the possible misleading subheader, and reworded the text to be more imperative.

JF002:
Thanks for this PR!
However, this change is not easily maintainable : the file `jetbrains_mono_bold_20.c` is generated by the LVGL font generator from and should not be edited manually (as those changes will be lost the next time someone will generate the font).

Adding the instructions to manually fix the generated file in the documentation is certainly a good idea but those manual operations are cumbersome and not error proof. In the long term, I would like the fonts to be automatically generated by CMake during the project generation. The idea is to ensure the fonts will always be generated correctly without any manual manipulation.

Would it be possible to fix the 0 character in the source file of the font (`JetBrainsMono-Bold.ttf`) instead of the generated file? The font is also available on Github (https://github.com/JetBrains/JetBrainsMono) so we might also try to 'mainline' those changes. It might have already been fixed since the last release of the font is newer than the file used in the project. 

If editing the source of the font is not possible, another possibility would be to automatically (by CMake) patch the generated file each time the file is re-generated. 

Can we explore such alternatives before sticking to additional manual operations?

medeyko:
I tried to edit the font, and it seems that there's no any problem in the original TTF-file. The edit that fixed this particular 0 at 20px makes the symbol in other sizes look wrong. So it seems to be an artifact of LVGL font convertor that I don't see how to handle.

So it seems that we should patch the `jetbrains_mono_bold_20.c` file either during build or in the repository (as it is proposed in the patch we discuss). I afraid that the latter is more manageable, because the former may be harder to understand.

JF002:
Mhm ok, it's more complicated than what I expected. Thanks for checking to possibilities!
So, yeah, in the short term, we can manually patch the generated file and in the long term, try to generate and patch the font automatically during the project generation :)

medeyko:
There's theoretically also an option to propose changes in LVGL font converter. Its sources could be found here: https://github.com/lvgl/lv_font_conv
However I looked into the lib/freetype/build/ft_render.js (571 KB long) there and found a wasm binary code in it.
But I haven't found the wasm binary code sources yet...

medeyko:
@JF002, @Riksu9000:
Could you please advise should I do anything else with this PR, or just wait?

Riksu9000:
@medeyko Check my review, or reply to it if you disagree.

Avamander:
Anyone more familiar with the font generator, is there a way to alter the glyph generation at all? E.g. downscale from a custom size that might fix the zero, pixel offset for downscaling or something similar?

We should probably ask on the LVGL font converter issue tracker if there are any ways if someone here doesn't know.

JF002:
@medeyko Thanks again for your contribution! I'll merge it, and I hope we'll be able to automate the generation and fix of the font soon :)

PR #1000, opened 2022-02-16
Remove unused and not compiling DropDownDemo

NeroBurner:
For ease of use the simulator uses a globbing expression to get all
screens source files. This one was picked up as well and lead to a
compilation error.


JF002:
Yep, this is a very old demo I wrote to test the touch panel, but it's not needed anymore!

JF002:
Congratulations on the 1000th PR on the project! :1st_place_medal:

PR #994, opened 2022-02-13
#992 Added installing cbor packet into the Dockerfile

AlexXZero:
# Ticket:
https://github.com/InfiniTimeOrg/InfiniTime/issues/992

# Description:
This PR should fix the error of building FW using docker:
```
Traceback (most recent call last):
  File "../../tools/mcuboot/imgtool.py", line 17, in <module>
    from imgtool import main
  File "/sources/tools/mcuboot/imgtool/main.py", line 23, in <module>
    from imgtool import image, imgtool_version
  File "/sources/tools/mcuboot/imgtool/image.py", line 22, in <module>
    from .boot_record import create_sw_component_data
  File "/sources/tools/mcuboot/imgtool/boot_record.py", line 17, in <module>
    import cbor
ModuleNotFoundError: No module named 'cbor'
```

# Other
I can't push docker image to dockerhub so @JF002 could you do it to get docker image be fixed.
I have tested the result docker image locally, but it would be good if it will be tested by someone else :-)

Issue #992, opened 2022-02-10
Docker build fails -- no module 'cbor' when running tools/mcuboot/imgtool/boot_record.py.  Importing 'cbor2' instead works.


stoehraj:
### Verification

- [X] I searched for similar bug reports and found none was relevant.

### What happened?

Docker build fails, with "no module named 'cbor'"

### What should happen instead?

Docker build succeeds.

### Reproduction steps

1. From root directory, build infinitime-build docker image:

    `docker image build -t infinitime-build ./docker`

    When building the image, it looks at `mcuboot/requirements.txt` and downloads `cbor2` to fulfill the requirement of `cbor>=1.0.0` with this output:

    ```
    Collecting cbor2 (from -r /opt/mcuboot/scripts/requirements.txt (line 4))
    Downloading https://files.pythonhosted.org/packages/49/74/2b9f9b76e2831f54d70e177a4e602e72cb86441632d1bcc15be1f04a73f6/cbor2-5.4.2.post1.tar.gz (85kB)
    ```

2. Alternatively, pull latest pre-built image:

    `docker pull pfeerick/infinitime-build`

3. Run the docker build:

    `docker run --rm -it -v $(pwd):/sources infinitime-build`

    or

    `docker run --rm -it -v $(pwd):/sources pfeerick/infinitime-build`

4. The build fails with this output:

    ```
    Traceback (most recent call last):
      File "../../tools/mcuboot/imgtool.py", line 17, in <module>
        from imgtool import main
      File "/sources/tools/mcuboot/imgtool/main.py", line 23, in <module>
        from imgtool import image, imgtool_version
      File "/sources/tools/mcuboot/imgtool/image.py", line 22, in <module>
        from .boot_record import create_sw_component_data
      File "/sources/tools/mcuboot/imgtool/boot_record.py", line 17, in <module>
        import cbor
    ModuleNotFoundError: No module named 'cbor'
    ```

### More details?

Modifying `tools/mcuboot/imgtool/boot_record.py` to use `import cbor2` rather than `import cbor` fixes the issue and allows the build to finish successfully.

I'm not sure where exactly this should be fixed (or if this impacts non-Docker builds) so I documented it here for somebody more familiar with the build system.

### Version

develop

### Companion app

N/A

AlexXZero:
I found that same issue on my PC using docker container. I have fixed it locally just installing `cbor` packet into the docker container:
[...]

PR #1001, opened 2022-02-20
#992 Update docker/build.sh to use specific requirements.txt version from mcutools

AlexXZero:
# Ticket:
https://github.com/InfiniTimeOrg/InfiniTime/issues/992

# Description:
This PR should fix the error of building FW using docker:
```
Traceback (most recent call last):
  File "../../tools/mcuboot/imgtool.py", line 17, in <module>
    from imgtool import main
  File "/sources/tools/mcuboot/imgtool/main.py", line 23, in <module>
    from imgtool import image, imgtool_version
  File "/sources/tools/mcuboot/imgtool/image.py", line 22, in <module>
    from .boot_record import create_sw_component_data
  File "/sources/tools/mcuboot/imgtool/boot_record.py", line 17, in <module>
    import cbor
ModuleNotFoundError: No module named 'cbor'
```

# Other
This pull request was created instead [PR-944](https://github.com/InfiniTimeOrg/InfiniTime/pull/994). I used @stoehraj the idea of the fix (See [link](https://github.com/InfiniTimeOrg/InfiniTime/issues/992#issuecomment-1038378230)), since I could not find his PR.
I can't push docker image to dockerhub so @JF002 could you do it to get docker image be fixed.
I have tested the result docker image locally.

### DIFF
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -43,8 +43,10 @@ GetGcc() {
 }
 
 GetMcuBoot() {
-  git clone https://github.com/mcu-tools/mcuboot.git "$TOOLS_DIR/mcuboot"
-  pip3 install -r "$TOOLS_DIR/mcuboot/scripts/requirements.txt"
+  mkdir -p "$TOOLS_DIR/mcuboot"
+  wget -q https://raw.githubusercontent.com/mcu-tools/mcuboot/71b8f981df4f4e65df1757c6ba14ce8d2b7e7696/scripts/requirements.txt \
+    -O "$TOOLS_DIR/mcuboot/requirements.txt"
+  pip3 install -r "$TOOLS_DIR/mcuboot/requirements.txt"
 }
 
 GetNrfSdk() {


JF002:
Did you check that `GetMcuBoot` is really needed ? My point in [my comment](https://github.com/InfiniTimeOrg/InfiniTime/issues/992#issuecomment-1045990749) is that MCUBoot is already included in the InfiniTime projecet, so it shouldn't be needed to download it from another place.

AlexXZero:
Even if we use existing MCUBoot tools we still need to install its dependencies from `tools/mcuboot/requirements.txt`
The main issue that we can't use existing `requirements.txt` since it placed outside of the `docker` folder.

If you add the following code to the Dockerfile
```
COPY ../tools/mcuboot/requirements.txt /opt/mcuboot/requirements.txt
RUN pip3 install -r "/tmp/mcuboot/requirements.txt"
```
Then you receive the following error:
```
Step 6/17 : COPY ../tools/mcuboot/requirements.txt /opt/mcuboot/requirements.txt
COPY failed: forbidden path outside the build context: ../tools/mcuboot/requirements.txt ()
```
So in my opinion downloading specific version of `requirements.txt` is good option.

Note: The updated version of `GetMcUBoot` simply installs the dependency packets, and it does not install the McUBoot tool into the Docker container.

stoehraj:
I apologize for not getting a PR up quicker haha -- but you seem eager to take this on so you can certainly have this one.

Another alternative would be just putting a copy of Infinitime's `requirements.txt` into the docker directory, but I'm not sure that's a good practice.

As it stands, in the current form, `requirements.txt` needs to be downloaded from somewhere -- either from the `mcuboot` GitHub repo or the Infinitime GitHub repo would work.  At the same time, having to depend on GitHub is a negative in its own way when the file is already part of the source... maybe having another copy of the file in the docker directory would actually be the cleanest solution.

AlexXZero:
In my opinion, have a copy or even two copies of the file worse than to download it using a reference to a specific commit of the file in the repository of a third party. At a minimum because someone can change the local copy not knowing what is the original file. So I suggest to follow DRY principle in this case.

JF002:
Oh I understand : We need to know the dependencies during the creation of the container, and not at build time... Sorry for the confusion!

So, the original issue is that `cbor2` is missing in the container. As we don't plan to update mcuboot any time soon, wouldn't the simplest solution be to just add the installation command line in the docker file?
Yes, I know, it won't automatically update when (if) we ever update mcuboot and `imgtool`, but it's simple and it avoids us from depending on the availability of a file online.

AlexXZero:
The original issue is that `cbor` is missing in the container, since MCUBoot repo updated their `requirements.txt` to use `cbor2` instead. But we use old version of MCUBoot, so we should install dependencies of its used version. Initially we download latest version of `requirements.txt` from MCUBoot repo, but after this fix it will be used the same version of `requirements.txt` as MCUBoot tool.  So we have at least three options:
1. Don't use external `requirements.txt` file and just write a list of installed packets in the `Dockerfile`
Cons: it is not cleanest solution and it will be hard to figure our from where these packets come. 

2. Update `build.sh` to download proper version of `requirements.txt` using proper SHA number of commit from MCUBoot repo
Cons: We have a copy of proper version of `requirements.txt` in the InfiniTime repo (`tools/mcuboot/requirements.txt`), but we can't use it for building docker image because it isn't placed in the `docker` folder.

3. Make a copy of proper version of`requirements.txt` in InfiniTime repo.
Cons: it produce two copy of the same file in InifiniTime repo and it break DRY principal.

I like option 2 since it follow previous way when we download MCUBoot tools from their repo, just right now it will download proper version.

pfeerick:
4. 3, using a symbolic link to remove con/duplication? Assuming docker will accept that.

JF002:
To be honest, I prefer the 1st solution (just add the dependency in the Dockerfile). There are already some dependencies that are specified in the this file (git, cmake, python, adafruit-nrfutil,...) and I don't think adding one would hurt anything.

Also, there are very few chances we update MCUBoot any time soon so... let's keep it simple! Remember : this `requirements.txt` file is just a 4 lines file, let's try no to overthink how to correctly handle it.

AlexXZero:
@JF002 I reopen my old pull request: [PR-994](https://github.com/InfiniTimeOrg/InfiniTime/pull/994). I guess it is what you actually want as 1st solution. So we need to decide if we want to merge this PR or another PR, both PR are look fine for me since they both fix the issue.

PR #1002, opened 2022-02-20
terminal watchface: remove icons and other fixes

Riksu9000:
The battery icon is redundant and the power icon can be replaced by "Charging" text. Also tweaked some label functions.

terminal

PR #1005, opened 2022-02-22
Terminal watchface: replace notifications icon with a text entry

aveeryy:
The notification icon looked a bit weird on the terminal watchface, so i changed it to a text entry like the other stuff.
Here's how it looks on real hardware:
* Without notifications
no-notifications
* With notifications
with-notifications

This PR also changes the "[STAT]" text color to the Bluetooth brand one


trman:
the icon of battery is weird as well on the watchface , you should remove it as well...

aveeryy:
@trman since #1002 already takes care of that, i don't think it's necessary to do it here, since both PRs could be merged, but I don't know a lot about git, so I could be wrong.

Anyways, I agree it looks way better without it

NeroBurner (approved):

Riksu9000:
It could also say "You have mail." at the top, like it might say on login.

trman:
Is it possible to give to the battery 100% value the same color of the icon ?
in fact it would be interesting to change the color of l battery label with :
bright green with 100% to 75%
yellow with 74 to 40% 
and red with 39% to 0%

trman:
> It could also say "You have mail." at the top, like it might say on login.

not bad  idea @Riksu9000  , i would be fitting the pc way 

sadly , i don't thinks that we have enough space  with the login to write it instead of "now" ... :(

so the notification label should be prefered

Riksu9000:
This is what I had in mind. I think it will fit.

```
You have mail.
user@watch:~ $ now
[TIME]01:23:45
...
```

Regarding the battery color, let's open a new PR if that needs changing.

PoorPocketsMcNewHold:
> @Riksu9000 how does this look? If it looks good i'll commit it when i wake up
> youhavemail
> 

It does look great. Just not sure if some might not litterally think it's actually new mail and not an AOL mail notification reference.

Riksu9000:
Looks good to me.

aveeryy:
@Riksu9000 alright, committed it

Riksu9000:
@aveeryy Can you fix the conflicts, so this can be merged?

aveeryy:
@Riksu9000 should be good now

src/displayapp/screens/WatchFaceTerminal.cpp
```diff
@@ -38,10 +38,11 @@ WatchFaceTerminal::WatchFaceTerminal(DisplayApp* app,
 
   connectState = lv_label_create(lv_scr_act(), nullptr);
   lv_label_set_recolor(connectState, true);
+  lv_label_set_text(connectState, "[STAT]#0082fc Disconnected#");
```
Riksu9000, line 41: I removed this line in #1002, because this gets overridden later in the call to `Refresh()`, making this initial text unnecessary. I think it came back during the merging. It should be safe to remove.

Riksu9000 (approved):

trman:
when it will be merged  for 1.90 @JF002 ?
it would near finalize the terminal before 1.9.0 release !

JF002:
Looks good to me, thanks!

PR #1007, opened 2022-02-23
Workout App

PilotOfSparrow:
#### Description

App which helps to track heart rate (max, min, current), longevity of the training session, and setting 2 timers for resting (two timers because one for rest between sets and another for rest between exercises)

#### How does it work

Separate app, which reuses multiple existing controllers (and 'borrows' code from other apps).

User can select from 3 modes: Workout, Timer 1, Timer 2
Workout - user can start and stop workout, when workout started screen sleep are disabled and total workout time is shown in right top corner, heart rate updates are shown.
Timer 1 - start/pause timer 1
Timer 2 - start/pause timer 2

Rest time can be edited by long press on timer mode button. App saves state to FS, so it's relatively safe to exit, progress wouldn't be lost.

#### How it is designed

<details>
<summary>First open</summary>

image

</details>

<details>
<summary>Workout Started</summary>

image

</details>

<details>
<summary>Rest Timer started</summary>

image

</details>

<details>
<summary>Rest Timer edit</summary>

image

</details>

#### Limitations

1. Impossible to disable screen while workout in progress, because it seems like heart rate measurements can only be collected when screen is lit
2. Breaks behavior of Timer App, because timer end event is used to notify user about rest period end. Can be fixed by adding separate event, but it will probably require separate controller (which will heavily 'borrow' from timer controller)
3. Vibration time is very low, despite the value given.
4. Not enough space in the menu, so I had to remove some commented code (which is not cool, so tell me how to preserve it better)
5. After pulling current develop brunch debug build begin to crash with 'Error: Image size (0x73bc8) + trailer (0x630) exceeds requested size 0x74000'. Release build still works fine

I will write the documentation if maintainers give green light to the app.


trman:
is it possible to add a way select the hr frequency (ex : every 1 / 2 /  /5 /15 /30/60 min , or continuous ) per workout?
because if i do i hit training , iwould be more interested in countious hr data ( because each execrcie is less than 5 min ) 
but for for jogging 2 /5 / 15 min freq is not bad ...

RageGamerBoi:
Gadgetbridge has passive run/walk/jog tracking by simply tracking how many steps happen over a time period and determining if you're running or jogging or just walking. This would be great to have integrated onto the watch, as it currently relies too heavily on the real-time step data sent over BLE which is flaky at best (and if you don't have your phone on you it gets even worse...)

This could be combined with different manual workout categories like a lot of other watches on the market, and synced to Gadgetbridge (I think they already have a way to do it here: https://codeberg.org/Freeyourgadget/Gadgetbridge/wiki/Sports-Activities-Workouts but it needs the watch to have ways to track specific workouts)

This concept is a good start, but it's currently mostly a multiple timer app, and dedicated workout tracking is almost necessary for this kind of thing. Some workout-specific tracking options to include are biking, weight lifting, and manual switching to specific jog, run, and sprint modes.

PR #1013, opened 2022-02-26
feat(firmwareupdate): vibrate when the firmware update starts

FrazerClews:
A haptic alert would be very useful for the user before the update process starts


JF002:
Hi and thanks for this PR! 
Could you elaborate why this vibration when the update process starts would be useful?

FrazerClews:
Thanks

My main concern was not getting any sort of alert from the watch when an external device was doing something as important as flashing firmware to the watch. The external device could possibly flash a piece of firmware I did not request, and I would have no way of knowing unless I looked at my watch at the perfect time. It's not a perfect solution, but I feel it is better than nothing.
I know measures such as a confirmation screen on the watch popping up when the firmware update starts would be better, which is something I am experimenting with, but that is still in the works and I would want the InfiniTime team's thoughts on it before I go any further. This patch would work well alongside the confirmation screen too.

In my opinion, it also felt more natural, and it really came in handy when testing new images as I would know when the watch was starting the firmware upgrade. As there were times when it would not initiate for whatever reason.

Riksu9000 (commented):
Whether we end up adding this or not, the vibration should probably be added in SystemTask instead in the message `BleFirmwareUpdateStarted` and then this would be just a single line change.

Also I think 255ms vibration is too much. I would set it to 50 at most.

FrazerClews:
> Whether we end up adding this or not, the vibration should probably be added in SystemTask instead in the message `BleFirmwareUpdateStarted` and then this would be just a single line change.
Thanks for the suggestion @Riksu9000 , changes seem to be working. 

> Also I think 255ms vibration is too much. I would set it to 50 at most.
Personally, I prefer a stronger vibration especially for a task as important as a firmware update. But if most people think 50ms or any other length of time is better then I will make the change.

PR #1026, opened 2022-03-08
Fix SettingChimes cbOption array size

NeroBurner:
There are 3 options, but the array-size is set to 2. This leads to
memory corruption in the initialization of the SettingChimes screen when
assigning the third option object pointer.

Found in https://github.com/InfiniTimeOrg/InfiniSim/issues/11 by @edo0

### DIFF
--- a/src/displayapp/screens/settings/SettingChimes.h
+++ b/src/displayapp/screens/settings/SettingChimes.h
@@ -20,7 +20,7 @@ namespace Pinetime {
       private:
         Controllers::Settings& settingsController;
         uint8_t optionsTotal;
-        lv_obj_t* cbOption[2];
+        lv_obj_t* cbOption[3];
       };
     }
   }


NeroBurner:
@geekbozu @Avamander please review this memory corruption fix. It somehow doesn't crash the firmware, but broke the simulator. Either way, some memory is getting corrupted, which is never good

geekbozu (approved):

NeroBurner:
@JF002 if possible please merge this memory corruption fix before v1.9.0

JF002:
@edo0 @NeroBurner Good catch! Thanks for reporting the issue!

I'll merge this PR to fix the memory corruption issue. However, I think the code could be improved so that static analysis code can detect the issue: In SettingChimes.cpp, the variable [optionsTotal](https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/src/displayapp/screens/settings/SettingChimes.cpp#L43) is useless : it will always be equal to 0, then 1 and then 2. A constant would better show the intent to build an array that always contain the same items.

A static analysis tool should be able to detect the buffer overflow at that point, and I'm not sure why the code didn't trigger any warning.

However, if you explicitly index the array with constant, clang tidy warns about the buffer overflow :
image

NeroBurner:
I was thinking about replacing the C-style array with `std::array`. Then we can use `cbOptions.size()` instead of `optionsTotal`. What's your stance on this?

JF002:
Yep, I would just replace the C-array with an std::array. Then, in the ctor(), just index the array with constant (better: constexpr variable to give a name to those constants). 
And I would just modify `updateSelected` to not use a loop. There're only 3 states, just enable/disable the comboboxes according to those 3 values.

PR #1027, opened 2022-03-09
SettingChimes: replace raw array with std::array

NeroBurner:
~~As suggested in https://github.com/InfiniTimeOrg/InfiniTime/pull/1026#issuecomment-1063394505
replace the C-sytle array with a `std::array` and use constexpr variables to access the array to clarify the accessed combobox.~~

ping @JF002 

---

Use a custom `Option` class containing the chimesOption the the corresponding name to render on the display. Create a `std::array` for the possible chimes options and iterate over that to simplify the code (simple loop instead of hard coding all three possibilities)

### DIFF
--- a/src/displayapp/screens/settings/SettingChimes.cpp
+++ b/src/displayapp/screens/settings/SettingChimes.cpp
@@ -14,6 +14,8 @@ namespace {
   }
 }
 
+constexpr std::array<SettingChimes::Option, 3> SettingChimes::options;
+
 SettingChimes::SettingChimes(Pinetime::Applications::DisplayApp* app, Pinetime::Controllers::Settings& settingsController)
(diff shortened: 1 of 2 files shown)


NeroBurner:
@JF002 please review :)

trman:
@Avamander  maybe be able to review ?
or even @Riksu9000 ?

Riksu9000:
I don't really like the detached indexes. I would instead do it like this 5a2971b064c6f2f751ec5b863a98c0a7af8dbade

NeroBurner:
@Riksu9000 I really like your changes! Just added two changes
- make the `Option` struct a private class, as it isn't needed outside the `SettingChimes` class
- use a `std::array` for the `cbOptions` member

JF002 (approved):
LGTM!

Issue #825, opened 2021-11-11
Music app issues when playing video in Firefox on Android


mruss77:
### What Happened?
The Music app stops responding to screen and button inputs, or sometimes InfiniTime crashes
### What should happen instead?
The Music app should still respond to button and screen input and InfiniTime should not crash
### Reproduction Steps
Using Firefox on Android with Gadgetbridge:

- Go to a page with an embedded video, e.g.: https://boingboing.net/2021/11/10/kitty-is-an-early-realistic-computer-animation-from-1968.html
- Start playing the video
- On PineTime, open the Music app

The music app shows the video title and URL (scrolling since they are long) but no longer responds  to button or screen presses. The watch must be restarted by holding the button. Or sometimes InfiniTime crashes altogether and reboots.

### More Details
I'm on Android 12. The issue only happens with Firefox; watching the same videos in Chrome or the YouTube app does not impact the Music app. It seems these apps also don't send the video metadata to the watch (or at least the Music app doesn't display it).

I see the same issue on other sites with embedded videos, and not just YouTube-hosted videos.

This may be considered low priority since Chrome is the default browser and seems to be used by almost all Android users. I don't generally use the watch to control embedded video playback, but if I accidentally  open Music while video is playing, I'm stuck having to reboot the watch.

### Version
72900ca

### Companion App
Gadgetbridge v0.62.0

Avamander:
InfiniTime probably shouldn't freeze because of this, but Gadgetbridge should avoid causing this as well.

JF002:
I could easily reproduce the issue and analyze it : DisplayTask is not looping anymore because it's stuck in `lv_task_handler`. It seems that processing the refresh of the screen when displaying very long text takes too much time. If this time is > than the refresh time, `lv_task_handler` will simply loop again to refresh the screen again. 
In this case, it's looping indefinitely because the refresh takes too much time.

According to [this](https://github.com/lvgl/lvgl/issues/1839), this is an expected behavior in LVGL7. Even if they agree that it might not be the best behavior, it hasn't been improved in LVGL7 (but it is in 8, see [last comment](https://github.com/lvgl/lvgl/issues/1839#issuecomment-739884326).

While analyzing this issue, I found 2 new issues :
 - The button is not taken into account when DisplayTask consumes too much CPU time. This is probably because the TIMER task has a lower priority than DisplayTask
[...]

PR #1036, opened 2022-03-14
Limit the size of the track and album name received by MusicService.

JF002:
This should work around this bug : https://github.com/InfiniTimeOrg/InfiniTime/issues/825 and prevent heap over-allocation.

### DIFF
--- a/src/components/ble/MusicService.cpp
+++ b/src/components/ble/MusicService.cpp
@@ -47,6 +47,8 @@ namespace {
   constexpr ble_uuid128_t msRepeatCharUuid {CharUuid(0x0b, 0x00)};
   constexpr ble_uuid128_t msShuffleCharUuid {CharUuid(0x0c, 0x00)};
 
+  constexpr uint8_t MaxStringSize {40};
+
   int MusicCallback(uint16_t conn_handle, uint16_t attr_handle, struct ble_gatt_access_ctxt* ctxt, void* arg) {
     return static_cast<Pinetime::Controllers::MusicService*>(arg)->OnCommand(conn_handle, attr_handle, ctxt);
   }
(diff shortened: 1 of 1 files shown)


src/components/ble/MusicService.cpp
```diff
@@ -125,6 +127,11 @@ void Pinetime::Controllers::MusicService::Init() {
 int Pinetime::Controllers::MusicService::OnCommand(uint16_t conn_handle, uint16_t attr_handle, struct ble_gatt_access_ctxt* ctxt) {
   if (ctxt->op == BLE_GATT_ACCESS_OP_WRITE_CHR) {
     size_t notifSize = OS_MBUF_PKTLEN(ctxt->om);
+
+    if(notifSize > MaxStringSize) {
```
Avamander, line 131: Missing space :stuck_out_tongue:

Avamander:
Do you think 40 is enough, even with non-ASCII names? Or should we just look at it when it becomes an issue?

Also when truncated, maybe it should have an ellipsis at the end?

src/components/ble/MusicService.cpp
JF002: Done :)

JF002:
> Do you think 40 is enough, even with non-ASCII names? Or should we just look at it when it becomes an issue?
> 
> Also when truncated, maybe it should have an ellipsis at the end?

According to my tests, the rendering time in LVGL takes too much time with more than 40-45 characters. That's why I chose this value :)

Regarding the ellipsis : yes, why not. How would you do it? Simply replace the last 3 characters by `.`?

Avamander:
> Regarding the ellipsis : yes, why not. How would you do it? Simply replace the last 3 characters by `.`?

If original exceeds 40, yes?

JF002:
> If original exceeds 40, yes?

Yes, of course :)

JF002:
Here is the result with the ellipsis :)

https://user-images.githubusercontent.com/2261652/158887933-ef806548-d8ee-48e8-9eab-4d76e67cde74.mp4

PR #1037, opened 2022-03-14
Fix task priorities to ensure that button presses are correctly detected

JF002:
In current configuration, the timer task (the one from FreeRTOS) has the lowest priority (0). Both display and system tasks are also set on priority 0.

In cases where any other task takes too much time to execute (it can happen in Display Task, see https://github.com/InfiniTimeOrg/InfiniTime/issues/825), the timer task does not have the opportunity to run fast enough to detect and debounce presses on the button.

This commit sets the following priorities:
 - [0] : Display  Task
 - [1] : Timer and System tasks
 - [2] : BLE Host
 - [3] : BLE LL

This way, we ensure that button presses will always be detected, even if the rendering of the display takes a huge amount of time.

### DIFF
--- a/src/FreeRTOSConfig.h
+++ b/src/FreeRTOSConfig.h
@@ -60,7 +60,7 @@
 #define configUSE_TICKLESS_IDLE_SIMPLE_DEBUG    0 /* See into vPortSuppressTicksAndSleep source code for explanation */
 #define configCPU_CLOCK_HZ                      (SystemCoreClock)
 #define configTICK_RATE_HZ                      1024
-#define configMAX_PRIORITIES                    (4)
+#define configMAX_PRIORITIES                    (3)
 #define configMINIMAL_STACK_SIZE                (120)
 #define configTOTAL_HEAP_SIZE                   (1024 * 17)
 #define configMAX_TASK_NAME_LEN                 (4)
--- a/src/FreeRTOSConfig.h
+++ b/src/FreeRTOSConfig.h
@@ -60,7 +60,7 @@
 #define configUSE_TICKLESS_IDLE_SIMPLE_DEBUG    0 /* See into vPortSuppressTicksAndSleep source code for explanation */
 #define configCPU_CLOCK_HZ                      (SystemCoreClock)
 #define configTICK_RATE_HZ                      1024
-#define configMAX_PRIORITIES                    (3)
+#define configMAX_PRIORITIES                    (4)
 #define configMINIMAL_STACK_SIZE                (120)
 #define configTOTAL_HEAP_SIZE                   (1024 * 17)
 #define configMAX_TASK_NAME_LEN                 (4)
@@ -93,7 +93,7 @@
 
 /* Software timer definitions. */
 #define configUSE_TIMERS             1
-#define configTIMER_TASK_PRIORITY    (0)
+#define configTIMER_TASK_PRIORITY    (1)
 #define configTIMER_QUEUE_LENGTH     32
 #define configTIMER_TASK_STACK_DEPTH (300)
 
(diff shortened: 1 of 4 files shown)


src/FreeRTOSConfig.h
```diff
@@ -60,7 +60,7 @@
 #define configUSE_TICKLESS_IDLE_SIMPLE_DEBUG    0 /* See into vPortSuppressTicksAndSleep source code for explanation */
 #define configCPU_CLOCK_HZ                      (SystemCoreClock)
 #define configTICK_RATE_HZ                      1024
-#define configMAX_PRIORITIES                    (3)
+#define configMAX_PRIORITIES                    (4)
```
NeroBurner, line 63: why the change to 4 priority levels? (or do I understand this line wrong?)

JF002: Oops! I forgot to commit 1 file ([src/libs/mynewt-nimble/porting/npl/freertos/src/nimble_port_freertos.c](https://github.com/InfiniTimeOrg/InfiniTime/pull/1037/files#diff-c8bbccd055837f9c9c627160edeccc6dcd4f398ad38c433b6fa2c290bdb7fc51)), sorry for this!

So, we need 4 priorities  : 
 - 0 : Display
 - 1 : System & timer
 - 2 : BLE host
 - 3 : BLE LL
This way, we ensure that BLE tasks priorities are still higher than anything else, and that Display won't prevent system and timer tasks from executing from time to time.

src/libs/mynewt-nimble/porting/npl/freertos/src/nimble_port_freertos.c
```diff
@@ -47,5 +47,5 @@ nimble_port_freertos_init(TaskFunction_t host_task_fn)
      * default queue it is just easier to make separate task which does this.
      */
     xTaskCreate(host_task_fn, "ble", configMINIMAL_STACK_SIZE + 600,
-                NULL, tskIDLE_PRIORITY + 1, &host_task_h);
+                NULL, 2, &host_task_h);
```
Riksu9000, line 50: I'm not sure it's a good idea to hard code or change these nimble tasks. Is it necessary?

NeroBurner (approved):
LGTM, works on my dev-kit, tested the most important apps (Paddle.h and Twos.h ;) )

Riksu9000 (commented):
Even if ButtonHandler can handle the button, it's still DisplayApp's responsibility to read the button event and change screens, so if `lv_task_handler` gets stuck, this won't help escape it.

src/libs/mynewt-nimble/porting/npl/freertos/src/nimble_port_freertos.c
JF002: Before the change, those tasks had priorities 1 and 2. Now, they are 2 and 3.
I hard-coded the constant so it's easier to figure out the actual value of the priority. I can change them to const variables or to values relative to tskIDLE_PRIORITY as it was done before, if it improve the readability?

Riksu9000: I consider the other task to be "max priority" and not any number like 2 or 3, so keeping it `configMAX_PRIORITY - 1` seems more appropriate to me. The host task priority was picked by people working on nimble, and since they don't consider this host task high priority, why change it?

JF002: Fair enough. I set the priority of the Host task higher than the other ones out of habits, but that's probably not mandatory! So I could use only 3 priorities : 0 for display, 1 for system, timer and ble host and 2 for ble LL, then!

PR #1040, opened 2022-03-16
Date formats: Use ISO 8601 to avoid confusion

earboxer:
On the firmware page, the date format was DD/MM/YYYY, which is easily misread by american users as being completely wrong, and read by people who know better as being ambiguous.

(for more information, see ["Date format by country" on Wikipedia](https://en.wikipedia.org/wiki/Date_format_by_country)).

This PR standardizes using YYYY-MM-DD, from ISO 8601, which is internationally unambiguous.

InfiniSim_2022-03-16_180156
InfiniSim_2022-03-16_180145

(note that the "Build date", as given by [`__DATE__`](https://www.cprogramming.com/reference/preprocessor/__DATE__.html) is not changed).


Avamander:
As this is very dependant on users' preferences, it should be a configuration option. As a parallel - most of the world uses Celsius, people in the States use Fahrenheit then a good default isn't going to be Kelvin.

earboxer:
On March 16, 2022 4:21:34 PM EDT, Avamander ***@***.***> wrote:
>As this is very dependant on users' preferences, it should be a configuration option. As a parallel - most of the world uses Celsius, people in the States use Fahrenheit then a good default isn't going to be Kelvin.
>

This is a pretty infrequent screen, I don't think customization would help anyone here. Also, YYYY-MM-DD is not anything like Kelvin, it's easy to understand and universally accepted (as opposed to many of the variations with slashes).

Riksu9000:
For the SystemInfo screens this format is fine. For the terminal watchface it can be debatable. Everywhere else we have months in a written format, and we should keep doing it where possible, so the terminal watchface would be the only place that needed this setting. By using this format, we could avoid adding this option for now, and one could also say it fits the aesthetic. On the other hand, even if it's easier to read than Kelvin, this format does feel weird to me, although I wouldn't use this watchface anyway, so make of that what you will.

trman:
> This is a pretty infrequent screen, I don't think customization would help anyone here. Also, YYYY-MM-DD is not anything like Kelvin, it's easy to understand and universally accepted (as opposed to many of the variations with slashes).

sorry but  :

first :  the yy-mm-dd is already implemented but when the time is in not in 24h (aka with pm  )   that is coherent because 
the time written with pm is  for english country (usa / England)

second : as written in the link you provided , aka https://en.wikipedia.org/wiki/Date_format_by_country
, YYYY-MM-DD is **not** unversally  accepted ! 

in fact if you see thz grid by country , the one with header All-numeric date form , it's the  DMY format that is 
universaly accepted and in use in more 70 % of country of the world !

NeroBurner (approved):

JF002:
I don't think we'll ever agree on a single date format : each of us is used to using a specific format and will find other formats unusable, even if this format is standardized by ISO.
In my part of the world, we use DD/MM/YYYY, and that's probably why most of the dates are formatted this way for now. Changing it to MM/DD/YYYY or YYYY/DD/MM will just feel strange to me. 

In the case of the Terminal watchface, the time/date format is a design choice from [the author of the watchface](https://github.com/InfiniTimeOrg/InfiniTime/pull/932#discussion_r784509042).

I would suggest we try to find a more generic solution to this date format? Should we just add a setting (just like there's already a setting for 12/24h, for example)?

13werwolf13:
let me enter this discussion.
I set the yyyy.mm.dd format because and only because I personally like it, it is in this form that I use it wherever possible.
if everyone thinks that this is wrong and you need to replace the points with the hyphens, then I can understand this. but the use of a slash in the terminal (as in something that looks like a terminal) seems completely wrong to me and at least requires the addition of quotation marks.
but I personally consider the change in the yyyy-mm-dd format to the dd-mm-yyyy format a mistake. I live in a country where it is customary to write dd.mm.yyy and it seemed strange to me from my since childhood, especially when it goes in one line with time. first we write from less to more (from day to year) and then from more to less (from hour to second). I think this is wrong and if time goes down then the date should go the same.
P.S.: I hope my level of English and the possibilities of an online translator did not make this text incomprehensible.

Riksu9000 (approved):
For the terminal watchface this is only changing the dot to a hyphen which makes it less ambiguous. For SystemInfo localizing the format isn't important. The current format is very ambiguous since in the US they also use slashes between the digits, but swap the months and the days, thus changing this is an improvement.

Please fix the conflicts, so we can merge this if we ever come to an agreement.

JF002 (approved):

Alias001:
I know it's closed already but I like the ISO 8601 standard.  It makes file organization a snap.  Say you want the pinetime to write a timestamp on some future app like a sleep tracking app or step tracking, etc.  I think the use of ISO standard would better enable cross development and companion app, or watch app functions.
Personally I like to use the ISO standard 2022-10-03T20:16:32Z for most anything I do.  It's easier for me to think linear in my time and I work at an airport so UTC is an easy swap for me to local time.  As for the date on the Pinetime I wish it had the option for ISO 8601 standard in any watch face.

PR #1042, opened 2022-03-20
Extract thread safe functions from the DisplayApp task

Riksu9000:
These functions can be executed directly, instead of queueing them up for the DisplayApp task to execute.
Simplifies the DisplayApp task by removing unnecessary messages from it.

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -194,16 +194,6 @@ void DisplayApp::Refresh() {
         brightnessController.Restore();
         state = States::Running;
         break;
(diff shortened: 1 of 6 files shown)


Riksu9000:
I'll have to fix DisplayAppRecovery. In DispAppRecovery, `UpdateBleConnection` message should display a logo, but I can't find anywhere where this message is pushed. Is it unused?

Riksu9000:
DisplayApp being interchangeable with DisplayAppRecovery is problematic, since it limits the amount of public methods that they can reasonably have. I think the recovery firmware needs to be reworked before these kinds of changes can be made.

PR #1043, opened 2022-03-20
Global color picker

kieranc:
This PR implements a color and opacity selector for the settings menu and app drawer icons, and also uses the app drawer icon color for the buttons in the Music app. It's based on the PTS color picker code, which I think is fairly well tested by now, and adds about 1.5kB to the overall firmware size.
The opacity is stored as an integer, 51,102,153,204,255, rather than LV_OPA values, but they line up exactly so it didn't seem worth the effort of complicating it further. Black and white are skipped in the color selection to avoid confusion/invisible buttons/text.

Peek 2022-03-20 13-55

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -450,6 +450,7 @@ list(APPEND SOURCE_FILES
         displayapp/screens/settings/SettingChimes.cpp
         displayapp/screens/settings/SettingShakeThreshold.cpp
         displayapp/screens/settings/SettingAirplaneMode.cpp
+        displayapp/screens/settings/SettingColor.cpp
 
         ## Watch faces
         displayapp/icons/bg_clock.c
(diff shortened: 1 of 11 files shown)


Riksu9000:
The UI you've created looks great, but unfortunately I think there's an issue with this idea. I recently explained why I believe adding too much customization isn't a good idea https://github.com/InfiniTimeOrg/InfiniTime/pull/1030#issuecomment-1070866569. A color picker like this is better, but the issue I see is rather that these screens probably shouldn't be colored in the first place. I think different apps should look more distinct, with unique icons or different colors for example, and I don't think the settings screen should be so colorful. It should probably be more grayish and maybe with colored icons, and at that point customizing just a settings screen wouldn't be so interesting.

I do like the idea of an accent color picker, but I don't think it should affect the app icons or the settings entries.

kieranc:
I confess I hadn't seen #1030, if nothing more this can serve as a basis for a color picker for that PR. I didn't really make it as a perfect long term solution to theming Infinitime but because I was annoyed at my watchface colors clashing with the menu and this was a solution within my abilities :-)

joseph58tech:
I feel like that would use to much ram

kieranc:
Closing in preference of the global theming efforts

PR #1044, opened 2022-03-20
Replace lv_label_set_text where possible

Riksu9000:
Improvements from #521, without conflicts.

### DIFF
--- a/src/displayapp/screens/Music.cpp
+++ b/src/displayapp/screens/Music.cpp
@@ -62,7 +62,7 @@ Music::Music(Pinetime::Applications::DisplayApp* app, Pinetime::Controllers::Mus
   lv_obj_align(btnVolDown, nullptr, LV_ALIGN_IN_BOTTOM_LEFT, 0, 0);
   lv_obj_add_style(btnVolDown, LV_STATE_DEFAULT, &btn_style);
   label = lv_label_create(btnVolDown, nullptr);
-  lv_label_set_text(label, Symbols::volumDown);
+  lv_label_set_text_static(label, Symbols::volumDown);
   lv_obj_set_hidden(btnVolDown, true);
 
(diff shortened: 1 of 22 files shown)


src/displayapp/screens/Music.cpp
```diff
@@ -211,31 +211,27 @@ void Music::Refresh() {
...
-    char timer[12];
-    sprintf(timer,
-            "%02d:%02d/%02d:%02d",
-            (currentLength / 60) % 100,
-            (currentLength % 60) % 100,
-            (totalLength / 60) % 100,
-            (totalLength % 60) % 100);
-    lv_label_set_text(txtTrackDuration, timer);
+    lv_label_set_text_fmt(txtTrackDuration,
+                          "%02d:%02d/%02d:%02d",
```
NeroBurner, line 230: non-blocking, just personal preference and curiosity: I would indent the lines with just one more indent than the parent function. Do we have a style guideline to match the indent to the function length? I personally like the "one indent more" approach better, because the formatting of the arguments doesn't need to change when the function name changes. Like it is currently all the arguments would have to be touched when the function-name is changed

NeroBurner (approved):
thanks for the nice cleanup! Tested in the simulator, working great!

src/displayapp/screens/Music.cpp
Avamander: > Do we have a style guideline to match the indent to the function length?

Is this something that `clang-format` is not fixing?

Riksu9000: Running `clang-format` will set the indentation like this, so that's the style guideline currently.

NeroBurner: ah ok, thanks for the info

Avamander (approved):

PR #1050, opened 2022-03-25
Add initial P8b smartwatch support

StarGate01:
This PR extends the existing P8 smartwatch support (eg. from https://github.com/InfiniTimeOrg/InfiniTime/commit/e614af1c4a53e943256b3b2acee4bfe7d7204bbf) to better support more recent models of the P8 series; the P8b. This is a feature desired by the community (see https://github.com/InfiniTimeOrg/InfiniTime/issues/62).

The P8 smartwatches and the Pinetime have a few small differences:
- Slightly different pins are used
- Some variants have a `SC7A20` acceleration sensor, some have a `BMA421` (like the Pinetime)
- Some variants have a `CST716` touch screen sensor, some have a `CST816S` (like the Pinetime)
- Some variants have no external low frequency crystal
- Different SPI flash chips are used (Not relevant as no whitelist is enforced in InfiniTime)

I added support for the `SC7A20` sensor based on the work by @hubmartin and @ildar at https://github.com/ildar/InfiniTime/tree/p8 . A CMake variable (`DRIVER_ACC`) can be used to select which sensor is used at compile time. The acceleration driver code was split into an generic interface definition and two hardware-specific implementations.

In addition, the touch driver now detects which touch sensor chip is used, and adjusts its logic accordingly. The `CST716` does not support an automatic sleep state, and cannot wake from the explicit manual sleep state. On these devices, the `SC7A20` accelerometer is configured to recognize tap and double-tap events. The `CST816S` is not explicitly put to sleep, instead it uses its auto-sleep functionality. On my Pinetime dev kit, I had some issues concerning the initialization of the touch sensor after a reset, so I implemented a small workaround for variants with that issue as well.

It must be said that while this improved touch driver works great on both my P8b watch as well as on my Pinetime dev kit, there are older P8(a) variants which might have incompatible touch sensor configurations. The  `CSTX16` series can be used in two modes - reporting mode and gesture mode - which control when and how often an event is reported to the main CPU. The InfiniTime code currently assumes that both the `CST816S` as well as the `CST716` are configured in reporting mode. The `CST816S` is able to switch at runtime to this mode, which is what the code does. The `CST716` can only be configured once in the factory. In my P8b it came in reporting mode, however there might be other P8 variant with a (yet) incompatible `CST716` in gesture mode. However, I do not have such a watch available, nor have I found a vendor that still sells these (reliably). Adding support for a sensor in gesture mode should be done in a future PR.

[...]

PR #1051, opened 2022-03-25
CMake: Improve debug configuration

StarGate01:
This PR has been broken out of https://github.com/InfiniTimeOrg/InfiniTime/pull/1050.

- Extended debug configuration
- ~~Use correct linker script in RECOVERY_MCUBOOT~~ merged via #256
- ~~Use iHEX format for mcuboot image creation~~ merged via #256

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,7 @@
 cmake_minimum_required(VERSION 3.10)
+
+set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose Debug or Release")
+
 project(pinetime VERSION 1.9.0 LANGUAGES C CXX ASM)
 
 set(CMAKE_C_STANDARD 99)
@@ -70,8 +73,10 @@ string(STRIP "${PROJECT_GIT_COMMIT_HASH}" PROJECT_GIT_COMMIT_HASH)
 
 message("PROJECT_GIT_COMMIT_HASH_SUCCESS? " ${PROJECT_GIT_COMMIT_HASH_SUCCESS})
 
+message("")
 message("BUILD CONFIGURATION")
 message("-------------------")
+message("    * Mode : " ${CMAKE_BUILD_TYPE})
 message("    * Version : " ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH})
 message("    * Toolchain : " ${ARM_NONE_EABI_TOOLCHAIN_PATH})
 message("    * GitRef(S) : " ${PROJECT_GIT_COMMIT_HASH})
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -9,21 +9,21 @@ set(NRF_BOARD pca10040)
 
 # check if all the necessary tools paths have been provided.
 if (NOT NRF5_SDK_PATH)
-    message(FATAL_ERROR "The path to the nRF5 SDK (NRF5_SDK_PATH) must be set.")
+  message(FATAL_ERROR "The path to the nRF5 SDK (NRF5_SDK_PATH) must be set.")
 endif ()
 if (DEFINED ARM_NONE_EABI_TOOLCHAIN_PATH)
-    set(ARM_NONE_EABI_TOOLCHAIN_BIN_PATH ${ARM_NONE_EABI_TOOLCHAIN_PATH}/bin)
+  set(ARM_NONE_EABI_TOOLCHAIN_BIN_PATH ${ARM_NONE_EABI_TOOLCHAIN_PATH}/bin)
 endif ()
 
 if (NOT NRF_TARGET MATCHES "nrf52")
-    message(FATAL_ERROR "Only rRF52 boards are supported right now")
+  message(FATAL_ERROR "Only rRF52 boards are supported right now")
 endif ()
 
 # Setup toolchain
 include(${CMAKE_SOURCE_DIR}/cmake-nRF5x/arm-gcc-toolchain.cmake)
 
 if (NOT DEFINED ARM_GCC_TOOLCHAIN)
-    message(FATAL_ERROR "The toolchain must be set up before calling this macro")
+  message(FATAL_ERROR "The toolchain must be set up before calling this macro")
 endif ()
 set(CMAKE_OSX_SYSROOT "/")
 set(CMAKE_OSX_DEPLOYMENT_TARGET "")
@@ -934,11 +934,11 @@ add_custom_command(TARGET ${EXECUTABLE_MCUBOOT_NAME}
         )
 
 if(BUILD_DFU)
-    add_custom_command(TARGET ${EXECUTABLE_MCUBOOT_NAME}
-            POST_BUILD
-            COMMAND adafruit-nrfutil dfu genpkg --dev-type 0x0052 --application ${IMAGE_MCUBOOT_FILE_NAME} ${DFU_MCUBOOT_FILE_NAME}
-            COMMENT "post build (DFU) steps for ${EXECUTABLE_MCUBOOT_FILE_NAME}"
-            )
+  add_custom_command(TARGET ${EXECUTABLE_MCUBOOT_NAME}
+          POST_BUILD
+          COMMAND adafruit-nrfutil dfu genpkg --dev-type 0x0052 --application ${IMAGE_MCUBOOT_FILE_NAME} ${DFU_MCUBOOT_FILE_NAME}
+          COMMENT "post build (DFU) steps for ${EXECUTABLE_MCUBOOT_FILE_NAME}"
+          )
 endif()
 
 # InfiniTime recovery firmware (autonomous)
@@ -1003,11 +1003,11 @@ add_custom_command(TARGET ${EXECUTABLE_RECOVERY_MCUBOOT_NAME}
         )
 
 if(BUILD_DFU)
-    add_custom_command(TARGET ${EXECUTABLE_RECOVERY_MCUBOOT_NAME}
-            POST_BUILD
-            COMMAND adafruit-nrfutil dfu genpkg --dev-type 0x0052 --application ${IMAGE_RECOVERY_MCUBOOT_FILE_NAME} ${DFU_RECOVERY_MCUBOOT_FILE_NAME}
-            COMMENT "post build (DFU) steps for ${EXECUTABLE_RECOVERY_MCUBOOT_FILE_NAME}"
-            )
+  add_custom_command(TARGET ${EXECUTABLE_RECOVERY_MCUBOOT_NAME}
+          POST_BUILD
+          COMMAND adafruit-nrfutil dfu genpkg --dev-type 0x0052 --application ${IMAGE_RECOVERY_MCUBOOT_FILE_NAME} ${DFU_RECOVERY_MCUBOOT_FILE_NAME}
+          COMMENT "post build (DFU) steps for ${EXECUTABLE_RECOVERY_MCUBOOT_FILE_NAME}"
+          )
 endif()
 
 # Build binary that writes the recovery image into the SPI flash memory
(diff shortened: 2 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -818,13 +818,43 @@ add_definitions(-DFREERTOS)
...
-#add_definitions(-DDEBUG)
-#add_definitions(-DDEBUG_NRF_USER)
-
-if (NOT CMAKE_BUILD_TYPE)
-    set(CMAKE_BUILD_TYPE "Release")
-endif ()
+# Target hardware configuration options
+add_definitions(-DTARGET_DEVICE_${TARGET_DEVICE})
+if(LF_CLK STREQUAL "RC")
+        add_definitions(-DCLOCK_CONFIG_LF_SRC=0)
```
NeroBurner, line 824: in the top level CMakeLists.txt we have 2 spaces per tab, this file has sometimes 4 spaces per tab and for file indentations 8 spaces per tab. The C++ source has 2 spaces per tab. The doc/coding-convention.md file also specifies 2 spaces for indentation.

I think it would be a good idea to use 2 spaces per tab for new/changed code

src/CMakeLists.txt
```diff
@@ -924,7 +954,7 @@ add_custom_command(TARGET ${EXECUTABLE_NAME}
 # Build binary intended to be used by bootloader
 set(EXECUTABLE_MCUBOOT_NAME "pinetime-mcuboot-app")
 set(EXECUTABLE_MCUBOOT_FILE_NAME ${EXECUTABLE_MCUBOOT_NAME}-${pinetime_VERSION_MAJOR}.${pinetime_VERSION_MINOR}.${pinetime_VERSION_PATCH})
-set(IMAGE_MCUBOOT_FILE_NAME ${EXECUTABLE_MCUBOOT_NAME}-image-${pinetime_VERSION_MAJOR}.${pinetime_VERSION_MINOR}.${pinetime_VERSION_PATCH}.bin)
+set(IMAGE_MCUBOOT_FILE_NAME ${EXECUTABLE_MCUBOOT_NAME}-image-${pinetime_VERSION_MAJOR}.${pinetime_VERSION_MINOR}.${pinetime_VERSION_PATCH}.hex)
```
NeroBurner, line 957: why this change of the extension from `bin` to `hex`?

NeroBurner (changes requested):

src/CMakeLists.txt
StarGate01: The hex format stores the target offset as well, which enables flashing without explicitly having to specify into which memory region (usually `0x8000` for mcuboot images). This is especially useful because the nrfjprog tool does not support specifying manual offsets. 

For testing the bootloader and main firmware interaction, I often leave the bootloader on the device and only flash a new mcuboot image (via SWD), instead of erasing the entire flash memory.

In addition, tools like `mergehex` operate on the hex format as well.

Does the bin format offer any advantages? Or are there other disadvantages of the hex format in the current toolchain?

src/CMakeLists.txt
StarGate01: Okay, I adjusted the file to 2 spaces per indentation.

CMakeLists.txt
```diff
@@ -1,4 +1,11 @@
 cmake_minimum_required(VERSION 3.10)
+
+if(DEFINED CMAKE_BUILD_TYPE)
+   set(CMAKE_BUILD_TYPE ${CMAKE_BUILD_TYPE} CACHE STRING "Choose Debug or Release")
+else()
+   set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose Debug or Release")
```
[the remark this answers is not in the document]
StarGate01: Thanks for the tip, I changed it.

src/CMakeLists.txt
```diff
@@ -1089,81 +1120,82 @@ add_custom_command(TARGET ${EXECUTABLE_MCUBOOT_RECOVERYLOADER_NAME}
...
+  add_custom_command(TARGET ${EXECUTABLE_MCUBOOT_RECOVERYLOADER_NAME}
+          POST_BUILD
+          COMMAND adafruit-nrfutil dfu genpkg --dev-type 0x0052 --application ${IMAGE_MCUBOOT_RECOVERYLOADER_FILE_NAME} ${DFU_MCUBOOT_RECOVERYLOADER_FILE_NAME}
+          COMMENT "post build (DFU) steps for ${EXECUTABLE_MCUBOOT_RECOVERYLOADER_FILE_NAME}"
+          )
 endif()
 
 
 # FLASH
 if (USE_JLINK)
```
StarGate01, line 1139: The diff looks a bit odd, I only corrected the indentation here as well.

src/CMakeLists.txt
StarGate01: Original source of the commit comes from at https://github.com/ildar/InfiniTime/tree/p8 at https://github.com/ildar/InfiniTime/commit/1a4798f816f9d38c4da1efb9a8e37253511fa6f5 .

NeroBurner: This seems like a valid reasoning, unfortunately this is not my expertise. Maybe @JF002 @Avamander @Riksu9000 have more insight to this bin to hex change

doc/buildAndProgram.md
```diff
@@ -35,7 +35,9 @@ CMake configures the project according to variables you specify the command line
 **GDB_CLIENT_BIN_PATH**|Path to arm-none-eabi-gdb executable. Used only if `USE_GDB_CLIENT` is 1.|`-DGDB_CLIENT_BIN_PATH=/home/jf/nrf52/gcc-arm-none-eabi-9-2019-q4-major/bin/arm-none-eabi-gdb`
 **GDB_CLIENT_TARGET_REMOTE**|Target remote connection string. Used only if `USE_GDB_CLIENT` is 1.|`-DGDB_CLIENT_TARGET_REMOTE=/dev/ttyACM0`
 **BUILD_DFU (\*\*)**|Build DFU files while building (needs [adafruit-nrfutil](https://github.com/adafruit/Adafruit_nRF52_nrfutil)).|`-DBUILD_DFU=1`
-**WATCH_COLMI_P8**|Use pin configuration for Colmi P8 watch|`-DWATCH_COLMI_P8=1`
+**TARGET_DEVICE**|Target device, used for pin map. Allowed: `PINETIME, P8`|`-DTARGET_DEVICE=PINETIME` (Default)
+**LF_CLK**|Configures the LF clock source. Allowed: `RC, XTAL, SYNT`|`-DLF_CLK=XTAL` (Default)
+**DRIVER_ACC**|Acceleration sensor driver selection. Allowed: `BMA421, SC7A20`|`-DDRIVER_ACC=BMA421` (Default)
```
Riksu9000, line 40: I think the entire hardware configuration should just be based on the target device instead of separate options. I find them unnecessary and it allows for configurations that don't exist. Also what are the LF clock sources for?

StarGate01:
I rebased this PR on the latest development version, and removed any hardware target configuration options. These will be moved to #1050 .

src/CMakeLists.txt
```diff
@@ -799,9 +799,27 @@ add_definitions(-D__HEAP_SIZE=4096)
 #add_definitions(-DDEBUG)
 #add_definitions(-DDEBUG_NRF_USER)
 
-if (NOT CMAKE_BUILD_TYPE)
-    set(CMAKE_BUILD_TYPE "Release")
-endif ()
+# Debug configuration
+if (${CMAKE_BUILD_TYPE} STREQUAL "Debug")
+  add_definitions(-DDEBUG)
+  add_definitions(-DDEBUG_NRF_USER)
```
Riksu9000, line 805: The commented debug options can be removed, since it's controlled below.

Riksu9000 (commented):
The bin to hex change is blocking this currently, since there's concern that it may break compatibility with some companion apps. #256 also has this same change, so we'll reconsider that first.

src/CMakeLists.txt
StarGate01: > The bin to hex change is blocking this currently, since there's concern that it may break compatibility with some companion apps. #256 also has this same change, so we'll reconsider that first.

The `objcopy` command is invoked by the CMake script as well, such that both the `bin` as well as the `hex` version are created. The binary file is exactly the same as before. Or are there concerns of tools depending on the inner working of the CMake compilation chain?

Riksu9000:
>>The bin to hex change is blocking this currently, since there's concern that it may break compatibility with some companion apps. https://github.com/InfiniTimeOrg/InfiniTime/pull/256 also has this same change, so we'll reconsider that first.
>
>The objcopy command is invoked by the CMake script as well, such that both the bin as well as the hex version are created. The binary file is exactly the same as before. Or are there concerns of tools depending on the inner workings of the CMake compilation chain?

Maybe something about the DFU. I'm not entirely sure. Someone else will need to take a closer look at it.

StarGate01:
The image creating tool `tools/mcuboot/imgtool.py` accepts both bin as well as hex files to create a DFU package.

JF002:
I unfortunately do not have much time to think about those changes right now. I'm currently working on improving the memory usage and monitoring, and I do not have much brain bandwidth for other things :)

This PR is not easy to read as it serves multiples goals (change the debug configuration behavior, switch from bin to hex, change indentation,...). 

As I am the one who wrote the original script, I decided to use .bin file because that's what I'm used to use... I'm in full control of where the data will be flashed in memory and that fine for me. I know other people prefer the hex file because they don't have to worry about the offset in memory. I don't think one is better than the other, they are just different.

Now, other than this practical aspect, do we have other valid reasons to change how the files (firmwares, DFUs, mcuboot images) are generated?
Also, we need to pay attention to not break any companion app. I'm not sure all of them rely on the NRF library to parse the DFU file. I think at least Amazfish reads the .bin file directly for the OTA functionality.

This PR also changes the behavior of the `Debug` configuration by settings new variables. Not every developer will need them. Also, they are already defined in `sdk_config.h`. I'm wondering if these duplicate declarations/definitions won't be confusing...

StarGate01:
@JF002 Thanks for your feedback. I see that there are valid concerns about third party tools. How about I revert the process of the DFU generation back to its original state, but add a `objcopy` statement (unrelated to the DFU creation) to generate a hex file from the bin file in addition?

StarGate01:
The hex format change as well as the linker fix included in this PR are the same as in #256, since my P8 fork is based in the fork by @ildar . This is also noted in the commit messages.

src/CMakeLists.txt
```diff
@@ -787,13 +787,27 @@ add_definitions(-DFREERTOS)
...
+
+  # NRF SDK Logging
+  add_definitions(-DNRF_LOG_ENABLED=1)
```
JF002, line 796: Is this a good idea to define constants that might also be defined in other configuration files ([sdk_config.h](https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/src/sdk_config.h) and [syscfg.h](https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/src/libs/mynewt-nimble/porting/nimble/include/syscfg/syscfg.h)) ?
In my experience, that can be confusing, especially when those constants are defined twice with different values. What do you think?

JF002 (commented):
@StarGate01 I've finally taken the time to check a few things : the .bin and .hex files are correctly generated, and the DFU file still contains a .bin file. So I guess my concerns were unfounded. Sorry about that!

src/CMakeLists.txt
StarGate01: As far as I understand these files, they only provide default fallback values for variables which are not explicit defined (note the `ifndef` guards).

Riksu9000: For Nimble at least I definitely think they should be defined in our files instead of "modifying the lib" as I see it. That way it's easier to see the configuration changes and maybe use upstream. We could remove the copy of sdk_config.h if we did the same with it, since from the 12000 lines only a couple dozen at most are changed as far as I can tell, making it largely redundant.

StarGate01: Isn't the `sdk_config.h` file required by the NRF SDK? Does it supply a default one?

Riksu9000: The file's copied from config/nrf52832/config/sdk_config.h in the SDK. I'm assuming it could be used somehow.

StarGate01: Since the SDK specifically looks for a file named `sdk_config`, we could add `config/nrf52832/config/` to the include path, or put just `#include "config/nrf52832/config/sdk_config.h"` into our `sdk_config.h`.

JF002: > As far as I understand these files, they only provide default fallback values for variables which are not explicit defined (note the ifndef guards).

> For Nimble at least I definitely think they should be defined in our files instead of "modifying the lib" as I see it. That way it's easier to see the configuration changes and maybe use upstream. We could remove the copy of sdk_config.h if we did the same with it, since from the 12000 lines only a couple dozen at most are changed as far as I can tell, making it largely redundant.

Those are good points! I guess we need to ensure that default fallback definitions contain valid values, or to properly override them in this file.

JF002:
@Riksu9000 @StarGate01 Which PR should we merge? This one or #256?

Riksu9000:
@JF002 I would merge #256 first, but it doesn't make a big difference in the end.

src/CMakeLists.txt
Riksu9000: Sounds good. Let's work on that in another PR, so as not to delay merging of this PR.

JF002:
I've just merged #265 ;-)
But I guess the remaining changes in this PR still make sense? If yes, I'll merge it too !

NeroBurner:
@JF002 I think you mean #256 instead of #265

NeroBurner (approved):
I'm not using on device debugging, so I can't comment on that.

But the changes are simple and seem to be sound

Avamander (approved):
Looks better than before, I think.

JF002 (approved):

JF002:
Thanks for this PR, @StarGate01 ! Looks good to me :-)

By the way, as you've touched the flash commands in the CMake file, may I ask your opinion about #801 ? It removes those custom targets as I think they are not generic/portable enough. I haven't dared to merge it as I'm afraid to confuse some developers who might use them.

PR #1053, opened 2022-03-27
Fix display corruption when the timer is triggered

### LINKED ISSUE #445: Stopwatch+timer bug

JF002:
Fixes #445
This display was "corrupted" because the app do not have a background label that cover the whole screen. Because of that, LVGL would not refresh the whole display at once, but would refresh it in 3 non consecutive parts, which confused the HW scrolling transition.


Riksu9000 (approved):
Everything looks good to me.

PR #1054, opened 2022-03-27
Fix app loading and event management in DisplayApp

### LINKED ISSUE #1023: Long pressing paint app moves the screen
### Verification

### What happened?

When long the paint app is long pressed it moves the screen a bit up and shows content under

It should open the app

### Reproduction steps

Scroll through the app menu and long press the paint app

### Version

1.8.0

### Companion app

_No response_

JF002:
In display app, call the event handler of the current app before loading the new one. This way, we ensure that lv_task_refresh() is called before sending event to the newly loaded app.

Fix #1023

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -306,13 +306,13 @@ void DisplayApp::Refresh() {
     }
   }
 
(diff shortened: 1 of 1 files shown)


Riksu9000 (approved):
This function is only used in paint and paddle and both work perfectly.

PR #1057, opened 2022-03-28
Update track progress in MusicService

### LINKED ISSUE #127: Broken song current length when re-entering music app.
Pinetime: 0.9.0-dev
Gadgetbridge: 0.48
Android: 10

If you start playing from music app, then click side button and enter music display again, the current playing time will be wrong. And it is stays still.
If display will be turned off with timeout and turned on with button, the time will be shown properly.

It looks like this time the same as uptime:

<a href="url"></a>

<a href="url"></a>

Riksu9000:
Update the progress in the music service, instead of the app.
I'm unsure if this is the best way to fix it, and this probably warrants more thorough testing, but in my limited testing everything seems to work correctly.
Fixes #127

### DIFF
--- a/src/components/ble/MusicService.h
+++ b/src/components/ble/MusicService.h
@@ -47,7 +47,7 @@ namespace Pinetime {
 
       std::string getAlbum() const;
 
-      int getProgress() const;
+      int getProgress();
 
       int getTrackLength() const;
 
@@ -81,6 +81,7 @@ namespace Pinetime {
       int trackLength {0};
       int trackNumber {};
       int tracksTotal {};
+      TickType_t trackProgressUpdateTime {0};
 
       float playbackSpeed {1.0f};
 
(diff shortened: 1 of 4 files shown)


Avamander:
Yep this makes sense as the app might be closed by some users.

Avamander (approved):

src/components/ble/MusicService.h
```diff
@@ -47,7 +47,7 @@ namespace Pinetime {
 
       std::string getAlbum() const;
 
-      int getProgress() const;
+      int getProgress();
```
JF002, line 50: Any reason why this method is not `const` anymore?

Riksu9000: Nope, it was a leftover change :) thanks.

JF002 (approved):

PR #1070, opened 2022-04-04
Change settings to use CBOR

Arsen6331:
This PR makes the InfiniTime settings file use CBOR rather than saving as a binary file. This is useful for companion apps to be able to change settings on the watch.

This is my first time using C++, and I've not used C for more than a couple months a while ago, so my code most likely has some issues. If someone points them out, I'd be happy to fix them.

Currently, this is able to automatically migrate the old settings to the new CBOR settings file, but after that, it crashes. I do not have a devkit so I don't know why it does that. I would appreciate it if someone tested this on a devkit and helped me fix it.

This will probably destroy `settings.dat` in its current state, so back that up if you want to keep it.


NeroBurner:
lfs_info.size is only defined for REG files

```cpp
    // Size of the file, only valid for REG files. Limited to 32-bits.
    lfs_size_t size;
```

I've no idea what a REG file is in the context of littlefs, but maybe this can cause the crash if the `size` isn't set and thus either a random value or 0

I've fixed InfiniSim locally and the cbor file is created and still works after a restart. I'll commit and update my InfiniSim fixes as soon as possible

Arsen6331:
> lfs_info.size is only defined for REG files

Interesting. I wonder if there is a better way to obtain the size of the file.

NeroBurner:
currently (before this PR) the filesize was implicitly known by the binary size of the settings struct. The bits at the position where the file should be is read into memory and the read version number is compared to the current one. The read in memory is only used if the version numbers do match. This way a filesize was never actually needed

Arsen6331:
> currently (before this PR) the filesize was implicitly known by the binary size of the settings struct. The bits at the position where the file should be is read into memory and the read version number is compared to the current one. The read in memory is only used if the version numbers do match. This way a filesize was never actually needed

Yes, I saw that. Unfortunately, there is no easy way to calculate the size of the CBOR data, which is why I need to get the file size so I can make an appropriately-sized buffer for it. If the file size being 0 is the issue, then I may need to find a different way to get the size, or as a last resort, store the size when I write the file and then read it when I decode.

Arsen6331:
I looked in the LFS source code, and it seems there are only two file types: REG and DIR. Since the file is not a directory, it is a REG file, so the size should be set.

Arsen6331:
It appears that my code does work. I believe the crash may be a RAM issue. I tried applying some of the changes from #911, and it no longer crashed, but it was no longer responsive to touch, the side button, or BLE, and it never went to sleep. However, the clock was still changing and the screen dimmed after a delay, so something was happening.

When I flashed the develop branch back onto the watch, I checked the FS and the settings were successfully migrated to CBOR, so my actual code works. I have no idea how to further debug this.

src/components/settings/Settings.cpp
```diff
@@ -23,25 +32,228 @@ void Settings::SaveSettings() {
...
+  SettingsData bufferSettings;
+  lfs_file_t settingsFile;
+
+  if (fs.FileOpen(&settingsFile, "/settings.dat", LFS_O_RDONLY) != LFS_ERR_OK) {
+    return;
+  }
+  fs.FileRead(&settingsFile, reinterpret_cast<uint8_t*>(&bufferSettings), sizeof(settings));
+  fs.FileClose(&settingsFile);
+
+  settings = bufferSettings;
```
NeroBurner, line 254: You're missing the settingsVersion check when migrating the settings. This could lead to random data read in as Settings, when upgrading from a firmware with changed settings-struct and version. This leads to MUCH wonkiness
```suggestion
  if ( bufferSettings.version == settingsVersion ) {
    settings = bufferSettings;
  }
```

src/components/settings/Settings.cpp
```diff
@@ -23,25 +32,228 @@ void Settings::SaveSettings() {
...
+
+  SaveSettingsToFile();
+
+  fs.FileDelete("/settings.dat");
+}
```
NeroBurner, line 259: please add a newline at the end of source files
```suggestion
}

```

NeroBurner (changes requested):

Avamander:
I have only two questions.

What happens if the settings file is corrupt? On decode errors it returns but how does it recover?

What happens if a companion app writes a value out of bounds, the code casts, but is that sufficient to avoid values that cause major issues elsewhere? Say watchface selection, due to OOB reads/writes?

src/components/settings/Settings.cpp
```diff
@@ -7,6 +11,11 @@ using namespace Pinetime::Controllers;
 Settings::Settings(Pinetime::Controllers::FS& fs) : fs {fs} {
 }
 
+void CleanUpQcbor(QCBORDecodeContext* decodeCtx) {
```
Avamander, line 14: Maybe you wish to split it off to some Util class so that both Weather and Settings can use it?

Arsen6331:
> What happens if the settings file is corrupt? On decode errors it returns but how does it recover?

If I understand how this works correctly, then since the settings are not replaced with the bufferSettings, it should use the default settings as defined in the [SettingsData struct](https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/src/components/settings/Settings.h#L217) until the next time SaveSettingsToFile() is called.

> What happens if a companion app writes a value out of bounds, the code casts, but is that sufficient to avoid values that cause major issues elsewhere? Say watchface selection, due to OOB reads/writes?

I will add an if statement after everything where this could cause an issue to check that the value is valid and only set it when it is.

src/components/settings/Settings.cpp
Arsen6331: That's probably a good idea, but I have no idea how to add a new class to the code and use it from both places. As I mentioned, I do not know C++ very well.

Arsen6331:
I just tried creating a CBOR settings file on the simulator, copying it to the develop branch on my PineTime, and then flashing this PR. I can confirm that the settings file is read and the settings are properly applied, and you can see the themed PineTimeStyle (White text, Light green sidebar) just before it crashes. That confirms that the crash is most likely caused by something outside of my code.

Arsen6331:
I am not sure how to check whether the `wakeUpMode` bitset is in range.

src/components/settings/Settings.cpp
```diff
@@ -23,25 +32,254 @@ void Settings::SaveSettings() {
...
 void Settings::LoadSettingsFromFile() {
-  SettingsData bufferSettings;
+  MigrateSettingsToCBOR();
+
+  lfs_info cborFileInfo;
+  if (fs.Stat("/settings.cbor", &cborFileInfo) != LFS_ERR_OK) {
+    return;
+  }
+
+  uint8_t* cborBuf = new uint8_t[cborFileInfo.size]();
```
NeroBurner, line 42: instead of doing the memory management yourself use `std::vector`. You'll need to add `#include <vector>` at the includes and initialize the buffer with

```suggestion
  std::vector<uint8_t> cborBuf(cborFileInfo.size);
```

and when you need to pass a pointer to the fileRead function (or other functions) use `cborBuf.data()`

Then you won't have to do the `delete[] cborBuffer;` calls everywhere. The standard-library will do that for you

NeroBurner (changes requested):

src/components/settings/Settings.cpp
Arsen6331: Done, thank you.

Issue #1056, opened 2022-03-27
Linking not working on Windows 10


watertrainer:
### Verification

- [X] I searched for similar bug reports and found none was relevant.

### What happened?

When trying to build with Windows 10 after following [buildInVsCode.md](https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/doc/buildWithVScode.md) on WIndows, everything builds but the linking fails and throws an error.

### What should happen instead?

Linking should work and building complete without error

### Reproduction steps

On Windows 10 follow [buildInVsCode.md](https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/doc/buildWithVScode.md) and [buildAndProgram.md](https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/doc/buildAndProgram.md).

I used cmake 3.19.4 and make 3.81, installed with mingw64 and gnuwin32 respectively.
When trying to build using ```make -j pinetime-app``` it builds to 100% and then fails on the following message.
```
c:/program files (x86)/gnu arm embedded toolchain/9 2020-q2-update/bin/../lib/gcc/arm-none-eabi/9.3.1/../../../../arm-none-eabi/bin/ld.exe: unrecognized option '--major-image-version'
c:/program files (x86)/gnu arm embedded toolchain/9 2020-q2-update/bin/../lib/gcc/arm-none-eabi/9.3.1/../../../../arm-none-eabi/bin/ld.exe: use the --help option for usage information      
collect2.exe: error: ld returned 1 exit status
make[3]: *** [src/pinetime-app-1.8.0.out] Error 1
make[2]: *** [src/CMakeFiles/pinetime-app.dir/all] Error 2
make[1]: *** [src/CMakeFiles/pinetime-app.dir/rule] Error 2
make: *** [src/CMakeFiles/pinetime-app.dir/rule] Error 2
```

### More details?

I tried to remove the argument in question and found, that ld.exe is called in [CMakeList.txt:913](https://github.com/InfiniTimeOrg/InfiniTime/blob/a8b7fbe48b4a86238f38ed0f084b277b44c428fb/src/CMakeLists.txt#L913). There doesn't seem to be an argument ```--major-image-version``` being passed to ld.exe.

When I try it in linux, everything works, although the counterpart to ld.exe doesn't accept and ```major-image-version``` argument either, even though it shows up on the man page.

### Version

newest Version, development Branch

Avamander:
[...]

PR #1073, opened 2022-04-06
Documentation for building on Windows (see #1056)

watertrainer:
Building on Windows needs some small tweaks as @[ITCactus](https://github.com/ITCactus) noticed. After those changes building on Windows worked for me.


Avamander:
I think it should be clarified that it isn't supported in the sense that nobody has gone trough the trouble of figuring the process out. It's somewhat unlikely that it's Windows, the toolchain or the project that makes it impossible.

Avamander (approved):

PR #1074, opened 2022-04-07
New sharper battery icon. Remove old unused batteryicons

Riksu9000:
New hand drawn battery icon with separately drawn capacity indicator, which can display many more steps than the previous icon.
I made it so this icon doesn't turn green when the battery is full, since it should be precise enough that it isn't necessary.
Meanwhile I removed the old and unused custom battery icons.

One of the font awesome battery symbols is still used in the settings list, so I didn't remove any of them yet.

InfiniSim_2022-04-07_143331
InfiniSim_2022-04-07_143554
InfiniSim_2022-04-07_143713
InfiniSim_2022-04-07_143337
InfiniSim_2022-04-07_143339
InfiniSim_2022-04-07_143344
InfiniSim_2022-04-07_143348


trman:
does this new battery  use more memory or less @Riksu9000  ?
since it seem to be drawn , it most likely take more , ram memory than the previous.

by the way , you call it sharp but, since the height of the new battery icon seem to be less that the width of the previous one ,
i don't think it would be more easier to see the difference between 40 and 60%.

medeyko:
Maybe to add black strains?

Like here:
изображение

That's just an example. The exact distance between strains depends on the exact number of the steps. Each step should have one strain more.

Avamander:
> i don't think it would be more easier to see the difference between 40 and 60%..

I don't think it matters much.

Avamander (approved):

medeyko:
> > i don't think it would be more easier to see the difference between 40 and 60%..
> 
> I don't think it matters much. This is a quick indicator, not a precise diagnostic tool.

The difference between 40% and 60% matters and should be distinguishable, I believe.

Are there any problems that I missed with the stripes I proposed above?
As far as I see, they won't take additional memory, we won't lose anything with them.
Also, the pictogram looks better with them from aestethics point of view, in my opinion.

Riksu9000:
@medeyko It's not a bad suggestion, but I think I prefer it without the stripes. It might be hard to see on a very small screen.

JF002:
> does this new battery use more memory or less

It should use less memory as previously, there were more than 20 icons (1 for each "step") while this PR need only 1 icon that is customized according to the current battery level. It's quite smart!

JF002 (approved):

Danimations:
I'd like to suggest an alternative to the stripes to visually indicate battery level.

What about _tinting_ the whole battery icon to rapidly indicate its state? The end result would be that the whole battery icon would change colour as the battery is either depleted or charged. Assuming there are five visible "steps" here's a potential system... though it could easily be more gradual:

0-20% = red
20-40% = orange
40-60% = yellow 
60-80% = pale yellow/green 
80-100% = mid/bright-green

Riksu9000:
The battery icon already indicates the level of charge by how full the battery is, so I don't consider coloring it as being necessary. It could turn deep orange when the battery is low as a warning, but that would require the percentage conversion to be improved first, since currently 40% is in reality already almost empty.

Danimations:
Thanks @Riksu9000 . I agree that a colour change to indicate low battery, or perhaps even a flashing icon, would be useful... perhaps moreso than the graded colour concept that I raised above.

In my opinion, flashing the icon when battery is low would alert the wearer and encourage them to get their Pinetime onto a charger. Certainly worthwhile.

The concept of changing the icon's colour to reflect the % of charge does have some supporting logic behind it. For example, if a user has less than 20/20 vision, telling the difference between 40 and 60% charged may be difficult. The vertical bar within the battery icon is a useful fine-scale indicator, but the colour coding provides more visual information faster, and at coarser scale. It could be rapidly and reliably interpreted with a quick glance or by someone with imperfect eye-sight.

The combination of fine-scale (column within battery icon) and colour of icon could benefit the user base broadly.

It sounds like percentage conversion could be worth further developing. Everyone appreciates an accurate battery meter, regardless of how the data is visualised. Thanks for all your work as always.

PR #1077, opened 2022-04-09
Fix Timer layout

### LINKED ISSUE #924: Timer text and +/- buttons not centered horizontally
### Verification

### What happened?

Was working on some changes to the timer app and noticed something was off and found it was like 5 pixels to the left!

The timer text and +/ minus buttons should be centered horizontally

### Reproduction steps

Open the timer app

### More details?

I may work on this after I finish what I'm doing but I don't want to interrupt what I'm doing rn.

### Version

1.8.0

### Companion app

_No response_

Riksu9000:
Make Alarm and Timer layouts the same.
Also made the start button 50px tall, since every other button on the bottom edge is 50px tall.

InfiniSim_2022-04-09_095616

Fixes #924


Avamander (approved):
Seems like a small fix to quickly merge, future progress can be built on top.

JF002 (approved):

PR #1079, opened 2022-04-09
Stopwatch fixes

Riksu9000:
`calculateDelta()` isn't necessary, because `TickType_t` handles overflow naturally.
Use centiseconds in a conversion function instead of milliseconds, so the math uses smaller numbers and the last digit wasn't used anyway.
Other small fixes.


Avamander (approved):

JF002:
> `calculateDelta()` isn't necessary, because TickType_t handles overflow naturally.

~~Are you sure about this?~~ For anyone wondering if it really works : https://godbolt.org/z/o7fjcrTxo. 
TL;DR : Yes, overflows are handled correctly, it's the magic of unsigned arithmetic :D

JF002 (approved):

PR #1080, opened 2022-04-09
Firmware update screen fixes

Riksu9000:
Avoid using floats.
Increase bar resolution.
Other small fixes.

### DIFF
--- a/src/displayapp/screens/FirmwareUpdate.cpp
+++ b/src/displayapp/screens/FirmwareUpdate.cpp
@@ -16,18 +16,17 @@ FirmwareUpdate::FirmwareUpdate(Pinetime::Applications::DisplayApp* app, Pinetime
 
   titleLabel = lv_label_create(lv_scr_act(), nullptr);
   lv_label_set_text_static(titleLabel, "Firmware update");
-  lv_obj_set_auto_realign(titleLabel, true);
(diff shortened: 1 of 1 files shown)


src/displayapp/screens/FirmwareUpdate.cpp
```diff
@@ -78,22 +78,20 @@ void FirmwareUpdate::Refresh() {
 }
 
 void FirmwareUpdate::DisplayProgression() const {
-  float current = bleController.FirmwareUpdateCurrentBytes() / 1024.0f;
-  float total = bleController.FirmwareUpdateTotalBytes() / 1024.0f;
-  int16_t pc = (current / total) * 100.0f;
-  sprintf(percentStr, "%d %%", pc);
-  lv_label_set_text(percentLabel, percentStr);
+  const uint32_t current = bleController.FirmwareUpdateCurrentBytes();
```
Avamander, line 81: Is this const correct?

Riksu9000: Since it's a local variable, it's initialized each time the function is called. I believe it's correct, because the variable isn't changed after initialization.

Avamander (approved):

JF002 (approved):

PR #1083, opened 2022-04-12
Turn off chimes when notifications are silenced

minacode:
Hi everyone :)

This PR turns off chimes, when notifications are silenced.
I think the current behaviour is counter-intuitive.
It surprised me, when I turned off notifications to sleep.

I tested the changes and they work for me.

Best regards and thanks for all the work.


SteveAmor:
The trouble is, if you get 10s/100s of notifications a day, you might want to silence them but still have chimes. Ideally, you could turn notifications on and off independently of a global "vibrate" toggle. IMHO

minacode:
Then maybe we could have a third mode: "sleep"?
The modes could just cycle through on the same button.

PR #1085, opened 2022-04-12
Heartrate ring buffer

watertrainer:


watertrainer:
sorry that was an accident i misclicked

Issue #836, opened 2021-11-18
Torch settings


medeyko:
It's great that the torch is now ajustable, thanks!
But personally for me it has became much less usable due to the dim initial state.

I often use it to routinely light a room. For this I need a torch that lights at maximum level with minimum gestures.

The old version (firmware 1.6.0) required only TWO gestures: 1. swipe left for menu; 2. click on the torch icon.
Now (1.7.1) I have to make FIVE gestures: 1. swipe left for menu; 2. click on the torch icon; 3. swipe left to increase light; 4. swipe left second time to increase light; 5. click to turn the torch on. That's really a lot!

This is quite inconvient for my use case. I would like to either have an opportuninty to set the default light level to maximum and the torch state to on, or have some other short sequence (of no more than two gestures) that allows to turn the torch on maximum (for example, long button press or long click on the torch icon in the menu).

Also it's a pity that requests to support RED color were rejected. There are lot cases when low-energy photons are preferable, like photograhy or astronomy. Also, the red color is used to draw attention. Namely, I would use red torch when walking by a roadside. It might be just a fourth "light level" option, a red one after three white ones (in the torch configuration or in the torch app itself).

For the latter use case, it would be nice to have a strobe light option (it draws more attention and conserves battery power), but probably it's another topic for discussion.

Riksu9000:
The default brightness is based on the backlight level, but maybe it makes more sense to default it to the highest level as that's what is most likely to be used.

The reason the flashlight is off by default is to not shine bright light in the users face when they launch it. The idea is that you need to look at the watch to open the app, but when the app is open, you can point it away and tap the screen to turn it on.

The long press idea is interesting. We could use that for something if we need to, but for now I think we can get by with just a bit of tweaking.

Red wasn't necessarily rejected. Red on the PineTime is very dim and it's not very pure I think, so I'm waiting for someone to actually use red light on the PineTime and tell me it's useful. Drawing attention when walking in the dark may be a good idea, but maybe not enough to justify having red when you can just use white.

medeyko:
Just a small comment, why I'd prefer a red light for the roadside walks: a red color is used to denote edge of the same-side roadside (to the right side for most countries, including my one). The white color is used to denote opposite roadside/opposite direction lanes (to the left side). So using white color for objects walking on the same-side roadside might be a bit misleading for the drivers. Also red color is less frequent than white one, and it is linked to drawing attention. Despite of course white light is still much better than nothing!
[...]

PR #1087, opened 2022-04-14
Keep preset backlight setting for flashlight app in off mode

rev22:
Currently the flashlight app starts off and the user can turn it on by tapping the screen.

This modification is meant to reduce battery usage by restoring the previously-set backlight level when the user taps the screen to turn the light off.

In addition, for ease of usage the app is started with the flashlight on.

I tried to simplify and streamline the code a bit, saving 12 bytes in the compiled binary.

### DIFF
--- a/src/displayapp/screens/FlashLight.cpp
+++ b/src/displayapp/screens/FlashLight.cpp
@@ -9,6 +9,8 @@ namespace {
     auto* screen = static_cast<FlashLight*>(obj->user_data);
     screen->OnClickEvent(obj, event);
   }
+
+  typedef Pinetime::Controllers::BrightnessController::Levels Levels;
 }
 
 FlashLight::FlashLight(Pinetime::Applications::DisplayApp* app,
@@ -19,8 +21,6 @@ FlashLight::FlashLight(Pinetime::Applications::DisplayApp* app,
     brightnessController {brightnessController}
 
 {
-  brightnessController.Backup();
-
   brightnessLevel = brightnessController.Level();
 
   flashLight = lv_label_create(lv_scr_act(), nullptr);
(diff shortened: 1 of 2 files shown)


Riksu9000:
The app was specifically changed to start with the light off, though not everyone agrees with it.

Perhaps using the default brightness when flashlight turned off is fine.

I would argue the previous code is simpler, though I guess I would be biased. I don't think combining separate functions into one big function is useful.

#836 Would be the place to discuss changes to the app.

Shatur:
I like the rework, previous behavior was annoying for me.
But @Riksu9000 also suggested to remove brightness levels at all in https://github.com/InfiniTimeOrg/InfiniTime/issues/836#issuecomment-1099274998. But this could be a part of a separate PR.

rev22:
I was not aware of the previous discussion, thanks for the link.  Indeed there are a lot of potential use cases and preferences.
So I suppose we can retain the battery-saving feature for this PR and eventually continue discussing other options in the other PR.

Riksu9000:
Calling it battery saving is a bit of a stretch, since I don't think anyone would keep the flashlight at the off position for long enough that it makes a real difference. Also technically it will increase battery use when toggling at the lowest brightness setting and bleed more light at the off setting, if I understand correctly.

rev22:
Yes that's exact Riksu.  However in the most common use cases I think the user would enter the flashlight app in low environment light so of course the initial backlight setting would be normally Low or Medium, while the most common settings for the backlight in the on state would be Medium or High.  The backlight is a major battery drainer at high settings.
The main intent of this PR is to reduce battery usage slightly for common use, without increasing UI complexity or binary size.

rev22:
Starting the app in the off state would make my code more complex, because it would introduce an additional initial state or make more lines of code necessary, I'm a bit of a perfectionist, so I'll keep my modification in the original form in my fork for now, and close this PR, at least for now, unless someone wants to work on it.  Thank you Shatur and Riksu9000 for your feedback!

PR #1089, opened 2022-04-14
Flashlight app update

### LINKED ISSUE #836: Torch settings

Riksu9000:
~Start with the flashlight on.~
Remove middle brightness option.
~Long press the app icon to launch with low brightness.~

When the flashlight is off in high brightness mode, the backlight level will be the system backlight level. In low brightness mode, the backlight brightness will always be low. The flashlight icon dims slightly in low brightness mode, so you can distinguish the modes when the system brightness is low.

This uses the tap touch event instead of an LVGL click to toggle the flashlight, since even though it's slightly slower, it doesn't accidentally activate when swiping.

This video is outdated. The flashlight will default to off and the long press option is removed

https://user-images.githubusercontent.com/37774658/163480241-ace88433-873f-4545-b623-9e267e6cd10c.mp4

Fixes #836

### DIFF
--- a/src/displayapp/screens/FlashLight.cpp
+++ b/src/displayapp/screens/FlashLight.cpp
@@ -4,51 +4,27 @@
 
 using namespace Pinetime::Applications::Screens;
 
-namespace {
-  void event_handler(lv_obj_t* obj, lv_event_t event) {
-    auto* screen = static_cast<FlashLight*>(obj->user_data);
-    screen->OnClickEvent(obj, event);
-  }
-}
-
 FlashLight::FlashLight(Pinetime::Applications::DisplayApp* app,
                        System::SystemTask& systemTask,
                        Controllers::BrightnessController& brightnessController)
-  : Screen(app),
-    systemTask {systemTask},
-    brightnessController {brightnessController}
-
-{
-  brightnessController.Backup();
-
-  brightnessLevel = brightnessController.Level();
+  : Screen(app), systemTask {systemTask}, brightnessController {brightnessController} {
 
   flashLight = lv_label_create(lv_scr_act(), nullptr);
   lv_obj_set_style_local_text_font(flashLight, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, &lv_font_sys_48);
   lv_label_set_text_static(flashLight, Symbols::highlight);
   lv_obj_align(flashLight, nullptr, LV_ALIGN_CENTER, 0, 0);
 
-  for (auto & i : indicators) {
-    i = lv_obj_create(lv_scr_act(), nullptr);
-    lv_obj_set_size(i, 15, 10);
-    lv_obj_set_style_local_border_width(i, LV_OBJ_PART_MAIN, LV_STATE_DEFAULT, 2);
-  }
+  brightnessController.Backup();
 
-  lv_obj_align(indicators[1], flashLight, LV_ALIGN_OUT_BOTTOM_MID, 0, 5);
-  lv_obj_align(indicators[0], indicators[1], LV_ALIGN_OUT_LEFT_MID, -8, 0);
-  lv_obj_align(indicators[2], indicators[1], LV_ALIGN_OUT_RIGHT_MID, 8, 0);
+  SetBrightness();
 
-  SetIndicators();
   SetColors();
 
-  backgroundAction = lv_label_create(lv_scr_act(), nullptr);
-  lv_label_set_long_mode(backgroundAction, LV_LABEL_LONG_CROP);
-  lv_obj_set_size(backgroundAction, 240, 240);
-  lv_obj_set_pos(backgroundAction, 0, 0);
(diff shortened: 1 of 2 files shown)


rev22:
I haven't tested this yet but the feature set seems optimal to me and it reduces lines of code significantly.

Riksu9000:
I ended up walking back on this a bit, since I'm still not 100% convinced that defaulting to on is the way to go. Pressing the button while not looking, which although is a quarter of the screen, is still a very small button, isn't something that we should expect the user to do IMO, since there's a high likelyhood of pressing other buttons.

I believe even when defaulting to off, these changes will reduce the cognitive load of using the app significantly. Defaulting to off may even be a lower cognitive load than on, since you don't need to be cautious about shining bright light in your eye.

KoalaV2:
+1 for defaulting to off. Accidentally pressing the flashlight app and having a bright light shine in your eyes sounds awful.

rev22:
In my develop branch, the compromise is to start the app with the light on without changing the backlight level, so you are less likely to be accidentally showered by a bright light at night, but still have a flashlight that turns on without too much effort.
I like the idea of alternate starts, for example red light by long press or bright light by double tap, though that would make more sense as a UX choice if alternate starts are supported by other apps too.
I also like the long-swipe proposal for the watch face, and other proposals I have read so far which should make it less likely to start apps or change settings accidentally.

medeyko:
Maybe you could consider to at least turn it on at full brightness by double click or long press the app icon? It's quite unlikely that one will make long press accidentally when clicking other buttons.

Ok, there may be another cause, why the default state for simple click should be off: because you can run the torch app in sleep. I managed to do this one time. And then it drains battery. It may be avoided, if the default state is off AND it the display turns off the display at standard display timeout when the light state is off.

But it's unlikely to make a long click in sleep, so it doesn't mean problems for turning it on at full brightness by double click or long press on the app icon.

Riksu9000:
> Maybe you could consider to at least turn it on at full brightness by double click or long press the app icon? It's quite unlikely that one will make long press accidentally when clicking other buttons.

Having an option like that wouldn't be a problem, but I don't think it would be any more efficient than just opening the app and turning it on.

medeyko:
Unfortunately, there is a difference that I have experienced in the current version.

First, one tries to turn the light on as quickly as possible (it's a natural desire, people don't like to wait). But if one makes a second tap too early, it will not turn on, because of the scrolling animation before starting the app.
Second, even if one waits long enough, sometimes the tap is registers as a swift instead of a tap.

This requires concentration and reduces the chances of turning it on witout looking at the screen.

However, after writing this, I realized that this a wrong approach anyway. And even a long tap would be just a bad compromise in this case.
So let's reserve long tap for red light or strobe light.

And I will make a separate feature request for quick launches.

Riksu9000:
> But if one makes a second tap too early, it will not turn on

Currently in the release LVGL clicks are used to turn the flashlight on. It is fast, but very prone to accidentally toggling the flashlight while swiping. The tap events are slower, but work more consistently. I prioritized minimizing unintended actions over speed in this case. These issues might get fixed over time.

medeyko:
I created the feature request I talked above: #1096

medeyko:
> > But if one makes a second tap too early, it will not turn on
> 
> Currently in the release LVGL clicks are used to turn the flashlight on. It is fast, but very prone to accidentally toggling the flashlight while swiping. The tap events are slower, but work more consistently. I prioritized minimizing unintended actions over speed in this case. These issues might get fixed over time.

As for the unintended actions, I would like to repeat my note above: You can run the torch app in sleep unintentionally. I managed to do this one time. And then it drains battery. It may be avoided, if the display turns off after the standard display timeout when the flashlight's light state is off. Display on should be persisted only when the light is on.

Riksu9000:
I agree the app shouldn't stay on forever, but that's outside the scope of this PR and unrelated to the issues I've mentioned.

PR #1090, opened 2022-04-15
Adding a reset button for the Timer app

Ertain:
Sometimes a PineTime user would like to quickly reset the timer to zero. The addition of this reset button helps with that. The button zeros out the `minutesToSet` and the `secondsToSet` variables, calls the `timerController.StopTimer()` method, and sets the `time` label to zero. This pull also adds the `Reset()` member function to the `TimerController` class.

Reset button in InfiniSim
Timer app with reset button on the PineTime


medeyko:
In which cases in may be useful? I don't see when a user may be so eager to to quickly clear the timer.
If one exits the timer app, and then runs it again, it already clears.

This additional button may be confusing, and also the overall compositions of buttons looks ugly with it. Also, one could accidentally click it when setting the time up. Therefore its addition should be seriouly substantiated, I think.

Ertain:
You have a point there. The whole Timer app interface should be redesigned. I shall close this pull request.

Issue #749, opened 2021-10-16
Fitness data protocol


JF002:
For now, we only expose the *live* heart rate values. Exposing steps the same way won't be difficult.

For more advanced features, companion app need more than just the current value, they'll need an historic of all the data acquired during the day/night, even if the watch was not connected to the companion app when the data was acquired.

This goal of this discussion is to define how the data will be stored and the API InfiniTime will expose to the companion app.

I asked Adam from Amazfish how fitness data were managed on other devices : 

- The watch stores the data with a specific granularity along with a timestamp. Ex : number of steps counted every 10 minutes, HR value per minute,...
- The data should be stored in some kind of circular buffer so that we ensure we don't fill the memory
- Using BLE, the companion app request the data from a specific date-time, and of course, InfiniTime sends them if they are available in the buffer.

Questions:

- What would be an ideal granularity for the data/timestamp?
- Should the data be stored in RAM? External flash? Temporarily in RAM and flushed in external flash from time to time (pay attention to wear of the flash memories) ?

## Settings
In settings, we define the granularity (the delay between 2 samples). Ex : 10 minutes

## Data acquisition
InfiniTime monitores the data (steps + HR if enabled) and accumulates them during the "granularity" period.

When the period elapses, the accumulated data are appended in the circular buffer in the form

```
struct FitnessData {
  uint64_t timestamp;
  uint8_t nbSteps;
  uint8_t heartRate;
}
```

We could also add average/min/max for the heart rate value. We can also store a single reference timestamp, and then store a delta from this timestamp in `FitnessData` to save a few bytes for each sample.

## BLE API

Service : FitnessDataService 
Characteristics:

- Period (Read/Write) : the companion app can read the current acquisition period and set a new one.
- Commands (Write)
  - Request data from a specified timestamp
  - Clear data (from / to a specific timestamp)
- Data (Read) : requested data will be written on this characteristic
- Data available (Notify) : notifies the companion app that new values are available

The protocol should be designed in such a way it will be possible to extend it (add a `version` characteristic or field in the data sent to the companion app).

## Let's talk about this ;)
These are just some ideas out of my head, feel free to give your opinion.
I'm not sure that 'fitnessData' is the most appropriate name for the data...

infinitytec:
10 minutes seems like a pretty reasonable time to me. Depending on some exercise routines people may do there may be reasons to do more frequent logging, but I imagine those could be implemented later.

[...]

Issue #532, opened 2021-07-28
Heart rate measurement accuracy


Riksu9000:
I have done some experiments with the heart rate sensor. I think one of the issue is that small movements while measuring can produce noise that could throw off the measurement. I don't think there's any code that detects noisy data and ignores it. The accelerometer could also be used to ignore data while there is movement. For some reason decreasing the gain improves accuracy for people. Why is that?

Setting the gain from 64 to 8 may have improved the results a bit. It worked at least as well as before.

Setting the LED current to 40mA didn't seem to make a difference.

The ambient light sensor could be used to detect if the watch isn't on the wrist, or if it's too loose, which could cause incorrect measurements as well.

Below you can see the SPL value graphed. This was with gain set to eight. The blinking light I added might actually be more accurate than the number shown, implying there might be an issue with the conversion to BPM. EDIT: Maybe its just HRV?

The data looks clean enough so it should be possible to get accurate measurements.

Here's the branch if anyone wants to test this.
https://github.com/Riksu9000/InfiniTime/tree/heartrate_test

https://user-images.githubusercontent.com/37774658/127296540-e5944101-0156-4c61-a764-d90913efc684.mp4

Riksu9000:
I just read that the sensor actually gives better results with less pressure. I think everyones first instinct is to tighten the strap to get better contact, myself included. Perhaps this is why people are having issues with it, and if so, we should make it known to everyone. Maybe even add instructions to the app.

JF002:
The algorithm for the HR measurement comes from [WaspOS]'https://github.com/daniel-thompson/wasp-os/blob/master/wasp/ppg.py). I ported it from Python to C++. When I talked to Daniel about it, he told me that it was quite simple and should/could probably be much improved :)

I'm certainly not a specialist in signal processing, so I'm glad someone takes the time to have a look at this!
And the graph with the SPL values on the background of the HR app looks really good!!

hatmajster:
Yes, very nice Riksu, I don't know how you guys do it here, but I would personally even merge this graph, its really nice to debug behavior of sensor :) Though maybe its feature creep in me;)
I've combined it with my own PR to change gain via options (https://github.com/JF002/InfiniTime/pull/531) and was testing it a bit.
[...]

PR #1092, opened 2022-04-15
Ring Buffer for Heart Rate Data

watertrainer:
This implements a simple Ring Buffer for the Preprocessed Heart Rate Data, so that actually processing it can always take the last 200 datapoints. Before, after 200 datapoints were collected, all the data was discarded and new data was collected. This had, imo, a few issues.
1. High frequenzy Heart Rate calculation/measurment wasn't possible (there only was one value very 4 seconds) (may be relevant for #749), this may also be useful to average Heart Rate values in the future and reduce noise (#532)
2. There was no very recent "on demand" Heart Rate avaiable - it always was lacking behind somewhat. This makes it much easier to implement e.g. an instant Heart Rate Update on screen wake.
3. Everytime 200 datapoints were collected, the Heart Rate HAD to be calculated and saved in the HeartRateController, because on the next tick the whole data array was invalid again. With this the PPG can always provide a HeartRate if required

Note that this Pull Request doesn't change any of the user experience/behaviour of the watch yet. A new Heart Rate value is still only calculated and provided every 200 datapoints (4 Seconds at 25Hz, which is the speed [I think](https://github.com/InfiniTimeOrg/InfiniTime/issues/532#issuecomment-1099043431), although I read [here](https://github.com/InfiniTimeOrg/InfiniTime/blob/2607c3d79947e900ce4c5ded296f649677511a34/src/drivers/Hrs3300.cpp#L28), that each cycle takes 12,5ms, which would mean 80Hz). This pull request would just make future implementations for working with HeartRate data much easier in my opinion.

I also cleaned up and fixed all the clang-tidy complaints in the file I wrote code in, I don't know if that belongs in a different request.
This is also basically my first cpp code, so any help/critique regarding code style etc. is greatly appreciated.


watertrainer:
The change also requires very little extra processing power, one add and one modulo more per array access, the array access is much more expensive in that case.

medeyko:
> The change also requires very little extra processing power, one add and one modulo more per array access, the array access is much more expensive in that case.

I guess nothing prevents us from making UPDATE_HEARTRATE_AFTER and DATA_SIZE powers of two, because i think that 200 is an arbitrary number. Most compilers will automatically optimize the "modulo" operation into an "and" operation.

watertrainer:
DATA_SIZE is not arbitrary I think, because the range of valid bpms depends on how many datapoints there are. That's how I read this comment

medeyko:
As far as I undestand, this is because the heart rate is unlikely to exceed 200 bpm.
However, it can. According to this article, https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3273956/ they registered 600 bpm in a case of an extreme non-fatal tachyarrhythmia.
220 bpm is recorded for healthy athletes.
So I don't see the 256 constant is any worse than 200.

watertrainer:
The fast bpms would show up in the low indices I think.
If the algorithm works any similar to [this one](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6651860/#!po=15.8333)and I understood it correctly.
It looks for the first peak in the PPG signal an then just assumes that all othe peaks are evenly spaced after that (with some noise reduction etc.)  A very quick peak would be in the first few indices and only for very slow bpms more indices would be needed (the data points are collected at (24Hz)[https://github.com/daniel-thompson/wasp-os/blob/f29a6013f5580fd12955563702981707543e23f1/wasp/ppg.py#L109] so for 200 that would be a last peak at 4 seconds and 12 bpm with the slowest heartbeats being around [25](https://en.m.wikipedia.org/wiki/Bradycardia) that's even too high of a number.

I think the number 200 is chosen because in the process function of PPG the max resulting index could be around 200 (see [here](https://github.com/daniel-thompson/wasp-os/blob/f29a6013f5580fd12955563702981707543e23f1/wasp/ppg.py#L159), t3+4 would be the highest index (which I got at a max of 210, don't really know why)

Making it bigger won't break anything, it might not be very useful though and uses a few bytes of memory. We should maybe still do it though because if t3+4 goes to 210 the heart rate measurement will be completely off (and it will access memory outside of the array before this pr).

src/components/heartrate/Ppg.cpp
```diff
@@ -52,48 +49,59 @@ int8_t Ppg::Preprocess(float spl) {
...
+  int t0 = Trough(data.data(), DATA_SIZE, 7, 48);
+  if (t0 < 0) {
     return 0;
+  }
 
-  float t1 = t0 * 2;
-  t1 = Trough(data.data(), dataIndex, t1 - 5, t1 + 5);
-  if (t1 < 0)
+  int t1 = t0 * 2;
+  t1 = Trough(data.data(), DATA_SIZE, (t1 - 5), (t1 + 5));
```
Riksu9000, line 73: The added parenthesis aren't necessary here, or the line few lines down.

src/components/heartrate/Ppg.cpp
```diff
@@ -5,45 +5,42 @@
...
-  }
-}
-
 Ppg::Ppg()
   : hpf {0.87033078, -1.74066156, 0.87033078, -1.72377617, 0.75754694},
     agc {20, 0.971, 2},
     lpf {0.11595249, 0.23190498, 0.11595249, -0.72168143, 0.18549138} {
 }
 
+int Ppg::Compare(int8_t* d1, int shift, size_t count) {
```
Riksu9000, line 19: Why was this function changed, apart from adding `getRingIndex()`?

Riksu9000 (commented):
A couple observations.

The small changes to `dataIndex` are hard to notice and review. They're also not really necessary.

In `ProcessHeartRate()`, `dataIndex` and `DATA_SIZE` would be the same value without the slight `dataIndex` change. `DATA_SIZE` may be slightly more correct, but only before the actual behaviour is changed.

But the more I think about it, this PR doesn't actually change much apart from adding `getRingIndex()` does it? Maybe it would be better to add and start using the ring buffer in one PR. As it is, these changes are hard to justify. The code improvements would be a nice separate PR.

src/components/heartrate/Ppg.cpp
watertrainer: in the original implementation the pointer was moved forward and then accessed as a 0-based array. That wouldn't work in the ring buffer, because the actual index has to be known to modulo it.

src/components/heartrate/Ppg.cpp
watertrainer: I just added them because clang-tidy was complaining, don't know whats the preferred way.

watertrainer:
First of all thanks for the review :)

I agree that the PR doesn't really have a justification to be merged yet, I just implemented it because it might be useful for other future features, which I wanted to start working on (e.g. #1094).
And your right the only real thing changed is `getRingIndex`, I could have written that before ^^>

At the moment `dataIndex` and `DATA_SIZE` don't have a difference in the `Process` method, but only because it's called when `dataIndex` is 200. That's the whole "point" of the PR, that `Process` can be called with any value of `dataIndex` and return a valid HR, not as before, where `dataIndex` HAD to be equal to `DATA_SIZE` or else the whole function wouldn't have worked anymore.

src/components/heartrate/Ppg.cpp
Riksu9000: It doesn't complain about that for me.

src/components/heartrate/Ppg.h
```diff
@@ -10,22 +10,28 @@ namespace Pinetime {
   namespace Controllers {
     class Ppg {
     public:
+      static const uint8_t DATA_SIZE = 200;
```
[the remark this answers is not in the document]
watertrainer: Done with c6d76739ff18585eac6d41720b82ea4d24bf5b27

src/components/heartrate/Ppg.cpp
watertrainer: removed in faf2faabea986b1efebb204beab3a4f439008a56

Mindavi:
This definitely seems useful for e.g. higher frequency updates on the display.

khimaros:
@watertrainer are you still planning to work on this? i'm available to help.

watertrainer:
Yes, if you have any ideas how to make this useful for the user I'd be more than happy to work on it together.

khimaros:
@watertrainer has this been solved elsewhere?

watertrainer:
not that I'm aware of, although there were some heart rate changes a few versions ago, not to sure haven't been following to closely

PR #1093, opened 2022-04-16
Increase Timer app max from 59 to 99 mins

SteveAmor:


watertrainer:
This might make it a little hard to set a timer for 45 or any a bit higher number of minutes. I personally still think it's a good idea, but maybe there'd need to be a faster way to change the amount of minutes than as well.

medeyko:
> This might make it a little hard to set a timer for 45 or any a bit higher number of minutes. I personally still think it's a good idea, but maybe there'd need to be a faster way to change the amount of minutes than as well.

I agree both your points. And I propose double click to change the respective value by 5 or 10.
(I don't know how subsequent clicks are treated. I think, if four clicks are treated as a two double clicks, the double click step should be 5, if four clicks are treated as one double click and two single clicks, then the double click step should be 10. If a tripple click is a separate event, then the double click step should be 5, and the tripple click step should be 30, I think).

However, it's applicable to other numbers settings as well, and should be done consistently in all places, not only with a timer.
But it would be cool, because it's irritating to set up numbers for alarm currently.

Riksu9000:
The double click idea wouldn't work, because whatever way subsequent clicks are treated, it would still be confusing. The buttons need to support `LV_EVENT_LONG_PRESSED_REPEAT` before this change can be made. Also 69:55 and longer times don't work correctly on the PineTime. This issue would need to be resolved as well.

medeyko:
What's about long taps instead of double clicks, that will change the value by 10? It wouldn't work either before the `LV_EVENT_LONG_PRESSED_REPEAT` is supported?

Riksu9000:
Long press repeat and long press to add 10 can't both be used. I would prefer the simple long press repeat over inputting multiple subsequent long presses, even if it was a bit slower, because of it's ease of use.

medeyko:
I'd prefer multiple subsequent long presses, due to speed and predicability, but it probably just a question of taste.
Both solution would be cool, becuase currently it's bothersome to set time to +30 minutes.

kieranc:
> Also 69:55 and longer times don't work correctly on the PineTime.

Do you know why this is? I can't figure it out.

Itai-Nelken:
> > Also 69:55 and longer times don't work correctly on the PineTime.
> 
> Do you know why this is? I can't figure it out.

I think it is a timer limitation, but don't quote me on that.

Itai-Nelken:
> > > Also 69:55 and longer times don't work correctly on the PineTime.
> > 
> > Do you know why this is? I can't figure it out.
> 
> I think it is a hardware timer limitation, but don't quote me on that.

kieranc:
> I think it is a hardware timer limitation, but don't quote me on that.
I don't think so, the timer uses a uint32 to store the timer value as ticks which means a maximum value of 4,294,967,295

 `timerController.StartTimer((secondsToSet + minutesToSet * 60) * 1000);`

This starts the timer, so 69:55 would be 7,440,000, well within the limit. It does feel like some sort of a rollover though, since setting a timer for 70:05 results in it counting down from 10 seconds.
The alarm is capable of handing much longer periods but it looks like it stores the time until alarm as a number of seconds rather than ticks/ms...

Riksu9000:
I'm closing this due to it not working and not being updated.

Issue #895, opened 2021-12-30
Expose the HeartRate service UUID in BLE ADV packet


JF002:
### Verification

- [X] I searched for similar feature request and found none was relevant.

### Pitch us your idea!

Expose the HR service UUID in ADV so that 3rd party application detects the PineTime as a HR device

### Description

OpenTracks reported that a PineTime running InfiniTime is not detected as a HR device because the UUID of the HR service is not exposed in the ADV packet.

See https://github.com/OpenTracksApp/OpenTracks/issues/1063 and https://github.com/OpenTracksApp/OpenTracks/pull/1069.

Adding this UUID in ADV would allow companion apps to detect the PineTime as a HR device without needed to connect and discover the services running on the PineTime.

Basically, we would need to add the specific UUID here : https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/src/components/ble/NimbleController.cpp#L162

- [ ] What are the implications of adding multiple UUIDs to the ADV packet?

dennisguse:
Providing multiple service UUIDs in the ADV packet increases the package size and thus consumes more energy.

trman:
i don't thinks @JF002  , that the only things to do : 
even if you add the packet ,the heart rate task must be able to function even when the screen is off otherwise
(it's not the case right now)
 , this change will be  misleading since infinitime would say it's a  HR device **but** it would not show hr data only for the sporadic case when the screen is on (and for maybe only 3 fc too , since the measurement take nearly 20 sec and the screen up time is at most 30 sec)


by the way , since the waspos main dev didn't have received issue about the fact that putting the gain to 8 is bad , maybe it's time for infinitime to put the HR gain to 8 as well (in order to have better and faster measurement)i

vdmkenny:
@trman the latest infinitime firmware addresses this HR gain issue.
https://github.com/InfiniTimeOrg/InfiniTime/releases/tag/1.9.0

trman:
hi @vdmkenny 

that's true at very least  , the easiest part have been addressed (the hr gain)  and that's good !

Now it would be great if the main part would be adressed as well (add the UUI and the heart rate task function all time , even screen off)

jmaris:
No movement on this issue? Also is there anywhere we can submit bug bounties for this project?

dennisguse:
From OpenTracks perspective, the problem is resolved as it now supports to search for any BLE device as heartrate sensor (i.e., not filtering using the advertised services). 
[...]

Issue #307, opened 2021-04-21
App idea: Sleep tracking


myxor:
### New app idea: Sleep tracking

**Things to do:**
- Estimate start and end of sleep depending on movement detection (from acceleration sensor). 
     - Data from heart rate sensor and touch screen activity could be considered as well
- Local logging of data for now
- Store and show data on the phone via companion app (Gadgetbridge has support for this)

He is a write up on how this works in general: https://en.wikipedia.org/wiki/Actigraphy

What do you think?

danipozo:
Two things to note:
 - as far as I know, battery life with HR sensor constantly on hasn't been tested,
 - HR sensor processing needs to improve yet, since measurements can be very unstable.

I think this is a cool thing to have.

pfeerick:
I would not have thought HR monitoring would be constant when in a sleep mode... more like every 15 minutes to half hour... Plus, if this were being done in sleep mode, unless the accuracy is fixed some other way, it could be by doing a running averaging of the samples over a 30 period, and chucking any outliers that are say 30 above that running average to try and eliminate the spurious readings that sometimes come in. 

Either way, This would be nice to see.. someday soon ;)

MysteriousLog6:
I would rather turn on the Heart Rate every five minutes rather than 15Minutes to half an hour...., we could try and calibrate the heart rate readings with a smartwatch we know works. (Someone who has both)

Raupinger:
here's a paper on accelerometer based sleep tracking https://www.nature.com/articles/s41598-018-31266-z

MysteriousLog6:
> here's a paper on accelerometer based sleep tracking https://www.nature.com/articles/s41598-018-31266-z

Thanks! I will read it and see what we can do.

MysteriousLog6:
We could try and have raw accelerometer data with a video of the person wearing it (Creepy I know) so that we can program it to be as accurate as possible , the video could have the person turning on the accelerometer logging (An app for this can be made that logs accelerometer data.) helping us sync the two of them together.

The same person could also wear another smartwatch/fitness tracker (Preferably Apple Watch or Fitbit as they have accurate sleep detection compared to others)

This will give us a small hope , battery life is another concern.

theuser18:
Could it potentially interface with Sleep as Android to offload some of the effort required in measuring sleep? [The devs are interested supporting PineTime,](https://forum.urbandroid.org/t/support-for-pinetime/7431) but say the APIs required are not in place yet.

[...]

PR #1094, opened 2022-04-16
Heart Rate measurements regularly every 60 seconds

watertrainer:
This pull requests starts a Heart Rate measurement every 60 seconds. 
It's a bit wanky if the user wants to stop the heart rate sensor in the app while a measurement is running. At the moment it turns off the sensor, but will turn it back on the next time it checks if it needs to go on (currently 1 tenth of a second I think), because I thought it would be confusing if the sensor wouldn't turn off at all.
It's also not detected by the Heart Rate App yet if the sensor is running before the app is started, so it still gives the option to start the sensor, even though it is on, the press has no effect though (except that the app starts to show the current heart Rate). The digi watch face was able to detect the heart rate and showed that a measurement is going on.

In the future it might be a good idea to average out the heart rate over a prolonged period instead of taking the first avaible value to reduce noise, this would depend on #1092.

This might be useful for general fitness tracking (#749), sleep tracking (#307) or just to keep track of your heart rate (#895).
The intervall could easily be implemented as a setting, I don't know how to implement new settings though, especially if they are supposed to be persistent.
60 Seconds does seem quite often when my watch was just laying aroung on the table, so the standard might be something more like 90/120. 
I'll let the watch run overnight to see how much power would be consumed by this feature.

### DIFF
--- a/src/heartratetask/HeartRateTask.cpp
+++ b/src/heartratetask/HeartRateTask.cpp
@@ -6,12 +6,13 @@
 using namespace Pinetime::Applications;
 
 HeartRateTask::HeartRateTask(Drivers::Hrs3300& heartRateSensor, Controllers::HeartRateController& controller)
-  : heartRateSensor {heartRateSensor}, controller {controller}, ppg{} {
+  : heartRateSensor {heartRateSensor}, controller {controller}, ppg {} {
 }
 
 void HeartRateTask::Start() {
   messageQueue = xQueueCreate(10, 1);
   controller.SetHeartRateTask(this);
+  lastRegularMeasurementTick = xTaskGetTickCount();
 
   if (pdPASS != xTaskCreate(HeartRateTask::Process, "Heartrate", 500, this, 0, &taskHandle))
     APP_ERROR_HANDLER(NRF_ERROR_NO_MEM);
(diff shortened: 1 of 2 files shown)


watertrainer:
I let it run on my watch for 8h and 40min which was charged to 93% and it is now on 80%. A night before gadgetbridge shows a power consumption of 3-4% in the same period. So this should definitely be a setting, I still think users should have the option to regularly let their heart rate be measured, especially in the farer future when sleep tracking is a thing.

kieranc:
This is certainly something that needs implementing, I can't speak for the quality of this implementation but I hope someone else can? In terms of adding a setting it's reasonably straightforward, you define them in src/components/Settings.h and increment the settingsVersion, then add a settings app to change the value. You could look at #1043 as an example, or the 12/24h, ble disable etc for a simple boolean toggle.
Nice work in any case, I think 60s is too short an interval but that can easily be tweaked, you could even add the interval as a setting if you wanted to!

Riksu9000:
I'll close this as I don't understand the intended behaviour here since it seems to start measuring without ever going into the app let alone starting a measurement, and the author acknowledges there are issues which haven't been resolved.

khimaros:
@Riksu9000 offering a proposal: add a new setting named "Heart Rate" with description "Measurement frequency:" with options options: 1) Never, 2) 60s, 3) 90s, 4) 120s

when the setting is enabled, a background service will be started which checks heart rate at this frequency. the most recent 30 measurements are preserved and transmitted to eg. Gadgetbridge.

is that a compelling end goal?

Issue #1088, opened 2022-04-14
Increasing confidence in alarm


medeyko:
### Verification

- [X] I searched for similar bug reports and found none was relevant.

### What happened?

Awful user experience with 1.9.0 alarm version

### What should happen instead?

Alrms should work like in 1.8.0 

### Reproduction steps

It's not a technical bug, but awful user experiece.

Just set up an alarm in 1.9.0, then sleep hard, and miss the alarm.

### More details?

After shifting to 1.9.0 I missed alarms all five times I tried.

I had absolutely no problems with 1.8.0 alarms.

So this limited vibration period introduced in 1.9.0 makes alarm not only unreliable, but **completely unusable** for hard-sleepers.

I had to fall back to 1.8.0 version.
Probably I will try to learn how to make a custom build, but no doubts I'm not the only hard-sleeper on the Earth.

### Version

1.9.0

### Companion app

-

Riksu9000:
Duplicate of #1086

Riksu9000:
https://github.com/InfiniTimeOrg/InfiniTime/issues/1086#issuecomment-1099466892

According to this I may have misunderstood, but are you sure it's not the same? How do you know if it rang?

medeyko:
Unfortunately, I can't be sure that it vibrated. But I all the times when I was already alert at the time of the alarms, it vibrated properly.

I guess, I should do more experiments. Ok, I will do them for about a week, and then will report back with the detailed stats.

Riksu9000:
Did you ever happen to check how long it had to vibrate until you woke up with 1.8.0 behaviour? What was the longest time?

medeyko:
I didn't check, as it worked fine, but definitely less than 10 minutes.
I will check this with 1.8.0 as well!

medeyko:
Preliminary information: it seems that my problem is indeed duplicate, and differences of outcome in case of "sleeping experiments" and "alert experiments" was caused by the distance beween the watch and the smartphone it is paired with.
It alert state they were near, but when I was sleeping, the smartphone was laying in another room, so the bluetooth connection was unstable.

As for the wake time, I registered 5 to 25 seconds, so 60 seconds seem to be too few to confidently alert hard-sleepers.
I would ask to increase it for 300 seconds. Confidence in alarms is very, very, very important. The lack of confidence in alarm makes nervous and harms quiality of sleep...

Riksu9000:
[...]

PR #1095, opened 2022-04-17
Alarm reliability and switch to freertos timers

rev22:
Addresses observations and issues reported in #1088 and possibly others.

This switches alarm, timer and motor controller to use FreeRTOS timers instead of the app_timer component.

The logic was changed slightly for MotorController, to avoid two timers being started simultaneosily but being executed in an unintended order.

An event is introduced to differentiate between time settings and time adjustments, this is to make alarms resilient to time zone or daylight saving changes, and to prevent companion apps from canceling alarms by periodically adjusting the time in the watch.

I have positively tested the changes, though testing was done in my personal branch mainly so far, and it's still preliminary.  I only own one PineTime at the moment, and testing alarms and recurring alarms can take days.


src/components/timer/TimerController.cpp
```diff
@@ -4,32 +4,29 @@
...
-void TimerController::Init() {
-  app_timer_create(&timerAppTimer, APP_TIMER_MODE_SINGLE_SHOT, TimerEnd);
+void TimerController::Init(System::SystemTask* systemTask) {
+  this->systemTask = systemTask;
+  timerAppTimer = xTimerCreate("timerAppTm", 1, pdFALSE, this, TimerEnd);
 }
 
 void TimerController::StartTimer(uint32_t duration) {
-  app_timer_stop(timerAppTimer);
+  xTimerStop(timerAppTimer,0);
```
Avamander, line 24: Please use clang-format to correctly format all modified files.

Avamander (changes requested):

src/components/timer/TimerController.cpp
rev22: Oops missed that space.  I'll try to reformat each commit.

rev22: Done now

medeyko:
It's great! Because alarm reliability is a very important thing... A unreliable alarm is may be even worse than absense of any alarm.

src/components/alarm/AlarmController.cpp
```diff
@@ -27,30 +27,45 @@ using namespace std::chrono_literals;
...
+  }
+
+  xTimerStop(alarmAppTimer, 0);
+  auto now = dateTimeController.CurrentDateTime();
+
+  if (now > alarmTime) {
+    xTimerChangePeriod(alarmAppTimer, 1, 0);
+  } else {
+    auto mSecToAlarm = std::chrono::duration_cast<std::chrono::milliseconds>(alarmTime - now).count();
+    xTimerChangePeriod(alarmAppTimer, APP_TIMER_TICKS(mSecToAlarm), 0);
```
Riksu9000, line 60: Use `pdMS_TO_TICKS()` instead of `APP_TIMER_TICKS()`. With this the rest of the app_timer includes should be possible to remove across all files.

src/components/alarm/AlarmController.cpp
```diff
@@ -27,30 +27,45 @@ using namespace std::chrono_literals;
...
+  void SetOffAlarm(TimerHandle_t xTimer) {
+    auto controller = static_cast<Pinetime::Controllers::AlarmController*>(pvTimerGetTimerID(xTimer));
+    controller->SetOffAlarmNow();
   }
 }
 
 void AlarmController::Init(System::SystemTask* systemTask) {
-  app_timer_create(&alarmAppTimer, APP_TIMER_MODE_SINGLE_SHOT, SetOffAlarm);
   this->systemTask = systemTask;
+  alarmAppTimer = xTimerCreate("alarmAppTm", 1, pdFALSE, this, SetOffAlarm);
```
Riksu9000, line 39: I would call this timer and variable just alarmTimer. `app_timer` is just what the nRF timer is called.

src/components/alarm/AlarmController.cpp
```diff
@@ -102,6 +118,12 @@ void AlarmController::SetOffAlarmNow() {
 
 void AlarmController::StopAlerting() {
   systemTask->PushMessage(System::Messages::StopRinging);
+}
+
+void AlarmController::OnStopRinging() {
+  if (state != AlarmState::Alerting) {
+    return;
+  }
```
Riksu9000, line 126: The SystemTask message `StopRinging` is supposed to just stop the MotorController ringing. Combining the behaviour with the Alarm is confusing. Why was this change made?

Riksu9000 (commented):
I find the "new time" and "adjust time" a bit confusing. I feel like there may be a way to simplify it. I'll have to think about it.

TimerController was recently updated to use FreeRTOS timers. Please fix the conflicts.

src/components/alarm/AlarmController.cpp
rev22: pdMS_TO_TICKS() can't be used in that context because the 32-bit arithmetic it performs would overflow for time durations over ~69 minutes.
APP_TIMER_TICKS uses 64-bit arithmetic and doesn't have that limitation.
Perhaps we can define a custom/function macro to replace it?

src/components/alarm/AlarmController.cpp
rev22: This postpones processing and makes sure that ScheduleAlarm is called for recurring alarms by the system task.

For reliability, in my modifications I tried to make sure that timers are started by the system task in response to system messages.

src/components/alarm/AlarmController.cpp
rev22: Agreed, brief names for timers are better now as they are in the class scope.

rev22:
Thank you @Riksu9000 for reviewing the code!  I have now rebased and updated the commits for the latest develop branch, and applied few cosmetical changes.

src/components/alarm/AlarmController.cpp
Riksu9000: I see. Maybe we could change the mSecToAlarm to seconds, so it would only need to be multiplied by `configTICK_RATE_HZ` to get the ticks? In an alarm this should be enough precision. In MotorController `pdMS_TO_TICKS` can be used.

src/components/alarm/AlarmController.cpp
Riksu9000: I'm not sure SystemTask is more or less reliable in this case, but requiring it to go through another hoop and making the code less readable could be considered to make it less reliable.

src/components/alarm/AlarmController.cpp
rev22: That would sound acceptable to me too, we may lose second-splitting accuracy, but save some bytes in the compiled code.  The alarm can be set only with minute resolution at present.

src/components/alarm/AlarmController.cpp
rev22: I have tried to enforce this discipline in my code because the FreeRTOS timer primitives are sensitive to their calling context, as they use mutexes and limited-size queues.
It is possible that this change is not necessary but it may require rather complex testing or analysis (stopAlerting currently has 5 call sites including event handlers, destructors and callbacks) and possibly more complex code changes.

Riksu9000: If there's concern that the queue can get full, the timer commands can be blocked for some time to let the timer task make room in the queue, or even read the return value, which is pdFAIL in case the command couldn't be sent. This is the intended and preferred way to handle this issue, however I wouldn't even be concerned about this right now, since the queue is length 32 and won't get full under normal circumstances.

rev22: I see, the timer queue shouldn't get full if the timer task is awaken often enough and not accidentally blocked by a callback.
What about two tasks trying to access the timer queue simultaneously?

Riksu9000: https://freertos.org/Embedded-RTOS-Queues.html It says here that queues are used as thread safe buffers, so I'm pretty sure any concerns are handled in FreeRTOS.

rev22: Indeed they are thread safe as they are protected by a mutex.  So what could happen is that one of the two tasks trying to access the queue in wait-free mode will fail.  (This may not happen if the access to the timer queue is further protected by a critical section to prevent preemption by the scheduler, but I have not been able to confirm that yet by looking at documentation or at FreeRTOS source code yet)

rev22: Using blocking mode for timer primitives as you hinted before may not be an option in this case because StopAlerting is called by an event handler, among other contexts, where blocking may not be recommended. 

Recovery options in case of failure have their own drawbacks or problems, for example even though you can introduce a new system message to recover from a failed timer command, you will not be able to push it the system task because StopAlerting is already sending a message to the system task, and the system task queue has size 1.

I think it's possible that the timer queue can get full in some cases and cause such failures, for example in times of high cpu load which can cause the timer task to execute callbacks slowly and become congested, such a congestion may last only a few millisecond and not be noticeable by the user, but it can still be relevant to correct code execution.  Delaying the execution of the timer commands and letting the system task wake up again gives a better chance for the timer task to process events too and to become less congested.

Riksu9000: The documentation says this:
> xBlockTime | Specifies the time, in ticks, that the calling task should be held in the Blocked state to wait for the reset command to be successfully sent to the timer command queue, should the queue already be full when xTimerReset() was called.

There's no mention that the commands couldn't be called simultaneously because of accessing the queue. I'm sure it would have been mentioned somewhere if it was an issue. I don't see any reason to doubt FreeRTOS working as expected in this case. Without digging too deep into it, I can see that in `xQueueGenericSend()` in queue.c, it enters critical before doing anything.

SystemTask queue is length 10.

> in times of high cpu load which can cause the timer task to execute callbacks slowly and become congested

https://freertos.org/implementation/a00008.html I don't see how that could happen, but if there's still concern, let's move that to a new issue or PR, so we can move forward with this PR.

rev22: Thank you for researching about FreeRTOS and correcting about the size of the system task, somehow I misread that in the source code,  About FreeRTOS that's great because it makes timers more reliable and easy to use in wait-free mode, as they are in the current codebase.
As far as I know nobody understood yet why the app_timer component was malfunctioning for alarms.  The implementation of app_timer uses FreeRTOS timer commands internally.  Perhaps the timer queue was getting full.  For averting such issues I think that trying to space out timer commands using the system message queue is a good discipline, and it's what worked in my tests while developing this solution.  However I also understand concerns about code clarity.  It's not easy to make asynchronous code clear, but I tried to keep code changes minimal.

rev22: Since it's possible to send more than one message to the system task, I think I can try to develop a more clear solution than the current one, by introducing a RESCHEDULE_ALARM event.  This should also not increase the firmware size, compared to the present solution.  Do you think it's a good idea?

rev22: That seems to explain the observed behaviour, it seemed that alarms were being disabled after ScheduleAlarm occurred when the alarm was already set, at least in some cases.

Higher priority for the timer task, compared to the task(s) sending timer commands, should also prevent in general the timer queue from getting full, as long as timer callbacks are not performing blocking operations.

Perhaps there is an existing rationale for the timer and the system task having the same priority?

If not, even if we may need to increase the number of allowed task priorities, it may become safe to reduce the length of the timer task queue, and hopefully achieve a slight reduction of RAM usage.

Riksu9000: I did another experiment where I read the return value of `app_timer_start()` and set off the alarm immediately depending on the error. If the queue is full, it returns `NRF_ERROR_NO_MEM`. If it thinks the timer is alrady running, it returns NRF_SUCCESS. When changing the time, the alarm wasn't set off, which proves that app_timer_start returned NRF_SUCCESS, queue was not full, yet it didn't alert.
```
-  app_timer_start(alarmAppTimer, APP_TIMER_TICKS(mSecToAlarm), this);
+  if (app_timer_start(alarmAppTimer, APP_TIMER_TICKS(mSecToAlarm), this) != NRF_SUCCESS) {
+    SetOffAlarmNow();
+  }
```

So with this in mind, please revert the extra changes from this PR and strictly focus on replacing app_timers. Let's also move the time adjustment change to a new PR. Thanks.

rev22:
I have rebased the code again and excluded the commit about resilience to time adjustments, so that it's possible to focus better on the switch to FreeRTOS timers (per code review).   A new commit was added that may address code clarity.

src/components/alarm/AlarmController.cpp
```diff
@@ -102,14 +99,14 @@ void AlarmController::SetOffAlarmNow() {
...
-  systemTask->PushMessage(System::Messages::StopRinging);
-
   // Alarm state is off unless this is a recurring alarm
   if (recurrence == RecurType::None) {
     state = AlarmState::Not_Set;
   } else {
-    state = AlarmState::Set;
     // set next instance
-    ScheduleAlarm();
+    systemTask->PushMessage(System::Messages::ScheduleAlarm);
```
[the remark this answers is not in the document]
rev22: Alright, I replaced that with a direct call to ScheduleAlarm in this PR.

src/components/motor/MotorController.cpp
```diff
@@ -1,43 +1,40 @@
...
-  app_timer_create(&longVibTimer, APP_TIMER_MODE_REPEATED, Ring);
+  shortVib = xTimerCreate("shortVib", 1, pdFALSE, nullptr, StopMotor);
+  longVib = xTimerCreate("longVib", 1, pdTRUE, this, Ring);
 }
 
-void MotorController::Ring(void* p_context) {
-  auto* motorController = static_cast<MotorController*>(p_context);
+void MotorController::Ring(TimerHandle_t xTimer) {
+  auto motorController = static_cast<MotorController*>(pvTimerGetTimerID(xTimer));
+  xTimerChangePeriod(motorController->longVib, pdMS_TO_TICKS(1000), 0);
```
Riksu9000, line 18: `longVib` can be initialized to the correct period and then changing the period afterwards can be avoided. `StartRinging()` needs to be tweaked slightly

Riksu9000 (changes requested):

src/components/motor/MotorController.cpp
rev22: I can make that change, but that would delay by one second the start of the vibration.
An alternative could be starting the two timers at the same time, but that didn't work in my previous testing, I think it's because of unintended execution order when you start two timers simultaneously.

Riksu9000: The two timers should be started simultaneously like before. It seems to work fine for me. `Ring(this)` can't be used, because the function now expects a `TimerHandle_t` instead of a simple pointer. Just duplicating the RunForDuration(50) should be good enough for now, or alternatively a new function in an unnamed namespace can be created.

rev22: I see, can you push the modification that you tested?

Riksu9000 (approved):

src/components/alarm/AlarmController.cpp
```diff
@@ -102,14 +98,13 @@ void AlarmController::SetOffAlarmNow() {
 }
 
 void AlarmController::StopAlerting() {
-  systemTask->PushMessage(System::Messages::StopRinging);
-
   // Alarm state is off unless this is a recurring alarm
   if (recurrence == RecurType::None) {
     state = AlarmState::Not_Set;
   } else {
-    state = AlarmState::Set;
```
NeroBurner, line 111: I'm unclear why is this state setting removed? Doesn't it do anything?

src/components/alarm/AlarmController.cpp
```diff
@@ -102,14 +98,13 @@ void AlarmController::SetOffAlarmNow() {
...
     // set next instance
     ScheduleAlarm();
   }
+
```
NeroBurner, line 108: two newlines, I think we only need one line

```suggestion
```

src/components/alarm/AlarmController.cpp
```diff
@@ -81,8 +76,9 @@ void AlarmController::ScheduleAlarm() {
 
   // now can convert back to a time_point
   alarmTime = std::chrono::system_clock::from_time_t(std::mktime(tmAlarmTime));
-  auto mSecToAlarm = std::chrono::duration_cast<std::chrono::milliseconds>(alarmTime - now).count();
-  app_timer_start(alarmAppTimer, APP_TIMER_TICKS(mSecToAlarm), this);
+  auto secondsToAlarm = std::chrono::duration_cast<std::chrono::seconds>(alarmTime - now).count();
+  xTimerChangePeriod(alarmTimer, secondsToAlarm * configTICK_RATE_HZ, 0);
```
NeroBurner, line 80: In MotorController the `pdMS_TO_TICKS` macro is used, and here the `configTICK_RATE_HZ` define is used directly. I think we should use the same conversion method in both cases

```suggestion
  auto mSecToAlarm = std::chrono::duration_cast<std::chrono::milliseconds>(alarmTime - now).count();
  xTimerChangePeriod(alarmTimer, pdMS_TO_TICKS(mSecToAlarm), 0);
```

NeroBurner (changes requested):

src/components/alarm/AlarmController.cpp
Riksu9000: This is intentional. See https://github.com/InfiniTimeOrg/InfiniTime/pull/1095#discussion_r871873957

src/components/alarm/AlarmController.cpp
Riksu9000: The state is also set in `ScheduleAlarm()`

src/components/alarm/AlarmController.cpp
Riksu9000: I'm indifferent about this. There are no guidelines about this.

NeroBurner (approved):

PR #1097, opened 2022-04-18
Auto generate font

yehoshuapw:
closes #964 (supersedes)

this adds a script `src/displayapp/fonts/generate.py` to generate the font files (using lv_font_conv) from a singular config file (`src/displayapp/fonts/fonts.json`).

the json has each font, and it's relevant options (size,bpp,compress, and "patches") and sources.
in addition there's also a "feature" feature, allowing extra sources with a extra flag (and the config has an example usage of adding hebrew, which won't do anything unless `-e hebrew` is given)

currently, this manages to generate exactly the same as what is already in place.

TODO:
integrate this into the CMake (and add lv_font_conv to the build images).
(my CMake-fu is not that strong, so I'll get to that tomorrow - but opening this to get comments about the rest of it from anyone interested)


yehoshuapw:
@Riksu9000 i'd like to get your thoughts about whats already here. thanks!

src/displayapp/fonts/generate.py
```diff
@@ -0,0 +1,80 @@
...
+        fonts_to_run = []
+        for font in args.font:
+            if font.endswith('.c'):
+                fonts_to_run.append(font[:-2])
+            else:
+                sys.exit(f'requested to remove .c extension, but {font} does not have it.')
+
+    for (name,font) in data['fonts'].items():
+        if fonts_to_run and name not in fonts_to_run:
+            continue
```
Riksu9000, line 61: I think there should be a warning when the font that has been entered as an option can't be found. This could happen just as a typo. Perhaps this could loop through the fonts_to_run instead of all fonts in the config.

src/displayapp/fonts/generate.py
```diff
@@ -0,0 +1,80 @@
...
+            sys.exit(f'Error: the requested feature {enabled_feature} does not exist in {args.config}.')
+
+    fonts_to_run = args.font
+    if args.removec and args.font:
+        fonts_to_run = []
+        for font in args.font:
+            if font.endswith('.c'):
+                fonts_to_run.append(font[:-2])
+            else:
+                sys.exit(f'requested to remove .c extension, but {font} does not have it.')
```
Riksu9000, line 57: This seems useful, but I think this could be automated and the argument removed.

src/displayapp/fonts/fix_jetbrains_mono_bold_20_zero.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python
```
Riksu9000, line 1: Maybe we could use a normal patch file for patches.

src/displayapp/fonts/fonts.json
```diff
@@ -0,0 +1,90 @@
...
+   "features": {
+      "hebrew": {
+         "jetbrains_mono_bold_20": [
+            {
+               "font": "SimpleCLM-Medium.ttf",
+               "range": "0x05D0-0x05EA,0x05F3,0x05F4"
+            }
+         ]
+      }
+   }
```
Riksu9000, line 89: Features should probably be inside the corresponding font section

Riksu9000 (commented):
I like this solution. The README should be updated as a part of this PR by removing the old instructions and writing something about this new method.

src/displayapp/fonts/generate.py
yehoshuapw: Added a check (which is only a warning, not an error).
not looping only thorough "fonts_to_run" since by default (without `-f`s) will run over all the fonts in the config file.

src/displayapp/fonts/generate.py
yehoshuapw: Indeed - and this is exactly such an automation. do you think it will be more clear if done inside the CMake files, and not here? (I added it in it's own commit after deciding it will make the CMake more readable - if you disagree, I'll remove it)

src/displayapp/fonts/fix_jetbrains_mono_bold_20_zero.py
yehoshuapw: this will work only so long as the relevant lines stay in the same place, and look exactly the same (including the comment, which just changed between versions of lv_font_conv), which is why I made this over-complicated patcher script.

(I almost made it even more of a mess, by allowing for space changes and such).

so I recommend to leave it like this, but this really is a mess - and I'll abide by your decision.

src/displayapp/fonts/fonts.json
yehoshuapw: that will simplify the script, but will make the config less readable. like this the "features" can be clearly ignored by a human reading the config.

yehoshuapw: Added commit which makes this change.

src/displayapp/fonts/generate.py
yehoshuapw: done

src/displayapp/fonts/generate.py
Riksu9000: `fonts_to_run` could also by default contain all the fonts in the config file.

src/displayapp/fonts/fix_jetbrains_mono_bold_20_zero.py
yehoshuapw: nvm about the comment (the diff does not need to include parts before and after), but the lines do still matter - which can change anytime a new symbol is added which ends up before.

src/displayapp/fonts/generate.py
Riksu9000: I meant it more like the script could automatically strip the extension, but I guess we can decide what to do when we have some CMake code to review. I see it as a possibiliity for this to be merged even without any CMake integration initially depending on the situation.

yehoshuapw: changed to that method (strip .c if it's there)

src/displayapp/fonts/fonts.json
yehoshuapw: it won't happen often, since most requests for font changes are either for a new language (which should be added in the same place as the hebrew example),
or for new watchfaces - which also use at most 1 font, and often a completely new one, and that need it (so it's not a toggle-able feature)

src/displayapp/fonts/README.md
```diff
@@ -6,94 +6,40 @@
 
 ## Generate the fonts:
 
-* Open the [LVGL font converter](https://lvgl.io/tools/fontconverter)
-* Enter the settings for the font that you wish to convert
-* Click on Convert, download the file and place it in `src/DisplayApp/Fonts`
+* Make sure you have installed lv_font_conv, and it's in the PATH
+  * `npm i lv_font_conv`
```
yehoshuapw, line 10: i'm not sure if adding npm instructions is necessary - or just annoying.

yehoshuapw:
and now the font .c files are generated at compile time with CMake.

Riksu9000:
I've been thinking that the font "features" probably isn't going to be very useful to us, since we only ship a single firmware, and that's likely not going to be changing. The other use case would be if fork maintainers pushed their changes in this repo, which I don't think should be our responsibility to handle either.

JF002 (changes requested):
Thank you for this PR, that's definitely something I want to add tot he project!

The generation of the fonts in CMake and then in Python and in Github Actions look good to me!

If you don't mind, I think we should make 2 addition:
 - Add  `lv_font_conv` (and how to install it via `npm`) as a needed dependency to build the project in `doc/buildAndProgram.md` as the project won't build without it.
 - Add the installation of `npm` and `lv_font_conv` in the Dockerfile so we can build the fonts inside the docker container. Note than I plan on switching the CI to docker so that we won't have to duplicate those changes in the future (https://github.com/InfiniTimeOrg/InfiniTime/pull/1045).
 - Oh and also maybe specify which version of `lv_font_conv` we support, and enforce this version in Docker/Action workflows to avoid broken builds because of a change in the tool ?

I can help with those points if you want to ;)

src/displayapp/fonts/README.md
Riksu9000: I think this entire section can be removed, because you're not intended to run `generate.py` manually. Since `lv_font_conv` is now a build dependency, it should just be added [here](https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/doc/buildAndProgram.md) with a link to the [repo](https://github.com/lvgl/lv_font_conv) with instructions being there.

Riksu9000 (commented):
I noticed that changing the font configuration will regenerate all the fonts, even if only one was changed. Not a big issue, but maybe worth considering if something can be done about that.

yehoshuapw:
> I've been thinking that the font "features" probably isn't going to be very useful to us, since we only ship a single firmware, and that's likely not going to be changing. The other use case would be if fork maintainers pushed their changes in this repo, which I don't think should be our responsibility to handle either.

since extra languages is a commonly requested feature, I decided this will make things easier (until a real solution is implemented), but I see your point.
How about this: adding a flag to add "features" from a separate json file? then the script is aware of such things, but the main files don't need to be changed.

yehoshuapw:
> I noticed that changing the font configuration will regenerate all the fonts, even if only one was changed. Not a big issue, but maybe worth considering if something can be done about that.

the options as I see them are:

* a config file per font
* some method to non regenerate if some part was not changed
* a single config file, and regenerate all the fonts.

since the generation is pretty fast, and changing the config does not happen often - I don't like option 1.
option 2 is doable, but will make it more complicated - and may not get much of an increase in speed.

therefore I just left it at a plain single-file. do you think one of the others is better? or do you have a better idea?
thanks!

src/displayapp/fonts/README.md
```diff
@@ -6,94 +6,35 @@
 
 ## Generate the fonts:
 
-* Open the [LVGL font converter](https://lvgl.io/tools/fontconverter)
-* Enter the settings for the font that you wish to convert
-* Click on Convert, download the file and place it in `src/DisplayApp/Fonts`
+* Make sure you have installed lv_font_conv, and it's in the PATH
+  * See `doc/buildAndProgram.md`
+* run the `generate.py` script, while in the folder.
+  * it has a `-h` flag to show usage, nontheless: `python generate.py fonts.json`
```
Riksu9000, line 12: I think this entire section can be removed now that the dependency note has been added elsewhere and generate.py shouldn't be run manually.

Riksu9000:
> > I've been thinking that the font "features" probably isn't going to be very useful to us, since we only ship a single firmware, and that's likely not going to be changing. The other use case would be if fork maintainers pushed their changes in this repo, which I don't think should be our responsibility to handle either.
> 
> since extra languages is a commonly requested feature, I decided this will make things easier (until a real solution is implemented), but I see your point. How about this: adding a flag to add "features" from a separate json file? then the script is aware of such things, but the main files don't need to be changed.

It's a neat feature, but adding the extra characters and keeping it up to date with git manually isn't difficult either, and it's already much easier than before since the font is generated automatically. I don't think we'd want to maintain a feature that we don't use.

yehoshuapw:
does anyone know if `libssl-dev` as opposed to `libssl1.0-dev` matters in the docker?

src/displayapp/fonts/fonts.json
```diff
@@ -0,0 +1,78 @@
...
+            "range": "0x20-0x7e, 0x410-0x44f"
+         },
+         {
+            "file": "FontAwesome5-Solid+Brands+Regular.woff",
+            "range": "0xf293, 0xf294, 0xf244, 0xf240, 0xf242, 0xf243, 0xf241, 0xf54b, 0xf21e, 0xf1e6, 0xf54b, 0xf017, 0xf129, 0xf03a, 0xf185, 0xf560, 0xf001, 0xf3fd, 0xf069, 0xf1fc, 0xf45d, 0xf59f, 0xf5a0, 0xf029, 0xf027, 0xf028, 0xf6a9, 0xf04b, 0xf04c, 0xf048, 0xf051, 0xf095, 0xf3dd, 0xf04d, 0xf2f2, 0xf024, 0xf252, 0xf569, 0xf201, 0xf06e, 0xf015"
+         }
+      ],
+      "bpp": 1,
+      "size": 20,
+      "patches": [["patch", "{file}", "{file}_zero.patch"]]
```
Riksu9000, line 15: I think supplying just the name of the patch file should be enough here. It doesn't need to be able to run any command.

src/displayapp/fonts/README.md
```diff
@@ -4,96 +4,29 @@
...
-* Click on Convert, download the file and place it in `src/DisplayApp/Fonts`
-
 ### How to add new symbols:
 
 * Browse [this cheatsheet](https://fontawesome.com/cheatsheet/free/solid) and pick symbols
-* For each symbol, add its hex code (0xf641 for the 'Ad' icon, for example) to the *Range* list (Remember to keep this
-  readme updated with newest range list)
-* Convert this hex value into a UTF-8 code
+* For each symbol, add its hex code (0xf641 for the 'Ad' icon, for example) to the *Range* list (or the symbol list when its simple enough) in the `fonts.json` file
+* * Convert this hex value into a UTF-8 code
```
Riksu9000, line 11: There's an extra asterisk.

```suggestion
* Convert this hex value into a UTF-8 code
```

Riksu9000 (commented):
I think the script should check that `lv_font_conv` is installed before running and show an error if it isn't, because otherwise it will show many errors, which may be confusing.

src/displayapp/fonts/fonts.json
yehoshuapw: done. (I left in that if the script fails using a simple `patch`, it will try using as a whole command)

yehoshuapw:
> because otherwise it will show many errors, which may be confusing.

now it will check and give a clear error message, however it will still display the same message for each - CMake runs each font by itself.

src/displayapp/fonts/fonts.json
Riksu9000: I don't think the `{file}` overriding is useful for filenames here, and I'm not sure falling back to executing as a command is necessary. I'd rather remove features that aren't necessary since they won't get maintained.

Riksu9000: Good question. As far as I can tell, that capability just requires one extra for loop, so it's not that bad. If a font required two changes, my immediate thought would be that they should be two separate patches, but I'm not 100% sure.

yehoshuapw: then I'll leave that in.

Riksu9000 (approved):
For InfiniSim to compile, generate.py currently must be run manually. I think that may be something that needs to be done on InfiniSim's side. Ping @NeroBurner.

I can't comment much on the CMake changes, since I'm not so familiar with that.

Other than that I really like this and it works well.

yehoshuapw:
> For InfiniSim

Added a bit of changes here (use CMAKE_CURRENT_LIST_DIR to reach things, instead of assuming paths),
and included this CMakeLists.txt there (and changed the GLOB there to use the list inside the CMake here, since globbing on filenames won't work anymore)

yehoshuapw:
> > For InfiniSim

Also, I did (except when I forgot once) compile before pushing, and here watching the checks status, so it's weird that the simulator check succeeded.

so I had a look, and in the simulator workflow here I manually generated the scripts - I just removed it, so it will fail here too (until InfiniTimeOrg/InfiniSim#28 is merged).
(so both these PRs should be merged about the same time)

src/displayapp/fonts/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+   open_sans_light)
+find_program(LV_FONT_CONV "lv_font_conv" NO_CACHE REQUIRED)
+configure_file(${CMAKE_CURRENT_LIST_DIR}/jetbrains_mono_bold_20.c_zero.patch
+   displayapp/fonts/jetbrains_mono_bold_20.c_zero.patch COPYONLY)
+foreach(FONT ${FONTS})
+   set_source_files_properties(displayapp/fonts/${FONT}.c
+      PROPERTIES GENERATED TRUE)
+
+   add_custom_command(OUTPUT displayapp/fonts/${FONT}.c
+      COMMAND python ${CMAKE_CURRENT_LIST_DIR}/generate.py
```
JF002, line 14:
```suggestion
      COMMAND python3 ${CMAKE_CURRENT_LIST_DIR}/generate.py
```

JF002 (changes requested):
This looks totally fine to me! 
There's just 1 last issue : the build in the docker container does not work properly. See this log message:

```
[ 35%] Generating displayapp/fonts/jetbrains_mono_42.c
[ 35%] Generating displayapp/fonts/jetbrains_mono_76.c
[ 35%] Generating displayapp/fonts/jetbrains_mono_76.c
[ 35%] Generating displayapp/fonts/lv_font_navi_80.c
[ 35%] Generating displayapp/fonts/open_sans_light.c
[ 35%] Generating displayapp/fonts/open_sans_light.c
[ 35%] Generating displayapp/fonts/jetbrains_mono_extrabold_compressed.c
[ 35%] Generating displayapp/fonts/jetbrains_mono_extrabold_compressed.c
[ 35%] Generating displayapp/fonts/jetbrains_mono_bold_20.c
[ 35%] Generating displayapp/fonts/jetbrains_mono_42.c
[ 35%] Generating displayapp/fonts/lv_font_sys_48.c
[ 36%] Generating displayapp/fonts/jetbrains_mono_bold_20.c
[ 36%] Generating displayapp/fonts/lv_font_navi_80.c
  File "  File "/sources/src/displayapp/fonts/generate.py/sources/src/displayapp/fonts/generate.py", line ", line 21
21
    def gen_lvconv_line(dest: str, size: int, bpp: int, sources: typing.List[Source], compress:bool=False):
  File "/sources/src/displayapp/fonts/generate.py  File "/sources/src/displayapp/fonts/generate.py", line 21
  File "        def gen_lvconv_line(dest: str, size: int, bpp: int, sources: typing.List[Source], compress:bool=False):
                              ", line 21
    def gen_lvconv_line(dest: str, size: int, bpp: int, sources: typing.List[Source], compress:bool=False):
/sources/src/displayapp/fonts/generate.py      def gen_lvconv_line(dest: str, size: int, bpp: int, sources: typing.List[Source], compress:bool=False):
         File "  File "/sources/src/displayapp/fonts/generate.py", line  /sources/src/displayapp/fonts/generate.py21", line 
     21 def gen_lvconv_line(dest: str, size: int, bpp: int, sources: typing.List[Source], compress:bool=False):

           File "/sources/src/displayapp/fonts/generate.py", line ", line 21
 21    def gen_lvconv_line(dest: str, size: int, bpp: int, sources: typing.List[Source], compress:bool=False):

                                             def gen_lvconv_line(dest: str, size: int, bpp: int, sources: typing.List[Source], compress:bool=False):
                                ^
  SyntaxError :   invalid syntax 
     ^
SyntaxError                       ^
    SyntaxError        ^
     File "SyntaxError :  invalid syntax/sources/src/displayapp/fonts/generate.py
 ", line  21 
     def gen_lvconv_line(dest: str, size: int, bpp: int, sources: typing.List[Source], compress:bool=False):
      :   def gen_lvconv_line(dest: str, size: int, bpp: int, sources: typing.List[Source], compress:bool=False):
invalid syntax  
:  invalid syntax
          ^
    ^
SyntaxError^
: SyntaxErrorinvalid syntaxSyntaxError: 
:   File "invalid syntaxinvalid syntax
/sources/src/displayapp/fonts/generate.py
", line 21
    def gen_lvconv_line(dest: str, size: int, bpp: int, sources: typing.List[Source], compress:bool=False):
                      File "   /sources/src/displayapp/fonts/generate.py ", line   21 
     ^
def gen_lvconv_line(dest: str, size: int, bpp: int, sources: typing.List[Source], compress:bool=False):
                            ^
SyntaxError: invalid syntax
  File "/sources/src/displayapp/fonts/generate.py", line 21
    def gen_lvconv_line(dest: str, size: int, bpp: int, sources: typing.List[Source], compress:bool=False):
                            ^
SyntaxError: invalid syntax
                   ^
SyntaxError: invalid syntax
src/CMakeFiles/pinetime-mcuboot-app.dir/build.make:65: recipe for target 'src/displayapp/fonts/jetbrains_mono_extrabold_compressed.c' failed
make[2]: *** [src/displayapp/fonts/jetbrains_mono_extrabold_compressed.c] Error 1
make[2]: *** Waiting for unfinished jobs....
    src/CMakeFiles/pinetime-app.dir/build.make:73: recipe for target 'src/displayapp/fonts/jetbrains_mono_76.c' failed
 SyntaxErrormake[2]: *** [src/displayapp/fonts/jetbrains_mono_76.c] Error 1
    :   File " [ 36%] Generating displayapp/fonts/lv_font_sys_48.c
invalid syntaxmake[2]: *** Waiting for unfinished jobs....
/sources/src/displayapp/fonts/generate.py 
src/CMakeFiles/pinetime-mcuboot-app.dir/build.make:73: recipe for target 'src/displayapp/fonts/jetbrains_mono_76.c' failed
make[2]: *** [src/displayapp/fonts/jetbrains_mono_76.c] Error 1
", line 21 
    def gen_lvconv_line(dest: str, size: int, bpp: int, sources: typing.List[Source], compress:bool=False):
    src/CMakeFiles/pinetime-app.dir/build.make:77: recipe for target 'src/displayapp/fonts/jetbrains_mono_42.c' failed
          make[2]: *** [src/displayapp/fonts/jetbrains_mono_42.c] Error 1
               ^
    SyntaxError :   invalid syntax 
       ^
SyntaxError: invalid syntax
src/CMakeFiles/pinetime-app.dir/build.make:65: recipe for target 'src/displayapp/fonts/jetbrains_mono_extrabold_compressed.c' failed
make[2]: *** [src/displayapp/fonts/jetbrains_mono_extrabold_compressed.c] Error 1
src/CMakeFiles/pinetime-mcuboot-app.dir/build.make:77: recipe for target 'src/displayapp/fonts/jetbrains_mono_42.c' failed
make[2]: *** [src/displayapp/fonts/jetbrains_mono_42.c] Error 1
src/CMakeFiles/pinetime-mcuboot-app.dir/build.make:61: recipe for target 'src/displayapp/fonts/lv_font_navi_80.c' failed
make[2]: *** [src/displayapp/fonts/lv_font_navi_80.c] Error 1
src/CMakeFiles/pinetime-app.dir/build.make:69: recipe for target 'src/displayapp/fonts/jetbrains_mono_bold_20.c' failed
make[2]: *** [src/displayapp/fonts/jetbrains_mono_bold_20.c] Error 1
src/CMakeFiles/pinetime-app.dir/build.make:85: recipe for target 'src/displayapp/fonts/open_sans_light.c' failed
make[2]: *** [src/displayapp/fonts/open_sans_light.c] Error 1
src/CMakeFiles/pinetime-app.dir/build.make:81: recipe for target 'src/displayapp/fonts/lv_font_sys_48.c' failed
make[2]: *** [src/displayapp/fonts/lv_font_sys_48.c] Error 1
src/CMakeFiles/pinetime-mcuboot-app.dir/build.make:85: recipe for target 'src/displayapp/fonts/open_sans_light.c' failed
make[2]: *** [src/displayapp/fonts/open_sans_light.c] Error 1
src/CMakeFiles/pinetime-mcuboot-app.dir/build.make:69: recipe for target 'src/displayapp/fonts/jetbrains_mono_bold_20.c' failed
make[2]: *** [src/displayapp/fonts/jetbrains_mono_bold_20.c] Error 1
src/CMakeFiles/pinetime-app.dir/build.make:61: recipe for target 'src/displayapp/fonts/lv_font_navi_80.c' failed
make[2]: *** [src/displayapp/fonts/lv_font_navi_80.c] Error 1
CMakeFiles/Makefile2:409: recipe for target 'src/CMakeFiles/pinetime-app.dir/all' failed
make[1]: *** [src/CMakeFiles/pinetime-app.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
  File "/sources/src/displayapp/fonts/generate.py", line 21
    def gen_lvconv_line(dest: str, size: int, bpp: int, sources: typing.List[Source], compress:bool=False):
                            ^
SyntaxError: invalid syntax
src/CMakeFiles/pinetime-mcuboot-app.dir/build.make:81: recipe for target 'src/displayapp/fonts/lv_font_sys_48.c' failed
make[2]: *** [src/displayapp/fonts/lv_font_sys_48.c] Error 1
CMakeFiles/Makefile2:214: recipe for target 'src/CMakeFiles/pinetime-mcuboot-app.dir/all' failed
make[1]: *** [src/CMakeFiles/pinetime-mcuboot-app.dir/all] Error 2
```

I think this is caused by the version of Python : the docker image uses Python 2.7.17, while my computer is running Python 3.10.

Do you think it would cause any issue if we called `python3` instead of `python` in the cmake command that generates the font?

src/displayapp/fonts/CMakeLists.txt
yehoshuapw: the typing should be solved by using python3 instead of 2, but this will break some newer linux builds - which don't have a "python3" executable anymore.
Also, in old enough distributions perhaps the python3 is not new enough to have f-strings,

so Instead I'll convert things to version agnostic code.

yehoshuapw: that worked, but ended up being a bit more of a mess then expected,
so I changed to use python3, and when it's not found to use python. (if there's only python2, it will find it - and fail when running.)

JF002: Or maybe we should update the version of the base image of the docker (to a version that uses python3 by default)? 18.04 is probably quite old.

yehoshuapw: changed to use ubuntu 20.04 (the current LTS), which does indeed solve the issue without any of the other parts (which I removed)

JF002:
@yehoshuapw I've just tried the new docker image based on ubunutu 20.04, and I still got the same issues with the Python script. 
It looks like the container based on Ubuntu 20.04 is still based on Python 2.7:

 ```
infinitime@cacd01d36e16:/$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.4 LTS
Release:        20.04
Codename:       focal
infinitime@cacd01d36e16:/$ python --version
Python 2.7.18
infinitime@cacd01d36e16:/$ python3 --version
Python 3.8.10
```
I'm not sure why it works for you and not for me... Did I miss anything?

PS : Please excuse me for the time it takes to finish this review... Those little details about Docker and the CI are taking more time than I expected...

JF002:
> Ubuntu 22.04 is released and it has a package called `python-is-python3` which symlinks python to python3. Would this be acceptable?

Yes, probably! As long as it still builds the firmware and generates the DFU images, that fine for me :)

yehoshuapw:
> I'm not sure why it works for you and not for me... Did I miss anything?

I must have made some mistake (left the python2 version? left .c files in the build dir?), because indeed the image I built here does have the "python" version as 2.7.

updated as you suggested, and seems to work. (building the image and then infinitime, from the start)

(I'm also wondering if any dist except for arch (which I don't even use) doesn't have a "python3" anymore without installing some extra things. nonetheless having a newer version which works tends to be worth it)

JF002:
> (I'm also wondering if any dist except for arch (which I don't even use) doesn't have a "python3" anymore without installing some extra things. nonetheless having a newer version which works tends to be worth it)

On Manjaro, I still have python and python3, but since python 3 is support to replace python 2, it's probably better to assume that python is Python 3

> updated as you suggested, and seems to work. (building the image and then infinitime, from the start)

Thanks, that looks very good to me :)

JF002 (approved):

JF002:
Now... How do we handle that in InfiniSim ?

yehoshuapw:
> Now... How do we handle that in InfiniSim ?

I opened a PR there too, https://github.com/InfiniTimeOrg/InfiniSim/pull/28 which seems to work.

NeroBurner:
I think nearly identical to here. Need to install npm and the lv-font generator in the CI. And need to add those generated fonts to InfiniSim. I'll need to have a thorough look at this PR and see how and where to fonts.c files are generated.
For my personal guideline, every generated file should be in the `${CMAKE_BINARY_DIR}` and not in the source directory.

Maybe I can make the font-generating CMakeLists.txt file generic enough so I just need to do `add_subdirectory` in InfiniSim and link to a infinisim-fonts target or something like that :thinking:

yehoshuapw:
> where to fonts.c files are generated. 
in `build/displayapp/fonts` (or wherever the build dir is)

> Maybe I can make the font-generating CMakeLists.txt file generic enough so I just need to do `add_subdirectory` in InfiniSim and link to a infinisim-fonts target or something like that thinking

I'm new at CMake, so go for it.

JF002:
I'll merge this PR as it's working as expected in InfiniTime. We'll follow up with the integration in InfiniSim in anothe PR (probably in the InfiniSim project).

> Maybe I can make the font-generating CMakeLists.txt file generic enough so I just need to do add_subdirectory in InfiniSim and link to a infinisim-fonts target or something like that thinking

It should be possible to create a 'font' target or a GenerateFont() function that we'll be able to include in InfiniSim, so we don't have to duplicate the whole code needed to generate the fonts.

@yehoshuapw I'm happy to see you've already started work in InfiniSim too! Thanks again for this nice PR! :1st_place_medal:

PR #1100, opened 2022-04-20
Force full screen refresh with a rounder function

Riksu9000:
When the screen switches, the full screen needs to be refreshed for the
hardware scrolling to work. This was enforced with backgroundLabels, but
is simpler to do with a rounder function.


JF002 (approved):
Nice, I didn't know about this rounder function! This is quite smart and allows us to remove all those counter-intuitive "background labels" in all apps!

PR #1103, opened 2022-04-22
Tweak gamma on ST7789 and match UI colors

Riksu9000:
This change will increase the color accuracy of the PineTime and make UI development with the simulator easier.

I recreated #971 to minimize unrelated changes to hopefully make reviewing simpler.

I think the gamma should fixed as soon as possible before any significant UI rework.

I've tried to keep the colors looking the same. Anyone with two PineTimes can help by highlighting differences caused by this change.

### DIFF
--- a/src/displayapp/Colors.cpp
+++ b/src/displayapp/Colors.cpp
@@ -10,17 +10,17 @@ lv_color_t Pinetime::Applications::Convert(Pinetime::Controllers::Settings::Colo
     case Pinetime::Controllers::Settings::Colors::Gray: return LV_COLOR_GRAY;
     case Pinetime::Controllers::Settings::Colors::Black: return LV_COLOR_BLACK;
     case Pinetime::Controllers::Settings::Colors::Red: return LV_COLOR_RED;
-    case Pinetime::Controllers::Settings::Colors::Maroon: return LV_COLOR_MAROON;
+    case Pinetime::Controllers::Settings::Colors::Maroon: return lv_color_lighten(LV_COLOR_MAROON, LV_OPA_20);
     case Pinetime::Controllers::Settings::Colors::Yellow: return LV_COLOR_YELLOW;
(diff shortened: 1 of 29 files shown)


Avamander (approved):
Let's merge, if any subtle problems appear we have time to react.

(Anyone merging please use "Rebase & Merge", it's much nicer)

kieranc:
I've just tested side-by-side with 2 watches and the most obvious change is that orange is now orange, rather than slightly dark yellow. I have a few photos but it's difficult to capture, this is about the best I got, develop on the left, this PR on the right.


PR 1103

Riksu9000:
Indeed, orange was intentionally "changed" by not changing it, so it looks the way it always should've IMO.

I agree using `LV_COLOR_MAKE` everywhere isn't great in the long run, but using the LVGL colors isn't either. I'm not sure what the best way to store the colors would be. The biggest issue being that the theme file should also use the same color definitions, but it is currently a C file, so it can't pull the colors from a C++ file. I can create a new file with just color `#defines` if that's acceptable. Otherwise we can change these things later once we figure out the best way to do it.

JF002:
> I agree using LV_COLOR_MAKE everywhere isn't great in the long run, but using the LVGL colors isn't either. I'm not sure what the best way to store the colors would be. The biggest issue being that the theme file should also use the same color definitions, but it is currently a C file, so it can't pull the colors from a C++ file. I can create a new file with just color #defines if that's acceptable. Otherwise we can change these things later once we figure out the best way to do it.

I also think that using 'standard' color instead of custom LV_COLOR_MAKE will make the code easier to read and maintain. We should probably have a look at `lv_pinetime_theme.h`, `lv_pinetime_theme.c`, `Colors.h` and `Colors.cpp` to figure out what's the best place/way to define colors.

I'm OK with merging this PR as is and figure that out in a future PR, though!

JF002 (approved):

PR #1106, opened 2022-04-24
List: add missing array include

NeroBurner:
List.h uses `std::array` as container, but is missing the `<array>`
include. Add it to make the header self contained.

### DIFF
--- a/src/displayapp/screens/List.h
+++ b/src/displayapp/screens/List.h
@@ -2,7 +2,7 @@
 
 #include <lvgl/lvgl.h>
 #include <cstdint>
-#include <memory>
+#include <array>
 #include "displayapp/screens/Screen.h"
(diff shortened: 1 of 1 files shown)


src/displayapp/screens/List.h
```diff
@@ -2,6 +2,7 @@
 
 #include <lvgl/lvgl.h>
 #include <cstdint>
+#include <array>
 #include <memory>
```
Riksu9000, line 6: I noticed that `<array>` is currently included through `<memory>`, which is otherwise unused, so `<memory>` should be removed.

NeroBurner: indeed, removed, compiled locally. Fixed now

Riksu9000 (approved):

JF002 (approved):

PR #1109, opened 2022-04-26
Timer UI redesign

kieranc:
The main goal of this PR is to increase the button size in the timer app to make incrementing and decrementing the time easier. To do this I've made the buttons transparent and put a gradient behind them to give a roller-type look, the buttons are now 100x90 instead of 60x40. 

I've changed the text color to black on the light background to increase contrast/readability, and added a long-press fast increment in preparation for #1093 

I had planned to split the minutes and seconds into separate labels to allow the colon between to be a different color, but using the letter_spacing setting I was able to get the alignment pretty good, and I overlaid a white colon - bit of a hack but I hope it's forgivable.

I think this would probably be usable on the alarm app, the info button might be a bit tricky but it should work. It could be theme-able later but I aimed for maximum contrast for now.

It looks like this:

image

### DIFF
--- a/src/displayapp/screens/Timer.cpp
+++ b/src/displayapp/screens/Timer.cpp
@@ -96,8 +96,9 @@ Timer::Timer(DisplayApp* app, Controllers::TimerController& timerController)
 
   uint32_t seconds = timerController.GetTimeRemaining() / 1000;
   lv_label_set_text_fmt(time, "%02lu:%02lu", seconds / 60, seconds % 60);
+  lv_obj_set_style_local_text_letter_space(time, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, -3);
 
-  lv_obj_align(time, lv_scr_act(), LV_ALIGN_IN_LEFT_MID, 0, -20);
+  lv_obj_align(time, lv_scr_act(), LV_ALIGN_IN_LEFT_MID, 10, -20);
   
 
   btnPlayPause = lv_btn_create(lv_scr_act(), nullptr);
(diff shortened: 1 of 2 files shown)


kieranc:
Quick mockup of the alarm:

image

Zandengoff:
Much better, I rarely use the timer or alarm because it is so hard to press the tiny buttons when needed. I usually end up breaking out the phone as it is faster. This looks much more functional for folks with larger hands.

Riksu9000:
Making the button click area larger is useful, but I'm afraid it's quite contrasting to the rest of InfiniTIme with the whites and the gradients. It may be useful to simply extend the click area with `lv_obj_set_ext_click_area()`.

Ertain:
This is looking promising. I hope my addition of a reset button will fit in. It's a pain having to reset the numbers to zero.

kieranc:
> Making the button click area larger is useful, but I'm afraid it's quite contrasting to the rest of InfiniTIme with the whites and the gradients. It may be useful to simply extend the click area with `lv_obj_set_ext_click_area()`.

I agree that it doesn't fit perfectly with the current 'theme' but I would argue that currently there isn't really a theme, it's just default elements which are functional but not much more. This change adds some style, I'm not 100% sure if it is the best style but I think it's an improvement, it improves contrast/visibility, and improves the ease of setting the timer/button size. I can strip it back to just increase the button hit area and leave it visually the same but I'd like to hear a few more people say they prefer the visual style of the old version to convince me that's the best approach.

Avamander:
It seems like a nice improvement, but the gradient seems a bit intense?

Maybe it could be replicated with a rounded rectangle with a lighter rectangle in the middle behind the numbers?

Rough ASCII illustration:
```
 _____
|  +  |
 _____
|  10 |
 _____
|  -  |
 _____
```

Riksu9000:
I had been working on something similar a bit ago as well, but couldn't get it to look just right. I think the font needs tweaking as well, and with the automatic font generation, I can continue experimenting.

InfiniSim_2022-05-13_170603

PR #1113, opened 2022-05-01
Add clang-format check

Riksu9000:
Add a GitHub action that clang-formats the changed cpp files, checks them for changes, and if it finds any, create patches for them and upload them as artifacts.

Automating code review will save a lot of time.

An example of this can be seen here https://github.com/Riksu9000/InfiniTime/pull/7

```
Checking src/displayapp/DisplayApp.cpp
Error: Formatting error in src/displayapp/DisplayApp.cpp
Creating patch 0001-Apply-clang-format-to-DisplayApp.cpp.patch
Error: Issues found. You may use the patches provided as artifacts to format the code.
Error: Process completed with exit code 1.
```
[Patches.zip](https://github.com/InfiniTimeOrg/InfiniTime/files/8598505/Patches.zip)

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,0 +1,34 @@
+name: Code formatting
+
+on:
+  pull_request:
+    branches: [ master, develop ]
+
+jobs:
+  test-format:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


JF002:
Nice use-case for github actions, thanks!

PR #1115, opened 2022-05-02
Workflow cleanup and artifact renaming

### LINKED ISSUE #795: Automatically rename workflow artifacts after the PR

Riksu9000:
Remove unnecessary comments in workflows.
Rename uploaded artifacts with the source branch name. The InfiniSim binary doesn't get renamed, only the zip.

Fixes #795

### DIFF
--- a/.github/workflows/lv_sim.yml
+++ b/.github/workflows/lv_sim.yml
@@ -60,7 +60,7 @@ jobs:
           cmake --build build_lv_sim
 
       - name: Upload simulator executable
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
         with:
-          name: infinisim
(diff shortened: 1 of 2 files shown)


JF002:
#1138 has just broke this PR, but it should be fairly easy to apply them on the new workflow files. @Riksu9000 do you want me to have a look?

Riksu9000:
@JF002 I've rebased this and applied a couple new fixes.

JF002:
@Riksu9000  Great, thanks!

PR #1117, opened 2022-05-02
Spaces added to Terminal face for readability

PReDiToR:
It's looks cleaner with a gap between the [label] and the data.
IMHO.


Riksu9000:
It looks quite different. I think the blockiness of the left side ends up emphasizing the added space.

InfiniSim_2022-05-03_102019
InfiniSim_2022-05-03_100903

Riksu9000:
This is probably not worth changing now.

PR #1121, opened 2022-05-08
Apply clang-format to all files in src dir

FintasticMan:
Applies clang-format to all src files except those in src/libs dir.


src/components/ble/DfuService.cpp
```diff
@@ -165,10 +165,10 @@ int DfuService::WritePacketHandler(uint16_t connectionHandle, os_mbuf* om) {
...
       if ((nbPacketReceived % nbPacketsToNotify) == 0 && bytesReceived != applicationSize) {
         uint8_t data[5] {static_cast<uint8_t>(Opcodes::PacketReceiptNotification),
-                         (uint8_t)(bytesReceived & 0x000000FFu),
-                         (uint8_t)(bytesReceived >> 8u),
-                         (uint8_t)(bytesReceived >> 16u),
-                         (uint8_t)(bytesReceived >> 24u)};
+                         (uint8_t) (bytesReceived & 0x000000FFu),
+                         (uint8_t) (bytesReceived >> 8u),
+                         (uint8_t) (bytesReceived >> 16u),
+                         (uint8_t) (bytesReceived >> 24u)};
```
Riksu9000, line 171: There may be a difference between `clang-format` versions that's causing the error here. We should be using static_cast anyway, so let's change them now.

src/displayapp/icons/music/disc.cpp
```diff
@@ -20,91 +20,40 @@
...
+  #define LV_ATTRIBUTE_MEM_ALIGN
 #endif
 
 #ifndef LV_ATTRIBUTE_IMG_DISC
-#define LV_ATTRIBUTE_IMG_DISC
+  #define LV_ATTRIBUTE_IMG_DISC
 #endif
 
 const LV_ATTRIBUTE_MEM_ALIGN LV_ATTRIBUTE_IMG_DISC uint8_t disc_map[] = {
-    0xbd, 0xc1, 0xbe, 0xff,  /* Color of index 0: foreground */
```
Riksu9000, line 31: These disc files are generated by LVGL image converter. I think they should be .c files and not formatted. Can you rename them instead, so they get ignored?

src/components/ble/DfuService.cpp
Avamander: Yep, I wanted to highlight that C-style casts should be replaced as well when formatting.

src/components/ble/MusicService.cpp
```diff
@@ -56,63 +54,35 @@ namespace {
...
-                                  .flags = BLE_GATT_CHR_F_WRITE | BLE_GATT_CHR_F_READ};
-  characteristicDefinition[11] = {.uuid = &msRepeatCharUuid.u,
-                                  .access_cb = MusicCallback,
-                                  .arg = this,
-                                  .flags = BLE_GATT_CHR_F_WRITE | BLE_GATT_CHR_F_READ};
-  characteristicDefinition[12] = {.uuid = &msShuffleCharUuid.u,
-                                  .access_cb = MusicCallback,
-                                  .arg = this,
-                                  .flags = BLE_GATT_CHR_F_WRITE | BLE_GATT_CHR_F_READ};
+  characteristicDefinition[0] = {
```
Avamander, line 57: This is suspicious. What configuration are you using? What clang-format version?

Avamander (changes requested):

src/components/ble/MusicService.cpp
FintasticMan: I'm just using clang-format version 13.0.1. I'm just using the clang-format used by this repository. The exact command-line arguments are `clang-format -i --style=file "$file"`.

FintasticMan: It might be differences between versions, I'll see if clang-format 12 (the version used by the clang-format GH Actions) formats it any differently.

FintasticMan: The GH Actions run has finished, and it doesn't recognise that as a mistake in formatting, so it isn't a difference between versions.

Riksu9000: This is controlled by AllowAllArgumentsOnNextLine. I never liked it. I think we should change it.

FintasticMan: Done, I think it looks better now.

src/displayapp/screens/settings/Settings.cpp
```diff
@@ -49,35 +50,30 @@ std::unique_ptr<Screen> Settings::CreateScreen1() {
...
 std::unique_ptr<Screen> Settings::CreateScreen2() {
-  std::array<Screens::List::Applications, 4> applications {{
-    {Symbols::shoe, "Steps", Apps::SettingSteps},
-    {Symbols::clock, "Set date", Apps::SettingSetDate},
-    {Symbols::clock, "Set time", Apps::SettingSetTime},
-    {Symbols::batteryHalf, "Battery", Apps::BatteryInfo}}};
+  std::array<Screens::List::Applications, 4> applications {{{Symbols::shoe, "Steps", Apps::SettingSteps},
+                                                            {Symbols::clock, "Set date", Apps::SettingSetDate},
+                                                            {Symbols::clock, "Set time", Apps::SettingSetTime},
+                                                            {Symbols::batteryHalf, "Battery", Apps::BatteryInfo}}};
```
Riksu9000, line 56: Inserting trailing commas to these arrays makes the format like it is in `CreateScreen1()`, which would be preferred.

Riksu9000 (changes requested):

src/systemtask/SystemMonitor.cpp
```diff
@@ -1,9 +1,9 @@
 #include "systemtask/SystemTask.h"
 #if configUSE_TRACE_FACILITY == 1
-// FreeRtosMonitor
-#include <FreeRTOS.h>
-#include <task.h>
-#include <nrf_log.h>
+  // FreeRtosMonitor
+  #include <FreeRTOS.h>
+  #include <task.h>
+  #include <nrf_log.h>
```
FintasticMan, line 6: This doesn't seem right... Should the preprocessor directives really be indented?

Riksu9000: IndentPPDirectives is set to BeforeHash in .clang-format, so it is working as intended at least. Do you think it should be changed?

FintasticMan: Just tried it out with IndentPPDirectives set to None, and it made many preprocessor if statements worse, so I'll leave it as is.

Riksu9000 (approved):

src/components/ble/HeartRateService.cpp
```diff
@@ -72,11 +73,11 @@ void HeartRateService::OnNewHeartRateValue(uint8_t heartRateValue) {
 }
 
 void HeartRateService::SubscribeNotification(uint16_t connectionHandle, uint16_t attributeHandle) {
-  if(attributeHandle == heartRateMeasurementHandle)
+  if (attributeHandle == heartRateMeasurementHandle)
```
Avamander, line 76: We should probably add braces to these at some point, but this can be done in a separate PR at some later point

Avamander (approved):

Avamander:
@JF002 Yep, seems good.

src/components/ble/HeartRateService.cpp
FintasticMan: Once clang-format 15 is out, that can be done automatically with the `InsertBraces` option.

PR #1127, opened 2022-05-10
Format workflow improvements

Riksu9000:
Don't run the workflow when not necessary.
Ignore libs folder.
Don't allow test-format.sh to be run locally.

Tested here https://github.com/Riksu9000/InfiniTime/pull/7

### DIFF
--- a/tests/test-format.sh
+++ b/tests/test-format.sh
@@ -2,11 +2,13 @@
 
 set -e
 
-[ -z "$1" ] && exit
+if [ -z "$GITHUB_BASE_REF" ]
+then
+  echo "This script is only meant to be run in a GitHub Workflow"
+  exit 1
+fi
 
-basebranch=$1
-
-CHANGED_FILES=$(git diff --name-only "$basebranch"...HEAD)
+CHANGED_FILES=$(git diff --name-only "$GITHUB_BASE_REF"...HEAD)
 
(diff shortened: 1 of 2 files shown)


JF002 (approved):

tests/test-format.sh
```diff
@@ -2,18 +2,21 @@
 
 set -e
```
NeroBurner, line 3: the shebang sets `#!/bin/sh`. I'm not sure, but is `set -e` a bash feature?

The script output prints the following

```
Run tests/test-format.sh
  tests/test-format.sh
  shell: /usr/bin/bash -e {0}
Checking src/displayapp/DisplayApp.cpp
```

so I thought maybe the first line should be changed to `#!/bin/bash`

FintasticMan: I don't think that `set -e` is a bashism, or at least `shellcheck` doesn't complain about it.

Riksu9000: As far as I can tell, `set -e` is present in posix sh. The script passes shellcheck with no issues and works with `dash`. Ubuntu doesn't ship a posix shell by default, so it uses bash.

NeroBurner: then sorry about the noise

NeroBurner (approved):

PR #1128, opened 2022-05-10
Add basic hardware configuration options for P8

StarGate01:
This PR has been broken out of https://github.com/InfiniTimeOrg/InfiniTime/pull/1050.

This enables the compile-time configuration of the LFCLK source, as well as the target hardware board pin configuration.

The `WATCH_COLMI_P8` compilation option has been replaced with `TARGET_DEVICE`, see `doc/buildAndProgram.md`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,13 +54,8 @@ if(BUILD_DFU)
   set(BUILD_DFU true)
 endif()
 
-option(WATCH_COLMI_P8 "Build for the Colmi P8" OFF)
-set(TARGET_DEVICE "PineTime")
-
-if(WATCH_COLMI_P8)
-  set(TARGET_DEVICE "Colmi P8")
-  add_definitions(-DWATCH_P8)
-endif()
+set(TARGET_DEVICE "PINETIME" CACHE STRING "Target device")
+set_property(CACHE TARGET_DEVICE PROPERTY STRINGS PINETIME MOY-TFK5 MOY-TIN5 MOY-TON5 MOY-UNK)
 
 set(PROJECT_GIT_COMMIT_HASH "")
 
--- a/doc/buildAndProgram.md
+++ b/doc/buildAndProgram.md
@@ -39,7 +39,7 @@ CMake configures the project according to variables you specify the command line
 **GDB_CLIENT_BIN_PATH**|Path to arm-none-eabi-gdb executable. Used only if `USE_GDB_CLIENT` is 1.|`-DGDB_CLIENT_BIN_PATH=/home/jf/nrf52/gcc-arm-none-eabi-9-2019-q4-major/bin/arm-none-eabi-gdb`
 **GDB_CLIENT_TARGET_REMOTE**|Target remote connection string. Used only if `USE_GDB_CLIENT` is 1.|`-DGDB_CLIENT_TARGET_REMOTE=/dev/ttyACM0`
 **BUILD_DFU (\*\*)**|Build DFU files while building (needs [adafruit-nrfutil](https://github.com/adafruit/Adafruit_nRF52_nrfutil)).|`-DBUILD_DFU=1`
-**WATCH_COLMI_P8**|Use pin configuration for Colmi P8 watch|`-DWATCH_COLMI_P8=1`
+**TARGET_DEVICE**|Target device, used for hardware configuration. Allowed: `PINETIME, MOY-TFK5, MOY-TIN5, MOY-TON5, MOY-UNK`|`-DTARGET_DEVICE=PINETIME` (Default)
 
 ####(**) Note about **CMAKE_BUILD_TYPE**:
 By default, this variable is set to *Release*. It compiles the code with size and speed optimizations. We use this value for all the binaries we publish when we [release](https://github.com/InfiniTimeOrg/InfiniTime/releases) new versions of InfiniTime.
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -779,6 +779,35 @@ add_definitions(-DFREERTOS)
 add_definitions(-D__STACK_SIZE=1024)
 add_definitions(-D__HEAP_SIZE=4096)
 
+# Note: Only use this for debugging
+# Derive the low frequency clock from the main clock (SYNT)
+# add_definitions(-DCLOCK_CONFIG_LF_SRC=2)
+
+# Target hardware configuration options
+add_definitions(-DTARGET_DEVICE_${TARGET_DEVICE})
+if(TARGET_DEVICE STREQUAL "PINETIME")
+  add_definitions(-DDRIVER_PINMAP_PINETIME)
+  add_definitions(-DCLOCK_CONFIG_LF_SRC=1) # XTAL
+elseif(TARGET_DEVICE STREQUAL "MOY-TFK5") # P8a
+  add_definitions(-DDRIVER_PINMAP_P8)
+  add_definitions(-DCLOCK_CONFIG_LF_SRC=1) # XTAL
+elseif(TARGET_DEVICE STREQUAL "MOY-TIN5") # P8a variant 2
+  add_definitions(-DDRIVER_PINMAP_P8)
+  add_definitions(-DCLOCK_CONFIG_LF_SRC=1) # XTAL
+elseif(TARGET_DEVICE STREQUAL "MOY-TON5") # P8b
+  add_definitions(-DDRIVER_PINMAP_P8)
+  add_definitions(-DCLOCK_CONFIG_LF_SRC=0) # RC
+  add_definitions(-DMYNEWT_VAL_BLE_LL_SCA=500)
+  add_definitions(-DCLOCK_CONFIG_LF_CAL_ENABLED=1)
+elseif(TARGET_DEVICE STREQUAL "MOY-UNK") # P8b mirrored
+  add_definitions(-DDRIVER_PINMAP_P8)
+  add_definitions(-DCLOCK_CONFIG_LF_SRC=0) # RC
+  add_definitions(-DMYNEWT_VAL_BLE_LL_SCA=500)
+  add_definitions(-DCLOCK_CONFIG_LF_CAL_ENABLED=1)
+else()
+  message(FATAL_ERROR "Invalid TARGET_DEVICE")
+endif()
+
 # Debug configuration
 if (${CMAKE_BUILD_TYPE} STREQUAL "Debug")
   add_definitions(-DDEBUG)
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -312,10 +312,15 @@ int main(void) {
   nrf_drv_clock_init();
   nrf_drv_clock_lfclk_request(NULL);
 
-// The RC source for the LF clock has to be calibrated
-#if (CLOCK_CONFIG_LF_SRC == NRF_CLOCK_LFCLK_RC)
+  // When loading the firmware via the Wasp-OS reloader-factory, which uses the used internal LF RC oscillator,
+  // the LF clock has to be explicitly restarted because InfiniTime uses the external crystal oscillator if available.
+  // If the clock is not restarted, the Bluetooth timers fail to initialize.
+  nrfx_clock_lfclk_start();
   while (!nrf_clock_lf_is_running()) {
   }
+
+// The RC source for the LF clock has to be calibrated
+#if (CLOCK_CONFIG_LF_SRC == NRF_CLOCK_LFCLK_RC)
   nrf_drv_clock_calibration_start(0, calibrate_lf_clock_rc);
 #endif
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -300,10 +300,24 @@ void nimble_port_ll_task_func(void* args) {
 }
 }
 
+void calibrate_lf_clock_rc(nrf_drv_clock_evt_type_t event) {
+  // 16 * 0.25s = 4s calibration cycle
+  // Not recursive, call is deferred via internal calibration timer
+  nrf_drv_clock_calibration_start(16, calibrate_lf_clock_rc);
+}
+
 int main(void) {
   logger.Init();
 
   nrf_drv_clock_init();
+  nrf_drv_clock_lfclk_request(NULL);
+
+// The RC source for the LF clock has to be calibrated
+#if (CLOCK_CONFIG_LF_SRC == NRF_CLOCK_LFCLK_RC)
+  while (!nrf_clock_lf_is_running()) {
+  }
+  nrf_drv_clock_calibration_start(0, calibrate_lf_clock_rc);
+#endif
 
   // Unblock i2c?
   nrf_gpio_cfg(Pinetime::PinMap::TwiScl,
--- a/src/drivers/PinMap.h
+++ b/src/drivers/PinMap.h
@@ -4,8 +4,8 @@
 namespace Pinetime {
   namespace PinMap {
 
-#ifdef WATCH_P8
-    // COLMI P8
+#if defined(DRIVER_PINMAP_P8)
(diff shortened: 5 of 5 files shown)


doc/buildAndProgram.md
```diff
@@ -35,7 +35,8 @@ CMake configures the project according to variables you specify the command line
 **GDB_CLIENT_BIN_PATH**|Path to arm-none-eabi-gdb executable. Used only if `USE_GDB_CLIENT` is 1.|`-DGDB_CLIENT_BIN_PATH=/home/jf/nrf52/gcc-arm-none-eabi-9-2019-q4-major/bin/arm-none-eabi-gdb`
 **GDB_CLIENT_TARGET_REMOTE**|Target remote connection string. Used only if `USE_GDB_CLIENT` is 1.|`-DGDB_CLIENT_TARGET_REMOTE=/dev/ttyACM0`
 **BUILD_DFU (\*\*)**|Build DFU files while building (needs [adafruit-nrfutil](https://github.com/adafruit/Adafruit_nRF52_nrfutil)).|`-DBUILD_DFU=1`
-**WATCH_COLMI_P8**|Use pin configuration for Colmi P8 watch|`-DWATCH_COLMI_P8=1`
+**TARGET_DEVICE**|Target device, used for pin map. Allowed: `PINETIME, P8`|`-DTARGET_DEVICE=PINETIME` (Default)
+**LF_CLK**|Configures the LF clock source. Allowed: `RC, XTAL, SYNT`|`-DLF_CLK=XTAL` (Default)
```
Riksu9000, line 39: As I wrote in https://github.com/InfiniTimeOrg/InfiniTime/pull/1051#discussion_r862996876, I think the hardware configuration should be based on the target device. What I mean by this, is that enabling PineTime target device should automatically enable XTAL for example. Having `LF_CLK` for example is good abstraction, but my idea is that they don't need to be directly controllable by the user/developer.

src/CMakeLists.txt
```diff
@@ -795,6 +795,20 @@ add_definitions(-DFREERTOS)
...
+if(LF_CLK STREQUAL "RC")
+  add_definitions(-DCLOCK_CONFIG_LF_SRC=0)
+  # The LF RC clock source has a precision of 500ppm
+  add_definitions(-DMYNEWT_VAL_BLE_LL_SCA=500)
+  add_definitions(-DCLOCK_CONFIG_LF_CAL_ENABLED=1)
+elseif(LF_CLK STREQUAL "XTAL")
+  add_definitions(-DCLOCK_CONFIG_LF_SRC=1)
+elseif(LF_CLK STREQUAL "SYNT")
+  add_definitions(-DCLOCK_CONFIG_LF_SRC=2)
+else()
```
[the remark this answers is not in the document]
StarGate01: Fair point. The synthetic clock has a much higher precision than the RC clock, and does not need to be recalibrated. It is useful for debugging the BLE stack. Do you thing the third option is confusing? I agree that there is no current end-user use for it.

doc/buildAndProgram.md
StarGate01: This abstraction could be solved entirely in CMake, which would then set the required "fine-grained" configuration definitions. Alternatively, I could rewrite the various drivers etc. to depend only on a tingle target selection parameter. What would you suggest?

Riksu9000: The first option sounds good.

src/CMakeLists.txt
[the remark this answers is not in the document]
Riksu9000: We can keep the option if we add a comment like this.
```suggestion
elseif(LF_CLK STREQUAL "SYNT")
  # Only use this for debugging
  add_definitions(-DCLOCK_CONFIG_LF_SRC=2)
else()
```

StarGate01:
As discussed, I reduced the CMake configuration to `TARGET_DEVICE`, which sets all the required preprocessor variables for the selected device.

I rebased #1130 and #1132 to use this system as well.

src/main.cpp
```diff
@@ -302,11 +301,29 @@ void nimble_port_ll_task_func(void* args) {
 }
 }
 
+void calibrate_lf_clock_rc(nrf_drv_clock_evt_type_t event) {
+  // 16 * 0.25s = 4s calibration cycle
+  // Not recursive, call is deferred via internal calibration timer
+  nrf_drv_clock_calibration_start(16, calibrate_lf_clock_rc);
+}
+
```
Riksu9000, line 309: What's the purpose of this handler?

src/main.cpp
```diff
@@ -302,11 +301,29 @@ void nimble_port_ll_task_func(void* args) {
...
 int main(void) {
   logger.Init();
 
   nrf_drv_clock_init();
 
+  // In some cases (eg. when a watchdog is enabled),
+  // the NRF SDK incorrectly assumes the LF clock is already running,
```
Riksu9000, line 316: Can you elaborate on this and the purpose of `while (!nrf_clock_lf_is_running()) {}` assuming it's related?

src/main.cpp
StarGate01: Yes, see https://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.sdk5.v15.0.0%2Fgroup__nrf__drv__clock.html . The RC clock drifts fast and source has to be re-calibrated every ~4 seconds, in order to stay accurate to the 500ppm (`MYNEWT_VAL_BLE_LL_SCA`) maximum allowed by the BLE stack.

The timer used for this loop is a dedicated hardware timer (`CT`), see https://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.nrf52832.ps.v1.1%2Fclock.html . It does not interfere with any other timers.

src/main.cpp
StarGate01: The `nrf_drv_clock_calibration_start` requires the clock to be running, if it is not running the call fails and the calibration loop is not started. Starting the clock takes a bit of time in the background.

I'll move this check into the `#if (CLOCK_CONFIG_LF_SRC == NRF_CLOCK_LFCLK_RC)` guard, since it only concerns that case.

src/main.cpp
```diff
@@ -302,11 +301,29 @@ void nimble_port_ll_task_func(void* args) {
...
+  nrf_drv_clock_lfclk_request(NULL);
+  nrfx_clock_lfclk_start();
```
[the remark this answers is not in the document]
StarGate01: You are correct. This is probably a relic of debugging. Calling `nrf_drv_clock_init` to initialize the NRF clock SDK control memory block  then and nrf_drv_clock_lfclk_request(NULL) to set a "client" semaphore which ensures that the clock stays active seems to be sufficient. 

In the past, I had some troubles with the watchdog. The watchdog internally starts the low frequency clock as well, and I had some situations in which the SDK assumed an incorrect clock state.

StarGate01:
I modified the clock setup, as discussed.

Riksu9000 (approved):
Checks should be passed before merging.

StarGate01:
@Riksu9000 I figured out why I included `nrfx_clock_lfclk_start` previously.

The call is absolutely required, in the case that InfiniTime is isntalled via reloader-facory, coming from Wasp-OS. This is because Wasp-OS uses the internal RC oscillator, while Inifinitime switches to the external crystal osciallator if one is available.

Since there is no cold boot between the installation, there are some leftovers in the clock peripheral configuration, which confuse the NRF SDK to incorrectly assume the clock state.

I tested it, and without this call the InfiniTime recovery fails to establish a Bluetooth connection after being installed by the Wasp-OS reloader. I added the line back in c5a3d8cb8d783bf7374067b1c4211faf2ffe5f9f .

Riksu9000:
Am I understanding this correctly that this still only applies to devices that use the RC oscillator? If that is the case, maybe those lines only needs to be included on devices that use it.

StarGate01:
Not quite, because Wasp-OS (at least its bootloader and reloader) *always* use the RC oscillator for compatibility reasons. This error even happens on a genuine Pinetime when sideloading InfiniTime from Wasp-OS, I tested it.

Riksu9000:
https://wiki.pine64.org/wiki/Switching_your_PineTime_between_InfiniTime_and_Wasp-os

TBH I haven't used Wasp OS and I'm not familiar with the reloader things, however this site claims that that file is broken and instructs using some other files. @JF002 Do you know more about this https://github.com/InfiniTimeOrg/InfiniTime/pull/1128#issuecomment-1137213136?

StarGate01:
Yes, that's why I wrote my own toolchain to build reloaders that actually work and use the current version of InfiniTime, see https://github.com/StarGate01/p8b-infinitime .

JF002:
@StarGate01 Your work on porting InfiniTime to the P8 smartwatches is impressive! 

As you may know, porting InfiniTime to other hardware is not on my todo list as I probably won't have the time to maintain and support it. However, I'm totally open to accept changes to support new hardware if they don't break InfiniTime on the PineTime. I would also appreciate if you would be willing to help with maintenance and support of those change. Seeing how advanced your fork for the P8 is, this will probably not be an issue :)
 
> Do you know more about this https://github.com/InfiniTimeOrg/InfiniTime/pull/1128#issuecomment-1137213136?

Not really. The reloader was mostly built by Daniel from Waspos. I tested them (from infinitime to waspos and revert) at that time and they seemed to work fine. But I haven't done any new test or maintenance recently. Those reloader apps probably need to be updated and tested again!

src/main.cpp
```diff
@@ -302,10 +301,29 @@ void nimble_port_ll_task_func(void* args) {
...
   logger.Init();
 
   nrf_drv_clock_init();
+  nrf_drv_clock_lfclk_request(NULL);
+
+  // When loading the firmware via the Wasp-OS reloader-factory, which uses the used internal LF RC oscillator,
+  // the LF clock has to be explicitly restarted because InfiniTime uses the external crystal oscillator if available.
+  // If the clock is not restarted, the Bluetooth timers fail to initialize.
+  nrfx_clock_lfclk_start();
+  while (!nrf_clock_lf_is_running()) {
```
JF002, line 320: This is an infinite loop. Is there any chance we never get out of this loop?

StarGate01: The loop runs until the LF clock peripheral is confirmed to be running. Without this hardware peripheral, the long timers, watchdog, and Bluetooth stack won't work (on any device).

I don't know how this loop could fail to exit, except on bad or damaged silicon, I guess? In any case, this kind of busy waiting is even used in the NRF SDK, e.g. at https://github.com/cmdwtf/nRF5_SDK/blob/master/modules/nrfx/drivers/src/nrfx_clock.c#L252= .

If this loop is considered risky, I can implement a timeout, but it should not be necessary imho. In all my testing across three different devices this loop never failed.

StarGate01:
@JF002 
 > Your work on porting InfiniTime to the P8 smartwatches is impressive!

Thank you! I think there are quite a lot of P8s in circulation which would profit from the support, since it is a cheap and available watch.

> However, I'm totally open to accept changes to support new hardware if they don't break InfiniTime on the PineTime

I routinely test all my changes against a genuine PineTime in order to ensure that nothing breaks.

>  I would also appreciate if you would be willing to help with maintenance and support of those change

Of course. Just having the support upstreamed would be a massive help for me personally, since maintaining and rebasing a fork and ~10 PRs is kind of complex. Since most of my changes are hardware drivers, I don't expect much maintenance load. Ideally, all future changes would just run on the unified hardware interface.

> Those reloader apps probably need to be updated

I provide a `reloader-factory.zip` built from the most recent code of the module for all there watches with every release on https://github.com/StarGate01/p8b-infinitime/releases . Once the changes are upstreamed, I can work on updating the reloaders and the wiki page.

src/main.cpp
JF002: Thanks for the explanation! You're right : without clock, we cannot expect anything to work properly... And the watchdog will take care to reset the device!

Riksu9000:
I realized that there should probably be some instructions on how users can identify their P8 models. I also found this issue about a third model https://github.com/StarGate01/p8b-infinitime/issues/2. Is there any way to reliably determine the exact hardware a user has without experimentation, like for example the information available in the stock firmware? Any reason not to use the original MOY-*** labels to identify the devices?

JF002 (approved):

StarGate01:
@JF002 I'll set this to a draft for today - I'll convert the `P8A` / `P8B` stings into the "official" `MOY` ones before merging.

StarGate01:
I have adapted the identifiers. The watch by @ildar (https://github.com/StarGate01/p8b-infinitime/issues/2), the `MOY-TIN5` has been added, as well as the watch by @izzeho (https://github.com/StarGate01/p8b-infinitime/issues/3). The identifier of that watch is not known until I receive my order, so in the meantime it is called `MOY-UNK`.

Riksu9000 (changes requested):

doc/buildAndProgram.md
```diff
@@ -39,7 +39,7 @@ CMake configures the project according to variables you specify the command line
 **GDB_CLIENT_BIN_PATH**|Path to arm-none-eabi-gdb executable. Used only if `USE_GDB_CLIENT` is 1.|`-DGDB_CLIENT_BIN_PATH=/home/jf/nrf52/gcc-arm-none-eabi-9-2019-q4-major/bin/arm-none-eabi-gdb`
 **GDB_CLIENT_TARGET_REMOTE**|Target remote connection string. Used only if `USE_GDB_CLIENT` is 1.|`-DGDB_CLIENT_TARGET_REMOTE=/dev/ttyACM0`
 **BUILD_DFU (\*\*)**|Build DFU files while building (needs [adafruit-nrfutil](https://github.com/adafruit/Adafruit_nRF52_nrfutil)).|`-DBUILD_DFU=1`
-**WATCH_COLMI_P8**|Use pin configuration for Colmi P8 watch|`-DWATCH_COLMI_P8=1`
+**TARGET_DEVICE**|Target device, used for the pin map and the low frequency clock source. Allowed: `PINETIME, MOY-TFK5, MOY-TIN5,MOY-TON5, MOY-UNK`|`-DTARGET_DEVICE=PINETIME` (Default)
```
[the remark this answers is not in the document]
StarGate01: Good catch! Thanks, fixed.

Riksu9000 (approved):

CMakeLists.txt
```diff
@@ -54,13 +54,8 @@ if(BUILD_DFU)
...
 
-option(WATCH_COLMI_P8 "Build for the Colmi P8" OFF)
-set(TARGET_DEVICE "PineTime")
-
-if(WATCH_COLMI_P8)
-  set(TARGET_DEVICE "Colmi P8")
-  add_definitions(-DWATCH_P8)
-endif()
+set(TARGET_DEVICE "PINETIME" CACHE STRING "Target device")
+set_property(CACHE TARGET_DEVICE PROPERTY STRINGS PINETIME P8A P8B)
```
Riksu9000, line 58: I don't know CMake that well, but I think this needs updating as it has the old P8A and P8B labels.

Riksu9000 (changes requested):

CMakeLists.txt
StarGate01: Correct, fixed. Thank you.

Riksu9000 (approved):

ildar:
On Tue, Jun 14, 2022 at 5:17 PM Christoph Honal wrote:

> However, I am unsure on what the MOY* labels actually mean - could be a
> reseller / software version as well (?) .
>
No. The FW version is in the different line. The reseller info is totally
missing: the watch is White Label per s.e.

> it is afaik not possible to know whether an external LF crystal exists
> without looking at the PCB.
>
 Easy:
1. Turn on the LF osc
2. Check the status bit with timeout.
3. If timeout then it's missing

Typical lag is in the docs.

PR #1129, opened 2022-05-10
Convert ScreenList from array to vector for dynamic generation of Screens

JeffmeisterJ:
For some applications, such as the ApplicationList, it's easier to make the ScreenList a vector instead of a precompiled array. This makes it a lot easier to dynamically add or remove screens, for example based on incoming BLE data or user input. 
Some examples include creating multiple alarms/timers, home automation apps, multiple cities in the Weather display etc.

### DIFF
--- a/src/displayapp/screens/ApplicationList.h
+++ b/src/displayapp/screens/ApplicationList.h
@@ -25,7 +25,7 @@ namespace Pinetime {
         Pinetime::Controllers::Battery& batteryController;
         Controllers::DateTime& dateTimeController;
 
-        ScreenList<2> screens;
+        ScreenList screens;
         std::unique_ptr<Screen> CreateScreen1();
         std::unique_ptr<Screen> CreateScreen2();
         // std::unique_ptr<Screen> CreateScreen3();
--- a/src/displayapp/screens/ScreenList.h
+++ b/src/displayapp/screens/ScreenList.h
@@ -11,11 +11,11 @@ namespace Pinetime {
     namespace Screens {
 
       enum class ScreenListModes { UpDown, RightLeft, LongPress };
-      template <size_t N> class ScreenList : public Screen {
+      class ScreenList : public Screen {
       public:
         ScreenList(DisplayApp* app,
                    uint8_t initScreen,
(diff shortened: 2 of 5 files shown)


Riksu9000:
Making the screenlists more dynamic is a good idea, however I'm not sure we need them all to be vectors right now, since the number of screens doesn't change. We may return to this idea once there's a bigger need for it. What I would like to see currently is, for example, for the code automatically create enough screens to fit all 12 applications which would be stored in a single array, since the number of screens needed can be determined before creating them.

JeffmeisterJ:
I'm using it now for the [HomeAssistant screen](https://github.com/JeffmeisterJ/InfiniTime/tree/homeassistant_screen) branch I'm working on, where the goal is to receive the rooms, devices and states via BLE and generate a screen for every room dynamically from the received data (within limits for RAM reasons of course). This is a usecase where at least at compile time the number of screens (for rooms) is not yet known, so it's easier to use a vector. 

The ApplicationList is another good example, and one I'd like to implement, so I'll give that a go with this as well. It might be nice to just determine the amount of screens based on a single list of available applications, which may even be installable or be chosen at compile time ([Modpacks](https://github.com/ZephyrLabs/InfiniTime-Builder#Modpacks) being a bit of a rough but illustrative example).

JF002:
Using a fix sized std::array was a deliberate choice : I want to avoid dynamic memory allocation and memory fragmentation as much as possible. The PineTime has very few RAM memory, and we have no "swap" space to use in case the firmware runs out of memory.
With static allocation, memory is allocated at compile time, and the memory usage won't change at run-time. This allows to avoid a lot of bugs that will be very difficult to diagnose (ex : infinitime crashes when you have 5 rooms in your home assistant screen and you receive a BLE notitication that is a bit too long).

I don't think using a std::vector is a good idea here (or any "infinite-sized" containers) : vectors uses dynamic memory allocation (malloc()) and copy the whole buffer in memory when it needs to be resized. Instead, you should use a fixed size array, which size will specify the maximum number of rooms your applications will be able to handle.

But, as @Riksu9000 said, it might be a good idea to automatically determine the size of the arrays at compile time.

PR #1130, opened 2022-05-10
Add support for CST716/CST718S touch sensor used in P8

StarGate01:

- Add support for the CST716 touch driver chip, as well as the factory fused operating modes of the CST816S.
- Fused mode can be configured using the compile-time variable `DRIVER_TOUCH`.


src/touchhandler/TouchHandler.cpp
```diff
@@ -75,15 +100,22 @@ bool TouchHandler::GetNewTouchInfo() {
...
   return true;
 }
 
 void TouchHandler::UpdateLvglTouchPoint() {
   if (info.touching) {
+#if defined(DRIVER_TOUCH_GESTURE)
+    // GESTURE config only generates a single event / state change
+    // so the LVGL wrapper is used to generate a successive release state update
+    lvgl.SetNewTap(info.x, info.y);
+#else
```
Riksu9000, line 114: Does this mean that LVGL isn't able to detect different events like long presses, or does this function get called multiple times before LVGL could read `tapped == false`? It's not very widely used currently, but it would become a problem I'm sure.

StarGate01: I added the function `SetNewTap` after the discussion in https://github.com/InfiniTimeOrg/InfiniTime/pull/1050#discussion_r869455953 . It is essentially the same as `SetNewTouchPoint`, but instead of setting the "pressed" state inside LVGL to either "pressed" or "released", it generates a "pressed" event, followed by an auto-generated "release" event: https://github.com/InfiniTimeOrg/InfiniTime/pull/1130/files#diff-ff98991d1db50496bd6609ea99f1bb9b2717d32a2f65f2838b77d8d689530813= . 

This is required because the display touch driver in the P8a watch (see https://github.com/StarGate01/p8b-infinitime/blob/master/HardwareVariants.md#p8a=) only generates a singular interrupt event when the display is touched in any way. Compare that to the PineTime driver chip, which generates frequent intermediary interrupts, as well as a "event start", i.e. "press" and a "event end", i.e. "release" event. For more info see the comment at https://github.com/InfiniTimeOrg/InfiniTime/pull/1130/files#diff-841fc25ab12089b6b1f623755778d18b387663ddd80c6ffb2054bfcdb972b7cb= .

Since LVGL relies on the successive events "press" and "release" to generate a tap action event, this synthetic event is required for chips which only generate one interrupt signal.

Riksu9000: I think this is how InfiniTime worked before. It's quite a bit more sluggish isn't it? Especially in the timer app. That's unfortunate. Unless this issue gets solved somehow, I'm not sure P8a can get more than experimental support, since we're mainly targeting PineTime and we'll keep using the features that it supports.

StarGate01: I see. Unfortunately, the P8a interrupt behavior is fused into the chip and cannot be changed, the only alternative would be to poll the chip (similar to InfiniTime 1.3). But I did not want to implement this, it seems clunky and I did not want to modify the existing code too much.

In my opinion the users of a P8a know that this port is experimental. Just having the support in the code base and allowing users to compile it for their watches would be a huge win, even if some features don't fully work (yet). I understand that InfiniTime is a PineTime-centric project, and running on other watches would be "at your own risk".

src/touchhandler/TouchHandler.cpp
```diff
@@ -75,15 +100,22 @@ bool TouchHandler::GetNewTouchInfo() {
...
     if (!isCancelled) {
       lvgl.SetNewTouchPoint(info.x, info.y, true);
     }
+#endif
```
Riksu9000, line 118: While I don't have these devices to test, I think this function may be able to be simplified further. On the PineTime, a swipe immediately switches the screen and `isCancelled` is used to ignore the rest of the touch until it has been released. On the other devices the swipe event only occurs on release if I understand correctly, so parts of this could be removed.

Having preprocessor in the middle of a function isn't very nice, so depending on how different the drivers end up being, we can see if it is worth creating separate files for them or create new functions that are only included in certain builds.

Riksu9000: I'm not sure it would lead to a lot of code duplication. Firstly `ConvertGesture()` could be moved to Cst816s.cpp for example. Then there's not that much code left in the first place and while there are *only* two pre-processor macros, they end up touching a large portion of the code. It would also be easier to find optimizations if they were separated and freely let them diverge. For example removing `isCancelled` from `GESTURE` and perhaps `REPORT` if possible.

If the end of `UpdateLvglTouchPoints()`, which is responsible for the release event, doesn't do anything on `GESTURE`, the entire function could be replaced with a simpler one for better visibility.

Riksu9000 (changes requested):
TouchHandler received some changes recently. TouchHandler should probably be a part of the driver, since it's just processing the touch screen data into an InfiniTime format, but that's not necessarily in the scope of this PR. This PR should be rebased, all unnecessary changes should be removed, and preprocessor shouldn't modify the behavior of functions, but can be used to pick which function to call. Then we may be able to merge this.

ildar:
hmm, not ready yet (?)

PR #1131, opened 2022-05-10
generalize lv-font creation

NeroBurner:
In https://github.com/InfiniTimeOrg/InfiniTime/pull/1097 new font
generation capabilites were added. Generalize the font creation to
make it possible to reuse the `displayapp/fonts/CMakeLists.txt` file
for `InfiniSim` and just add the new cmake file to the project and
link against the new `infinitime_fonts` target.

In the following a list of changes.

Allow non-global installed `lv_font_conv` executable installed with

```sh
npm install lv_font_conv@1.5.2
```

In CMake we search for `lv_font_conv` executable. Add the found
executable to the python script `generate.py`, to remove the need for
`lv_font_conv` to be in the path.

Search for `python3` executable, if CMake version 3.12 is available.
Otherwise use `python` as hard coded executable.

Instead of adding the generated fonts to `SOURCE_FILES` variable, create
a static library `infinitime_fonts` instead. Link this library to the
executables instead.

### DIFF
--- a/src/displayapp/fonts/CMakeLists.txt
+++ b/src/displayapp/fonts/CMakeLists.txt
@@ -1,19 +1,35 @@
-cmake_minimum_required(VERSION 3.10)
-
 set(FONTS jetbrains_mono_42 jetbrains_mono_76 jetbrains_mono_bold_20
    jetbrains_mono_extrabold_compressed lv_font_navi_80 lv_font_sys_48
    open_sans_light)
-find_program(LV_FONT_CONV "lv_font_conv" NO_CACHE REQUIRED)
+find_program(LV_FONT_CONV "lv_font_conv" NO_CACHE REQUIRED
+   HINTS "${CMAKE_SOURCE_DIR}/node_modules/.bin")
+message(STATUS "Using ${LV_FONT_CONV} to generate font files")
 configure_file(${CMAKE_CURRENT_LIST_DIR}/jetbrains_mono_bold_20.c_zero.patch
-   displayapp/fonts/jetbrains_mono_bold_20.c_zero.patch COPYONLY)
-foreach(FONT ${FONTS})
-   set_source_files_properties(displayapp/fonts/${FONT}.c
-      PROPERTIES GENERATED TRUE)
+   ${CMAKE_CURRENT_BINARY_DIR}/jetbrains_mono_bold_20.c_zero.patch COPYONLY)
+if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.12)
+   # FindPython3 module introduces with CMake 3.12
+   # https://cmake.org/cmake/help/latest/module/FindPython3.html
+   find_package(Python3 REQUIRED)
+else()
+   set(Python3_EXECUTABLE "python")
+endif()
 
-   add_custom_command(OUTPUT displayapp/fonts/${FONT}.c
-      COMMAND python ${CMAKE_CURRENT_LIST_DIR}/generate.py
-      -f ${FONT} ${CMAKE_CURRENT_LIST_DIR}/fonts.json
-      DEPENDS ${CMAKE_CURRENT_LIST_DIR}/fonts.json
-      WORKING_DIRECTORY displayapp/fonts
-      )
+# create static library building fonts
+add_library(infinitime_fonts STATIC)
+# add include directory to lvgl headers needed to compile the font files on its own
+target_include_directories(infinitime_fonts PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}/../../libs")
+foreach(FONT ${FONTS})
+   add_custom_command(
+      OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/${FONT}.c
+      COMMAND "${Python3_EXECUTABLE}" ${CMAKE_CURRENT_SOURCE_DIR}/generate.py
+      --lv-font-conv "${LV_FONT_CONV}"
+      --font ${FONT} ${CMAKE_CURRENT_SOURCE_DIR}/fonts.json
+      DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/fonts.json
+      WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+   )
+   add_custom_target(infinitime_fonts_${FONT}
+      DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/${FONT}.c
+   )
+   target_sources(infinitime_fonts PRIVATE "${CMAKE_CURRENT_BINARY_DIR}/${FONT}.c")
+   add_dependencies(infinitime_fonts infinitime_fonts_${FONT})
 endforeach()
--- a/src/displayapp/fonts/generate.py
+++ b/src/displayapp/fonts/generate.py
@@ -18,8 +18,8 @@ class Source(object):
         self.symbols = d.get('symbols')
 
 
-def gen_lvconv_line(dest: str, size: int, bpp: int, sources: typing.List[Source], compress:bool=False):
-    args = ['lv_font_conv', '--size', str(size), '--output', dest, '--bpp', str(bpp), '--format', 'lvgl']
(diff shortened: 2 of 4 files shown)


NeroBurner:
@yehoshuapw could you review my changes

And maybe also the sister PR in InfiniSim https://github.com/InfiniTimeOrg/InfiniSim/pull/29

Riksu9000 (commented):
Looks good as far as I can tell.

src/displayapp/fonts/generate.py
```diff
@@ -35,10 +35,11 @@ def main():
     ap = argparse.ArgumentParser(description='auto generate LVGL font files from fonts')
     ap.add_argument('config', type=str, help='config file to use')
     ap.add_argument('-f', '--font', type=str, action='append', help='Choose specific fonts to generate (default: all)', default=[])
+    ap.add_argument('--lv-font-conv', type=str, help='Path to "lv_font_conf" executable (default: lv_font_conf)', default="lv_font_conv")
```
yehoshuapw, line 38: don't bother with the "(default: lv_font_conf)" in the help, if there's a default argparse will do that for you.

yehoshuapw, line 38: it doesn't really matter, except for consistency - but consider adding a short flag too, and using it in the CMake (or converting all of them to the long versions)

src/displayapp/fonts/generate.py
```diff
@@ -35,10 +35,11 @@ def main():
...
     args = ap.parse_args()
 
-    if not shutil.which('lv_font_conv'):
-        sys.exit(f'Missing lv_font_conv. (make sure it is installed and in PATH)')
+    if not shutil.which(args.lv_font_conv):
+        sys.exit(f'Missing lv_font_conv. (make sure it is installed and in PATH or provide full path with --lv-font-conv parameter)')
```
yehoshuapw, line 42: don't write here how, the `-h` flag does that.
Also, to make this happen in a neater way, you can use argparse's validate to make sure the path exists.

src/displayapp/fonts/CMakeLists.txt
```diff
@@ -1,19 +1,34 @@
-cmake_minimum_required(VERSION 3.10)
-
 set(FONTS jetbrains_mono_42 jetbrains_mono_76 jetbrains_mono_bold_20
    jetbrains_mono_extrabold_compressed lv_font_navi_80 lv_font_sys_48
    open_sans_light)
-find_program(LV_FONT_CONV "lv_font_conv" NO_CACHE REQUIRED)
+find_program(LV_FONT_CONV "lv_font_conv" NO_CACHE REQUIRED
+   HINTS "${CMAKE_SOURCE_DIR}/node_modules/.bin")
```
yehoshuapw, line 5: I like this better then trusting PATH. once it's here, may as well add the global node_modules bin's too.

yehoshuapw (commented):
overall, it looks excellent (and the CMake much better) - thanks.
I added a few comment for consistency.

src/displayapp/fonts/CMakeLists.txt
```diff
@@ -1,19 +1,34 @@
...
+message(STATUS "Using ${LV_FONT_CONV} to generate font files")
```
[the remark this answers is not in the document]
yehoshuapw: it is. npm is a mess, and trusting versions of things to play nice is optimistic.
so having each thing you need in it's own env works better. (and also, won't clutter up normal paths)

src/displayapp/fonts/generate.py
NeroBurner:
```suggestion
    ap.add_argument('--lv-font-conv', type=str, help='Path to "lv_font_conf" executable', default="lv_font_conv")
```

src/displayapp/fonts/CMakeLists.txt
NeroBurner: do you know where this global `node_modules` folder is?

Riksu9000 (commented):
I noticed that this always re-generates the font, whereas before it didn't if it wasn't necessary.

NeroBurner:
valid concern, yes I need to fix that! Thanks for pointing that out

JF002 (approved):
That looks really good to me! I like the new static library that contains all the fonts! :+1: 
We'll be able to merge when the remaining reviews on the Python scripts are solved :)

src/displayapp/fonts/CMakeLists.txt
yehoshuapw: seems to be `$HOME/.local/lib64/node_modules`, and the binary ended up in `$HOME/.local/bin/lv_font_conv` too.

src/displayapp/fonts/CMakeLists.txt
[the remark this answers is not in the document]
NeroBurner: updated documentation to just mention `npm install lv_font_conv`. No need to mention `-g` and adding `lv_font_conv` to path anymore.

src/displayapp/fonts/CMakeLists.txt
NeroBurner: is that path in the global `PATH` variable? Maybe already added by `npm`? Then we won't need to add a `HINT` for it

src/displayapp/fonts/generate.py
NeroBurner: I now use the long flags in cmake

src/displayapp/fonts/generate.py
NeroBurner: thanks, fixed

NeroBurner:
@Riksu9000 fixed the always regenerating fonts. Now its generated only once (and if the `fonts.json` file is modified)

src/displayapp/fonts/generate.py
yehoshuapw: I mean the "--lv-font-conv parameter" part. leaving at "make sure in path" is also confusing. (sorry, I wasn't very clear. the point was to tell the user he can specify any other path, just now _how_, since that is done in the `-h`.)

src/displayapp/fonts/CMakeLists.txt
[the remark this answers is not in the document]
yehoshuapw: that works.

src/displayapp/fonts/CMakeLists.txt
yehoshuapw: `.local/bin` should be, but sometimes isn't. however, it is generic enough that trusting that it is seems fine to me.

src/displayapp/fonts/generate.py
NeroBurner: could you provide a new error message with a better formulation?

yehoshuapw: how about: `"Missing lv_font_conv. (make sure it is installed. and make sure it's in PATH or specify the path to it manually)"`
or just `"Missing lv_font_conv. Make sure it's findable (in PATH) or specify it manually"`  (this does not say anything about it being installed, but that can be implicitly understood perhaps)

NeroBurner: done, used the second one :)

yehoshuapw: excellent. thanks!

yehoshuapw (approved):

Riksu9000 (approved):

ITCactus:
and maybe it's make sense to note, that "node" version must be a fresh one...
tried to build on Ubuntu 20LTS, and faced the issue:
> TypeError: this.src.glyphs.map(...).flat is not a function

and it looks like the known issue like [this](https://github.com/lvgl/lv_font_conv/issues/61) and [this](https://github.com/lvgl/lv_font_conv/issues/73).
so, i had to update the Node to make it work :)

yehoshuapw:
> > TypeError: this.src.glyphs.map(...).flat is not a function

I actually knew that, (which is why the docker installs npm's repos), and really should have kept track for such an issue.. (and added some comment somewhere about it)

PR #1132, opened 2022-05-10
Abstract acceleration sensor and add support for SC7A20 chip used in P8

StarGate01:

- Abstract the acceleration sensor driver to inherit from the new `AccelerationSensor` base class
- Add driver for the SC7A20 chip
- Use FIFO buffer for smoothing for both chips. This improves the quality of the data.
- The CMake configuration has been extended to support driver selection via `DRIVER_ACC`


StarGate01:
The simulator fails to build because the signature of the constructor `Pinetime::System::SystemTask::SystemTask` was changed, it now expects a `Pinetime::Drivers::AccelerationSensor` instead of `Pinetime::Drivers::Bma421`.

StarGate01:
Ok, I moved the FIFO changes for the BMA421 to #1145 .

PR #1136, opened 2022-05-13
Add a Counter widget and use it in Timer

Riksu9000:
The Counter widget is a reusable self contained counter, which is easy to use in apps.

It is a big blocky object. This is slightly contrasting with other apps, which mostly have smaller objects separated by black, however I think it makes more sense for the counter widget to be connected.

The font was made lighter, to look more "balanced". For the counter this is important. It affects other apps very slightly. It may be worth making the font slightly smaller in the future.

The Timer app's look was updated along with adding the Counters. The text is now pure white. The buttons are pure gray. Instead of a symbol, the button has text now. I found that for wide buttons, text looks better, and for square buttons a symbol looks better. This could become a part of our new style. I wanted to colorize the big button, but in the end I kept this first PR tamer.

This may not be perfect in every way, but we can start a UI overhaul from here.

InfiniSim_2022-05-13_185844
InfiniSim_2022-05-13_185850

Compare the font thickness.
InfiniSim_2022-05-13_192726
InfiniSim_2022-05-13_192755


Riksu9000:
I have locally made this work on the SetTime screen. I think it'll be better for other screens to be separate PRs after this one though.

InfiniSim_2022-05-14_104335

src/displayapp/screens/Timer.h
```diff
@@ -19,23 +20,15 @@ namespace Pinetime::Applications::Screens {
...
-    lv_obj_t* btnMinutesUp;
-    lv_obj_t* btnMinutesDown;
-    lv_obj_t* btnSecondsUp;
-    lv_obj_t* btnSecondsDown;
-    lv_obj_t* txtMUp;
-    lv_obj_t* txtMDown;
-    lv_obj_t* txtSUp;
-    lv_obj_t* txtSDown;
     lv_task_t* taskRefresh;
+    Widgets::Counter minuteCounter = Widgets::Counter(0, 59);
```
NeroBurner, line 31: is there a reason why we cant have for example 99 minutes in the counter?

src/displayapp/widgets/Counter.cpp
```diff
@@ -0,0 +1,105 @@
...
+  counterContainer = lv_obj_create(lv_scr_act(), nullptr);
+  lv_obj_set_style_local_bg_color(counterContainer, LV_BTN_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_MAKE(0x38, 0x38, 0x38));
+
+  number = lv_label_create(counterContainer, nullptr);
+  lv_obj_set_style_local_text_font(number, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, &jetbrains_mono_76);
+  lv_obj_align(number, nullptr, LV_ALIGN_CENTER, 0, 0);
+  lv_obj_set_auto_realign(number, true);
+  lv_label_set_text_static(number, "00");
+
+  static const uint8_t padding = 5;
```
NeroBurner, line 57: we could make this variable constexpr

```suggestion
  constexpr uint8_t padding = 5;
```

src/displayapp/widgets/Counter.cpp
```diff
@@ -0,0 +1,105 @@
...
+  const uint8_t width = lv_obj_get_width(number) + padding * 2;
+  static const uint8_t btnHeight = 50;
```
NeroBurner, line 59: we could make this variable constexpr

```suggestion
  constexpr uint8_t btnHeight = 50;
```

src/displayapp/widgets/Counter.cpp
```diff
@@ -0,0 +1,105 @@
...
+  UpdateLabel();
+}
+
+void Counter::UpdateLabel() {
+  lv_label_set_text_fmt(number, "%.2i", value);
+}
+
+void Counter::Create() {
+  counterContainer = lv_obj_create(lv_scr_act(), nullptr);
+  lv_obj_set_style_local_bg_color(counterContainer, LV_BTN_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_MAKE(0x38, 0x38, 0x38));
```
NeroBurner, line 49: we could define the `LV_COLOR_MAKE(0x38, 0x38, 0x38)` only once and give it a name

src/displayapp/widgets/Counter.cpp
```diff
@@ -0,0 +1,105 @@
...
+  lv_obj_align(downBtn, nullptr, LV_ALIGN_IN_BOTTOM_MID, 0, 0);
+  downBtn->user_data = this;
+  lv_obj_set_event_cb(downBtn, downBtnEventHandler);
+
+  lv_obj_t* downLabel = lv_label_create(downBtn, nullptr);
+  lv_obj_set_style_local_text_font(downLabel, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, &jetbrains_mono_42);
+  lv_label_set_text_static(downLabel, "-");
+  lv_obj_align(downLabel, nullptr, LV_ALIGN_CENTER, 0, 0);
+
+  static lv_point_t line_points[] = {{0, 0}, {width, 0}};
```
NeroBurner, line 90: why is this variable static? Could you please explain this usage to me (`static` keyword isn't my strong side)

src/displayapp/widgets/Counter.h
```diff
@@ -0,0 +1,39 @@
...
+        Counter(int min, int max);
+
+        void Create();
+        static void upBtnEventHandler(lv_obj_t* obj, lv_event_t event);
+        static void downBtnEventHandler(lv_obj_t* obj, lv_event_t event);
+        void Increment();
+        void Decrement();
+        void SetValue(int newValue);
+
+        int GetValue() const {
```
NeroBurner, line 18: I'd move the body of the function into the cpp file to prevent multiple definition errors (just best practice, for our smallish project with just one binary it should be no problem, but better be safe)

src/displayapp/widgets/Counter.h
```diff
@@ -0,0 +1,39 @@
...
+          return value;
+        }
+
+        lv_obj_t* GetObject() const {
```
NeroBurner, line 22: I'd move the body of the function into the cpp file

src/displayapp/widgets/Counter.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
+#include <lvgl/lvgl.h>
+
+namespace Pinetime {
+  namespace Applications {
+    namespace Widgets {
+      class Counter {
+      public:
+        Counter(int min, int max);
```
NeroBurner, line 9: if the counter can only hold values with 2 digits. Then we could change this to `int8_t`. If we further assume to not have negative numbers we could use `uint8_t`

src/displayapp/screens/Timer.cpp
```diff
@@ -10,62 +10,32 @@ static void btnEventHandler(lv_obj_t* obj, lv_event_t event) {
...
-
-Timer::Timer(DisplayApp* app, Controllers::TimerController& timerController)
-  : Screen(app), running {true}, timerController {timerController} {
+Timer::Timer(DisplayApp* app, Controllers::TimerController& timerController) : Screen(app), timerController {timerController} {
 
   time = lv_label_create(lv_scr_act(), nullptr);
   lv_obj_set_style_local_text_font(time, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, &jetbrains_mono_76);
-  lv_obj_set_style_local_text_color(time, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_MAKE(0xb0, 0xb0, 0xb0));
+  lv_obj_set_style_local_text_color(time, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_WHITE);
+  lv_label_set_text_static(time, "00:00");
```
NeroBurner, line 18: we could replace the text by adding a function `Counter::HideControls()` and `Counter::ShowControls()` to not hide the whole counter-object, but tell the counter it should just hide the controls. Then we don't need the `time` object anymore

NeroBurner (changes requested):

src/displayapp/screens/Timer.h
Riksu9000: There's an overflow that happens in `pdMS_TO_TICKS()` in TimerController. There's also the fact that reaching certain values will take longer when the limit is higher, so I don't want to change it in this PR unnecessarily.

src/displayapp/widgets/Counter.cpp
Riksu9000: Would this be preferrable? I don't think there's a difference in the result.

src/displayapp/widgets/Counter.cpp
Riksu9000: `lv_line_set_points()` only copies the pointer to the array, so line_points must not be a local variable, however it should just be a member variable.

src/displayapp/widgets/Counter.h
Riksu9000: I think it needs to be in the header file for it to get inlined.

src/displayapp/widgets/Counter.cpp
NeroBurner: You're right, the end result with optimization will most likely be the same

for the `static const` the compiler will create a global variable, which with high probability will be optimized out later

with `constexpr` the variable will not be created in the first place (as far as I understand). It's more explicit to write `constexpr` to express what we actually want instead of relying on the knowledge that the optimizer will remove the variable in the end

src/displayapp/widgets/Counter.cpp
NeroBurner: thanks for the explanation. Could the variable be `const static`?

src/displayapp/widgets/Counter.h
NeroBurner: I didn't know that, thanks!

Stackoverflow provided me an additional interesting read https://stackoverflow.com/questions/453372/writing-function-definition-in-header-files-in-c

src/displayapp/widgets/Counter.h
NeroBurner: small member functions can get easily inlined by the compiler. Thanks @Riksu9000 for pointing that out

Ertain:
Is there a way to reset the counter to zero?

src/displayapp/widgets/Counter.cpp
Riksu9000: I changed it to static constexpr to force compile time initialization, since just constexpr is technically just a suggestion.

Riksu9000:
> Is there a way to reset the counter to zero?

Not really in the scope of this PR.

src/displayapp/widgets/Counter.h
Riksu9000: The Counter can easily be made to hold larger values if there's a use for it. It looks like using int actually produces a tiny bit smaller binary. Would there be benefits to changing it?

NeroBurner: int makes is smaller? that's unexpected. It would have just been more explicit and matching the value range the counter is designed for better

Riksu9000: I think it's because int is the native integer size of the processor. This may be something worth looking into elsewhere as well. There's int_fast8_t for example, which may use a larger size if it is faster, and in this case using it would produce the same sized binary, assuming it uses the native size.

src/displayapp/screens/Timer.cpp
```diff
@@ -10,62 +10,32 @@ static void btnEventHandler(lv_obj_t* obj, lv_event_t event) {
...
 
-void Timer::CreateButtons() {
-  btnMinutesUp = lv_btn_create(lv_scr_act(), nullptr);
-  btnMinutesUp->user_data = this;
-  lv_obj_set_event_cb(btnMinutesUp, btnEventHandler);
-  lv_obj_set_size(btnMinutesUp, 60, 40);
-  lv_obj_align(btnMinutesUp, lv_scr_act(), LV_ALIGN_IN_LEFT_MID, 20, -85);
-  txtMUp = lv_label_create(btnMinutesUp, nullptr);
-  lv_label_set_text_static(txtMUp, "+");
+Timer::Timer(DisplayApp* app, Controllers::TimerController& timerController) : Screen(app), timerController {timerController} {
```
NeroBurner, line 13: nitpick: this line is SO LONG! Can we update the clang formatter to make constructor elements be on their own lines? Makes diffs easier as well in my opinion

src/displayapp/widgets/Counter.h
```diff
@@ -0,0 +1,44 @@
...
+
+        void Create();
+        static void upBtnEventHandler(lv_obj_t* obj, lv_event_t event);
```
NeroBurner, line 12: do we need the evenhandlers to be in the public interface? In Timer.cpp the button event handler is only in the cpp file

src/displayapp/widgets/Counter.cpp
```diff
@@ -0,0 +1,123 @@
...
+  lv_obj_set_style_local_line_color(upperLine, LV_LINE_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_WHITE);
+  lv_obj_set_style_local_line_opa(upperLine, LV_LINE_PART_MAIN, LV_STATE_DEFAULT, LV_OPA_20);
+  lv_obj_align(upperLine, upBtn, LV_ALIGN_OUT_BOTTOM_MID, 0, 0);
+
+  lowerLine = lv_line_create(counterContainer, nullptr);
+  lv_line_set_points(lowerLine, linePoints, 2);
+  lv_obj_set_style_local_line_width(lowerLine, LV_LINE_PART_MAIN, LV_STATE_DEFAULT, 1);
+  lv_obj_set_style_local_line_color(lowerLine, LV_LINE_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_WHITE);
+  lv_obj_set_style_local_line_opa(lowerLine, LV_LINE_PART_MAIN, LV_STATE_DEFAULT, LV_OPA_20);
+  lv_obj_align(lowerLine, downBtn, LV_ALIGN_OUT_TOP_MID, 0, -1);
```
NeroBurner, line 122: not required, but reducing duplicated code with local lambda

```suggestion
  auto counter_line_create = [&]() {
    lv_obj_t *line = lv_line_create(counterContainer, nullptr);
    lv_line_set_points(line, linePoints, 2);
    lv_obj_set_style_local_line_width(line, LV_LINE_PART_MAIN, LV_STATE_DEFAULT, 1);
    lv_obj_set_style_local_line_color(line, LV_LINE_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_WHITE);
    lv_obj_set_style_local_line_opa(line, LV_LINE_PART_MAIN, LV_STATE_DEFAULT, LV_OPA_20);
    return line;
  };
  upperLine = counter_line_create();
  lv_obj_align(upperLine, upBtn, LV_ALIGN_OUT_BOTTOM_MID, 0, 0);

  lowerLine = counter_line_create();
  lv_obj_align(lowerLine, downBtn, LV_ALIGN_OUT_TOP_MID, 0, -1);
```

NeroBurner, line 122: not required, just proposal to reduce code duplication using lambda. Would switch back to `static const linePoints` array, as it isn't needed outside of cpp file.

```suggestion
  auto counter_line_create = [&]() {
    static const lv_point_t linePoints[2] = {{0, 0}, {width, 0}};
    lv_obj_t *line = lv_line_create(counterContainer, nullptr);
    lv_line_set_points(line, linePoints, 2);
    lv_obj_set_style_local_line_width(line, LV_LINE_PART_MAIN, LV_STATE_DEFAULT, 1);
    lv_obj_set_style_local_line_color(line, LV_LINE_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_WHITE);
    lv_obj_set_style_local_line_opa(line, LV_LINE_PART_MAIN, LV_STATE_DEFAULT, LV_OPA_20);
    return line;
  };
  upperLine = counter_line_create();
  lv_obj_align(upperLine, upBtn, LV_ALIGN_OUT_BOTTOM_MID, 0, 0);

  lowerLine = counter_line_create();
  lv_obj_align(lowerLine, downBtn, LV_ALIGN_OUT_TOP_MID, 0, -1);
```

src/displayapp/screens/Timer.cpp
Riksu9000: This is controlled by ColumnLimit, which is 140. I might also like to decrease it, however that's not in the scope of this PR.

src/displayapp/widgets/Counter.h
```diff
@@ -0,0 +1,44 @@
...
+      private:
+        void UpdateLabel();
+
+        lv_obj_t* counterContainer;
+        lv_obj_t* upBtn;
+        lv_obj_t* downBtn;
+        lv_obj_t* number;
+        lv_obj_t* upperLine;
+        lv_obj_t* lowerLine;
+        lv_point_t linePoints[2];
```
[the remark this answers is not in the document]
Riksu9000: I'd rather keep the linePoints in the class, since the LVGL line expects it to be always available, so that would seem safer to me.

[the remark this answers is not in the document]
Riksu9000: It might become a concern when there are Counters with different widths.

NeroBurner (approved):
LGTM

kieranc:
Looks good, more in keeping with the current style. I'll break parts of my previous attempt out into a PR to add the long press rapid increment if you like?

Riksu9000:
Thanks for the offer, however long pressing is already handled by this.

kieranc:
> Thanks for the offer, however long pressing is already handled by this.

Whoops, I should have looked more closely. Very nice 😎

JF002 (approved):
That looks really good to me (both the code and the result on the display :) ).
And this is a great use of our new font generator by @yehoshuapw  :+1: 

And the long-press functionality is just awesome!

PR #1137, opened 2022-05-14
DigitalWatchFace & WatchFacePineTimeStyle: show clock icon when alarm is active

faxe1008:
When using the device I am often uncertain if I did infact set the alarm.
This patch adds display of the clock icon for the DigitalWatchFace and WatchFacePineTimeStyle.

I decided not to add it to the other onces since:
- Terminal is already quite full and I did not want to clutter it further
- Analog: It did not fit the style, since the bleIcon is also not shown I assume this is meant to be more minimalistic.

Here are some screenshots:
image
image
image
image

### DIFF
--- a/src/displayapp/screens/WatchFaceDigital.cpp
+++ b/src/displayapp/screens/WatchFaceDigital.cpp
@@ -22,7 +22,8 @@ WatchFaceDigital::WatchFaceDigital(DisplayApp* app,
                                    Controllers::NotificationManager& notificatioManager,
                                    Controllers::Settings& settingsController,
                                    Controllers::HeartRateController& heartRateController,
-                                   Controllers::MotionController& motionController)
+                                   Controllers::MotionController& motionController,
+                                   Controllers::AlarmController& alarmController)
   : Screen(app),
     currentDateTime {{}},
     dateTimeController {dateTimeController},
@@ -31,7 +32,8 @@ WatchFaceDigital::WatchFaceDigital(DisplayApp* app,
     notificatioManager {notificatioManager},
     settingsController {settingsController},
     heartRateController {heartRateController},
-    motionController {motionController} {
+    motionController {motionController},
+    alarmController {alarmController} {
 
   batteryIcon.Create(lv_scr_act());
   lv_obj_align(batteryIcon.GetObject(), lv_scr_act(), LV_ALIGN_IN_TOP_RIGHT, 0, 0);
@@ -45,6 +47,10 @@ WatchFaceDigital::WatchFaceDigital(DisplayApp* app,
   lv_obj_set_style_local_text_color(bleIcon, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(0x0082FC));
   lv_label_set_text_static(bleIcon, Symbols::bluetooth);
   lv_obj_align(bleIcon, batteryPlug, LV_ALIGN_OUT_LEFT_MID, -5, 0);
+
+  alarmIcon = lv_label_create(lv_scr_act(), nullptr);
+  lv_label_set_text_static(alarmIcon, Symbols::clock);
+  lv_obj_align(alarmIcon, bleIcon, LV_ALIGN_OUT_LEFT_MID, -5, 0);
 
   notificationIcon = lv_label_create(lv_scr_act(), nullptr);
   lv_obj_set_style_local_text_color(notificationIcon, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(0x00FF00));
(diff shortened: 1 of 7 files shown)


faxe1008:
After showing it on discord the idea came up to change the placement and ordering of icons, now PinetimeStyle watch face looks like this:
image
image
image

This way the battery icon lines up with the battery icon in the menu screen :^)

src/displayapp/screens/WatchFacePineTimeStyle.cpp
```diff
@@ -34,6 +34,7 @@
 #include "components/motion/MotionController.h"
 #include "components/settings/Settings.h"
 #include "displayapp/DisplayApp.h"
+#include "components/alarm/AlarmController.h"
```
NeroBurner, line 37: please move the new include up to the other component includes, to have them alphabetically ordered (just before the components/ble include please

NeroBurner (changes requested):

src/displayapp/screens/WatchFacePineTimeStyle.cpp
faxe1008: Done :^)

src/displayapp/screens/WatchFaceDigital.cpp
```diff
@@ -46,6 +48,11 @@ WatchFaceDigital::WatchFaceDigital(DisplayApp* app,
   lv_label_set_text_static(bleIcon, Symbols::bluetooth);
   lv_obj_align(bleIcon, batteryPlug, LV_ALIGN_OUT_LEFT_MID, -5, 0);
 
+  alarmIcon = lv_label_create(lv_scr_act(), nullptr);
+  lv_obj_set_style_local_text_color(alarmIcon, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(0x999999));
```
NeroBurner, line 52: I think the grey alarm icon doesn't feel quite right. I would expect it to be white
```suggestion
```

WatchFaceDigital white alarm icon

src/displayapp/screens/WatchFaceDigital.h
```diff
@@ -8,6 +8,7 @@
 #include "displayapp/screens/Screen.h"
 #include "components/datetime/DateTimeController.h"
 #include "components/ble/BleController.h"
+#include "components/alarm/AlarmController.h"
```
NeroBurner, line 11: reordering includes to be alphabetically

```suggestion
#include "components/alarm/AlarmController.h"
#include "components/ble/BleController.h"
#include "components/datetime/DateTimeController.h"
```

src/displayapp/screens/WatchFacePineTimeStyle.cpp
```diff
@@ -340,13 +347,22 @@ void WatchFacePineTimeStyle::SetBatteryIcon() {
...
+    lv_obj_align(alarmIcon, sidebar, LV_ALIGN_IN_TOP_MID, 9, 26);
+    lv_obj_align(notificationIcon, sidebar, LV_ALIGN_IN_TOP_MID, -9, 25);
+  } else if (notificationState.Get() && !alarmEnabled.Get()) {
     lv_obj_align(notificationIcon, sidebar, LV_ALIGN_IN_TOP_MID, 0, 25);
   } else {
-    lv_obj_align(bleIcon, sidebar, LV_ALIGN_IN_TOP_MID, 0, 25);
+    lv_obj_align(alarmIcon, sidebar, LV_ALIGN_IN_TOP_MID, 0, 26);
+  }
+
+  if (bleState.Get()) {
```
NeroBurner, line 359: please move the bluetooth alignment up before the notification/alarm alignment, to make the code better match the actual layout

NeroBurner (changes requested):
could you rebase your commits onto the current `develop` branch including the fonts to fix the InfiniSim build job

faxe1008:
Resolved all remarks :^)

NeroBurner (approved):
LGTM, clang-format found something, but not at your changes

Riksu9000 (changes requested):
The gaps between the icons are inconsistent. This is because even though an icon is hidden, it's aligned 5 pixels left, so effectively it increases the gap. This wasn't a big issue until now, but now it is noticable and I think something should be done about it.

NeroBurner:
the nice and bad thing about programming: you can do it in multiple ways! :D

it sounds like a good way to start. Try it out and see if it works :+1:

JM480:
I think there should be an icon for the timer as well.

Riksu9000:
With #1242 it's now easier to add new status icons. Will you update this PR?

faxe1008:
Yes I am, I tried fiddeling around fixing the issue but it did not work, simply did not have any lvgl experience to know whats going wrong. I will at least update the digital watchface to include the info. As for the PineTimeStyle one - I think if we want to have the logic for what icon to show under which conditions still in one spot, I think that having some "grid" layout of the icons be the best approach. That way for the DigitalWatchFace the icons can be 4 columns, 1 row and for the PineTimeStyle one 2x2?

Will first do the DigitalFace one before I try and get fancy with the other one tho :^)

Riksu9000:
Closing in favor of adding this in the StatusIcons widget.

PR #1138, opened 2022-05-15
Github Action with docker to build the firmware

JF002:
**NOTE** : This PR replaces #1045 (it contains many 'tentative' commits and a merge probably failed which resulted in missing commits).

The main build workflow for GitHub Actions now uses the infinitime-build docker image to build the firmware. This allows to avoid the duplication of the build logic in Docker and Github workflow format. It also ensures that the project is built exactly the same way on dev environment and on Github Actions.

The Dockerfile has been slightly adapted (remove the USER instructions). Docker images are now published in DockerHub (jf002/infinitime-build).

- [x] Create a Docker account for InfiniTime and publish the images using this account.


doc/buildWithDocker.md
```diff
@@ -44,19 +44,19 @@ docker run --rm -it -v $(pwd):/sources --user 1234:1234 infinitime-build
...
 
 ```bash
-docker pull pfeerick/infinitime-build
+docker pull jf002/infinitime-build
 ```
 
 The default `latest` tag *should* automatically identify the correct image architecture, but if for some reason Docker does not, you can specify it manually:
 
-* For AMD64 (x86_64) systems: `docker pull pfeerick/infinitime-build:amd64`
+* For AMD64 (x86_64) systems: `docker pull jf002/infinitime-build:amd64`
```
NeroBurner, line 57: According to https://stackoverflow.com/questions/60114854/pull-docker-image-for-different-architecture the `--platform` parameter is now the easiest way to explicitly specify the architecture

```sh
docker pull --platform linux/amd64 jf002/infinitime-build
```

```suggestion
* For AMD64 (x86_64) systems: `docker pull --platform linux/amd64 jf002/infinitime-build`
```

doc/buildWithDocker.md
```diff
@@ -44,19 +44,19 @@ docker run --rm -it -v $(pwd):/sources --user 1234:1234 infinitime-build
...
 
-* For ARM64v8 (ARM64/aarch64) systems: `docker pull pfeerick/infinitime-build:arm64v8`
+* For ARM64v8 (ARM64/aarch64) systems: `docker pull jf002/infinitime-build:arm64v8`
```
NeroBurner, line 59:
```suggestion
* For ARM64v8 (ARM64/aarch64) systems: `docker pull --platform linux/arm64 jf002/infinitime-build`
```

NeroBurner (changes requested):
small modernization with `--platform` parameter in docker readme

doc/buildWithDocker.md
JF002: Thanks for the suggestion. Now, is that parameter really needed? I should detect the platform automatically if it's not specified, I think.

NeroBurner: no it is not needed. Like stated in the paragraph above the `--platform` parameter can be used to specify the architecture manually if the auto-detect wouldn't work. I just updated the command stated for this purpose to be a bit more up to date. 

I wanted to change the provided commands, because I didn't like the way the tag is used to specify the architecture in those example commands (works, but feels wrong for me). The tag should be used for a version of the package (like for the ubuntu package it defines the release, like `ubuntu:jammy` or `ubuntu:focal`)

JF002: Thanks, I've just pushed those changes!

JF002:
I've just created a Docker account for InfiniTime : https://hub.docker.com/repository/docker/infinitime/infinitime-build

NeroBurner:
This statement is obsolete, isn't it? The image per default runs as `root` and we should run the `docker run` commands with the `--user` parameter 

> The image is built using 1000:1000 for the user id and group id. If this is different to your user or group ids (run `id -u` and `id -g` to find out what your id values are if you are unsure), you will need to override them via the `--user` parameter in order to prevent permission errors with the output files (and the cmake build cache).

JF002:
> This statement is obsolete, isn't it? The image per default runs as `root` and we should run the `docker run` commands with the `--user` parameter

@NeroBurner  Yes, right! I've just fixed that. Could you have a look?

doc/buildWithDocker.md
```diff
@@ -19,44 +20,38 @@ This example will build the firmware, generate the MCUBoot image and generate th
...
-The image is available via Docker Hub for both the amd64 and arm64v8 architectures at [jf002/infinitime-build](https://hub.docker.com/repository/docker/jf002/infinitime-build).
+The command line to build the project is the same as above. You just need to change the image name to `infinitime/infinitime-build`:
 
-It can be pulled (downloaded) using the following command:
-
-```bash
-docker pull jf002/infinitime-build
+```
+docker run --rm -it -v $(pwd):/sources --user $(id -u):$(id -g) infinitime/infinitime-build
 ```
```
NeroBurner, line 48: the mentioned image name we need to replace is exactly the same we use in the command above. I think this paragraph is obsolete

doc/buildWithDocker.md
```diff
@@ -66,10 +61,4 @@ The following commands must be run from the root of the project. This operation
 
 ```bash
 docker image build -t infinitime-build ./docker
```
NeroBurner, line 63: Nitpick: I think the `image` in the build command is redundant. I'm always using just `docker build` (could someone with working docker test the command without `image`?)

```suggestion
docker build -t infinitime-build ./docker
```

doc/buildWithDocker.md
```diff
@@ -19,44 +20,38 @@ This example will build the firmware, generate the MCUBoot image and generate th
 
 ```bash
 cd <project_root> # e.g. cd ./work/Pinetime
-docker run --rm -it -v $(pwd):/sources infinitime-build
+docker run --rm -it -v $(pwd):/sources --user $(id -u):$(id -g) infinitime-build
```
NeroBurner, line 23: optional: one less called command `pwd`. Use variable `${PWD}` instead
```suggestion
docker run --rm -it -v ${PWD}:/sources --user $(id -u):$(id -g) infinitime-build
```

NeroBurner (changes requested):

doc/buildWithDocker.md
```diff
@@ -19,44 +20,38 @@ This example will build the firmware, generate the MCUBoot image and generate th
...
 ```
 
+By default, the container runs as `root`, which is not convenient as all the file generated by the build will also belong to `root`. The parameter `--user` overrides that default behavior. The command above will run as your current user. 
+
 If you only want to build a single CMake target, you can pass it in as the first parameter to the build script. This means calling the script explicitly as it will override the `CMD`. Here's an example For `pinetime-app`:
```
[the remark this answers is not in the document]
JF002: Done!

doc/buildWithDocker.md
JF002: Yep, I tested that command, it runs perfectly!

doc/buildWithDocker.md
```diff
@@ -19,44 +20,38 @@ This example will build the firmware, generate the MCUBoot image and generate th
...
 
 ```bash
-docker run --rm -it -v $(pwd):/sources infinitime-build /opt/build.sh pinetime-app
+docker run --rm -it -v $(pwd):/sources --user $(id -u):$(id -g) infinitime-build /opt/build.sh pinetime-app
```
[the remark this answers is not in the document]
JF002: Done!

JF002:
@NeroBurner I pushed the changes you suggested :)

NeroBurner (approved):
LGTM

PR #1142, opened 2022-05-18
comment about common error "glyphs.map" in old nodejs

yehoshuapw:
Added a comment in doc/buildAndProgram.md about nodejs needing to be updated if the error 
> "TypeError: this.src.glyphs.map(...).flat is not a function"

 pops up.

### DIFF
--- a/doc/buildAndProgram.md
+++ b/doc/buildAndProgram.md
@@ -14,7 +14,7 @@ To build this project, you'll need:
  - A reasonably recent version of CMake (I use 3.16.5)
  - lv_font_conv, to generate the font .c files
    - see [lv_font_conv](https://github.com/lvgl/lv_font_conv#install-the-script)
-   - install npm (commonly done via the package manager)
+   - install npm (commonly done via the package manager, might need to upgrade node if you get `TypeError: this.src.glyphs.map(...).flat is not a function`)
    - install lv_font_conv: `npm install lv_font_conv`
 
 ## Build steps 


doc/buildAndProgram.md
```diff
@@ -14,7 +14,7 @@ To build this project, you'll need:
  - A reasonably recent version of CMake (I use 3.16.5)
  - lv_font_conv, to generate the font .c files
    - see [lv_font_conv](https://github.com/lvgl/lv_font_conv#install-the-script)
-   - install npm (commonly done via the package manager)
+   - install npm (commonly done via the package manager, might need to upgrade node if you get `TypeError: this.src.glyphs.map(...).flat is not a function`)
```
NeroBurner, line 17: when do you get the error message? is it during `npm install` or when running `lv_font_conv` executable?

for the first case the warning should be moved to the `npm install` command

if the warning is issued during `lv_font_conv` execution, then I'd add the warning as a new bulletpoint after the `npm install` command

NeroBurner (changes requested):

doc/buildAndProgram.md
Riksu9000: I think it should just say: install npm and make sure node is version 12 or newer. Having the note about the version requirement later doesn't seem right to me.

yehoshuapw: changed to "install npm (commonly done via the package manager. ensure the version is at least 12)" and that's it.
It does look much neater.

Riksu9000: Node needs to be mentioned however, since node and npm have different versions.

doc/buildAndProgram.md
```diff
@@ -14,7 +14,7 @@ To build this project, you'll need:
  - A reasonably recent version of CMake (I use 3.16.5)
  - lv_font_conv, to generate the font .c files
    - see [lv_font_conv](https://github.com/lvgl/lv_font_conv#install-the-script)
-   - install npm (commonly done via the package manager)
+   - install npm (commonly done via the package manager. ensure node's version is at least 12)
```
NeroBurner, line 17: A dot `.` inside a bracketed comment feels wrong, use a comma `,` instead

```suggestion
   - install npm (commonly done via the package manager, ensure node's version is at least 12)
```

NeroBurner (changes requested):

doc/buildAndProgram.md
NeroBurner: I wouldn't add it. For me "package manager" is `apt`, `dnf` or `pacman`

doc/buildAndProgram.md
yehoshuapw: it kind of looks like part of the installation step, which it really isn't. but it does look weird with the dot,
and moving it out just takes too much space for something so basic.

NeroBurner (approved):

PR #1133, opened 2022-05-10
Motion: Expose acceleration FIFO via BLE

StarGate01:
This PR has been broken out of https://github.com/InfiniTimeOrg/InfiniTime/pull/1050, and depends on / includes https://github.com/InfiniTimeOrg/InfiniTime/pull/1132 (which includes https://github.com/InfiniTimeOrg/InfiniTime/pull/1128) .

The motion BLE service now broadcasts the entire FIFO buffer, in order to expose a higher frequency measurement. A high enough MTU has to be negotiated to fit all max. 192 bytes. The format is backwards-compatible.

The internal sensor query rate is unchanged at 100 ms. The FIFO buffer fits about 20 samples per cycle.

An example Android app which uses this high frequency data can be found at https://github.com/StarGate01/PineTimeAcc .

PR #1145, opened 2022-05-18
BMA421: Use internal FIFO for smoothing

StarGate01:
This PR has been broken out of #1132 / #1050.

The BMA421 acceleration chip has an internal FIFO buffer. This buffer is now used to calculate the smoothed output signal. This generates a cleaner signal, and provides the FIFO data for further processing. The sampling rate has been increased to 200 Hz, which gives roughly 20 samples per Infinitime sensor loop (100ms cycle).


src/drivers/Bma421.cpp
```diff
@@ -68,11 +75,21 @@ void Bma421::Init() {
...
+  fifo_frame.fifo_header_enable = 0;
+
+  // Configure accelerometer
   struct bma4_accel_config accel_conf;
-  accel_conf.odr = BMA4_OUTPUT_DATA_RATE_100HZ;
+  accel_conf.odr = BMA4_OUTPUT_DATA_RATE_200HZ;
   accel_conf.range = BMA4_ACCEL_RANGE_2G;
   accel_conf.bandwidth = BMA4_ACCEL_NORMAL_AVG4;
-  accel_conf.perf_mode = BMA4_CIC_AVG_MODE;
+  accel_conf.perf_mode = BMA4_CONTINUOUS_MODE;
```
Riksu9000, line 92: Do these changes have an impact on power usage? Do we need such frequent data?

src/drivers/Bma421.cpp
```diff
@@ -32,30 +32,37 @@ Bma421::Bma421(TwiMaster& twiMaster, uint8_t twiAddress) : twiMaster {twiMaster}
...
+  bma.resolution = 12;
 }
 
 void Bma421::Init() {
-  if (not isResetOk)
+  if (!isResetOk)
     return; // Call SoftReset (and reset TWI device) first!
 
+  // Initialize interface
   auto ret = bma423_init(&bma);
```
Riksu9000, line 43: I personally don't think most of these added comments are required to explain what is happening.

Riksu9000 (commented):
I think the current way works fine, but accessing more data can be useful in the future.

More frequent data and some history would be useful in the raise to wake gesture for example. This also raises the question of how much data should be stored and at what frequency, considering the memory limitations.

I'm assuming the fifo array gets overridden each time the values are processed and the amount of data can vary, which can be problematic when using this data elsewhere. I would expect there to be something like a circular buffer.

src/drivers/Bma421.cpp
StarGate01: 200 Hz is the maximum data rate which can be read by the watch, since the FIFO is able to capture 32 samples at most, and the sensor cycle time is 100 ms. This gives 20 samples per cycle at a sample rate of 200 Hz. I chose this sample rate because I needed the best possible data for #1133 .

`BMA4_CIC_AVG_MODE` and `BMA4_CONTINUOUS_MODE` should make no difference in power consumption, since the internal sample rate stays the same, the AVG mode just averages the samples together. I theory, operating the sensor in continuous mode even offloads some calculations from the sensor chip.

According to the datasheet at https://datasheet.lcsc.com/lcsc/1912111437_Bosch-Sensortec-BMA425_C437656.pdf , the power consumption in normal mode is not correlated to the sampling frequency. This is only the case in low power mode, which was not used here before and still is not used.

StarGate01:
Hey @Riksu9000 I am picking this project back up, and slowly rebasing all my PRs which will take some time.

This one is quite easy. I have cleaned up the comments as you suggested.

Riksu9000:
Hi. I'm sorry, but I'm not active on this project anymore. You'll need to discuss this with @JF002.

PR #1150, opened 2022-05-19
Dismiss Notifications (Swipe right)

Tiggilyboo:
Hello all,

I've implemented dismissal of a notification by swiping right on each. This triggers an animation which throws the animation offscreen, then renders the next available notification.

### DIFF
--- a/src/components/ble/NotificationManager.cpp
+++ b/src/components/ble/NotificationManager.cpp
@@ -95,6 +95,7 @@ void NotificationManager::Dismiss(NotificationManager::Notification::Id id) {
 
   for(size_t i = foundIndex; i < TotalNbNotifications - 1; i++) {
     notifications[i] = notifications[i+1];
+    notifications[i].index = i;
   }
 
   if(count == 1) {
--- a/src/components/ble/NotificationManager.cpp
+++ b/src/components/ble/NotificationManager.cpp
@@ -88,19 +88,13 @@ void NotificationManager::Dismiss(NotificationManager::Notification::Id id) {
 
   size_t count = NbNotifications();
   size_t foundIndex = std::distance(notifications.begin(), currentIterator);
+  size_t foundId = notifications[foundIndex].id;
+
   for (size_t i = foundIndex; i < TotalNbNotifications - 1; i++) {
     notifications[i] = notifications[i + 1];
     notifications[i].index = i;
   }
   notifications[TotalNbNotifications - 1] = {};
-
-  // fill any gaps in ids
-  for (size_t i = 0; i < TotalNbNotifications; i++) {
-    if (!notifications[i].valid)
-      break;
-
-    notifications[i].id = i;
-  }
 
   if (count == 1) {
     readIndex = 0;
@@ -111,6 +105,12 @@ void NotificationManager::Dismiss(NotificationManager::Notification::Id id) {
     writeIndex = (writeIndex > 0) ? writeIndex - 1 : TotalNbNotifications - 1;
     readIndex = (readIndex > 0) ? readIndex - 1 : TotalNbNotifications - 1;
   }
+  for (size_t i = 0; i < TotalNbNotifications - 1; i++) {
+    if(notifications[i].id > foundId)
+      notifications[i].id--;
+  }
+
+  // Ensure any new id's are incrementing from the last notification id
   nextId = notifications[readIndex].id + 1;
 }
 
--- a/src/components/ble/NotificationManager.cpp
+++ b/src/components/ble/NotificationManager.cpp
@@ -35,7 +35,7 @@ NotificationManager::Notification NotificationManager::GetNext(NotificationManag
     return n.valid && n.id == id;
   });
   if (currentIterator == notifications.end() || currentIterator->id != id)
-    return Notification {};
+    return {};
 
   auto& lastNotification = notifications[readIndex];
 
@@ -45,9 +45,9 @@ NotificationManager::Notification NotificationManager::GetNext(NotificationManag
     result = *(notifications.begin());
   else
     result = *(currentIterator + 1);
-  
+
   if (result.id <= id)
-    return Notification{};
+    return {};
 
   result.index = (lastNotification.id - result.id) + 1;
   return result;
@@ -58,7 +58,7 @@ NotificationManager::Notification NotificationManager::GetPrevious(NotificationM
     return n.valid && n.id == id;
   });
   if (currentIterator == notifications.end() || currentIterator->id != id)
-    return Notification {};
+    return {};
 
   auto& lastNotification = notifications[readIndex];
 
@@ -68,37 +68,32 @@ NotificationManager::Notification NotificationManager::GetPrevious(NotificationM
     result = *(notifications.end() - 1);
   else
     result = *(currentIterator - 1);
-  
+
   if (result.id >= id)
-    return Notification {};
+    return {};
 
   result.index = (lastNotification.id - result.id) + 1;
   return result;
 }
 
 void NotificationManager::Dismiss(NotificationManager::Notification::Id id) {
-  if(empty)
+  if (empty)
     return;
 
-  bool found = false;
-  size_t foundIndex;
+  auto currentIterator = std::find_if(notifications.begin(), notifications.end(), [id](const Notification& n) {
+    return n.valid && n.id == id;
+  });
+  if (currentIterator == notifications.end() || currentIterator->id != id)
+    return;
+
   size_t count = NbNotifications();
-  for(size_t i = 0; i < TotalNbNotifications; i++) {
-    if(notifications[i].id == id) {
-      foundIndex = i;
-      found = true;
-      break;
-    } 
-  }
-  if(!found)
-    return;
-
-  for(size_t i = foundIndex; i < TotalNbNotifications - 1; i++) {
-    notifications[i] = notifications[i+1];
+  size_t foundIndex = std::distance(notifications.begin(), currentIterator);
+  for (size_t i = foundIndex; i < TotalNbNotifications - 1; i++) {
+    notifications[i] = notifications[i + 1];
     notifications[i].index = i;
   }
 
-  if(count == 1) {
+  if (count == 1) {
     readIndex = 0;
     writeIndex = 0;
     empty = true;
@@ -138,4 +133,3 @@ const char* NotificationManager::Notification::Title() const {
   }
   return {};
 }
-
--- a/src/components/ble/NotificationManager.cpp
+++ b/src/components/ble/NotificationManager.cpp
@@ -92,15 +92,26 @@ void NotificationManager::Dismiss(NotificationManager::Notification::Id id) {
     notifications[i] = notifications[i + 1];
     notifications[i].index = i;
   }
+  notifications[TotalNbNotifications - 1] = {};
+
+  // fill any gaps in ids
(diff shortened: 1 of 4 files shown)


src/displayapp/screens/Notifications.cpp
```diff
@@ -178,25 +228,38 @@ Notifications::NotificationItem::NotificationItem(const char* title,
...
+
+  container = lv_cont_create(lv_scr_act(), nullptr);
+  lv_obj_set_pos(container, 0, 0);
+  lv_obj_set_size(container, LV_HOR_RES, LV_VER_RES);
+  lv_obj_set_style_local_bg_color(container, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_BLACK);
+  lv_obj_set_style_local_pad_all(container, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, 0);
+  lv_obj_set_style_local_pad_inner(container, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, 0);
+  lv_obj_set_style_local_border_width(container, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, 0);
+  
+  page = lv_page_create(container, nullptr);
```
Tiggilyboo, line 243: Forgot to mention: Previously we had a container just for the message information. Now there is a parent container around the whole notification (category, number, count, and message). This container is used for the animation. I've changed the message contents to be a page to allow for scrolling for longer messages. This may end up being a useability struggle, as there are other SwipeUp / SwipeDown events handled on the parent... 
See what you guys think.

Riksu9000 (changes requested):
Usually a PR should be focused on one topic or a single feature for example. Some of the changes here should be split off to at least one more PR.

Tiggilyboo:
> Usually a PR should be focused on one topic or a single feature for example. Some of the changes here should be split off to at least one more PR.

No problem, I have made this PR into the notifications dismissal and now creating the other watch face related one.

Itai-Nelken:
I found a bug:
If i send 1 notification and dismiss it, the 'no notifications' "notification" shows and the counter is 0/0 as it should be.
but if I send another notifications and dismiss it, a black screen shows with the counter at 1/2.
sending and dismissing another notification causes the counter total to be incremented.
here is a video:

https://user-images.githubusercontent.com/70802936/169375596-29e53193-8194-4d2b-8601-577c44076d1f.mp4

Tiggilyboo:
Nice catch, taking a look at this now. Thanks @Itai-Nelken

NeroBurner:
Happened for me in InfiniSim (create a notification, swipe it away, go back to watchface, swipe down to get to notifications, see total counter is still counted up)

with your fix it doesn't happen on InfiniSim anymore

src/components/ble/NotificationManager.cpp
```diff
@@ -44,9 +45,9 @@ NotificationManager::Notification NotificationManager::GetNext(NotificationManag
     result = *(notifications.begin());
   else
     result = *(currentIterator + 1);
-
+
```
NeroBurner, line 48: please remove trailing whitespaces
```suggestion

```

src/components/ble/NotificationManager.cpp
```diff
@@ -44,9 +45,9 @@ NotificationManager::Notification NotificationManager::GetNext(NotificationManag
     result = *(notifications.begin());
   else
     result = *(currentIterator + 1);
-
+  
   if (result.id <= id)
-    return {};
+    return Notification{};
```
NeroBurner, line 50: just `return {};` should be enough, I think. No need to repeat the return type here
```suggestion
    return {};
```

src/components/ble/NotificationManager.cpp
```diff
@@ -67,14 +68,45 @@ NotificationManager::Notification NotificationManager::GetPrevious(NotificationM
     result = *(notifications.end() - 1);
   else
     result = *(currentIterator - 1);
-
+
```
NeroBurner, line 71: please remove trailing whitespaces
```suggestion

```

src/components/ble/NotificationManager.cpp
```diff
@@ -67,14 +68,45 @@ NotificationManager::Notification NotificationManager::GetPrevious(NotificationM
     result = *(notifications.end() - 1);
   else
     result = *(currentIterator - 1);
-
+  
   if (result.id >= id)
-    return {};
+    return Notification {};
```
NeroBurner, line 73: like above
```suggestion
    return {};
```

src/displayapp/screens/Notifications.cpp
```diff
@@ -20,6 +20,7 @@ Notifications::Notifications(DisplayApp* app,
     motorController {motorController},
     systemTask {systemTask},
     mode {mode} {
+
```
NeroBurner, line 23: no need to add empty line here
```suggestion
```

NeroBurner (changes requested):
didn't get through all of it, but ran out of time. Submitting notes up until now

Itai-Nelken:
Mostly fixed. Now if I send 5 notifications and then dismiss them all, same thing happens.

NeroBurner:
found a dismissal problem. Steps to reproduce

- start simulator
- hit `n` three times
- dismiss the second (2/3) notification
  InfiniSim_2022-05-19_195359
- swipe down to see the oldest notification with wrong index `3/2`
  InfiniSim_2022-05-19_195408


Doesn't happen if I swipe away the first `1/3` notification. Then the remaining notifications read `1/2` and `2/2`

NeroBurner:
Found a complication, which needs to be handled: The `notifications` array is used as a ring buffer

- In the beginning `readIndex` and `writeIndex` are 0
- after the first push `readIndex` is still 0, and `writeIndex` is 1
- `writeIndex` will always be one more advanced than `readIndex`
- after the fifth insert `readIndex` will be 4 and `writeIndex` has wrapped to be `0`
- with the next push we're at `readIndex` is 0, and `writeIndex` is 1

src/components/ble/NotificationManager.cpp
```diff
@@ -67,14 +68,46 @@ NotificationManager::Notification NotificationManager::GetPrevious(NotificationM
...
+  size_t count = NbNotifications();
+  for(size_t i = 0; i < TotalNbNotifications; i++) {
+    if(notifications[i].id == id) {
+      foundIndex = i;
+      found = true;
+      break;
+    } 
+  }
+  if(!found)
+    return;
```
NeroBurner, line 94: use the available stlib functions like in `GetNext()` and `GetPrevious()` functions

```suggestion
  auto currentIterator = std::find_if(notifications.begin(), notifications.end(), [id](const Notification& n) {
    return n.valid && n.id == id;
  });
  if (currentIterator == notifications.end() || currentIterator->id != id)
    return;
  size_t foundIndex = std::distance(notifications.begin(), currentIterator);
  size_t count = NbNotifications();
```

NeroBurner (changes requested):

Tiggilyboo:
> found a dismissal problem. Steps to reproduce
> 
>     * start simulator
> 
>     * hit `n` three times
> 
>     * dismiss the second (2/3) notification
>       InfiniSim_2022-05-19_195359
> 
>     * swipe down to see the oldest notification with wrong index `3/2`
>       InfiniSim_2022-05-19_195408
> 
> 
> Doesn't happen if I swipe away the first `1/3` notification. Then the remaining notifications read `1/2` and `2/2`

I think I've fixed this now in my latest commit. Sad news, it's really not the nicest work around. As you've also pointed out, it uses a ring buffer, which means the tail and head are shifting about and indexes and the indexes that are displayed for each notification depends on an ever increasing id.

So my solution was to reassign each valid notification, and also reassign the nextId incrementation based on the highest last id (aka. the count). 
I'll have more pokes at this, but at least it seems to be behaving more predictably now.

TLDR; Dirty but it works now.

Itai-Nelken:
Works well for me.

src/components/ble/NotificationManager.cpp
```diff
@@ -68,44 +68,50 @@ NotificationManager::Notification NotificationManager::GetPrevious(NotificationM
...
+  for (size_t i = foundIndex; i < TotalNbNotifications - 1; i++) {
+    notifications[i] = notifications[i + 1];
     notifications[i].index = i;
   }
+  notifications[TotalNbNotifications - 1] = {};
+
+  // fill any gaps in ids
+  for (size_t i = 0; i < TotalNbNotifications; i++) {
+    if (!notifications[i].valid)
+      break;
```
NeroBurner: If it can happen, that the first element is not valid, then the id-reassignment will do nothing. I think we should use `continue` instead of `break` to make sure we go through all the messages, no matter where the ring-buffer is filled currently

```suggestion
    if (!notifications[i].valid)
      continue;
```

Tiggilyboo: Take a look at the latest implementation in 9fedd9bf1b3ed1109fdb269b8d2c615617de8bc9
This logic no longer exists, instead the loop is checking any id's > dismissed notification id.

Good eye though, thanks!

NeroBurner:
Found a way to break the current code:
- start simulator
- hit `n` 7 times (create seven notifications)
- swipe down to get to the notifications
  showing 1/5 notifications before swiping away
- swipe away two times
- 0/0 notifications shown
  showing 0/0 notifications

Tiggilyboo:
> Found a way to break the current code:
> 
>     * start simulator
> 
>     * hit `n` 7 times (create seven notifications)
> 
>     * swipe down to get to the notifications
>       showing 1/5 notifications before swiping away
> 
>     * swipe away two times
> 
>     * 0/0 notifications shown
>       showing 0/0 notifications

Having a go at this... This is tripping me up quite a bit!
The issue also happens for n == 6.

It seems that:
- n == 7: readIndex = 1
- dismiss 1: readIndex = 0
- dismess 2: readIndex loops back to 4.
- Then notifications[4] is an empty element, and shows it. As notifications[TotalNbNotifications - 1] = {} every dismiss. Which seems to be the problem...

In order to fix this, I need to be able to dismiss based on readIndex's position, not just from dismissed element to end of array... I think.

```
11111  // n == 7
 ^
11110  // dismiss x1
^
11100  // dismiss x2
    ^
```

Cooking something up... If anyone has any ideas, let me know!

NeroBurner:
I've created a SmallRingBuffer class to hide the ringbuffer index handling inside. And outside the index from 0 to 4 can be requested. You're welcome to test it :)

It needs some restructure and cleanups, but I think it is working

NeroBurner:
@Tiggilyboo I liked the swipe to dismiss feature so much I opened a new PR https://github.com/InfiniTimeOrg/InfiniTime/pull/1173 with the code from this PR, rebased it and added the ring-buffer rework. Is it ok for you to move to the new PR, or would you like to update this PR instead and champion the changes yourself? Either one is fine by me, I just want the notification dismissal feature :grin:

Tiggilyboo:
> @Tiggilyboo I liked the swipe to dismiss feature so much I opened a new PR https://github.com/InfiniTimeOrg/InfiniTime/pull/1173 with the code from this PR, rebased it and added the ring-buffer rework. Is it ok for you to move to the new PR, or would you like to update this PR instead and champion the changes yourself? Either one is fine by me, I just want the notification dismissal feature :grin: 

Yeah for sure get it merged, afk this weekend otherwise I would. Thanks for the ring buffer fixes!

NeroBurner:
You're very welcome! :)

Then I'm closing this PR in favor of my new one.

PR #1024, opened 2022-03-07
Infineat Watchface

dmlls:
### This PR implements the new watchface Infineat.
**For anyone interested in trying it, it is already available [here](https://github.com/dmlls/InfiniTime/releases/tag/1.10.0-1-infineat).**
> Note: the linked version differs slightly from the one in this PR, since it adds extended Latin support.

<br>

demo

<br>

This PR also implements a handy `CheckboxList` (check the watchface list in the above demo). I kept this class as general as possible, along the lines of `Tile` or `ApplicationList`, so that it can be used for other purposes as well.

This watchface could be considered "heavy" in terms of flash storage, since it adds two new fonts and an image. However, I've tried to keep things at their absolute minimum. One of the fonts only includes numbers (Bebas Neue, the main font used for the time) and the other (Teko) includes the characters [A-Za-z0-9]. The range of the letters in the Teko font could even be reduced, since they are only used for the days of the week ("Mon", "Tue", "Wed", etc).

The image is the PINE64 logo and has a size of 23x31px. I tried to make its background transparent so that it could be used in other places, but I didn't get it to render correctly. So in the end the background is white. If someone wants to try to make it transparent, I can provide the original SVG.

For a comparison, the [latest build](https://github.com/InfiniTimeOrg/InfiniTime/actions/runs/1930222342) in `develop`:

```
 Memory region         Used Size  Region Size  %age Used
           FLASH:      403028 B       480 KB     82.00%
             RAM:       54872 B        64 KB     83.73%
```

And the Infineat build:

```
Memory region         Used Size  Region Size  %age Used
           FLASH:      416464 B       480 KB     84.73%
             RAM:       54984 B        64 KB     83.90%
```

PR #1151, opened 2022-05-19
Analog Face: Add stepmeter indicators around watch face, add missing icons, restyling

Tiggilyboo:
Hey all,

I've revamped the analogue watch face and would like to discuss what people think.
Here is a screenshot:
screenshot

More detailed changes:
- Removed rendering of background image.
- Step meter uses lv_linemeter with 12 indicator lines (one for each hour). So for instance you have 10000 steps goal, and have walked 5000, 6 hour indicator lines will be highlighted. Once you go above the goal, the indicator lines will remain highlighting all 12.
- Add missing icons for bluetooth and heartrate (as in Digital watchface)
- Re-style face and hands (lovely red dot).
- Day of week and date now render on a single line (like a lot of analogue watches out there!)

Open to discussion, like the ideas? I am fond of making the face minimal (no second indicators between the hours), and simple single lines for each hand - but obviously this is just my opinion...

### DIFF
--- a/src/displayapp/screens/WatchFaceAnalog.cpp
+++ b/src/displayapp/screens/WatchFaceAnalog.cpp
@@ -5,16 +5,16 @@
 #include "displayapp/screens/BleIcon.h"
 #include "displayapp/screens/Symbols.h"
 #include "displayapp/screens/NotificationIcon.h"
+#include "components/heartrate/HeartRateController.h"
+#include "components/motion/MotionController.h"
 #include "components/settings/Settings.h"
-
-LV_IMG_DECLARE(bg_clock);
 
 using namespace Pinetime::Applications::Screens;
 
 namespace {
-constexpr int16_t HourLength = 70;
-constexpr int16_t MinuteLength = 90;
-constexpr int16_t SecondLength = 110;
+constexpr int16_t HourLength = 60;
+constexpr int16_t MinuteLength = 80;
+constexpr int16_t SecondLength = 90;
 
 // sin(90) = 1 so the value of _lv_trigo_sin(90) is the scaling factor
 const auto LV_TRIG_SCALE = _lv_trigo_sin(90);
@@ -49,22 +49,28 @@ WatchFaceAnalog::WatchFaceAnalog(Pinetime::Applications::DisplayApp* app,
                                  Controllers::Battery& batteryController,
                                  Controllers::Ble& bleController,
                                  Controllers::NotificationManager& notificationManager,
-                                 Controllers::Settings& settingsController)
+                                 Controllers::Settings& settingsController,
+                                 Controllers::HeartRateController& heartRateController,
+                                 Controllers::MotionController& motionController)
   : Screen(app),
     currentDateTime {{}},
     dateTimeController {dateTimeController},
     batteryController {batteryController},
     bleController {bleController},
     notificationManager {notificationManager},
-    settingsController {settingsController} {
+    settingsController {settingsController},
+    heartRateController {heartRateController},
+    motionController {motionController} {
 
   sHour = 99;
   sMinute = 99;
   sSecond = 99;
 
-  lv_obj_t* bg_clock_img = lv_img_create(lv_scr_act(), NULL);
-  lv_img_set_src(bg_clock_img, &bg_clock);
-  lv_obj_align(bg_clock_img, NULL, LV_ALIGN_CENTER, 0, 0);
+  stepsMeter = lv_linemeter_create(lv_scr_act(), NULL);
+  lv_linemeter_set_range(stepsMeter, 0, 12);
+  lv_linemeter_set_scale(stepsMeter, 360, 13);
+  lv_obj_set_size(stepsMeter, 250, 250);
+  lv_obj_align(stepsMeter, NULL, LV_ALIGN_CENTER, 0, 0);
 
   batteryIcon.Create(lv_scr_act());
   lv_obj_align(batteryIcon.GetObject(), nullptr, LV_ALIGN_IN_TOP_RIGHT, 0, 0);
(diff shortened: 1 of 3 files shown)


src/displayapp/screens/WatchFaceAnalog.cpp
```diff
@@ -73,55 +79,63 @@ WatchFaceAnalog::WatchFaceAnalog(Pinetime::Applications::DisplayApp* app,
...
+  lv_style_set_line_color(&second_line_style, LV_STATE_DEFAULT, LV_COLOR_RED);
+  lv_style_set_line_rounded(&second_line_style, LV_STATE_DEFAULT, true);
+  lv_obj_add_style(second_body, LV_LINE_PART_MAIN, &second_line_style);
+  
+  dot = lv_line_create(lv_scr_act(), NULL);
+  lv_style_init(&dot_style);
+  lv_style_set_line_width(&dot_style, LV_STATE_DEFAULT, 10);
+  lv_style_set_line_color(&dot_style, LV_STATE_DEFAULT, LV_COLOR_RED);
+  lv_style_set_line_rounded(&dot_style, LV_STATE_DEFAULT, true);
+  lv_obj_add_style(dot, LV_LINE_PART_MAIN, &dot_style);
```
Riksu9000, line 138: This dot can be a simple `lv_obj` object with a radius of `LV_RADIUS_CIRCLE`

src/displayapp/screens/WatchFaceAnalog.cpp
```diff
@@ -49,22 +49,28 @@ WatchFaceAnalog::WatchFaceAnalog(Pinetime::Applications::DisplayApp* app,
...
   sSecond = 99;
 
-  lv_obj_t* bg_clock_img = lv_img_create(lv_scr_act(), NULL);
-  lv_img_set_src(bg_clock_img, &bg_clock);
-  lv_obj_align(bg_clock_img, NULL, LV_ALIGN_CENTER, 0, 0);
+  stepsMeter = lv_linemeter_create(lv_scr_act(), NULL);
+  lv_linemeter_set_range(stepsMeter, 0, 12);
+  lv_linemeter_set_scale(stepsMeter, 360, 13);
+  lv_obj_set_size(stepsMeter, 250, 250);
+  lv_obj_align(stepsMeter, NULL, LV_ALIGN_CENTER, 0, 0);
```
Riksu9000, line 73: This object goes beyond the screen area, yet it doesn't fill the entire screen. Maybe there's an invisible background. The clock looks a bit small for the screen, so this should be made as large as possible.
image

Riksu9000 (changes requested):
Please use `clang-format` to format the files and address the code formatting checks.

src/displayapp/screens/WatchFaceAnalog.cpp
Tiggilyboo: Nice one, still coming in fresh to this library, seems like a better solution, thanks!

src/displayapp/screens/WatchFaceAnalog.cpp
Tiggilyboo: Started looking into this, I have set the background to red in order to see it, and indeed there is a background that is larger than the indicators / rendered lines. However, I have tried setting the style with several properties but I cannot get the background to shrink...

```
  lv_obj_set_style_local_bg_color(stepsMeter, LV_LINEMETER_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_RED);
  lv_obj_set_style_local_pad_inner(stepsMeter, LV_LINEMETER_PART_MAIN, LV_STATE_DEFAULT, 0);
  lv_obj_set_style_local_margin_left(stepsMeter, LV_LINEMETER_PART_MAIN, LV_STATE_DEFAULT, 0);
```
Will keep poking away at it, but the larger than resolution was the quick and dirty as you have pointed out.

Itai-Nelken:
I like this watchface, but it's almost a completely different watchface compared to the original analog one.
I think it should be a separate watchface. Maube SimpleAnalog as it looks simpler.

Tiggilyboo:
> I like this watchface, but it's almost a completely different watchface compared to the original analog one. I think it should be a separate watchface. Maybe it can be called SimpleAnalog as it looks simpler.

I have implemented this face as the 5th face "Simple Analog" using the same setting method as in https://github.com/InfiniTimeOrg/InfiniTime/pull/1024 .As any of the other on hold faces are merged using the same settings menu, it should be easy enough to bump the index. Though I am aware that new faces are on hold whilst we sort memory usage issues?

Cheers

yehoshuapw:
while it's not quite the same, part of the point (as I understood it) is to free up memory - so just adding a new analog face isn't much help.

Tiggilyboo:
> while it's not quite the same, part of the point (as I understood it) is to free up memory - so just adding a new analog face isn't much help. so actually replacing the old memory heavy analog face does make sense.
> 
> once there's a better method to load faces, then the old one can be re-added.

If we do go this route, we should also remove the `bg_clock.c` inclusion for the old analog face background. I'm of the same opinion (replace the analog face), but not sure who decides on this, take it to a vote somehow? Otherwise, also happy to keep the PR as it stands once new faces are admitted.

calhix:
I'm for just replacing the current analog face with this. Why don't we have an option in the settings to remove the "smart" functions in this watch face so it's more like the current one if people prefer the style of the original?

FintasticMan:
I believe that the background image should definitely be removed, as it takes up a relatively large amount of memory. That would mean not being able to have a setting that makes it look like the old style.

yehoshuapw:
> That would mean not being able to have a setting that makes it look like the old style.

I think what @calhix meant was to disable the step meter indicator part.

calhix:
> > That would mean not being able to have a setting that makes it look like the old style.
> 
> I think what @calhix meant was to disable the step meter indicator part.

This, and without the heart rate. I would prefer it never be visible on any watch face though unless you actually are checking the rate by going into the app and pushing start.

Riksu9000:
I'm not sure we can justify having two very similar watch faces taking up precious memory. I'm also strongly opposed to the idea of replacing the current watch face, it's the one I've always used.

tigercoding56:
> I'm not sure we can justify having two very similar watch faces taking up precious memory. I'm also strongly opposed to the idea of replacing the current watch face, it's the one I've always used.

bg_clock.c == ~80 kb,  which probably gets compressed but that is still large

FintasticMan:
I think we should definitely move the background image to the external resources. That will free up some internal flash storage, and then we can decide what to do with this watch face.

I was thinking that this could be a selectable option in the analogue watch face, like with PTS.

PR #1153, opened 2022-05-20
Automatically create screens for applist and settings

Riksu9000:
Apps and settings are now stored in a single array (two arrays in total). ~~Adding more entries will automatically create more screens.~~

I'm assuming there's a reason for std::function, because I don't know exactly what it does.

### DIFF
--- a/src/displayapp/screens/ApplicationList.cpp
+++ b/src/displayapp/screens/ApplicationList.cpp
@@ -1,6 +1,6 @@
 #include "displayapp/screens/ApplicationList.h"
 #include <lvgl/lvgl.h>
-#include <array>
+#include <functional>
 #include "displayapp/Apps.h"
 #include "displayapp/DisplayApp.h"
 
--- a/src/displayapp/screens/ApplicationList.cpp
+++ b/src/displayapp/screens/ApplicationList.cpp
@@ -38,9 +38,9 @@ bool ApplicationList::OnTouchEvent(Pinetime::Applications::TouchEvents event) {
 }
 
 std::unique_ptr<Screen> ApplicationList::CreateScreen(unsigned int screenNum) const {
-  std::array<Tile::Applications, 6> apps;
-  for (int i = 0; i < 6; i++) {
-    apps[i] = applications[screenNum * 6 + i];
+  std::array<Tile::Applications, appsPerScreen> apps;
+  for (int i = 0; i < appsPerScreen; i++) {
+    apps[i] = applications[screenNum * appsPerScreen + i];
   }
 
   return std::make_unique<Screens::Tile>(screenNum, nScreens, app, settingsController, batteryController, dateTimeController, apps);
--- a/src/displayapp/screens/ApplicationList.cpp
+++ b/src/displayapp/screens/ApplicationList.cpp
@@ -1,12 +1,22 @@
 #include "displayapp/screens/ApplicationList.h"
 #include <lvgl/lvgl.h>
 #include <array>
-#include "displayapp/screens/Symbols.h"
-#include "displayapp/screens/Tile.h"
 #include "displayapp/Apps.h"
 #include "displayapp/DisplayApp.h"
 
 using namespace Pinetime::Applications::Screens;
+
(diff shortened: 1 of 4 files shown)


JF002:
This is a great use for `constexpr` :+1:

JF002 (approved):

src/displayapp/screens/settings/Settings.cpp
```diff
@@ -1,33 +1,27 @@
 #include "displayapp/screens/settings/Settings.h"
 #include <lvgl/lvgl.h>
 #include <array>
-#include "displayapp/screens/List.h"
-#include "displayapp/Apps.h"
 #include "displayapp/DisplayApp.h"
-#include "displayapp/screens/Symbols.h"
+#include "displayapp/Apps.h"
```
NeroBurner, line 5: please move before `DisplayApp.h` include for alphabetical ordering

src/displayapp/screens/ApplicationList.h
```diff
@@ -21,14 +23,31 @@ namespace Pinetime {
...
         Controllers::Settings& settingsController;
         Pinetime::Controllers::Battery& batteryController;
         Controllers::DateTime& dateTimeController;
 
-        ScreenList<2> screens;
-        std::unique_ptr<Screen> CreateScreen1();
-        std::unique_ptr<Screen> CreateScreen2();
-        // std::unique_ptr<Screen> CreateScreen3();
+        static constexpr int appsPerScreen = 6;
+        static constexpr std::array<Tile::Applications, appsPerScreen * 2> applications {{
```
NeroBurner, line 34: `nScreens` is explicitly mentioned here as 2. If I want to add another app to the list I need to add the `2` to `3`, which is equivalent to `nScreens`.

We could explicilty set `nScreens` and then calculate the `applications.size()` instead

If I just add an `+1` to add the new application as the only app on the third screen then I get an error at runtime when I swipe to the third screen

src/displayapp/screens/settings/Settings.h
```diff
@@ -17,14 +19,35 @@ namespace Pinetime {
...
 
-        ScreenList<4> screens;
+        Controllers::Settings& settingsController;
 
-        std::unique_ptr<Screen> CreateScreen1();
-        std::unique_ptr<Screen> CreateScreen2();
-        std::unique_ptr<Screen> CreateScreen3();
-        std::unique_ptr<Screen> CreateScreen4();
+        static constexpr int entriesPerScreen = 4;
+        static constexpr std::array<List::Applications, entriesPerScreen * 4> entries {{
```
NeroBurner, line 28: same as above, `nScreens` is already mentioned to be 4

NeroBurner (changes requested):

src/displayapp/screens/ApplicationList.cpp
```diff
@@ -1,13 +1,23 @@
 #include "displayapp/screens/ApplicationList.h"
 #include <lvgl/lvgl.h>
 #include <array>
```
NeroBurner, line 3: `std::function` and `std::unique_ptr` are used in this file. We should include the corresponding headers [<functional>](https://en.cppreference.com/w/cpp/header/functional) and [<memory>](https://en.cppreference.com/w/cpp/header/memory)

```suggestion
#include <array>
#include <functional>
#include <memory>
```

src/displayapp/screens/settings/Settings.cpp
```diff
@@ -1,33 +1,27 @@
 #include "displayapp/screens/settings/Settings.h"
 #include <lvgl/lvgl.h>
 #include <array>
```
NeroBurner, line 3: `std::function` and `std::unique_ptr` are used in this file. We should include the corresponding headers [<functional>](https://en.cppreference.com/w/cpp/header/functional) and [<memory>](https://en.cppreference.com/w/cpp/header/memory)

```suggestion
#include <array>
#include <functional>
#include <memory>
```

NeroBurner (changes requested):

NeroBurner (approved):
LGTM

PR #1158, opened 2022-05-27
Fix assertion failure in TimerController

FederAndInk:
0 is not valid for xTimerPeriodInTicks, changing it to 1 to pass the assertion


FintasticMan:
See #860

Riksu9000 (approved):

Issue #365, opened 2021-05-18
Proposal: Allow multiple times / timezones


km6thj:
The core points of this proposal:

- Add a few time zones
- Optionally display a second time zone's time on a watch face

This is particularly useful when communicating often with people in other time zones, or when one needs
to see UTC often.

FoxAviation:
I'd be very happy if I were able to display UTC times, which is very important in some fields like aviation, maritime and, I'm sure, many others.

km6thj:
Really useful for amateur radio too.

geekbozu:
I feel like a lot of this should/could fall under 2 settings
One being a third time format
```
12-hour
24-hour
UTC
```
The second being watch-face specific information. Infinitime would need a way to get/set the time zone, but the display of it will likely be dependent on the watch face outside of the above representation options.

Avamander:
I do not want the complexity, space, time or implementation wise, that is dealing with timezones. At best I'd allow an optional setting for an offset clock.

geekbozu:
> 
> 
> I do not want the complexity, space, time or implementation wise, that is dealing with timezones. At best I'd allow an optional setting for an offset clock.
> 
> Time format is a separate setting, that's fine.

Yeah I think this is what I was poorly saying. I only want an offset from UTC stored in InfiniTime nothing more. Anything more is a waste of our effort. Updating it/maintaining it is the User/Companion apps problem. We should be a dumb smart watch that just shows the time we are given....eheheh

calhix:
Just wanted to mention I previously opened #1168 and didn't see this issue was open when doing a search since I think I only searched for world clock or different iterations of it. Mine is more about a separate app but being able to also display on a watch face is great also. I would rather swipe left to a world clock app from the watch face though.

PR #1161, opened 2022-05-28
Extend CTS Server

cybuzuma:
## Motivation

I implemented a small TOTP Screen for my watch and thus needed access to UTC. When checking on how to get this information from my phone to the watch, I found and implemented the Local Time Characteristic of the Current Time Service BLE specification. This allows the companion app to transmit information about the current timezone as well as dst in effect.
~~When checking for related PRs and Issues I found #365 and thought that I'd implement parts of it allowing to show UTC on the watch.~~

### BLE

* implemented the Local Time Characteristic in `CurrentTimeService`
* added a check to both Local Time and Current Time to ignore too short data
* corrected the definition of the `CtsCurrentTimeData` struct in `CurrentTimeClient` as well as `CurrentTimeService` to match the specification

### `DateTimeController`

* added fields for timezone and dst offsets to store the values provided by the BLE characteristic
* added getter for UTC offset
* added getter for UTC

### ~~Settings~~

* ~~added UTC option to the clock formats
~~

### ~~WatchFaces~~

* ~~changed watch faces to respect UTC setting~~
* ~~UTC is always shown as 24h format (I do not think someone should display UTC in 12h format, change my mind ;)~~
* ~~WatchFace Analog: Found no good way to signal that UTC is active, did not want to impact the design~~
* ~~WatchFace Digital:  When displaying UTC the date is shown in ISO format (I guess whoever wants his watch to show UTC is also a big fan of ISO dates :)~~

* ~~WatchFace Style: Found no good way to signal that UTC is active, did not want to impact the design~~
* ~~WatchFace Terminal: Shows trailing "Z" on the time if UTC is shown. Also, changed the date to use dashes instead of dots.~~
~~

## Tests

I tested my code on the InfiniSim as well as my actual (sealed) watch. I used a patched version of Gadgetbridge to send the added characteristic. So far, I could not observe any problems.

## Notes

I added a check for the returncode of the `os_mbuf_copydata()` function (may for-money job consists a lot of making people not ignore return values, so I am afraid I can not not do it). The question was what to do with invalid data. A) just set the received, potentially invalid, data and let the user see that there is something wrong. B) ignore the data, potentially having the watch not work at all. Seeing that the user can not really do something in any case, I chose B) as it just feels wrong to use invalid data.

I changed the 16bit year field into two 8 bit fields, assembling them manually and making the code more portable. The previous implementation relied on the machine order being little endian. While for now, this code is not very likely to end up on a big endian machine, I really think it is great and could/should end up on a different platform, where this might become an annoying bug.

I kept the separate timezone and dst fields, even if the might not be extremely useful uncombined, but maybe some wants to show if DST is active or something.

~~For WatchFaceAnalog, I had to pull in `<date/date.h>` to be able to convert from the chrono information returned by `dateTimeController.UTCDateTime()`. I could have added `UTCHour()` and other accessors to DateTimeController, but thought it was cleaner to do the conversion in the WatchFace, seeing the other WatchFaces do as well.~~

This all depends on a companion app supporting the newly implemented Local Time Characteristic. As I said before, I am using a patched version of Gadgetbridge for myself and I hope I can open a PR there soon (and then look into other companion apps). ~~The question here is, what to do if the companion app does not support it. A) Disable/Hide the UTC option in settings B) leave it, but without effect (all values are initialized to 0, making UTC and local time equivalent). I went with B as I am not a fan of the complexity added by having to check if there was a Local Time update or not, if was valid at some point, and so on. I just think that it is not worth it and a user should be able to tolerate that the setting does not chang anything. Maybe a note can be placed in some documentation (I actually found no in tree documentation on the 24h setting).~~

### DIFF
--- a/src/components/ble/CurrentTimeService.cpp
+++ b/src/components/ble/CurrentTimeService.cpp
@@ -5,11 +5,23 @@
 using namespace Pinetime::Controllers;
 
 constexpr ble_uuid16_t CurrentTimeService::ctsUuid;
-constexpr ble_uuid16_t CurrentTimeService::ctChrUuid;
+constexpr ble_uuid16_t CurrentTimeService::ctsCtChrUuid;
+constexpr ble_uuid16_t CurrentTimeService::ctsLtChrUuid;
 
 int CTSCallback(uint16_t conn_handle, uint16_t attr_handle, struct ble_gatt_access_ctxt* ctxt, void* arg) {
   auto cts = static_cast<CurrentTimeService*>(arg);
-  return cts->OnTimeAccessed(conn_handle, attr_handle, ctxt);
+
+  return cts->OnCurrentTimeServiceAccessed(conn_handle, attr_handle, ctxt);
+}
+
+int CurrentTimeService::OnCurrentTimeServiceAccessed(uint16_t conn_handle, uint16_t attr_handle, struct ble_gatt_access_ctxt* ctxt) {
+  switch (ble_uuid_u16(ctxt->chr->uuid)) {
+    case ctsCurrentTimeCharId:
+      return OnCurrentTimeAccessed(conn_handle, attr_handle, ctxt);
+    case ctsLocalTimeCharId:
+      return OnLocalTimeAccessed(conn_handle, attr_handle, ctxt);
+  }
+  return -1; // Unknown characteristic
 }
 
 void CurrentTimeService::Init() {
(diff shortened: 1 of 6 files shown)


cybuzuma:
Opened corresponding [PR](https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/2687) at Gadgetbridge

FintasticMan:
This seems like a useful feature, but I'm not sure if people are going to want to display the time in UTC. I'd happily be proven wrong though!  
I can see it useful for, as you mentioned, apps that need access to a fixed time zone, and being able to set a 'home' time zone when travelling.

cybuzuma:
As I said, my main motivation was to make UTC available to apps.
On the other hand, some persons were supportive of the watch showing UTC in #365. Also, when working in aerospace, I had times where I was "living" in UTC so I can add my personal account to that.


> and being able to set a 'home' time zone when travelling.

Short clarification here: This feature does not change anything about the time displayed (if UTC is not selected). It merely adds the information about the current timezone to the watch's internals. The time displayed will always be the one set by the companion app (which in almost all cases is the local time of the phone/PC). To be able to have the watch display a different timezone than the one of the paired device is a different topic relying more on features of the companion app than the watch itself.

FintasticMan:
> On the other hand, some persons were supportive of the watch showing UTC in #365.

Ah ok, then forget I said anything :)

I do realise that this PR doesn't let you set your own time zone, I was just mentioning that it opens the door to that potential feature in the future.

src/components/ble/CurrentTimeService.cpp
```diff
@@ -18,47 +30,92 @@ void CurrentTimeService::Init() {
...
 
   if (ctxt->op == BLE_GATT_ACCESS_OP_WRITE_CHR) {
-    CtsData result;
-    os_mbuf_copydata(ctxt->om, 0, sizeof(CtsData), &result);
+    CtsCurrentTimeData result;
+    int res = os_mbuf_copydata(ctxt->om, 0, sizeof(CtsCurrentTimeData), &result);
+
+    if (res < 0) {
+      NRF_LOG_ERROR("Error reading BLE Data writing to CTS Current Time (too little data)")
+      // return OK/0 anyway, we did consume the request, we just ignored the invalid value´
```
Avamander, line 47: I think invalid packets should return an error code to the client.

src/components/ble/CurrentTimeService.cpp
```diff
@@ -18,47 +30,92 @@ void CurrentTimeService::Init() {
...
+  NRF_LOG_INFO("Setting timezone...");
+
+  if (ctxt->op == BLE_GATT_ACCESS_OP_WRITE_CHR) {
+    CtsLocalTimeData result;
+    int res = os_mbuf_copydata(ctxt->om, 0, sizeof(CtsLocalTimeData), &result);
+
+    if (res < 0) {
+      NRF_LOG_ERROR("Error reading BLE Data writing to CTS Local Time (too little data)")
+      // return OK/0 anyway, we did consume the request, we just ignored the invalid value´
+      return 0;
```
Avamander, line 86: Same here.

src/displayapp/screens/settings/SettingTimeFormat.cpp
```diff
@@ -74,6 +76,9 @@ void SettingTimeFormat::UpdateSelected(lv_obj_t* object, lv_event_t event) {
         if (i == 1) {
           settingsController.SetClockType(Controllers::Settings::ClockType::H24);
         };
+        if (i == 2) {
+          settingsController.SetClockType(Controllers::Settings::ClockType::UTC);
+        };
```
Avamander, line 81: This would probably look nicer as an `else if` or `switch`?

Avamander (changes requested):

Avamander:
I'm ok with having the ability to get UTC or offset, as long as we do not have to deal with timezones. So this PR seems quite fine.

src/displayapp/screens/settings/SettingTimeFormat.cpp
cybuzuma: Agreed.

src/components/ble/CurrentTimeService.cpp
cybuzuma: I guess I was mixed up in different BLE implementations and thought that returning !=0 means request was not handled.
Fixed.

Avamander (approved):

cybuzuma:
[PR] (https://github.com/alexr4535/siglo/pull/108) against siglo.

cybuzuma:
[PR](https://gitea.arsenm.dev/Arsen6331/itd/pulls/16) against itd

Arsen6331:
> [PR](https://gitea.arsenm.dev/Arsen6331/itd/pulls/16) against itd

Please look at my comments on your PR. I am putting this here because my Gitea instance does not send email notifications.

cybuzuma:
Testing found a missing break.

Riksu9000:
> On the other hand, some persons were supportive of the watch showing UTC in https://github.com/InfiniTimeOrg/InfiniTime/issues/365.

Maybe a world time clock app would be a better place for this rather than a time format setting. I'm assuming people don't use UTC constantly, so I think people using it would end up going to the settings to change it to UTC, then go to the watchface to look at the time, then go back to the settings to change the format back and finally return to the watchface again. In the end it would probably be easier to just learn the time difference from local time.

cybuzuma:
> Maybe a world time clock app would be a better place for this rather than a time format setting. I'm assuming people don't use UTC constantly, so I think people using it would end up going to the settings to change it to UTC, then go to the watchface to look at the time, then go back to the settings to change the format back and finally return to the watchface again. In the end it would probably be easier to just learn the time difference from local time.

I guess you are right. As I said before, my main point is getting UTC available internally.
It is just when reading #365, I thought I could implement one of the proposals on the way, but I agree that it is not the best UX.

I propose that I remove all changes to the watchfaces and keep this PR simple: Add Local Time characteristic to have UTC available internally.

As a matter of fact, I am already eperimenting with Gadgedbridge and how to get multiple timezones integrated there as well as synced to and displayed in InfiniTime.
What is the best course of action here, should I just implement my vision and create a PR to start a discussion? Or should I present my ideas beforehand for review, and if so, in what form, a WIP PR or an issue or something else?
Thanks for your Feedback!

FintasticMan:
> I propose that I remove all changes to the watchfaces and keep this PR simple: Add Local Time characteristic to have UTC available internally.

I think that sounds good, then a future PR can add something like a time offset from UTC or multiple times from the companion app.

A WIP PR would be a good way to start a discussion around how to handle things.

cybuzuma:
as I had to touch it again for clang-format, I squashed and rebased the PR against development.

cybuzuma:
sry for the failed test, ran clang-format but did not commit the changes -.-

cybuzuma:
I'm afraid I can not work out why the build failed.
I can build remotes/pull/1161/merge (a4900bf90f1a3df887a1f287d4a90dc0a69e2554) on my local VM (no docker) with 
```
$ ./arm-none-eabi-gcc --version
arm-none-eabi-gcc (GNU Toolchain for the Arm Architecture 11.2-2022.02 (arm-11.14)) 11.2.1 20220111
```
and `nRF5_SDK_15.3.0_59ac345` without problem.
There is a [bug](https://bugs.linaro.org/show_bug.cgi?id=5825) against `arm-none-eabi-11.2-2022.02` which hints at a problem with picking up something from the local build machine which might explain why my build succeeds while the docker build does not. Still, as far as I can see, other builds did not fail.

I will rebase my patch onto the current development branch to verify that the problem is not triggered by the implicit merge in the github action...

FintasticMan:
That bug you mentioned is the problem. Currently ARM recommends just downgrading if it affects you, so I've made a pull request: #1203. At the moment it's just a toss-up whether a build will work or not on GitHub Actions.

cybuzuma:
Thanks for the info, shall I rebase anyway?

FintasticMan:
If you rebase, it'll allow the build to run again, and it might work that time. But seeing as there are no conflicts and the simulator does build I think it's clear that your code does compile, so I think it's up to you.

If you rebase, a member of the InfiniTimeOrg organisation will have to manually trigger the action.

cybuzuma:
As I was in the process of doing so anyway, I did a rebase. Thanks for your continuing patience :)

medeyko:
Is there any problem with this pull request?

I suppose it is important to have the time zone information at least internally, in order to be used in a world clock app or/and for a secondary time zone at watchfaces.

JF002 (approved):

JF002:
> Is there any problem with this pull request?

Nope :) We just needed some time to review it :-)

PR #838, opened 2021-11-21
[WIP] Add configuration files for Sphinx/ReadTheDocs

JF002:
Following our discussion in #748, I open this PR to start working on the re-working of the documentation of the project:
 - The documentation (user documentation, developers documentation,...) is stored in this repo
 - It's processed by the documentation generator [Sphinx](https://www.sphinx-doc.org/en/master/) and hosted on [ReadTheDocs](https://readthedocs.org/).
 - The URL https://doc.infinitime.io will point to the documentation on RTD.

I created a [new team *doc-writer*](https://github.com/orgs/InfiniTimeOrg/teams) for contributors that focus on writing and maintaining the documentation of the project. I also invited @ncartron to join the project as the first member of this team

PR #1163, opened 2022-05-30
Minor Documentation Fix: Update imageFile.png path in SWD.md

sethitow:
Minor fix to documentation.


Riksu9000 (approved):
That doc looks outdated. Nevertheless fixing this in the mean time is fine.

JF002 (approved):
The plan is to write a whole new doc. We started the work here : https://github.com/InfiniTimeOrg/InfiniTime/pull/838. It still need a bit of work before we can switch to the new doc, though...

PR #1167, opened 2022-06-04
remove some unused symbols, free a little bit of space

mabuch:
I found some Symbols that are no longer used (or at least I found no usage of them).


Riksu9000 (approved):

src/displayapp/screens/Symbols.h
```diff
@@ -5,13 +5,8 @@ namespace Pinetime {
     namespace Screens {
       namespace Symbols {
         static constexpr const char* none = "";
-        static constexpr const char* batteryFull = "\xEF\x89\x80";
-        static constexpr const char* batteryEmpty = "\xEF\x89\x84";
-        static constexpr const char* batteryThreeQuarter = "\xEF\x89\x81";
         static constexpr const char* batteryHalf = "\xEF\x89\x82";
```
FintasticMan, line 8: It is a shame that the batteryHalf can't be replaced with the new icon without quite a lot of changes. Its only use is in the battery settings menu.

mabuch: many icons are used only once or twice (as App-Icon or Settings-Icon), therefore I see no problem with still including this icon for the usage in settings menu. 
Changing it to the the Battery-Icon used on the watch-faces would introduce other inconsistencies, as it would be the only settings menu icon not using a font-icon.

FintasticMan: Exactly why it can't be changed without a lot of changes.

NeroBurner (approved):

PR #1170, opened 2022-06-05
remove unused Brightness App

mabuch:
It seems to me, that Brightness App is an unnecessary remnant, as the brighness is now changed via the Quicksettings-Button. I found no usage of this App and therefore propose to remove it.


Riksu9000 (approved):

NeroBurner (approved):
Thank you for doing a spring clean on the codebase :bow:

Issue #176, opened 2021-01-22
Ability to dismiss notifications


Jab2870:
I have just received my pinetime, so it is very possible that I am missing
something, however, I can't find a way to dismiss notifications.

I am running version 0.10.0

Thanks in advance

JF002:
Hi and welcome among us :)

What do you mean by "dismiss the notification" ? Removing it from the list of notifications displayed in the notification app?

Jab2870:
Thank you.


Yes, that is what I was thinking

grayrattus:
I could try to implement this.

I was thinking about swipe right movement that could trigger a small **remove icon** which pops up from the right side of the screen and then second swipe right could remove the notification. A couple of swipes to the right could empty NotificationManager.

Swipe left would close **remove icon** and the user will not be able to delete notification till he swipes right again.

What do you think?

grayrattus:
FYI I started to work on that but I noticed that that first I need to learn a bit of LVGL :D 
I will still work on this feature but first I need to learn how to LVGL properly.

Is there a simpler way to preview/install changes than using the Gagetbridge app that installs my custom stuff? Currently, I'm putting built soft on so online hosting, then download it on phone, and from that, I install it on PineTime.

Avamander:
> Is there a simpler way to preview/install changes than using the Gagetbridge app that installs my custom stuff?

Yes, use a SWD programming dongle and use that to upload the firmware.

aymswick:
Is there currently no way to dismiss a notification? I've searched through a number of these issues which seem to indicate conflicting info about whether this is implemented or not. I've tried swiping left/right/up/down, long pressing, etc and can't remove a notification.

PR #1173, opened 2022-06-05
Dismiss notifications by swiping right

NeroBurner:
Original code by @Tiggilyboo at https://github.com/InfiniTimeOrg/InfiniTime/pull/1150 rebased on current `develop` branch. I just  reworked the ringbuffer and squashed to one clean commit

---

Rework notification ring buffer to have a begin-idx and a size_ internally. And add  a new interface to access the notifications by index. The `0` index is always the latest notification.

The second new function is a `Dismiss` function to remove the notification with a specified idx from the buffer.

Fixes: https://github.com/InfiniTimeOrg/InfiniTime/issues/176


FintasticMan (changes requested):
Could you keep the original commits, so that then the original author is credited properly?

NeroBurner:
I think the original author is properly credited: commit author is still Simon Willshire, and I'm mentioned as co-author in the commit, which is also shown correctly in the GitHub overview

And I like clean commits which work by itself.

NeroBurner:
fixed code formatting errors and rebased on current develop branch

FintasticMan:
> I think the original author is properly credited: commit author is still Simon Willshire, and I'm mentioned as co-author in the commit, which is also shown correctly in the GitHub overview

Ah ok, I didn't see @Tiggilyboo's profile picture by the commit, so I thought they weren't a co-author.

Tiggilyboo:
Nice work, testing it on my watch, so far everything's stable and working as expected!

Riksu9000:
If I have notifications like this:
[3/3] oldest
[2/3] 
[1/3] newest
If I swipe on the second one, it will change to the oldest notification, but the animation direction makes it look like it is coming from below, which I think is the wrong direction.

The PineTime LCD drawing is very slow and not well suited to display movement without making use of hardware scrolling. There are only a few frames of animation using this method. It would be nice if the sideways animation that's used between QuickSettings and watchface could be used here too.

https://user-images.githubusercontent.com/37774658/173192207-9c88b4e2-4b0e-4d6a-ac0b-500455d212b3.mp4

There's also this old notification dismissal pull which is probably worth examining if that hasn't been done already #323.

NeroBurner:
fixed the direction a new notifications is moved in. Didn't touch the swipe animation as quite like the swiping animation when dismissing the notification. Would it be a show stopper to keep the little laggy animation?

Thanks to https://github.com/InfiniTimeOrg/InfiniSim/pull/37 the screen transitions can also be shown in the simulator

2022-06-16_notification_dismiss_up_and_down

Tiggilyboo:
> fixed the direction a new notifications is moved in. Didn't touch the swipe animation as quite like the swiping animation when dismissing the notification. Would it be a show stopper to keep the little laggy animation?
> 
> Thanks to https://github.com/InfiniTimeOrg/InfiniSim/pull/37 the screen transitions can also be shown in the simulator
> 
> 2022-06-16_notification_dismiss_up_and_down
> 
> 

Agree, also I quite like the animation. Is the hardware scrolling only limited to the y direction? Will take a look and see for later MR.
I had a crash a little while back but it was from one of the earlier commits, unsure of causes as of yet. Will update to this commit and try to replicate. 

Good work!

Riksu9000:
I wouldn't say I like the laggy animation, but perhaps it's passable. I'd personally like for it to be changed soon after merging though if we don't change it beforehand. There is only hardware based vertical scrolling. This is why the horizontal scrolling is done in stripes.

I recently also had a crash related to notifications, but can't remember what branch I was running. It happened when notifications were already open, I think as history and not a popup, I think the screen was also on and I received a new notification.

devnoname120:
@Riksu9000 Would enabling VSync help make the animation more seamless?

Alternatively we could cheat by displaying a small thrash pop-up on top of the current notification and use vertical scroll to go to the next notification.

devnoname120:
See https://github.com/InfiniTimeOrg/InfiniTime/issues/1186 for smooth horizontal animations.

src/components/ble/NotificationManager.cpp
```diff
@@ -1,81 +1,86 @@
...
     return {};
+  }
+  return this->At(idx + 1);
+}
 
-  result.index = (lastNotification.id - result.id) + 1;
-  return result;
+void NotificationManager::Dismiss(NotificationManager::Notification::Id idx) {
+  if (this->IsEmpty())
+    return;
```
Avamander, line 65: Missing braces

src/displayapp/screens/Notifications.cpp
```diff
@@ -145,12 +194,13 @@ bool Notifications::OnTouchEvent(Pinetime::Applications::TouchEvents event) {
       }
 
       validDisplay = true;
-      currentId = nextNotification.id;
+      if (currentId > 0)
+        --currentId;
```
Avamander, line 198: Missing braces

src/components/ble/NotificationManager.cpp
NeroBurner: thx, fixed

JF002 (changes requested):
Thanks for this PR, this feature will make a lot of users happy ;-)

There are a few points I would like to discuss before merging.

I have to admit that the sliding animation looks far better than what I would have expected knowing that the whole screen is refreshed at every frame. However, this is still not on par with other animations in the project (the vertical animation is smooth thanks to the hardware scrolling effect, and the left/right animations we use for the quick menu hides the slowness of the display). On one hand, I like this animation because it's similar to what we are used to on other touch devices, and on the other side, it's still a bit laggy. I would like to keep the UI of InfiniTime as smooth as possible, so I would appreciate if we at least tried to improve this one. 

The changes in the struct `NotificationManager::Notification`. In the original code, the struct contained a field `id`. Its purpose was to uniquely identify a notification regardless of it position in the buffer. I added this field knowing that in the future, we will want to remove a specific notification either via BLE or the UI. As notification are received asynchronously, a new notification could be received while another one is being removed, which could result in the wrong notification being deleted. The purpose of this `id` field was to ensure that the correct notification is removed.
The new implementation removed that field and also mixes the concept of `id` (identifier) and `index` (position in the buffer). I think this implementation could be reviewed to re-add the concept of `id` : each notification has an unique ID, and, when the Notification app wants to delete a notification, it passes the `id` of the notification instead of the `index` of the notification in the buffer.

*Note : when writing this comment, I noticed that NotificationManager does not seem to be correctly protected against concurrent accesses from BLE and the UI (and was not before this PR)... that's probably something to keep in mind if we encounter crashes.*

Notice that `GetNext()`, `GetPrevious()` and `GetLastNotification()` return the Notification by value and I did it on purpose : If they returned a ref to the notification, the content of notification could be changed/deleted by a new notification received by BLE while it's currently being processed in the Notification app. I think the new methods `At()` should do the same. I also find the name `At()` a bit confusing and inconsistent. Couldn't we name this method `Get()` ?

Regarding the code formatting, the variables `begin_idx` and `read_idx` and `index_` do not follow the coding conventions of the project.

Well, I hope I didn't discourage you with this  rather long review, and feel free to discuss those comments if you do not agree with me ;-)

NeroBurner:
The easiest solution for the transition would be to use the "swipe right stripey" animation (the one that is used from clock to quicksettings) to "disintegrate" the current notification and instantly show the new one. Then the checks if a new message is shown to swipe in from below or of the top of the screen could be removed as well.

Another a bit more complicated way would be to "disintegrate" onto a black new screen, which in turn then swipes to the new message (from below or top).

---

The `id` field seemed to me like a way to have the ring buffer together with the `1/5` notification message, which I didn't need anymore after the restructure. As there was more to the `id` field, I'll need to re-add it again

---

The `At()` function was structured to work like the `at()` function of `std::vector` (or other random access stl-containers). If we don't want the reference a rename to `Get()` feels good for me (because I would expect a function named similar to a stl-container to do a similar thing)

---

uh, yes sorry, I need to rename those internal variables :sweat_smile: 

---

I like code reviews. They take effort and time, but the outcome is always better code and programmers who learned something new :)

NeroBurner:
ready for re-review

- dismiss is done by Notification::Id to really dismiss the shown message instead of the one at the current viewed index
- instead of the swipe animation use the `RightAnim` screen transition for more smoothness

JF002:
Thanks @NeroBurner. The result looks quite good, IMO : https://video.codingfield.com/w/p4kXKd3Bp9co4C6yvKvjQR

But I'm still a bit troubled by the new NotificationManager : it still exposes methods with "indexes" in the public API, which is against the encapsulation goal of the original implementation. 

In the original implementation, my goal was to provide a high level API to the Notification screen:
 - Get the most recent notification
 - Get the oldest notification
 - Get the next/previous notification relative to a given notification.
Such an interface allows to completely hide the underlying implementation. When developing the Notification app, you don't need to know that the NotificationManager stores the notification in a buffer, a queue, a circular buffer, a hashmap or anything else. You don't even need to know what a Notification ID is. It's up to the notification manager to ensure that notifications are returned in the correct chronological order.

In my mind, the only change needed to add support for notification dismissal should be to add a method `Dismiss(ID)` to NotificationManager. 

There might be good reasons to re-write a lot of the internals of NotificationManager, but I can't see them. Was the original implementation not adapted to this new use-case? Was it too complex or too difficult to understand?

However, I really think we should keep the encapsulation level as it was in the original implementation.

Avamander:
> However, I really think we should keep the encapsulation level as it was in the original implementation.

Could we do that refactor in a follow-up PR?

NeroBurner:
I've got a few hours today to see how much I can rewrite the encapsulation. The reason for the rewrite was the problems with the dismiss and reordering of the notifications after a dismiss. And the dependence of the "id" to the displayed index (the notification 1/5 part on the display).

A great part of the rewrite at the `Notifications.h` part is due to keeping track of the current index of the displayed message. And then it is much more straightforward to access the same index (like 3/5 to 3/4), or one decremented when at the oldest message (like 5/5 to 4/4).

JF002:
> Could we do that refactor in a follow-up PR?

Yes, probably :) And I'm even willing to write such PR, as I have a clear view of what I would like to achieve (and probably fail at explaining it).

JF002:
> I've got a few hours today to see how much I can rewrite the encapsulation. The reason for the rewrite was the problems with the dismiss and reordering of the notifications after a dismiss. And the dependence of the "id" to the displayed index (the notification 1/5 part on the display).
> 
> A great part of the rewrite at the `Notifications.h` part is due to keeping track of the current index of the displayed message. And then it is much more straightforward to access the same index (like 3/5 to 3/4), or one decremented when at the oldest message (like 5/5 to 4/4).

Thanks for the explanation! So yes, indeed, changes in NotificationManager were indeed needed! I didn't realize that when reading the PR.

NeroBurner:
That was easier than I expected :tada: 

InfiniSim_2022-06-25_182651


I still needed to introduce two new functions in total:
- `Dismiss(id)`
- `IndexOf(id)`: return the index of the notification with the specified `id`

This way `Notification.h` doesn't need to keep track of the `currentIdx`. The index can't be used for anything else, because the index based functions are all private

src/displayapp/screens/Notifications.cpp
```diff
@@ -285,6 +364,19 @@ void Notifications::NotificationItem::OnCallButtonEvent(lv_obj_t* obj, lv_event_
...
+
+bool Notifications::AnimationElapsed() const {
+  bool elapsed = dismissAnimTickCount != 0 && xTaskGetTickCount() > dismissAnimTickCount + dismissAnimLength;
+  return elapsed;
+}
+
+void Notifications::AnimationReset() {
+  dismissAnimTickCount = 0;
+}
+
```
Riksu9000, line 379: I don't think comparing ticks are necessary, as the sideways transition is blocking. I replaced this with just a boolean for should show a new notification and it seemed to work fine.

JF002: Yep, indeed, it seems like `dismissAnimTickCount` could be a simple boolean, as the whole animation will be handled in the lv task at once.

NeroBurner: renamed variable and made it to be a boolean

src/displayapp/screens/Notifications.cpp
```diff
@@ -285,6 +364,18 @@ void Notifications::NotificationItem::OnCallButtonEvent(lv_obj_t* obj, lv_event_
...
+bool Notifications::AnimationElapsed() const {
+  return dismissingNotification;
+}
+
+void Notifications::AnimationReset() {
+  dismissingNotification = false;
+}
+
```
Riksu9000, line 378: I don't think these functions are necessary now.

Avamander (approved):

src/displayapp/screens/Notifications.cpp
NeroBurner: removed/inlined those functions

src/displayapp/screens/Notifications.cpp
```diff
@@ -108,52 +136,82 @@ bool Notifications::OnTouchEvent(Pinetime::Applications::TouchEvents event) {
...
+          currentId = previousMessage.id;
+        } else if (nextMessage.valid) {
+          currentId = nextMessage.id;
+        } else {
+          // don't update id, won't be found be refresh and try to load latest message or no message box
+        }
+        currentItem.reset(nullptr);
+        app->SetFullRefresh(DisplayApp::FullRefreshDirections::RightAnim);
+        // create black "inTransition" screen
+        currentItem = std::make_unique<NotificationItem>(alertNotificationService, motorController, true);
```
Riksu9000, line 161: I think this can be simplified by just creating a black box that covers the entire screen.

```
lv_obj_t* blackBox = lv_obj_create(lv_scr_act(), nullptr);
lv_obj_set_size(blackBox, LV_HOR_RES, LV_VER_RES);
lv_obj_set_style_local_bg_color(blackBox, LV_OBJ_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_BLACK);
```

src/displayapp/screens/Notifications.h
```diff
@@ -33,31 +33,35 @@ namespace Pinetime {
...
           NotificationItem(const char* title,
                            const char* msg,
                            uint8_t notifNr,
                            Controllers::NotificationManager::Categories,
                            uint8_t notifNb,
-                           Modes mode,
                            Pinetime::Controllers::AlertNotificationService& alertNotificationService,
-                           Pinetime::Controllers::MotorController& motorController);
+                           Pinetime::Controllers::MotorController& motorController,
+                           bool isTransition = false);
```
Riksu9000, line 46: isTransition is never used because using the simple constructor creates an empty screen regardless, but the separate constructor wouldn't be needed either if the previous suggestion is implemented.

Riksu9000 (changes requested):

src/displayapp/screens/Notifications.cpp
NeroBurner: but wouldn't the black box be created immediately without the screen transition animation?

Riksu9000: No, because the screen drawing is done by a separate LVGL task, not immediately after object creation. I did a rough test by just replacing this and the transition worked like we want it to.

NeroBurner: just tried locally in the simulator and it works as you described. Very cool! Less hacky and straightforward solution. Thanks!

Riksu9000 (approved):
This'll be a great addition to the next release. The fast animations look great.

JF002 (approved):
It looks really good to me, thanks for making the extra effort on this PR, @NeroBurner

devnoname120:
@NeroBurner Thank you for this great feature! Could you consider to make the dismiss function also work for incoming notifications? (The screen that you see when you just received a notification and which shows is a time/progress bar until it goes back to the current app.)

NeroBurner:
@devnoname120 please create a separate issue (or maybe one is already open) and ping me there

It should be possible

PR #1174, opened 2022-06-06
Use Counter widget in SettingSetTime

Riksu9000:
The Counter widget was extended with custom font support, twelve hour mode and value changed callback.

SettingSetTime was reworked to use Counter, along with other small optimizations.

InfiniSim_2022-06-06_213540
InfiniSim_2022-06-06_213542

### DIFF
--- a/src/displayapp/widgets/Counter.cpp
+++ b/src/displayapp/widgets/Counter.cpp
@@ -6,35 +6,43 @@ namespace {
   void upBtnEventHandler(lv_obj_t* obj, lv_event_t event) {
     auto* widget = static_cast<Counter*>(obj->user_data);
     if (event == LV_EVENT_SHORT_CLICKED || event == LV_EVENT_LONG_PRESSED_REPEAT) {
-      widget->Increment();
+      widget->UpBtnPressed();
     }
   }
 
   void downBtnEventHandler(lv_obj_t* obj, lv_event_t event) {
     auto* widget = static_cast<Counter*>(obj->user_data);
     if (event == LV_EVENT_SHORT_CLICKED || event == LV_EVENT_LONG_PRESSED_REPEAT) {
-      widget->Decrement();
+      widget->DownBtnPressed();
     }
   }
 }
 
-Counter::Counter(int min, int max) : min {min}, max {max} {
+Counter::Counter(int min, int max, lv_font_t& font) : min {min}, max {max}, font {font} {
 }
 
-void Counter::Increment() {
+void Counter::UpBtnPressed() {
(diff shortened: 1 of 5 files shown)


Riksu9000:
Seconds are displayed to help roughly set the seconds by timing it when the minute changes.

The only other font size is the 76px one. Using the largest font would make it look just like the Timer app, and I think it's better for them to be more distinct by showing the title for example like in all other settings screens. The best we could do is increase the padding to make the widget larger, but I'd rather keep the padding fixed or a percentage of the font size. The padding is a bit inconsistent due to the differing line heights and widths of the fonts, but I may end up doing something about that.

There are more things to consider like changing font sizes, but I think it's best to take this one step at a time and just use the Counter widget in apps, which already has larger buttons, before making more changes.

Riksu9000:
I'm thinking of replacing the twelve hour mode with an over/underflow callback. This would be a more generic solution. Let me know if you have any thoughts about this, or other ideas about how the am/pm label could be controlled by the counter.

src/displayapp/widgets/Counter.cpp
```diff
@@ -58,7 +66,32 @@ void Counter::ShowControls() {
 }
 
 void Counter::UpdateLabel() {
-  lv_label_set_text_fmt(number, "%.2i", value);
+  if (twelveHourMode) {
+    switch (value) {
```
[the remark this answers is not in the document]
Riksu9000: Not a bad idea, but it doesn't seem to save memory over the current solution.

src/displayapp/widgets/Counter.cpp
```diff
@@ -58,7 +66,32 @@ void Counter::ShowControls() {
...
+      case 0:
+        lv_label_set_text_static(number, "12");
+        break;
+      case 1 ... 12:
```
JF002, line 74: Hey I didn't know this syntax! I can't find any doc about this in my books nor cppreference. Is this a new thing?

FintasticMan: It is a GCC extension, so it isn't standard C++.

Riksu9000: It was also used in `SettingSetTime.cpp`. It seems to be a GCC extension, not part of C++. Maybe we shouldn't use it then?

JF002: It's maybe not that important as we support only 1 compiler anyway, but yeah... I guess I prefer to stick with the standard C++. Thans :)

JF002 (approved):

PR #1176, opened 2022-06-07
Change format GH Actions workflow to use git-clang-format

FintasticMan:
This change uses the built-in git script from clang-format. I've tested it in FintasticMan#3 and FintasticMan#4.


.github/workflows/format.yml
```diff
@@ -17,22 +17,22 @@ jobs:
...
 
       - name: Install clang-format
         run: |
           sudo apt-get update
           sudo apt-get -y install clang-format-12
 
       - name: Check formatting
-        run: tests/test-format.sh
+        run: |
+          git clang-format-12 --extensions cpp,h --style file "$GITHUB_BASE_REF"
```
Riksu9000, line 29: On my system this would be `git-clang-format` with the space replaced by a hyphen. Do you know if either one's standard, and if the other one's something Ubuntu adds for example?

Riksu9000 (commented):
Looks like this could be a nice improvement. How does this script work? Does it work exactly the same as the previous script? I think this may try to format the libs if the workflow is triggered by changing another file.

.github/workflows/format.yml
FintasticMan: As far as I know, as long as tge git-clang-format executable is in you PATH, it can be called vie git clang-format. See the comment at the top of this file: https://github.com/llvm/llvm-project/blob/main/clang/tools/clang-format/git-clang-format.

Riksu9000: I see. I must've typo'd earlier when I tried it.

FintasticMan:
Just updated it, it should now ignore files in src/libs and src/FreeRTOS. The tests are in the PRs I mentioned earlier.

Riksu9000:
The output of the "Check formatting" step isn't as clear. Do you think it can be improved?

Also this only produces a single patch. Maybe not a bad thing but worth considering.

FintasticMan:
I've just copied the text it outputs when there are errors from your script.

I personally don't think having everything in one patch is an issue. I do think it's a shame that GitHub has to zip it, even if it's just one file.

Riksu9000:
I think there's some potential in `git-clang-format` and also creating the patch(es) without creating a dummy commit. If you want to try and improve the `test-format.sh` script with these ideas, feel free to open a new PR.

PR #1181, opened 2022-06-13
Flashlight default to max brightness

### LINKED ISSUE #836: Torch settings

Riksu9000:
Includes tiny code tweaks.

~~Depends on #1180~~

Fixes #836

### DIFF
--- a/src/displayapp/screens/FlashLight.cpp
+++ b/src/displayapp/screens/FlashLight.cpp
@@ -5,9 +5,11 @@
 using namespace Pinetime::Applications::Screens;
 
 namespace {
-  void event_handler(lv_obj_t* obj, lv_event_t event) {
-    auto* screen = static_cast<FlashLight*>(obj->user_data);
-    screen->OnClickEvent(obj, event);
+  void EventHandler(lv_obj_t* obj, lv_event_t event) {
+    if (event == LV_EVENT_CLICKED) {
+      auto* screen = static_cast<FlashLight*>(obj->user_data);
+      screen->Toggle();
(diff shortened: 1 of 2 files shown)


NeroBurner (changes requested):

NeroBurner (approved):

Avamander (approved):

JF002 (approved):

Danimations:
Great move... looking forward to this small but significant improvement!

PR #1182, opened 2022-06-14
Fix InfiniTime compiler warnings

Riksu9000:
`const` return type doesn't make sense.
~~Weather wasn't handled in LoadApp switch. Added default.~~
Removed unused function.

This doesn't fix warnings generated by nimble and QCBOR

### DIFF
--- a/src/BootloaderVersion.cpp
+++ b/src/BootloaderVersion.cpp
@@ -9,15 +9,15 @@ using namespace Pinetime;
 uint32_t BootloaderVersion::version = 0;
 char BootloaderVersion::versionString[BootloaderVersion::VERSION_STR_LEN] = "0.0.0";
 
-const uint32_t BootloaderVersion::Major() {
+uint32_t BootloaderVersion::Major() {
   return (BootloaderVersion::version >> 16u) & 0xff;
 }
 
-const uint32_t BootloaderVersion::Minor() {
+uint32_t BootloaderVersion::Minor() {
   return (BootloaderVersion::version >> 8u) & 0xff;
 }
 
-const uint32_t BootloaderVersion::Patch() {
+uint32_t BootloaderVersion::Patch() {
   return BootloaderVersion::version & 0xff;
 }
 
@@ -25,7 +25,7 @@ const char* BootloaderVersion::VersionString() {
   return BootloaderVersion::versionString;
 }
 
-const bool BootloaderVersion::IsValid() {
+bool BootloaderVersion::IsValid() {
   return BootloaderVersion::version >= 0x00010000;
 }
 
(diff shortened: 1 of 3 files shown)


src/displayapp/DisplayApp.cpp
```diff
@@ -316,6 +316,7 @@ void DisplayApp::LoadApp(Apps app, DisplayApp::FullRefreshDirections direction)
       break;
     case Apps::None:
     case Apps::Clock:
+    default:
```
NeroBurner, line 319: I think this change may hide warnings we want to have. Unhandled screens shouldn't just silently be ignored. A warning can show us where we have forgotten to change something

NeroBurner (changes requested):
I wanted to fix those `const` warnings for a while now, but never came around to do it. Thank you for doing it!

src/displayapp/DisplayApp.cpp
Riksu9000: Hmm, that's true, however entering an app instead going to the watchface is a pretty obvious issue as well, so it's not completely "silent". Maybe I can instead remove the unused Weather app from the Apps enum?

FintasticMan: I think I remember @Avamander saying that they didn't handle the Weather app in that switch on purpose. I can't remember the reasoning behind it anymore, but maybe they could shed some light?

NeroBurner: yes, removing the enum entry would feel "more right" for me

JF002: I think @Avamander did not add the weather app because it's mostly a work in progress and/or a demo application, and not something we want the user to use.

JF002 (approved):

src/displayapp/DisplayApp.cpp
Riksu9000: Why not add it to the switch and just not add it to the app list? It's more work to enable it otherwise. Should I add it there?

NeroBurner: I'd say, yes add it to the switch statement, and later when it is ready the screen will be made reachable through the interface

Riksu9000: Looks like WeatherService isn't accessible in DisplayApp currently as required by the Weather app. I'm not sure how this app is intended to be used or what's the best approach to resolve this issue.

NeroBurner: Then I'd say, don't touch the switch statement, to stop blocking the remaining fixed warnings. We can touch that at a later time

NeroBurner (approved):

src/displayapp/DisplayApp.cpp
Avamander: Yeah, JF002 is correct, the Weather app is very much a debug utility at this moment. 

If someone touched up the UI, it could be enabled by default, but I wanted to get the already dragged on for long PR to have a fixed and reachable feature scope.

PR #1184, opened 2022-06-15
Create GH Actions workflow to build Docker image

FintasticMan:
This PR adds a GitHub Actions workflow that builds the Docker image and pushes it to Docker Hub whenever anything in the `docker/` directory is changed. I've tested it [here](https://github.com/FintasticMan/InfiniTime/actions/runs/2505633550), and the images have been pushed [here](https://hub.docker.com/r/fintasticman/infinitime/tags).

Any pull requests should just build the image, but not upload it to Docker Hub. It should also be possible to publish to different tags than just latest by using docker/metadata-action, I just haven't gotten around to that yet.

Before merging, the secret `DOCKER_HUB_ACCESS_TOKEN` will be needed to be added to this repository.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -20,7 +20,7 @@ jobs:
         if: github.event_name != 'pull_request'
         uses: docker/login-action@v2
         with:
-          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          username: infinitime
           password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
 
       - name: Set up QEMU
@@ -37,6 +37,6 @@ jobs:
           platforms: linux/amd64,linux/arm64
           builder: ${{ steps.buildx.outputs.name }}
           push: ${{ github.event_name != 'pull_request' }}
-          tags: ${{ secrets.DOCKER_HUB_USERNAME }}/infinitime-build:latest
-          cache-from: type=registry,ref=${{ secrets.DOCKER_HUB_USERNAME }}/infinitime-build:buildcache
-          cache-to: type=registry,ref=${{ secrets.DOCKER_HUB_USERNAME }}/infinitime-build:buildcache,mode=max
+          tags: infinitime/infinitime-build:latest
+          cache-from: type=registry,ref=infinitime/infinitime-build:buildcache
+          cache-to: type=registry,ref=infinitime/infinitime-build:buildcache,mode=max
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,42 @@
+name: Build and push Docker image
+
+on:
+  push:
+    branches: [ master, develop ]
+    paths:
+      - 'docker/**'
+  pull_request:
+    branches: [ master, develop ]
+    paths:
+      - 'docker/**'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Log in to Docker Hub
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -0,0 +1,42 @@
...
+        uses: docker/setup-buildx-action@v2
+
+      - name: Build and push
+        uses: docker/build-push-action@v3
+        with:
+          context: ./docker/
+          file: ./docker/Dockerfile
+          platforms: linux/amd64,linux/arm64
+          builder: ${{ steps.buildx.outputs.name }}
+          push: ${{ github.event_name != 'pull_request' }}
```
Riksu9000, line 39: Pushing commits will trigger this workflow and this will be true. I feel like pushing the docker should be a separate workflow that only triggers when adding commits to develop

Riksu9000 (dismissed):

.github/workflows/docker.yml
Riksu9000: I think I was mistaken about the event_name, however I think this should instead explicitly require a push event rather than anything other than a pull_request, which relies on this workflow to run at intended times only.

JF002:
@FintasticMan That looks very good :+1:  And it even builds for amd64 **and** arm64 ! :medal_sports: 

Regarding the tags : is there anyway to handle them 'automatically'? If you have a look at the [docker repo for InfiniTime](https://hub.docker.com/r/infinitime/infinitime-build/tags), I pushed 2 tags when I updated the image : the tag `latest` and the tag `v2`. This will allow to pull an older version of the image to build an older version of the project. I'm not sure if this is how we are supposed to handle the versionning of the images, though...

FintasticMan:
You can definitely push to multiple tags at once, but I'm not sure of a good versioning system for it. Perhaps it should push to its commit hash and latest.

Another thing I was thinking of doing was pushing the builds to the GitHub Container Repository on pull requests. It would then be put under the username of whoever made the pull request.

JF002:
@FintasticMan Thanks for your PR #1023! Perfect timing to merge this one so we can test the automatic build :)

> You can definitely push to multiple tags at once, but I'm also not sure of a good versioning system for it. Perhaps it should push to its commit hash and latest.

For now, I'm just using an incrementing counter, but I'm doing it manually. Not sure how to automate that using the Action build. 

The idea was to keep older versions of the image on the Docker repo along with the new one (latest). Is this something we want/need? Or can I create this "version" tag manually after it has been pushed as "latest"?

postlund:
I have done similar things with just a single Dockerfile, where I use the SHA256 checksum of that file as a tag. That makes it easy to detect if anything a has changed. It can quite easily (via a simple python script for instance) be extended to calculate a checksum over multiple files as well.

FintasticMan:
@JF002 I have now pushed a change that makes it push to latest and the commit shasum. You can see how it looks [here](https://hub.docker.com/r/fintasticman/infinitime-build/tags).

JF002:
@FintasticMan that looks good to me! We'll use the commit sha as tags and see if we ever need older tags of the images!

JF002 (approved):

FintasticMan:
It needs to be a repository secret.

JF002:
@FintasticMan What will trigger the build and push of  a new image? If I understand correctly, the workflow will be triggered on push **and** pull-requests on both `master` and `develop`? Won't that generate a lot of unnecessary builds
 - When you create the pull-request that will be merge on develop
 - When the branch is merged in develop
 - When develop is merged in master 

I think we should try to generate (and publish) the image only when changes are merged in develop. What do you think?

JF002:
@FintasticMan So, I've just merged this branch and #1203. I forgot the setup the username in the settings of the project, so it took 1 or 2 trials.
Now, it worked ... except that the image was pushed to the wrong username. Let me explain : the image must be pushed to infinitime/infinitime-build. However, the account *infinitime* is an organization account, and I cannot create tokens from this account. That's why I created the token from my own user *jf002*. 
Now, I think the workflow must be modified to take that into account : login with *jf002* but push to *infinitime*. Could you have a look at this?

FintasticMan:
Thanks for the heads up. I'll have a look when I get back to my PC.

JF002:
@FintasticMan Awesome, thank you very much! I wasn't sure which username needed to be changed! I'll have a look at that PR!

PR #1190, opened 2022-06-20
Tweaked counter buttons and added icons to timer buttons

0x0000ff:
My reasoning behind this is to make the buttons on the counter easier to tap. I have had to tap the screen several times in the current release to change the minutes or seconds on the timer. I have also added icons to the buttons on both the timer and counter on a separate commit.
InfiniSim_2022-06-20_154041


Riksu9000:
The buttons are too small in the current release indeed, but the buttons on the Counter widget are already much larger, so this is not a good basis to make further changes to the widget. The reset button is a separate feature and should be a separate PR.

0x0000ff:
Should I put the icon changes in a separate branch as well?

Riksu9000:
If you think it's a necessary change to make, then yes, however I'm not confident everyone will agree with it.

Issue #343, opened 2021-05-13
"Find My Phone" functionality with Gadgetbridge


Raupinger:
I used that feature quite a bit when i was daily driving my Miband. Not sure about what the best place to put it would be though. It doesn't justify being a full watchapp, but the quick settings are full.

JF002:
Interesting feature, indeed! I think it would be nice to work together with the Gadgetbridge team to find the best BLE API for this functionality.
Then, this feature could also be added in other companion apps like Amazfish and Siglo :)

cincodenada:
This is one of the features I miss most after moving to a FitBit a while back, so I'd love to see this too! As for interface, it would kinda make sense long-pressing the vibration option on the quick settings screen, but that probably shouldn't be the only way to do it at least, since it's not very discoverable. And I'm not sure if long-pressing quick settings is something that's desirable to introduce in the first place.

Raupinger:
I don't think we can even detect long presses on individual items

MysteriousLog6:
> I used that feature quite a bit when i was daily driving my Miband. Not sure about what the best place to put it would be though. It doesn't justify being a full watchapp, but the quick settings are full.

As a Mi Band User , I cant exaggerate how MUCH I use this , kind of fun too it disturbs everyone!

huckvrni:
I also switched to infinitime from the mi band and enjoyed this feature, if I find the time, perhaps I will look into it from the gadgetbridge side...

yusufmte:
On 1.9.0 this is working flawlessly for me and appears as a call from GadgetBridge. Anyone still having hiccups with this feature? Or should the issue be closed?

![IMG_20220514_140048426~2](https://user-images.githubusercontent.com/45940010/168443441-b50da2c2-c5e3-462c-859f-16df0c7bb5ee.jpg)

huckvrni:
You got it wrong, what you're showing is 'find my watch', in some other smartbands there is an option to make the phone start ringing and gadgetbridge has support for this.

yusufmte:
My bad, misread the title-- you are right. Thank you for the correction.

WhyNotHugo:
There's actually a Bluetooth profile for this: https://www.bluetooth.org/docman/handlers/downloaddoc.ashx?doc_id=239389

I think it would be wise to implement the standard, so rather than require phones to have specific Infinitime support for this, they just need to support the standard.

Also relevant: https://www.bluetooth.com/specifications/specs/find-me-profile-1-0/

[...]

PR #1193, opened 2022-06-22
Add find phone feature

devnoname120:
Close https://github.com/InfiniTimeOrg/InfiniTime/issues/343.

- Add new `infinitime` service to make it much easier and noob-friendly to add new features both on GadgetBridge's side and InfiniTime's side. This is the approach used by many devices in GadgetBridge (e.g. [WASP-OS](https://codeberg.org/Freeyourgadget/Gadgetbridge/src/commit/0349b2530fd29c02ad4d7fa85c4144b8cf7c1457/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/waspos/WaspOSDeviceSupport.java#L145-L195), [Bangle.js](https://codeberg.org/Freeyourgadget/Gadgetbridge/src/commit/0349b2530fd29c02ad4d7fa85c4144b8cf7c1457/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/banglejs/BangleJSDeviceSupport.java#L414-L605), etc etc.)
- Change flashlight app button to trigger the find phone feature.
- Forked GadgetBridge to add support for this new feature. See [Freeyourgadget/Gadgetbridge#2737](https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/2737)

https://user-images.githubusercontent.com/2824100/175046440-97b1e8ad-67b9-4cf7-a338-0f54e1b6c2d0.mp4

**TODO**:
- Make the brightness/dnd/etc quick toggles vertically scrollable and add a toggle for finding the phone.
- Rely on `Pinetime::Controllers::InfinitimeService::isPhoneFinding()` to show the find phone toggle status in InfiniTime.
- Implement `Immediate Alert Service 1.0 revision V10r00` and remove the added service
  - Investigate possible blockers/pitfalls for implementing
  - Set-up an implementation plan (get UIDs, failure modes, etc.)
  - Implement in GadgetBridge
  - Implement in InfiniTime
  - Test feature on my dev PineTime
- Document this feature (README + wiki + in-code doc)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -448,6 +448,7 @@ list(APPEND SOURCE_FILES
         components/brightness/BrightnessController.cpp
         components/motion/MotionController.cpp
         components/ble/NimbleController.cpp
+        components/ble/InfinitimeService.cpp
         components/ble/DeviceInformationService.cpp
         components/ble/CurrentTimeClient.cpp
         components/ble/AlertNotificationClient.cpp
@@ -517,6 +518,7 @@ list(APPEND RECOVERY_SOURCE_FILES
         components/ble/NimbleController.cpp
         components/ble/DeviceInformationService.cpp
         components/ble/CurrentTimeClient.cpp
+        components/ble/InfinitimeService.cpp
         components/ble/AlertNotificationClient.cpp
         components/ble/DfuService.cpp
         components/ble/CurrentTimeService.cpp
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -448,13 +448,13 @@ list(APPEND SOURCE_FILES
         components/brightness/BrightnessController.cpp
         components/motion/MotionController.cpp
         components/ble/NimbleController.cpp
-        components/ble/InfinitimeService.cpp
         components/ble/DeviceInformationService.cpp
         components/ble/CurrentTimeClient.cpp
         components/ble/AlertNotificationClient.cpp
         components/ble/DfuService.cpp
         components/ble/CurrentTimeService.cpp
         components/ble/AlertNotificationService.cpp
+        components/ble/InfinitimeService.cpp
         components/ble/MusicService.cpp
         components/ble/weather/WeatherService.cpp
         components/ble/NavigationService.cpp
@@ -518,11 +518,11 @@ list(APPEND RECOVERY_SOURCE_FILES
         components/ble/NimbleController.cpp
         components/ble/DeviceInformationService.cpp
         components/ble/CurrentTimeClient.cpp
-        components/ble/InfinitimeService.cpp
         components/ble/AlertNotificationClient.cpp
         components/ble/DfuService.cpp
         components/ble/CurrentTimeService.cpp
         components/ble/AlertNotificationService.cpp
+        components/ble/InfinitimeService.cpp
         components/ble/MusicService.cpp
         components/ble/weather/WeatherService.cpp
         components/ble/BatteryInformationService.cpp
@@ -643,6 +643,7 @@ set(INCLUDE_FILES
         components/ble/ImmediateAlertService.h
         components/ble/ServiceDiscovery.h
         components/ble/BleClient.h
+        components/ble/InfinitimeService.h
         components/ble/HeartRateService.h
         components/ble/MotionService.h
         components/ble/weather/WeatherService.h
--- a/src/components/ble/InfinitimeService.cpp
+++ b/src/components/ble/InfinitimeService.cpp
@@ -1,0 +1,100 @@
+/*  Copyright (C) 2020-2022 JF, Adam Pigg, Avamander, devnoname120
+
+    This file is part of InfiniTime.
+
+    InfiniTime is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published
+    by the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    InfiniTime is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+#include "components/ble/InfinitimeService.h"
+#include "systemtask/SystemTask.h"
+#include <cstring>
+
+namespace {
+  // 0001yyxx-b5e2-40c4-b1d5-2c5c48529c84
+  constexpr ble_uuid128_t CharUuid(uint8_t x, uint8_t y) {
+    return ble_uuid128_t {.u = {.type = BLE_UUID_TYPE_128},
+                          .value = {0x84, 0x9c, 0x52, 0x48, 0x5c, 0x2c, 0xd5, 0xb1, 0xc4, 0x40, 0xe2, 0xb5, x, y, 0x01, 0x00}};
+  }
+
+  // 00010000-b5e2-40c4-b1d5-2c5c48529c84
+  constexpr ble_uuid128_t BaseUuid() {
+    return CharUuid(0x00, 0x00);
+  }
+
+  constexpr ble_uuid128_t infUuid {BaseUuid()};
+  constexpr ble_uuid128_t infFindPhoneCharUuid {CharUuid(0x01, 0x00)};
+
+  constexpr size_t MaxNotifSize {1024};
+
+  int InfinitimeCallback(uint16_t conn_handle, uint16_t attr_handle, struct ble_gatt_access_ctxt* ctxt, void* arg) {
+    return static_cast<Pinetime::Controllers::InfinitimeService*>(arg)->OnCommand(conn_handle, attr_handle, ctxt);
+  }
+}
+
+Pinetime::Controllers::InfinitimeService::InfinitimeService(Pinetime::System::SystemTask& system) : m_system(system) {
+  characteristicDefinition[0] = {.uuid = &infFindPhoneCharUuid.u,
+                                 .access_cb = InfinitimeCallback,
+                                 .arg = this,
+                                 .flags = BLE_GATT_CHR_F_NOTIFY | BLE_GATT_CHR_F_WRITE | BLE_GATT_CHR_F_READ,
+                                 .val_handle = &phoneFindingHandle};
+  characteristicDefinition[1] = {0};
+
+  serviceDefinition[0] = {.type = BLE_GATT_SVC_TYPE_PRIMARY, .uuid = &infUuid.u, .characteristics = characteristicDefinition};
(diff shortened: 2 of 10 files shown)


yehoshuapw:
I don't know anything about either of the implementations, morever its not _always_ true,
but "now we'll have A, and later we'll fix it and put B instead" tends to leave in A forever.
(bad solutions stick, even when they are inferior and temporary) 

in same cases this considerations means that its better to nor have some feature for longer - and do it better later.
(and in some cases its not. I have no idea which of those it is, just keep in mind in this discussion this issue)

Avamander:
> Are there realistic upsides to implementing it? 

It's standard, which is one upside more than the alternative.

> Are we sure enough that Operating Systems won't interfere with that Bluetooth service?

Yes.

> If we want in the future to implement this specification and supersede the current implementation

We don't want to supersede anything. BLE API changes are very cumbersome.

devnoname120:
> Yes.

Do you have any evidence of that? I'd appreciate if you could provide arguments & evidence for your claims.

Avamander:
> OS'es might mess up with that because just like reserved ports, reserved services can make sense from a security perspective.

Not really because it's not the OS itself creating the connection and server, usually. The exception might be BLE Audio or HID devices, even then they have no reason to create the service itself.

This is a far-fetched and unlikely concern.

JF002:
I agree with @Avamander and @yehoshuapw regarding the standard specification. Since the beginning of the development, I've always made sure to implement standard protocol if such a protocol exist before implementing a custom one. That's what allowed multiple fitness apps to support the heart rate sensor of the PineTime out of the box, without adding any code in these apps, for example.

Most of the devices supported by Gadgetbridge are probably commercial devices designed by companies that do not want their devices to be compatible with other devices/ecosystem so they implement their own custom protocol that will be supported only by their devices and apps.

We do not have such constraints. Let's make sure InfiniTime is easily integrated by other apps and devices! And let's encourage everyone to implement standard protocols!

devnoname120:
@JF002 Alright thank you for your sensible arguments. I'm not very convinced that it's worth it, but I'll likely give it a shot.

Avamander:
> The Immediate Alert Service (IAS) doesn't provide or suggest any category to make the device ring for a while and override DND. Due to this reason, I don't believe that implementing this service will help at all to make it work anywhere else.

If a companion device in the future decides that `High Alert` is not sufficient to ignore DND (or equivalent) then that's not InfiniTime's problem. Furthermore, it doesn't matter for currently existing companions and their future implementations.

devnoname120:
> There's zero need to deviate from the ten-page spec and the three specified alert levels.

Yeah sure so something sends a high priority alert for something completely unrelated and the whole phone lights up, ignores DND, and vibrates. Or it shows a small notification if the companion app / computer OS doesn't specifically implement this non-standard behavior. This is entirely useless.

I think I've given enough energy just to confirm Brandolini's law.

Avamander:
> Yeah sure so something sends a high priority alert for a completely unrelated notification and the whole phone lights up, ignores DND, and vibrates.

Such scenarios depend on multiple unlikely what-if's and make assumptions about future implementations that do not yet exist. It's not a reasonable concern of ours.

>  this non-standard behavior.

That behaviour is not specified by the standard, thus it is not non-standard.

I'm stopping this foot-dragging over what boils down to an UUID.

devnoname120:
@Avamander And ultimately with this dogmatic and irrational behavior you are baselessly depriving users from a feature that they request a lot.

Avamander:
> Non-standard = not specified/defined in a standard.

There's a difference between not standardised and non-standard.

> Just like the few past times you make a claim without providing any evidence.

Provide evidence that there's no implementation-specific quirks? There aren't any implementations.

> As it is, there is simply no good reason not to rely on the current implementation.

As it is there's simply no good reason why to not conform. Bringing up hypothetical scenarios that *might* some day happen, but probably won't because there's no indication that it might, is not helpful.

Ultimately the attempt to desperately justify an approach you chose to fulfil without proper discussion is true the obstacle here.

devnoname120:
Could you address these points?

**You**:

> It's standard, which is one upside more than the alternative.

> That behaviour is not specified by the standard

**Me**:

> If we want in the future to implement this specification and supersede the current implementation, we can do that just fine. Imho since it works great, it makes sense to merge it as is once ready and in the future it could be decommissioned without any issues.

> If it sticks around then it means that it's a good enough solution that it wasn't worth rewriting it. It's not a bad solution if it works & doesn't have major shortcomings. It's a pragmatic solution, the opposite is a dogmatic solution.

> The Immediate Alert Service (IAS) [doesn't provide or suggest any category](https://github.com/oesmith/gatt-xml/blob/4fd2ede1d3da9365fdc6dec89290c346581a03f9/org.bluetooth.characteristic.new_alert.xml#L56-L65) to make the device ring for a while and override DND. Due to this reason, I don't believe that implementing this service will help at all to make it work anywhere else

> We could decide our own UID but since it's a number in the range 0-255 we risk conflicting with another custom category

> This would require all devices and companion apps to specifically support this new category we introduce so I think it's useless to bother doing that.

> You said that we should piggyback on this standard but you didn't provide any reasons to do so

> there is simply no good reason not to rely on the current implementation

-------

> Provide evidence that there's no implementation-specific quirks? There aren't any implementations.

You're assuming that `High Alert` will be implemented as `Ringing the device & ignore DND` by default in implementations. This is baseless and there is literally no reason why implementations would behave like that (please find even a single counter-example). You need evidence that this approach does something else than just display a notification, which is what we would expect given what this service is usually used for. Conversely, there is absolutely no reasonable reason to believe that the default behavior of implementations.

-------

All of this despite the fact that literally none of the proprietary and non-proprietary devices in GadgetBridge use the service you mention for that purpose. They could just be all wrong like you expect, but then you should bring strong evidence that refutes this implicit consensus.

Riksu9000:
>You:
>    It's standard, which is one upside more than the alternative.
>then saying the opposite:
>    That behaviour is not specified by the standard

There's no contradiction.

In the spec:

> If the written alert level is “High Alert," the device shall alert in the strongest possible way.

We don't know if this will be implemented as `Ringing the device`, but I'm sure it will be enough to find the phone and ignore DND.

Avamander:
> then saying the opposite

IA is a standard, but what the device considers "strongest possible way" is not. I'm sorry if that was previously unclear.

> This is baseless and there is literally no reason why implementations would behave like that

Why not. Please find even a single counter-example. Better actually, let's please drop these Reddit-tier debate tactics.

> literally none of the proprietary and non-proprietary devices in GadgetBridge use the service you mention for that purpose

Proprietary devices do a lot of things in their own proprietary way, I don't think those are very relevant. From the non-proprietary ones, it's us and Bangle.js? Doesn't make a very overwhelming majority to set how things should be done.

devnoname120:
> > If the written alert level is “High Alert," the device shall alert in the strongest possible way.
> 
> We don't know if this will be implemented as `Ringing the device`, but I'm sure it will be enough to find the phone and ignore DND.

Do we have evidence for this though?

From my perspective we have an implementation that is tested and working. 

Why would we switch to another method without evidence that it will actually work like we hope it will?

I don't understand how rewriting the implementation to use the notification service would bring any upsides considering this. If there was a category `Find Phone` or `Locate Device` I would be all onboard to implement it.

But as it stands, e.g
 a pill reminder app could perfectly reasonably send an alert with this same exact category. `High Alert` is very vague and can encompass many uses and implementations.

devnoname120:
And ultimately you didn't address the fact that we have a working implementation that is a net improvement over not having this feature at all.

If you want to reimplement this feature later on, no extra effort is needed compared to if no implementation exists.

devnoname120:
Does the standard that the alert needs to be explicitly acknowledged on the phone? If not, what makes you think it wouldn't just do a single beep instead of a sustained 30-second ringing?

Avamander:
> Because as you mentioned the standard doesn't specify it should behave that way.

Goes both ways, if someone in the future does integrate it into their OS. But again, not really relevant for us.

> Why would we switch to another method without evidence that it will actually work like we hope it will?
> If not, what makes you think it wouldn't just do a single beep instead of a sustained 30-second ringing?

Because we're implementing it on companions for now. As said previously, OS's aren't providing or interfering with the service for now.

devnoname120:
I'm curious to learn what @JF002 thinks.

I think we're running in circles right now.

JF002:
> This is not a good use of my time compared to fixing other bugs and implementing other requested features.

I've just spent more than half an hour trying to write a nice and polite answer to this, trying to explain that I don't have the time to study the protocol from this PR against the other ones suggested by @Avamander and @Riksu9000, trying to understand what was the exact issue here and trying to suggest that we should at least test the suggested options before deciding on the best one...
 
*Is this a good use of my time?*

... Until I search on [Wikipedia for Brandolini's law](https://en.wikipedia.org/wiki/Brandolini%27s_law) : 

> Brandolini's law, also known as the bullshit asymmetry principle, is an internet adage that emphasizes the effort of debunking misinformation, in comparison to the relative ease of creating it in the first place. It states that "The amount of energy needed to refute **bullshit** is an order of magnitude larger than is needed to produce it.

> I'm curious to learn what @JF002 thinks.

I certainly don't think that @Avamander and @Riksu9000 opinions and ideas are bullshit. 

So uh... I think you should try to consider their opinions seriously and at least try to provide enough proofs and facts that yours is better instead of wasting my time with counter productive and unpleasant comments.

devnoname120:
> I think you should try to consider their opinions seriously and at least try to provide enough proofs and facts that yours is better

I did several times and but my comments weren't addressed. It became really annoying hence why I started having unpleasant comments. Nevertheless I will ask GadgetBridge if implementing this service wouldn't cause issues on their end. And switch the implementation in the future.

Given the unclear benefits of using a category that is not meant for a ‶find my phone″ feature, I will assign the lowest priority 
 to this though.

As far as I'm concerned this feature works great and as much as I want to give back to the community, I don't believe that the request is reasonable in light of the expanded argumented reasons I gave and which weren't addressed despite my asking several times.

devnoname120:
@Avamander

Given that the `Immediate Alert Service` specification [forbids reading the alerting status](https://www.bluetooth.org/docman/handlers/downloaddoc.ashx?doc_id=239390#page=7):

1) How would you go to have a visual indicator on the watch that shows if the phone is currently ringing or not when opening the new `find my phone` app? This is also required to set the correct state of the toggle button in this app.

Given that the `Immediate Alert Service` specification [forbids GATT notifications](https://www.bluetooth.org/docman/handlers/downloaddoc.ashx?doc_id=239390#page=7):

2) How would you go to have the phone notify the watch [once the phone has been found](https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/2737#issuecomment-517661) so as to update the state of a toggle button in a new `find my phone` app?

devnoname120:
@Avamander Any updates? 🙏

mark9064:
I don't think Avamander is working on InfiniTime at the moment, so I'm not sure you'll get a response

devnoname120:
@mark9064 Ah ok. Thanks for the heads-up.

PR #1194, opened 2022-06-22
Added reset button to timer

0x0000ff:
InfiniSim_2022-06-22_143415
Pressing the button will reset the time counters to 0


Riksu9000:
What do you think about #1214?

PR #1197, opened 2022-06-23
Counter icons

0x0000ff:
InfiniSim_2022-06-23_141708


0x0000ff:
Mislabeled the first commit. It didn't change the icons itself.

Riksu9000:
What's the motive for changing this?

0x0000ff:
It was part of the initial changes I made to the timer. I think it makes the screen a bit tidier and the buttons look like the standard button for moving graphical objects.
numberbox-spinbutton-inline

Riksu9000:
I don't think there's a standard. I'll close this as I don't think there's enough justification to change it and I think arrows may make it seem like it's draggable, which it's not.

Kuvakaappaus - 2022-07-22 07-39-08

PR #1198, opened 2022-06-23
Don't build the firmware when only docs were modified

Riksu9000:
No reason to run these checks in that case.

### DIFF
--- a/.github/workflows/lv_sim.yml
+++ b/.github/workflows/lv_sim.yml
@@ -5,8 +5,14 @@ name: Build PineTime LVGL Simulator
 on:
   push:
     branches: [ master, develop ]
+    paths-ignore:
+      - 'doc/**'
+      - 'images/**'
   pull_request:
     branches: [ master, develop ]
+    paths-ignore:
+      - 'doc/**'
+      - 'images/**'
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


FintasticMan (changes requested):
I think that the `.devcontainer`, `.idea`, `.vscode` and `hooks` directories can be excluded as well

JF002 (approved):

Riksu9000:
> I think that the `.devcontainer`, `.idea`, `.vscode` and `hooks` directories can be excluded as well

I've been trying to think of another solution than explicitly mention multiple directories in multiple places, but haven't figured out a solution yet. I think it's best to just ignore the few directories that definitely don't have to do with code for now.

FintasticMan:
> I've been trying to think of another solution than explicitly mention multiple directories in multiple places, but haven't figured out a solution yet.

All I can think of is still running the workflow, but checking if there is a file called something like `.nobuild` in the directory of the changes or any of its parents, and if so just saying that the test passed.

NeroBurner (approved):

pfeerick:
You can also ignore by file extension... i.e. 
```
    paths-ignore:
      - '**.md'
 ```
 ignores changes to markdown files regardless of where in the repo ;)

PR #1200, opened 2022-06-25
P8: Mirror display for P8b variant

StarGate01:

A new P8b hardware variant requires the display to be mirrored, and the colors to be inverted. Patch provided by @izzeho .

Related: https://github.com/StarGate01/p8b-infinitime/issues/3

### DIFF
--- a/src/drivers/St7789.cpp
+++ b/src/drivers/St7789.cpp
@@ -21,7 +21,10 @@ void St7789::Init() {
   MemoryDataAccessControl();
   ColumnAddressSet();
   RowAddressSet();
+// P8B Mirrored version does not need display inversion.
+#ifndef DRIVER_DISPLAY_MIRROR
   DisplayInversionOn();
+#endif
   NormalModeOn();
   SetVdv();
   DisplayOn();
@@ -62,7 +65,18 @@ void St7789::ColMod() {
 
 void St7789::MemoryDataAccessControl() {
   WriteCommand(static_cast<uint8_t>(Commands::MemoryDataAccessControl));
+#ifdef DRIVER_DISPLAY_MIRROR
+  // [7] = MY = Page Address Order, 0 = Top to bottom, 1 = Bottom to top
+  // [6] = MX = Column Address Order, 0 = Left to right, 1 = Right to left
+  // [5] = MV = Page/Column Order, 0 = Normal mode, 1 = Reverse mode
+  // [4] = ML = Line Address Order, 0 = LCD refresh from top to bottom, 1 = Bottom to top
+  // [3] = RGB = RGB/BGR Order, 0 = RGB, 1 = BGR
+  // [2] = MH = Display Data Latch Order, 0 = LCD refresh from left to right, 1 = Right to left
+  // [0 .. 1] = Unused
+  WriteData(0b01000000);
+#else
   WriteData(0x00);
+#endif
 }
 
 void St7789::ColumnAddressSet() {
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -804,6 +804,7 @@ elseif(TARGET_DEVICE STREQUAL "MOY-UNK") # P8b mirrored
   add_definitions(-DCLOCK_CONFIG_LF_SRC=0) # RC
   add_definitions(-DMYNEWT_VAL_BLE_LL_SCA=500)
   add_definitions(-DCLOCK_CONFIG_LF_CAL_ENABLED=1)
+  add_definitions(-DDRIVER_DISPLAY_MIRROR)
 else()
   message(FATAL_ERROR "Invalid TARGET_DEVICE")
 endif()


src/drivers/St7789.cpp
```diff
@@ -62,7 +65,12 @@ void St7789::ColMod() {
 
 void St7789::MemoryDataAccessControl() {
   WriteCommand(static_cast<uint8_t>(Commands::MemoryDataAccessControl));
+// P8B Mirrored version sets MX bit high in MADCTL register to flip about vertical axis.
+#ifdef DRIVER_DISPLAY_MIRROR
+  WriteData(0x40); // MY(0) MX(1) MV(0) ML(0) RGB(0) 0 0 0
+#else
```
Riksu9000, line 71: This is better represented in a binary format. I'd prefer the comments to be in a similar format as in Cst816s.cpp
```suggestion
// P8B Mirrored version sets MX bit high in MADCTL register to flip about vertical axis.
#ifdef DRIVER_DISPLAY_MIRROR
  // [6] = Mirror columns(?)
  // etc.
  WriteData(0b01000000);
#else
```

StarGate01: Okay, I adjusted the code.

src/drivers/St7789.cpp
```diff
@@ -62,7 +65,19 @@ void St7789::ColMod() {
...
+#ifdef DRIVER_DISPLAY_MIRROR
+  // [0 .. 1] = Unused
+  // [2] = MH = Display Data Latch Order, 0 = LCD refresh from left to right, 1 = Right to left
+  // [3] = RGB = RGB/BGR Order, 0 = RGB, 1 = BGR
+  // [4] = ML = Line Address Order, 0 = LCD refresh from top to bottom, 1 = Bottom to top
+  // [5] = MV = Page/Column Order, 0 = Normal mode, 1 = Reverse mode
+  // [6] = MX = Column Address Order, 0 = Left to right, 1 = Right to left
+  // [7] = MY = Page Address Order, 0 = Top to bottom, 1 = Bottom to top
+  WriteData(0b01000000);
+#else
```
Riksu9000, line 78: I think the order should be MSB first. Also the top comment may not be necessary, as this information can be deciphered from `DRIVER_DISPLAY_MIRROR` and the written data.

StarGate01: Okay, I adjusted the comments.

Riksu9000 (approved):

JF002 (commented):
This change looks good but... can we try to avoid `#ifdef`s? I always try to use them as a last resort as I think they make the code more difficult to read and to maintain. Also, as the number of setting increases, the number of `add_definitions` in CMake will also increase, making it difficult to ensure we'll set consistent values.

Another way to achieve this could be to extend the driver so that it support both modes (mirror and not mirror). The mode would be passed as a parameter of the constructor, and read from a constant (`constexpr`) when calling the ctor(). `constexpr` will allow the compiler to optimize (remove) the code that is not used.

Those constants could be listed in header files (1 for each supported variant) and included in main.cpp (the correct.h file will be included using CMake magic or `#ifdef`). 

I think this method will allow to more easily support and maintain multiple variants of the PineTime and the P8. What do you think about this?

StarGate01:
@JF002 I like the idea of reducing `#ifdefs`. I used them because they require minimal changes to the code and classes, but I agree that they kind of break the OOP model.

How would you go about writing these variant-specific header files? A file (e.g. `src/variants.h`), which essentially does a similar selection as the CMake config (https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/src/CMakeLists.txt#L789), i.e. check `TARGET_DEVICE`, and then define the needed constexpressions, similar to the Pins in `drivers/PinMap.h`?

This could also be used for control of the touch driver behaviour (https://github.com/InfiniTimeOrg/InfiniTime/pull/1130#discussion_r910661435), where @Riksu9000 was rightfully hesitant on littering the code with `#ifdefs` as well.

The acceleration driver (https://github.com/InfiniTimeOrg/InfiniTime/pull/1132) is already abstracted and split into multiple classes, so a simple flag can be used to slect which implementation to use. I believe link-time-optimization will then remove the unneded code.

JF002:
@StarGate01 Sorry for the late answer!
I have already investigated some ways to support multiple hardware using a more OOP approach, but I haven't finished those researches yet. The idea was to find a way to correctly encapsulate device drivers so that it's easy to add alternative implementations and support multiple hardware (simulator, other smartwatches like the P8, experiments with the PineDio STACK,...). 

I'll try to remember what I've already done (I mean... find the git stash with the code :D) and so we can move forward on that topic!

StarGate01:
@JF002 regarding encapsulation, what do you think of the approach of #1132 ? There I split the implementations of the device-specific drivers into two specifications of an abstract class. That would be quite a bit of overkill for this PR imho.

I will prepare a `variants.h` approach as mentioned above, then we can discuss that ?

Riksu9000:
I think altering a single line with #ifdefs is simple and acceptable. In the case of functions, two or more functions can be defined and #ifdefs only choose which one gets called, which is what I would do in #1130 for example.

JF002:
Yes, you are absolutely right @Riksu9000, those changes are pretty limited and harmless for the global architecture, and I think we can merge them.

I'm pretty sure we can find a better design than this to support multiple hardware variants, but I unfortunately don't have the time to work on that right now.

So, yeah, let's not block time any further. We'll still be able to improve the design in the future :)

JF002 (approved):

PR #1204, opened 2022-06-27
SystemInfo: Display target build variant

StarGate01:

This PR uses the last line on the second screen in the system information to display the target build variant, i.e. which device the firmware was compiled for. This is very useful to know which updates to download, once you have forgotten which variant your device is.

signal-2022-06-27-083124

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -785,6 +785,7 @@ add_definitions(-D__HEAP_SIZE=4096)
 
 # Target hardware configuration options
 add_definitions(-DTARGET_DEVICE_${TARGET_DEVICE})
+add_definitions(-DTARGET_DEVICE_NAME="${TARGET_DEVICE}")
 if(TARGET_DEVICE STREQUAL "PINETIME")
   add_definitions(-DDRIVER_PINMAP_PINETIME)
   add_definitions(-DCLOCK_CONFIG_LF_SRC=1) # XTAL
--- a/src/displayapp/screens/SystemInfo.cpp
+++ b/src/displayapp/screens/SystemInfo.cpp
@@ -136,6 +136,10 @@ std::unique_ptr<Screen> SystemInfo::CreateScreen2() {
   uptimeSeconds = uptimeSeconds % secondsInAMinute;
   // TODO handle more than 100 days of uptime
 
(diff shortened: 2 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -785,6 +785,7 @@ add_definitions(-D__HEAP_SIZE=4096)
 
 # Target hardware configuration options
 add_definitions(-DTARGET_DEVICE_${TARGET_DEVICE})
+add_definitions(-DTARGET_DEVICE_NAME="${TARGET_DEVICE}")
 if(TARGET_DEVICE STREQUAL "PINETIME")
```
Riksu9000, line 789: Non blocking. I'm wondering if the model should not be all caps if it will be visible, so `PineTime`

Riksu9000 (approved):

src/CMakeLists.txt
StarGate01: Fair point. Since the name is derived from the build configuration, there would have to be some kind of lookup / conditional in the CMake config. That would not be a problem, we could even prefix the `MOY` variants with e.g. `P8`. 

On the other hand, this info is purely for informational purposes (it is on page two of the system information ..), so aesthetic requirements are somewhat negotiable imho.

I am neutral on this, what do you think?

Riksu9000: I guess it's best practices to keep the definition all caps(, or all lowercase?). I don't think it's worth adding more code to get around this.

StarGate01: I agree that consistency is desirable. CMake enforces all-caps in https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/CMakeLists.txt#L58= , so we might as well keep it that way.

ajack2001my:
Curious to know what name shows up on the PineTime... Just PineTime, PineTime1 (or PineTime2, since later variants of the PineTime had some hardware changes, etc.)

StarGate01:
@ajack2001my since the string used is just the CMake target name, it will be "PINETIME". Other hardware differences will be listed in e.g. the accelerometer variant.

NeroBurner (approved):

JF002 (approved):

PR #1028, opened 2022-03-10
Add settings overview for digital watchface

Sykomaniac:
### What does it do?
Create a settings overlay for the digital watchface to add colourful battery icon and battery percentage display. Implementation is similar to PineTime to avoid adding entries into main settings pages. Closes #582 

### Screenshots
PXL_20220310_225147961 MP
PXL_20220310_225156661
PXL_20220310_225207912
PXL_20220310_225213511

PR #1205, opened 2022-06-28
feature: add battery percentage display

lpcvoid:
This commit adds the option to display
battery percentage in the top of the
screen, next to the battery icon.

The toggle for this option is in the
battery info screen.

Option

Active

**Please be aware that I only have a sealed watch. It works fine on the simulator, but please test this on real hardware**

### DIFF
--- a/src/components/settings/Settings.h
+++ b/src/components/settings/Settings.h
@@ -209,10 +209,18 @@ namespace Pinetime {
         return bleRadioEnabled;
       };
 
+      void SetShowBatteryPercentage(bool enabled) {
+        settings.showBatteryPercentage = enabled;
+      }
+
+      bool GetShowBatteryPercentage() {
+        return settings.showBatteryPercentage;
+      }
+
     private:
       Pinetime::Controllers::FS& fs;
 
-      static constexpr uint32_t settingsVersion = 0x0003;
+      static constexpr uint32_t settingsVersion = 0x0004;
       struct SettingsData {
         uint32_t version = settingsVersion;
         uint32_t stepsGoal = 10000;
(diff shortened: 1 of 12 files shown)


lpcvoid:
I wanted to copy android's behavior here - and I think keeping the battery is more indicative of what the number represents to new users.

yehoshuapw:
> I wanted to copy android's behavior here - and I think keeping the battery is more indicative of what the number represents to new users.

indeed, buy in android there can be lots of things on that line, where here there are much less - so it's a question of clearness vs bloatness. (and none of the options is clear cut as right)
how about the percentage, with a static battery icon under? (or just the top and bottom parts above and below)

lpcvoid:
I know what you envision - but I believe that the battery with a number inside would be pretty hard to read on the small display. Also, I believe that the font I am using is the smallest one generated currently.

lpcvoid:
Hi folks, I'd love to have this feature upstream at some point, is anybody up for discussion?

Riksu9000:
There have been other implementations of this as well, such as #1028, which I slightly prefer rather than adding a global option. However I'm considering adding a percentage display to the QuickSettings screen rather than clutter the watchface.

mashuptwice:
If I could add my 2 cents here:

Adding the percentage display to the quick settings would in my opinion not be a huge benefit to the current behavior, as the percentage could as well be viewed via the "about" screen or holding the button. The benefit of adding it directly to the watchface would be to view it with minimal user interaction. 

Also I don't see the necessity to add such options as quick-settings, as a user who prefers to have the percentage displayed will usually set this once and keep the settings indefinitely.

Riksu9000:
By QuickSettings, I mean the screen to the left of the watchface. The watchface options is where you would customize the watchface. The issue I'm facing with adding a percentage display is that they always add a lot of clutter.

PR #1207, opened 2022-06-29
Keep updating motion during sleep when Bluetooth is on

apilat:
Since connected Bluetooth devices can request motion data at any point, we should keep updating it when we are connected to one. Currently stale data will be delivered when the screen is turned off.
Some wake up settings already cause motion data to be updated with the screen off so this change should not have a significant impact on battery life.


StarGate01:
This is also enabled in https://github.com/InfiniTimeOrg/InfiniTime/pull/1133 .

Keep in mind that continuously updating the motion data consumes quite a bit of power, refer to https://github.com/StarGate01/p8b-infinitime/tree/master/poweranalysis .

apilat:
Continuously polling the accelerometer during sleep is also enabled by some features in order to detect wakeup so I assume that the power used for this is reasonably small or at least acceptable, but if not I'm willing to make this behavior optional.
If I'm reading the code correctly, [this if](https://github.com/apilat/InfiniTime/blob/d4a865af3ac9410bb75b84387094991dda0cf0b4/src/components/ble/MotionService.cpp#L94) will prevent broadcasting unless the companion specifically requested notifications, in which case I think using some power for this is reasonable (otherwise we would be providing incorrect data).

Riksu9000:
In general we should make the behaviour the most efficient and require the least input from the user. In this case the best solution to me seems to be updating motion only when it is needed, so when a motion based wakeup method is enabled or the service is subscribed. That way the user can save battery by disabling wakeup methods and the motion data is still accessible when needed.

mark9064:
Superseded by #2328

PR #1212, opened 2022-06-30
Limit backlight brightness when flashlight is off

### LINKED ISSUE #1210: Flashlight: only change brightness when turned on
I like that the default brightness of the flashlight is now (1.10) high, but I think it would be even better, if the brightness would only be changed, when the light is actually turned on. It is a little irritating, when the still black screen suddenly is very bright.

Riksu9000:
Backlight brightness is unlimited while the flashlight is on.
While it is off, the brightness is limited to the global backlight level.
Small code optimizations.

Fixes #1210

### DIFF
--- a/src/displayapp/screens/FlashLight.cpp
+++ b/src/displayapp/screens/FlashLight.cpp
@@ -18,17 +18,17 @@ FlashLight::FlashLight(Pinetime::Applications::DisplayApp* app,
                        Controllers::BrightnessController& brightnessController)
   : Screen(app), systemTask {systemTask}, brightnessController {brightnessController} {
 
-  brightnessController.Set(brightnessLevel);
+  brightnessController.Set(Controllers::BrightnessController::Levels::Low);
 
   flashLight = lv_label_create(lv_scr_act(), nullptr);
   lv_obj_set_style_local_text_font(flashLight, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, &lv_font_sys_48);
   lv_label_set_text_static(flashLight, Symbols::highlight);
   lv_obj_align(flashLight, nullptr, LV_ALIGN_CENTER, 0, 0);
 
-  for (auto& i : indicators) {
-    i = lv_obj_create(lv_scr_act(), nullptr);
-    lv_obj_set_size(i, 15, 10);
-    lv_obj_set_style_local_border_width(i, LV_OBJ_PART_MAIN, LV_STATE_DEFAULT, 2);
+  for (auto& indicator : indicators) {
+    indicator = lv_obj_create(lv_scr_act(), nullptr);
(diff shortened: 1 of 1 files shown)


JF002 (approved):

src/displayapp/screens/FlashLight.cpp
```diff
@@ -56,23 +57,20 @@ FlashLight::~FlashLight() {
   systemTask.PushMessage(Pinetime::System::Messages::EnableSleeping);
 }
 
+void FlashLight::SetBrightness() {
+  brightnessController.Set(isOn ? brightnessLevel : std::min(brightnessLevel, maxBrightnessLevel));
```
NeroBurner, line 61: suggestion: instead of limiting the off-brightness, why even change it? Let's keep it constant and only change the brightness when the flashlight is on?

Riksu9000: This is to minimize the light leakage when the flashlight is off and using a low brightness setting. I wonder if it could be always low in the off position. Is this what you mean?

NeroBurner: not quite, I propose to use the brightnesslevel from the system when the flashlight is turned off (instead of using the setting as upper bound). And only use the flashlight-set-brightnesslevel when the flashlight is on. This would reduce code complexity even more.

With the current upper bound approach the brightness of the off-flashlight may change, but not always. I think this could be confusing (even writing about it to explain the behaviour is rather hard and need much explanation)

Riksu9000: I don't think it's a good idea to just use the system backlight level because of the light leakage. Using medium when off and low when on for example doesn't make much sense. Keeping it always low would lower the visibility in daylight, but I guess you wouldn't use a flashlight then anyway.

Riksu9000: I don't know why it would be confusing. The brightness is optimal in each situation, it only changes inside the flashlight app and returns to normal when exiting, however now I'm leaning towards making it always low. Shall I make the change here?

minacode: Always low when off is the most reasonable, given that it is used in dark surroundings. It would then be weird, when one accidentally opened the flashlight in bright daylight, but this should be ok. 
I like the always low approach. 👍

NeroBurner: Ilike the always low when off solution as well

mashuptwice:
Is there any reason why the brightness is set to low instead of keeping the global brightness value?

Riksu9000:
@mashuptwice Please read the discussion here https://github.com/InfiniTimeOrg/InfiniTime/pull/1212#discussion_r913262970. The issue was the backlight switching to bright on starting the app and that has been fixed.

PR #1218, opened 2022-07-01
Add PageIndicator widget to reduce code duplication

Riksu9000:
Simply moved the code to a new class.

### DIFF
--- a/src/displayapp/widgets/PageIndicator.cpp
+++ b/src/displayapp/widgets/PageIndicator.cpp
@@ -2,7 +2,7 @@
 
 using namespace Pinetime::Applications::Widgets;
 
-PageIndicator::PageIndicator(int nCurrentScreen, int nScreens) : nCurrentScreen {nCurrentScreen}, nScreens {nScreens} {
+PageIndicator::PageIndicator(uint8_t nCurrentScreen, uint8_t nScreens) : nCurrentScreen {nCurrentScreen}, nScreens {nScreens} {
 }
 
 void PageIndicator::Create() {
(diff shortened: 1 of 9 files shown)


src/displayapp/widgets/PageIndicator.h
```diff
@@ -0,0 +1,23 @@
...
+namespace Pinetime {
+  namespace Applications {
+    namespace Widgets {
+      class PageIndicator {
+      public:
+        PageIndicator(int nCurrentScreen, int nScreens);
+        void Create();
+
+      private:
+        int nCurrentScreen;
```
JF002, line 13: Maybe use `uint8_t` instead of 32 bits integers ? That's not a lot, but we would spare a few bytes in RAM :)

JF002 (approved):

src/displayapp/widgets/PageIndicator.cpp
```diff
@@ -0,0 +1,31 @@
+#include "displayapp/widgets/PageIndicator.h"
+
+using namespace Pinetime::Applications::Widgets;
+
+PageIndicator::PageIndicator(uint8_t nCurrentScreen, uint8_t nScreens) : nCurrentScreen {nCurrentScreen}, nScreens {nScreens} {
+}
+
+void PageIndicator::Create() {
```
NeroBurner, line 8: is the `Create()` function needed? Is there a reason not to move it to the constructor?

Riksu9000: It's to make it possible to draw the widget above other objects. If the widget is created during construction, it will always be below everything else created in the screen.

NeroBurner (approved):

PR #1219, opened 2022-07-01
Bluetooth interface to control heart rate measurement

apilat:
Adds a bluetooth characteristic (`00050001-78fc-48fe-8e23-433b3a1942d0`) to read the state of the heart rate sensor and turn it on/off. This will be useful for companion apps that want to take periodic heart rate measurements.
It might be worth requiring that the watch is paired before being able to use this, but I'm not experienced enough with Bluetooth to do that.

I'm not sure what `HeartRateController::States::NoTouch` represents, but I've assumed the sensor is not running at this point.


lman0:
interesting , with which app did you have tested it @apilat ?

lman0:
interesting , with which app did you have tested it @apilat ?
i would like to test it since  it would allow heart rate without the need to make it on all the time , if i understand right.

PR #1220, opened 2022-07-02
Twos code cleanup

Riksu9000:
Made the interface slightly larger.
Removed magic row and col counts. The size can be increased to 5x5 for example just by incrementing a value.
Fixed many warnings.
Reduced for loop count. Each either save a few bytes or no change.
Reduced library dependency.

In `updateGridDisplay()` only a for loop was removed and the indentation changed. The diff looks more complicated.

InfiniSim_2022-07-02_063814

### DIFF
--- a/src/displayapp/screens/Twos.cpp
+++ b/src/displayapp/screens/Twos.cpp
@@ -122,7 +122,8 @@ bool Twos::tryMerge(int newRow, int newCol, int oldRow, int oldCol) {
   if (grid[newRow][newCol].value == grid[oldRow][oldCol].value) {
     if ((newCol != oldCol) || (newRow != oldRow)) {
       if (!grid[newRow][newCol].merged) {
-        score += grid[newRow][newCol].value *= 2;
+        grid[newRow][newCol].value *= 2;
+        score += grid[newRow][newCol].value;
         lv_label_set_text_fmt(scoreText, "Score #FFFF00 %i#", score);
         grid[oldRow][oldCol].value = 0;
         grid[newRow][newCol].merged = true;
--- a/src/displayapp/screens/Twos.cpp
+++ b/src/displayapp/screens/Twos.cpp
@@ -1,10 +1,7 @@
 #include "displayapp/screens/Twos.h"
-#include <array>
 #include <cstdio>
 #include <cstdlib>
 #include <lvgl/lvgl.h>
-#include <utility>
-#include <vector>
 
 using namespace Pinetime::Applications::Screens;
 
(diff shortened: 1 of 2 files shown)


Avamander (approved):

FintasticMan (commented):
Wouldn't it be better to use std::array instead of raw arrays? It should be possible to just replace the raw arrays with std::array because nCols, nRows and nCells are constexpr.

JF002:
I had the same thought, but I'm not sure it'll help to make the code more readable. For example `TwosTile grid[nRows][nCols]; ` would become `std::array<std::array<TwosTile, nRow>, nCol>`. Plus we are using this array as a simple table, we don't really need methods from std containers.

src/displayapp/screens/Twos.h
```diff
@@ -26,11 +26,14 @@ namespace Pinetime {
 
         lv_obj_t* scoreText;
         lv_obj_t* gridDisplay;
-        TwosTile grid[4][4];
+        static constexpr int nCols = 4;
+        static constexpr int nRows = 4;
+        static constexpr int nCells = nCols * nRows;
```
FintasticMan, line 31: It seems to me that a lot of the variables here are full (unsigned) ints, but they could just be (u)int8_ts.

Riksu9000: I've noticed that sometimes using smaller integers results in a larger binary size. These specifically don't make a difference, but if `tryMerge()` and `tryMove()` functions used uint8_t, the binary size would increase. This is why I'm not always sure which is best to use.

FintasticMan: Hmm interesting... Maybe with smaller integers the compiler isn't able to do as many optimisations? Anyway, if it increases the binary size, I don't see much benifit to using smaller integers.

src/displayapp/screens/Twos.cpp
```diff
@@ -151,28 +144,30 @@ bool Twos::tryMove(TwosTile grid[][4], int newRow, int newCol, int oldRow, int o
 
 bool Twos::OnTouchEvent(Pinetime::Applications::TouchEvents event) {
   bool validMove = false;
-  for (int row = 0; row < 4; row++) {
-    for (int col = 0; col < 4; col++) {
-      grid[row][col].merged = false; // reinitialize merge state
-    }
+  for (unsigned int i = 0; i < nCells; i++) {
+    const unsigned int row = i / nCols;
+    const unsigned int col = i % nCols;
```
NeroBurner, line 149: why do a single loop and a possibly costly modulo operation instead of a nested loop like done further down?

src/displayapp/screens/Twos.cpp
```diff
@@ -97,39 +92,37 @@ Twos::~Twos() {
...
 
-bool Twos::tryMerge(TwosTile grid[][4], int& newRow, int& newCol, int oldRow, int oldCol) {
+bool Twos::tryMerge(int newRow, int newCol, int oldRow, int oldCol) {
   if (grid[newRow][newCol].value == grid[oldRow][oldCol].value) {
     if ((newCol != oldCol) || (newRow != oldRow)) {
       if (!grid[newRow][newCol].merged) {
-        unsigned int newVal = grid[oldRow][oldCol].value *= 2;
-        grid[newRow][newCol].value = newVal;
-        score += newVal;
+        score += grid[newRow][newCol].value *= 2;
```
NeroBurner, line 125: I think this statement is a bit too optimized, can we split it up again?

```suggestion
        grid[newRow][newCol].value *= 2;
        score += grid[newRow][newCol].value;
```

src/displayapp/screens/Twos.cpp
```diff
@@ -97,39 +92,37 @@ Twos::~Twos() {
...
-        availableCells.push_back(std::make_pair(row, col));
-      }
+  unsigned int emptyCells[nCells];
+  unsigned int nEmpty = 0;
+  for (unsigned int i = 0; i < nCells; i++) {
+    const unsigned int row = i / nCols;
+    const unsigned int col = i % nCols;
+    if (grid[row][col].value == 0) {
+      emptyCells[nEmpty] = i;
+      nEmpty++;
```
NeroBurner, line 102: suggest using nested loop and less costly multiplications instead of costly division and modulo operations
```suggestion
  for (unsigned int row = 0; row < nRows; row++) {
    for (unsigned int col = 0; col < nCols; col++) {
      if (grid[row][col].value == 0) {
        emptyCells[nEmpty] = row * nCols + cols;
        nEmpty++;
      }
```

src/displayapp/screens/Twos.cpp
```diff
@@ -97,39 +92,37 @@ Twos::~Twos() {
 }
 
 bool Twos::placeNewTile() {
-  std::vector<std::pair<int, int>> availableCells;
-  for (int row = 0; row < 4; row++) {
-    for (int col = 0; col < 4; col++) {
-      if (!grid[row][col].value) {
-        availableCells.push_back(std::make_pair(row, col));
-      }
+  unsigned int emptyCells[nCells];
```
NeroBurner, line 95: using `std::array` feels more like modern C++ than C-raw-arrays. Any reason why it was removed? It should be basically syntactic sugar

NeroBurner (changes requested):

src/displayapp/screens/Twos.cpp
Riksu9000: Division and modulo get optimized to bitshift and `AND`. The current solution is a few bytes smaller than the suggestion.

FintasticMan: In this case, the modulo operations shouldn't be costly because nCols is a power of 2. Modulo operations with powers of 2 should be optimised by the compiler to an AND operation with (in this case) 3.

src/displayapp/screens/Twos.cpp
Riksu9000: It was using a vector and more complicated than necessary.

FintasticMan: See JF's reply above.

src/displayapp/screens/Twos.cpp
Riksu9000: It saves a few bytes of memory even when using a 5x5 grid.

src/displayapp/screens/Twos.cpp
Riksu9000: Actually it simply treats the array as one dimensional and eliminates the variables.

src/displayapp/screens/Twos.cpp
NeroBurner: for 2D data I understand JF's reply. But for 1D data like it is here I would say `std::array<unsigned int, nCells> emptyCells;` is just as readable

```suggestion
  std::array<unsigned int, nCells> emptyCells;
```

Riksu9000: It doesn't really make a difference, but since std::array doesn't have any benefits in this case, I'm not sure it's worth changing.

NeroBurner (approved):

PR #1222, opened 2022-07-04
Overhauled timeout setting screen

Riksu9000:
7s and 10s options added. Large buttons make the timeout easy to set.

InfiniSim_2022-07-04_064205

### DIFF
--- a/src/displayapp/screens/settings/SettingDisplay.cpp
+++ b/src/displayapp/screens/settings/SettingDisplay.cpp
@@ -2,35 +2,30 @@
 #include <lvgl/lvgl.h>
 #include "displayapp/DisplayApp.h"
 #include "displayapp/Messages.h"
-#include "displayapp/screens/Styles.h"
 #include "displayapp/screens/Screen.h"
 #include "displayapp/screens/Symbols.h"
 
 using namespace Pinetime::Applications::Screens;
 
 namespace {
-  void event_handler(lv_obj_t* obj, lv_event_t event) {
+  void IncreaseBtnHandler(lv_obj_t* obj, lv_event_t event) {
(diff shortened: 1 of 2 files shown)


JF002 (approved):

jakethelizard99:
I love this I believe I had an idea similar to this but forgot about it and never made a github post.  I have one suggestion could it be made to do every number from 5-30 I feel like that would add a bit more adjustability (I know how to code yet sorry if this is a stupid question). I've tested it out for about 20 mins now and it seems to be working great and as intended.

RageGamerBoi:
> I don't think single second precision matters in this case and it would make adjusting it much slower.

What about 5 second increments instead?

Riksu9000:
I believe these values are better. With a linear increment, the first step would be very large (2x), and the last step would be an insignificant difference. Here each step is roughly 1.5x the last step.

Riksu9000:
I should mention that I think the previous radio button menu is actually better, but we can't reasonably fit all opitions on screen at the same time and scrolling a full screen doesn't seem appealing. If hardware scrolling support was improved and this screen could be scrolled in a natural way, the radio buttons could return. I'll leave this open for a bit longer in case there are other ideas.

lman0:
maybe , @Riksu9000  , you could to half/half:

radio buton for 5 to max -1 radio button
and the end a button that send to this screen

mashuptwice:
I could imagine using two columns instead. The necessary horizontal space could be compensated by abbreviating seconds. That way there would be enough space for 8 options, with the ability to switch between them with a single click


inf
k

lvlgl:
Watch face should also have backlight timeout setting separate from the other screens

PR #1227, opened 2022-07-08
Docker workflow use separate username for login

FintasticMan:
@JF002 This pull request fixes the issue mentioned [here](https://github.com/InfiniTimeOrg/InfiniTime/pull/1184#issuecomment-1179127636) about that the username used for logging in to Docker Hub is different than the one that actually has the images. This works around it by having 2 secrets: `DOCKER_HUB_LOGIN_USERNAME`, which should be set to (in this case) `jf002`, and `DOCKER_HUB_IMAGE_USERNAME`, which should be set to `infinitime`.

I don't have an organisation on Docker Hub to test it out on, but I have tested it on my personal repository, with both secrets set to `fintasticman`, and it works fine.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -13,8 +13,6 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    env:
-      USERNAME: infinitime
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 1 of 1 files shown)


JF002 (approved):

JF002:
So, yesterday, I modified the workflow file so I can run the workflow manually.
Then, I ran it manually. It did the build but did not publish the image on docker hub (https://github.com/InfiniTimeOrg/InfiniTime/runs/7263114629?check_suite_focus=true). I guess I need to add a condition in those `if` statement. I'll try to have a look at it asap :-)

FintasticMan:
I think it would be best to change it from `github.event_name == 'push'` to `github.event_name != 'pull_request'`. This would mean that it would push to Docker Hub on everything other than pull_request, which is just pushes and manual triggers.

PR #1229, opened 2022-07-10
Fix Docker workflow not pushing on manual trigger

FintasticMan:
@JF002 This pull request fixes the issue mentioned [here](https://github.com/InfiniTimeOrg/InfiniTime/pull/1227#issuecomment-1179694202).

Also fix issue where the Docker workflow fails on pull requests when the user who made the pull request doesn't have the secrets set.  
Also allow other workflows to be triggered manually.  
Also don't trigger any workflows on pull requests to master, as there shouldn't be any pull requests against master.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -15,11 +15,13 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    env:
+      USERNAME: infinitime
     steps:
       - uses: actions/checkout@v3
 
       - name: Log in to Docker Hub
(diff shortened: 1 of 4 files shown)


JF002:
Yep, it seems to work fine : 
 - build : https://github.com/InfiniTimeOrg/InfiniTime/runs/7406727124?check_suite_focus=true
 - The new image is available on Docker Hub : https://hub.docker.com/repository/docker/infinitime/infinitime-build

Thank you very much, @FintasticMan !

Issue #883, opened 2021-12-20
Step to conversion


ghost:
### Verification

- [X] I searched for similar feature request and found none was relevant.

### Pitch us your idea!

A feature that converts steps to km/miles/calories

### Description

Whether it is by tapping on the step monitor or a seperate app that has conversions for km/miles/calories

ItzSwirlz:
The issue is, 'steps' can mean a lot; a slower person taking small steps is different from a fast person taking large steps. For someone, 1 mile could be 1000 steps but for another it could be 1500. It would have to be an 'average' conversion.

thefenriswolf:
So how do the likes of Fitbit or Apple Watch do it? Do they make you measure your steps or do they use GPS or maybe they just guess?

davidhealey:
> So how do the likes of Fitbit or Apple Watch do it? Do they make you measure your steps or do they use GPS or maybe they just guess?

They use GPS but have to be tethered to a phone - although some of the newer devices have GPS built in.

Riksu9000:
While it would definitely be nice to see the distance traveled, I'm not sure we have the capability to produce a sufficiently accurate estimation. In the end I think I'd rather not add this than add something that can potentially report wildly inaccurate distances.

JF002:
We could of course multiply the number of steps by the average step size (or even let the user configure their step size), but I'm not sure about the added value of this feature... since the step counter is triggered by many other motions than actual steps.

It might be possible to make a better estimations by processing the raw data from the accelerometer, but I'm not sure this would be in the scope of this project...

So I agree with @Riksu9000, we should not add a feature that will produce inaccurate and confusing results.

PR #1230, opened 2022-07-10
[WIP] Steps: Show distance traveled during trip (in kilometers)

ItzSwirlz:
The source for this data is https://www.uwyo.edu/wintherockies_edur/win%20steps/coordinator%20info/step%20conversions.pdf .

For now this is just an integer as it is the only way I can figure out how to format the string into the label. In the future, a more precise distance would be ideal in the future (up to hundredths).

------------------

I resized some things like the "reset" button so the item would fit. At first I thought to create a new screen (like Settings does, swipe down to next screen) but for one bit of data, unless a more advanced UI is made this is the easiest way.

### DIFF
--- a/src/displayapp/screens/Steps.cpp
+++ b/src/displayapp/screens/Steps.cpp
@@ -28,7 +28,7 @@ Steps::Steps(Pinetime::Applications::DisplayApp* app,
 
   stepsCount = motionController.NbSteps();
   currentTripSteps = stepsCount - motionController.GetTripSteps();
-  currentTripStepsDistance = motionController.GetTripSteps() * 0.0008;
+  currentTripStepsDistance = motionController.GetTripSteps() / 1250;
   // Step conversion data source: https://www.uwyo.edu/wintherockies_edur/win%20steps/coordinator%20info/step%20conversions.pdf
 
   lv_arc_set_value(stepsArc, int16_t(500 * stepsCount / settingsController.GetStepsGoal()));
--- a/src/displayapp/screens/Steps.cpp
+++ b/src/displayapp/screens/Steps.cpp
@@ -83,7 +83,7 @@ Steps::~Steps() {
 void Steps::Refresh() {
   stepsCount = motionController.NbSteps();
   currentTripSteps = motionController.GetTripSteps();
-  currentTripStepsDistance = motionController.GetTripSteps() * 0.0008;
+  currentTripStepsDistance = motionController.GetTripSteps() / 1250;
   // Step conversion data source: https://www.uwyo.edu/wintherockies_edur/win%20steps/coordinator%20info/step%20conversions.pdf
 
   lv_label_set_text_fmt(lSteps, "%li", stepsCount);
--- a/src/displayapp/screens/Steps.cpp
+++ b/src/displayapp/screens/Steps.cpp
@@ -28,6 +28,8 @@ Steps::Steps(Pinetime::Applications::DisplayApp* app,
 
   stepsCount = motionController.NbSteps();
   currentTripSteps = stepsCount - motionController.GetTripSteps();
+  currentTripStepsDistance = motionController.GetTripSteps() * 0.0008;
+  // Step conversion data source: https://www.uwyo.edu/wintherockies_edur/win%20steps/coordinator%20info/step%20conversions.pdf
 
   lv_arc_set_value(stepsArc, int16_t(500 * stepsCount / settingsController.GetStepsGoal()));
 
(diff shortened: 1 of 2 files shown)


src/displayapp/screens/Steps.cpp
```diff
@@ -28,6 +28,8 @@ Steps::Steps(Pinetime::Applications::DisplayApp* app,
 
   stepsCount = motionController.NbSteps();
   currentTripSteps = stepsCount - motionController.GetTripSteps();
+  currentTripStepsDistance = motionController.GetTripSteps() * 0.0008;
```
FintasticMan, line 31: This is doing conversion from an integer to a floating-point and then back, and that's not very efficient. It would be a lot better to replace the multiplication with 0.0008 with a division by 1250.
```suggestion
  currentTripStepsDistance = motionController.GetTripSteps() / 1250;
```

src/displayapp/screens/Steps.cpp
```diff
@@ -51,19 +53,25 @@ Steps::Steps(Pinetime::Applications::DisplayApp* app,
...
   lv_obj_set_event_cb(resetBtn, lap_event_handler);
-  lv_obj_set_height(resetBtn, 50);
+  lv_obj_set_height(resetBtn, 30);
   lv_obj_set_width(resetBtn, 115);
   lv_obj_align(resetBtn, lv_scr_act(), LV_ALIGN_IN_BOTTOM_MID, 0, 0);
   resetButtonLabel = lv_label_create(resetBtn, nullptr);
   lv_label_set_text_static(resetButtonLabel, "Reset");
 
   currentTripSteps = motionController.GetTripSteps();
+  currentTripStepsDistance = motionController.GetTripSteps() * 0.0008;
```
FintasticMan, line 63: Why are `currentTripSteps` and `currentTripStepsDistance` being set here again, when they were just set above?

src/displayapp/screens/Steps.cpp
```diff
@@ -28,6 +28,8 @@ Steps::Steps(Pinetime::Applications::DisplayApp* app,
 
   stepsCount = motionController.NbSteps();
   currentTripSteps = stepsCount - motionController.GetTripSteps();
```
FintasticMan, line 30: I think this should just be `motionController.GetTripSteps()`, not `stepsCount - motionController.GetTripSteps()`.
```suggestion
  currentTripSteps = motionController.GetTripSteps();
```

src/displayapp/screens/Steps.cpp
```diff
@@ -75,6 +83,8 @@ Steps::~Steps() {
 void Steps::Refresh() {
   stepsCount = motionController.NbSteps();
   currentTripSteps = motionController.GetTripSteps();
+  currentTripStepsDistance = motionController.GetTripSteps() * 0.0008;
```
FintasticMan, line 86: Same as above.
```suggestion
  currentTripStepsDistance = motionController.GetTripSteps() / 1250;
```

FintasticMan (changes requested):

Riksu9000:
Perhaps it could switch between showing steps and showing kilometers by tapping on the screen, as it's basically the same information.

yehoshuapw:
perhaps this number should be configurable. adding UI to do that, I think it overkill - but maybe just a number inside a file, and then it can be changed over the FS. (and read it only once per boot or something like that)

ItzSwirlz:
I'm on vacation, I just did this shortly because it was raining. I'll review everything eventually

src/displayapp/screens/Steps.cpp
ItzSwirlz: Eh, good point.

src/displayapp/screens/Steps.cpp
ItzSwirlz: I stumbled upon this in the code and honestly I had no idea. I thought it may be to refresh the tripsteps, so thats why I also added it again. I'll remove these two lines - the Refresh function can update it.

src/displayapp/screens/Steps.cpp
ItzSwirlz: The purpose of this (I believe) is to ensure steps are tracked efficiently in the time allocated. Let me explain.

Steps: 2000
I'm resetting my trip meter, so my steps are 0.
I walk 1000 steps.
Steps: 3000
Trip: 1000

I think this takes the current step count and ensures the trip is the portion of those steps correctly.

ItzSwirlz:
> perhaps this number should be configurable. adding UI to do that, I think it overkill - but maybe just a number inside a file, and then it can be changed over the FS. (and read it only once per boot or something like that)

I do not know how far we are in terms of BLE/FS. If I knew we were I would probably add a pdf/ebook reader lol.

But this number should be configurable via a setting. And overall I guess the Steps file could use a good cleanup.

src/displayapp/screens/Steps.cpp
FintasticMan: I think that that calculation should already be done in `motionController.GetTripSteps()`, as in the `Refresh()` function, `currentTripSteps` is just set the way I suggested.

RageGamerBoi:
> perhaps this number should be configurable. adding UI to do that, I think it overkill - but maybe just a number inside a file, and then it can be changed over the FS. (and read it only once per boot or something like that)

I would love to have this on the watch itself, because even though Gadgetbridge already does this (it lets you configure your pace length and displays distance walked among other things), I don't want to have to pair it to my phone for anything other then Music and Notifications.
I think a dedicated panel in the settings app would be the best option as I don't have to mess with anything over Bluetooth.

Riksu9000:
For the initial implementation, a static value is fine. Let's not expand the scope of this PR too much.

RageGamerBoi:
Makes sense...

Riksu9000:
We've decided to close the issue #883 and not add this feature.

Issue #1086, opened 2022-04-14
Alarms sometime do not trigger


brabalan:
### Verification

- [X] I searched for similar bug reports and found none was relevant.

### What happened?

A repeating alarm did not ring

### What should happen instead?

It should have rung

### Reproduction steps

- Set up a repeating alarm for the next day (it seems that spanning midnight is important)
- wait until the alarm rings

### More details?

This seems related to #774 but it does not seem to be about time change (unless something special happens at midnight).

My observations are:
- until the time of the alarm should ring, the estimated time ("i" button on alarm panel) is correct
- when the alarm does not ring, the estimated time becomes thousands of days in the future
- alarms that do not repeat are fine
- I’m testing a repeating alarm in more than 12 hours to see if the issue is related to midnight or the length of time to the alarm

### Version

v1.9.0

### Companion app

GadgetBridge v.0.66.0

rev22:
I've tried to reproduce the issue reported by using the Set Time function in the Settings, by using the following steps:
- disable bluetooth from the settings (this may be optional but it's intended to avoid interactions with companions apps)
- set a daily alarm for 00:02 time
- set the time to 23:59
- waiting for 4 minutes
When running 1.9.0 or 1.8.0 firmware on my Pinetime, no alarm sets off at 00:02, and in the Alarm app the time to alarm is reported as thousands of days into the future.
Oddily, when following the same steps on the 1.9.0 firmware running on InfiniSim, the alarm sets off normally at 00:02

Asdew95:
I can reproduce the issue with a non-repeating alarm, not passing midnight.

1. Disconnect or disable Bluetooth.
2. Set a non-repeating alarm to a few minutes from now.
3. Set the time to a minute before the alarm should go off.

Before the alarm should go off after setting the new time, the time to alarm is correct. It becomes tens of thousands of days (it always seems to become around 49710 days and 6 hours) after the alarm should have gone off. I can also reproduce this by not setting the time manually but having the watch connected to Gadgetbridge (v0.66.0) when setting the alarm and reconnecting from Gadgetbridge after the alarm is set.

medeyko:
I opened #1088, and I'm not sure it is a duplicate of this one.

[...]

PR #1233, opened 2022-07-11
changed the alarm timeout to a saner value of 3 minutes

mashuptwice:
In pull #945 the alarm functionality got expanded by a auto-dismiss after 60 seconds. As others pointed out, this makes the alarm hardly usable for hard sleepers. Also in **very** loud environments, for example on stage, 60 seconds could be to short to notice the alarm.

### DIFF
--- a/src/displayapp/screens/Alarm.cpp
+++ b/src/displayapp/screens/Alarm.cpp
@@ -253,7 +253,7 @@ void Alarm::UpdateAlarmTime() {
 void Alarm::SetAlerting() {
   lv_obj_set_hidden(enableSwitch, true);
   lv_obj_set_hidden(btnStop, false);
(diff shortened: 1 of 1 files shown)


Riksu9000:
I wouldn't say three minutes is more or less sane in this case. From what I've read, the issues people were having were most likely #1086 and not 60 seconds being too short. The use case on a watch is different than other alarms, as the watch should always be strapped to the wrist, whereas other alarms could be placed somewhere where they may not immediately get noticed. The timeout is to stop it from vibrating unnecessarily when the watch isn't worn. I think the proper solution to #1088 would be to have the alarm repeat and maybe with increasing strength.

JF002:
The current timeout of 60s was decided after a rather long discussion with many users. The conclusion of this discussion is that the perfect timeout value does not exist: it depends on many factors and on the use-case. So we decided to stick with 60s as the default timeout for now. 
It might take some time to implement a more advanced alarm, but you can be sure that if we switches to 180s, other users will open an issue similar to this one asking to change the timeout to another value :)

PR #1234, opened 2022-07-12
Show notification indicator when a notification is available

### LINKED ISSUE #1216: Enhancement Request: "New Mail" used for undismissed notifications
I'm describing the behavior as it is referred in the Terminal watchface, but I'm sure there are similar indicators in the other watch faces.

Current behavior: When notifications are set to silent on the watch, the presence of new notifications is indicated by the phrase "You have new mail" on the terminal watchface. Swiping up to read the notifications dismisses this message.

When silent is not set, new notifications appear on the screen briefly and then disappear, and the "new message" message does not show up.

Desired change: Now that InfiniTime, with 1.10, offers the ability to dismiss notifications with a gesture, I would like the "new notifications" message to appear when the number of notifications available to read is greater than 0/0, regardless of whether the watch is in silent or unsilent mode, and to only disappear after all the available notifications have been dismissed.

Thank you for the consideration of this entrancement.

0cc4m:
This change removes the newNotification flag and the clear flag function and replaces it with whether notifications are available (have not been swiped). Thus, the indicator is shown even when alerts are on and stays until all are dismissed.

Fixes #1216


JF002:
Previously, the notification indicator was displayed when a notification was missed (mostly when the "do not disturb" mode was enabled and when a notification was received while the user was in an app).

This PR changes the "meaning" of that indicator : instead of notifying the user that they missed a notification, it tells the user that there are notifications that are not dimissed yet.

I'm OK with the current behavior, but I know that some users were confused by this notification indicator. So... any other opinion about this? @InfiniTimeOrg/core-developers  ?

Riksu9000:
Yes, I've been thinking about this and it seems like a good idea, but the color of the icon should probably be changed to white, so it's less ugly to look at when it will be seen for extended periods of time.

0cc4m:
A combination would also be possible. Keep the green indicator for the missed notification when alerts are off, add a white one for notifications that haven't been swiped. I'm fine either way, I can change the PR depending on what you prefer.

DanieleVBella:
My two cents based on how I use InfiniTime, with the disclaimer that I have no clue if anyone uses the watch in a similar way to me:

Currently, dismissing a notification on the PineTime does not actually get rid of the notification on the paired device; it only discards it within InifiniTime. On top of this, notifications in InfiniTime don't work like in Android or iOS due to the fact that only 5 can be displayed, so older notifications get overwritten as you get more.

Because of this, I have always thought of the notification list on InfiniTime as a list of _recent_ notifications, rather than a list of _unread_ notifications. For me at least, it works really well like this, and honestly I've never felt the need to dismiss a notification; I just let the older ones get overwritten, and the vibration (or icon in "Do not Disturb" mode) have always been enough that I always knew when I got a new unread notification.

Of course, the _ability_ to dismiss notifications does not hinder the way I use my watch and thus is not an issue. However, for someone that uses the notification list like I do, merging this PR however would mean that I would have a constant notification icon in the status bar, which would make the symbol meaningless, becoming just visual clutter.

From a more objective point of view, if a more phone-like notification list is what InifiniTime wants to move towards, I don't know if it would be the right time to merge this PR; making this change encourages the user to dismiss all their notifications on the PineTime as soon as they see it, but as I mentioned this doesn't (yet) dismiss the notification on the paired device, and vice versa. This effectively means that every notification you get on the paired device would have to be dismissed twice (once on the PineTime and once on the paired device), which at least to me sounds like a lot of hassle.

With all this said, the dual colour notification idea proposed by @0cc4m would partially solve my issue (I'd still have a green notification icon for when I actually needed it), and if I'm not the only one that uses the notifications list like I do, a setting could be added in the worst case to accomodate everyone.

0cc4m:
Interesting perspective @DanieleVBella 
I did not consider this because I use the notifications differently. For me they don't act as full access to the device notification, since they only show a small part of the text anyways. For this reason I wouldn't like if they removed the notification on the host device either. I use them only as a small preview to an event on the phone, which is enough to know the broad urgency of the notification. That's why I dismiss them on the watch as soon as I've read them.

Looks like more consideration is required to find a solution that suits everyone or at least most people.

0cc4m:
I updated the PR with a version that keeps the NewNotification flag and indicator. I updated the watch faces to show the notification in green (white for PineTime watch face) when NewNotification is set or white (black for PineTime watch face) when the notification buffer is not empty.

I can look into creating a simple settings page to switch between only showing the indicator for new notifications (current behavior) and also showing when notifications have not been swiped. Please let me know what you think.

Riksu9000:
I agree with @DanieleVBella and I guess there isn't a good reason to change this behaviour right now after all. The dual color idea seems too complicated to me.

JF002:
My usage of the notification functionality is very similar to @DanieleVBella's : I mostly look at the overview of the notification to know what it's about so that I know if I have to check my phone. I rarely look at the notification history nor feel the need to dismiss them.
@0cc4m According to our [project vision and principle](https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/doc/InfiniTimeVision.md#core-principles), we prefer a solid and predictable default behavior over settings and customization, which means that we try to avoid adding new settings and options unless they are absolutely necessary. In this case, we'll probably keep the current behavior as is until (for example) notifications are also dismissed on the phone.

src/displayapp/screens/WatchFaceDigital.cpp
```diff
@@ -115,7 +114,15 @@ void WatchFaceDigital::Refresh() {
 
   notificationState = notificatioManager.AreNewNotificationsAvailable();
   if (notificationState.IsUpdated()) {
-    lv_label_set_text_static(notificationIcon, NotificationIcon::GetIcon(notificationState.Get()));
+    if (notificationState.Get()) {
+      lv_obj_set_style_local_text_color(notificationIcon, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(0x00FF00));
+      lv_label_set_text_static(notificationIcon, NotificationIcon::GetIcon(true));
+    } else if (notificatioManager.NbNotifications() > 0) {
```
Commenter25, line 120: This notificatio typo has been fixed since this PR was made.

```suggestion
    } else if (notificationManager.NbNotifications() > 0) {
```

src/displayapp/screens/WatchFacePineTimeStyle.cpp
```diff
@@ -370,8 +370,17 @@ void WatchFacePineTimeStyle::Refresh() {
   }
 
   notificationState = notificatioManager.AreNewNotificationsAvailable();
-  if (notificationState.IsUpdated()) {
-    lv_label_set_text_static(notificationIcon, NotificationIcon::GetIcon(notificationState.Get()));
+  notificationNbState = notificatioManager.NbNotifications() > 0;
```
Commenter25, line 373:
```suggestion
  notificationNbState = notificationManager.NbNotifications() > 0;
```

src/displayapp/screens/WatchFaceTerminal.cpp
```diff
@@ -103,6 +103,10 @@ void WatchFaceTerminal::Refresh() {
   notificationState = notificatioManager.AreNewNotificationsAvailable();
   if (notificationState.IsUpdated()) {
     if (notificationState.Get()) {
+      lv_obj_set_style_local_text_color(notificationIcon, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(0x00FF00));
+      lv_label_set_text_static(notificationIcon, "You have mail.");
+    } else if (notificatioManager.NbNotifications() > 0) {
```
Commenter25, line 108:
```suggestion
    } else if (notificationManager.NbNotifications() > 0) {
```

Commenter25 (changes requested):
Still functions on 1.11, just need to fix the old `notificatioManager` typo. Can't speak to the code quality as I am not fluent in C++, I can just tell it works on my watch :p

0cc4m:
Thanks! I'll take a look at the merge conflict soon.

Riksu9000:
I'll close this since we won't be changing this behavior until the notifications are synced with the phone. https://github.com/InfiniTimeOrg/InfiniTime/pull/1234#issuecomment-1192398826

rnwgnr:
> Previously, the notification indicator was displayed when a notification was missed (mostly when the "do not disturb" mode was enabled and when a notification was received while the user was in an app).

I think i get your reasoning why you don't want to change the behaviour if the notification icon.

From my point of view the current implementation is missing a point that is not related to "syncing notifications with the phone":
There will be moments when you miss a notification for some reason even when you have not manually switched to "do not disturb" or "sleep" mode.
E.g. you are too busy to notice the vibration (which can happen quite easily cause the vibration is very short and "weak" - but that may be another story) or you took the watch off for a few minutes to take a shower.

At the moment you need manual intervention to either change the watch mode (only suitable for the shower use case) or check manually for notifications by swiping down.

In the situations described above the "notification icon" almost completely misses it's purpose of notifying about something because it will most likely never show up.

PR #1243, opened 2022-07-22
Use Counter widget in Alarm

Riksu9000:
Use the Counter widget in alarm. The Info feature was retained by making it a transparent button that stretches above the time, so you can see the info by tapping the time.

InfiniSim_2022-07-22_053359
InfiniSim_2022-07-22_053411

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -367,7 +367,7 @@ void DisplayApp::LoadApp(Apps app, DisplayApp::FullRefreshDirections direction) 
       currentScreen = std::make_unique<Screens::Timer>(this, timerController);
       break;
     case Apps::Alarm:
-      currentScreen = std::make_unique<Screens::Alarm>(this, alarmController, settingsController, *systemTask);
+      currentScreen = std::make_unique<Screens::Alarm>(this, alarmController, settingsController.GetClockType(), *systemTask);
       break;
 
     // Settings
--- a/src/displayapp/screens/Alarm.cpp
+++ b/src/displayapp/screens/Alarm.cpp
@@ -22,6 +22,13 @@
 using namespace Pinetime::Applications::Screens;
 using Pinetime::Controllers::AlarmController;
 
+namespace {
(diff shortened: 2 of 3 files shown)


medeyko:
I think that thasparent button is not intuitive.
It is better to automatically display the info when the alarm is being turned on. If one needs  it, this will save them a click; if one doesn't need it, they would turn off the watch anyway, so the info won't require any additional actions from them.

Riksu9000:
I don't think it not being perfectly intuitive is an issue, as the feature is not critical. I wouldn't automatically pop it up in the face of the user.

medeyko:
> I don't think it not being perfectly intuitive is an issue, as the feature is not critical. I wouldn't automatically pop it up in the face of the user.

I afaid it is a doubtful decision from the user experience point of view. How is it supposed user to find out this feature? By a misclick, and than by test clicking to find an exact boundary of the area that causes the info text to pop? You mean that the gray color of the time fields might hint a user that they are clickable? But in the stop-watch app they are gray too, but NOT clickable. It is not intuitive, it is not consistent. There definitely should be some visible difference between clickable and non-clickable elements in the interface, I believe.
Maybe color?

Riksu9000:
I've now brought back the separate info button while we look for another solution if necessary.

InfiniSim_2022-07-27_123024

mashuptwice:
I propose to show the counter widget by swiping upwards, just as it is in the music app to show the volume controls. 

Swiping up might be a good default gesture for additional features anyways, as if there is one default people are more likely to "discover" these across different screens, even if there is no visible control element. It also doesn't clutter the screen and helps with the problem of hitting small buttons reliably.

Riksu9000:
> I propose to show the counter widget by swiping upwards, just as it is in the music app to show the volume controls.

@mashuptwice Do you mean that it shouldn't be adjustable by default or what do you mean?

mashuptwice:
@Riksu9000 I am not sure what you mean by adjustable. 

I mean to use a swipe-up instead of the info button to show the countdown until the alarm.

JF002 (approved):

PR #1245, opened 2022-07-22
Dismiss notifications in preview screen with SwipeRight

### LINKED ISSUE #1232: swipe to remove new notification
### Verification

### Pitch us your idea!

new notification dismissable by swipe while preview is shown

### Description

While dismissing notifications works fine while in the notification screen, I would like to be able to instantly dismiss a new notification during the time it gets shown on the display and (if possible) also return to the previous screen. 

Currently to dismiss a notification which just came in, I have to:
0. receive notification
1. swipe up to show home
2. swipe down to show notifications 
3. swipe right to dismiss the notification
4. swipe up to show home
5. (navigate to previous app)

My proposal would make it as following:
0. receive notification
1. swipe right to dismiss
2. (navigate to previous app)

0cc4m:
I often found myself trying to swipe away a new notification, but in the preview this wasn't possible. This small change allows to swipe away notifications from the preview.

I'm not sure if there's a more elegant way to do it, but I think this is the simplest.

Fixes #1232

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -226,6 +226,10 @@ void DisplayApp::Refresh() {
               default:
                 break;
             }
+          } else if (currentApp == Apps::NotificationsPreview && gesture == TouchEvents::SwipeRight) {
+            auto currentId = notificationManager.GetLastNotification().id;
+            notificationManager.Dismiss(currentId);
+            LoadPreviousScreen();
           } else if (returnTouchEvent == gesture) {
             LoadPreviousScreen();
           }


src/displayapp/DisplayApp.cpp
```diff
@@ -226,6 +226,10 @@ void DisplayApp::Refresh() {
               default:
                 break;
             }
+          } else if (currentApp == Apps::NotificationsPreview && gesture == TouchEvents::SwipeRight) {
+            auto currentId = notificationManager.GetLastNotification().id;
```
mashuptwice, line 230: Is it necessary to write the ID to a variable? 

Without testing my claim, `notificationManager.Dismiss(notificationManager.GetLastNotification().id);` would also work or am I missing something?

mashuptwice (changes requested):

src/displayapp/DisplayApp.cpp
0cc4m: You are right, I just think putting it into two lines makes it a little easier to read. In the end the compiler optimizes both versions to the same result I believe.

Riksu9000:
I think this behaviour should be implemented in Notifications.cpp rather than DisplayApp if possible.

Commenter25:
Should this be closed in favor of #1332?

NeroBurner:
I'd suggest yes, #1332 works very well, and it should be merged into develop soon-ish

0cc4m:
Alright. Closing in favor of #1332.

PR #1251, opened 2022-07-27
More timeout options and improved checkbox alignment

Riksu9000:
Added timeout options in a way which requires fewer clicks than #1222 as suggested by @mashuptwice. I want to make these larger, but I thought I'd do that in another PR.
Also improved the checkbox inner padding so the extra whitespace can be removed without visible changes.

InfiniSim_2022-07-27_171319

### DIFF
--- a/src/displayapp/lv_pinetime_theme.c
+++ b/src/displayapp/lv_pinetime_theme.c
@@ -207,7 +207,7 @@ static void basic_init(void) {
 
   lv_style_reset(&style_cb_bg);
   lv_style_set_radius(&style_cb_bg, LV_STATE_DEFAULT, LV_DPX(4));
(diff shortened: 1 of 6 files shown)


JF002 (approved):

lman0:
hi can add 45 and 60 second?
btw what is the purpose of the 7 second? 
it felt a placeholder if anything...

Itai-Nelken:
> [...]
> btw what is the purpose of the 7 second? 
> [...]

Personally it will be very useful for me. I currently use the 15s option because the 5s option is to shirt, but 15s is too long. 7s looks like a sweet spot.

trman:
well there is 10 second so @Itai-Nelken , the 7 felt truly out of place

PR #1252, opened 2022-07-27
Notification swap text colors

### LINKED ISSUE #972: More readable notifications for "old eyes"

Riksu9000:
Improves visibility. Also reduced code duplication slightly. Fixes #972.

InfiniSim_2022-07-27_175216

### DIFF
--- a/src/displayapp/screens/Notifications.cpp
+++ b/src/displayapp/screens/Notifications.cpp
@@ -272,7 +272,7 @@ Notifications::NotificationItem::NotificationItem(const char* title,
   lv_obj_align(alert_count, NULL, LV_ALIGN_IN_TOP_RIGHT, 0, 16);
 
   lv_obj_t* alert_type = lv_label_create(container, nullptr);
-  lv_obj_set_style_local_text_color(alert_type, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_MAKE(0xb0, 0xb0, 0xb0));
+  lv_obj_set_style_local_text_color(alert_type, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_MAKE(0xff, 0xb0, 0x0));
   if (title == nullptr) {
     lv_label_set_text_static(alert_type, "Notification");
   } else {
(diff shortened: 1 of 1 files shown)


mashuptwice:
Would it be an option to use an even darker background to further increase the contrast, eg. to #202020 ?

infini_notif_color

Riksu9000:
I think I'd prefer to keep it as it is for now to keep the contrast between the black background and according to the linked issue, this change should be good enough.

Sec42:
@mashuptwice I've made a version with larger text & improved contrast - check https://github.com/Sec42/InfiniTime/commit/bda8431671b1e0b5584def06cf07a2368b981e89 and the zip @ https://github.com/Sec42/InfiniTime/releases/tag/1.10.0

JF002 (approved):

calhix:
Is there a reason to have the grey box? White text with a black background (no text bubble) would be so much better than the current implementation IMO.

calhix:
> @mashuptwice I've made a version with larger text & improved contrast - check [Sec42@bda8431](https://github.com/Sec42/InfiniTime/commit/bda8431671b1e0b5584def06cf07a2368b981e89) and the zip @ https://github.com/Sec42/InfiniTime/releases/tag/1.10.0

This is really good but I would prefer a smaller font, a setting for adjustable fonts would be very helpful in the future.

jakethelizard99:
I second what @calhix said swapping to a grey background and black text box would look a lot better.

Riksu9000:
The background must be black to blend the screen edge to the thick border around the screen.

calhix:
> The background must be black to blend the screen edge to the thick border around the screen.

True, how about all black like @Sec42 build but a thin white line for the text box?

NeroBurner (approved):

Riksu9000:
Outlines aren't a part of the current InfiniTime style, so I think it might look out of place. Similar gray backgrounds are used on objects like on the QuickSettings screen and the Timer. I don't see the need to change anything more with this PR.

PR #1258, opened 2022-07-31
Centralize most color definitions

### LINKED ISSUE #559: Make color definitions more global in source code

Riksu9000:
Renamed lv_pinetime_theme files to InfiniTimeTheme and made them C++.
Most colors are now defined in InfiniTimeTheme.h or use the LVGL default colors. There were a few colors unique to a single screen, which I don't think should be defined in the theme at least for now.
A few very dark colors were rounded to bgDark.

Relates to #559

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -477,7 +477,7 @@ list(APPEND SOURCE_FILES
         FreeRTOS/port_cmsis.c
 
         displayapp/LittleVgl.cpp
-        displayapp/lv_pinetime_theme.c
+        displayapp/InfiniTimeTheme.cpp
 
         systemtask/SystemTask.cpp
         systemtask/SystemMonitor.cpp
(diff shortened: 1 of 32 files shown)


JF002 (approved):
LGTM. 
Looks like the simulator will need to be updated according to those changes too.

NeroBurner:
Supported in InfiniSim introduced with https://github.com/InfiniTimeOrg/InfiniSim/pull/50 :+1:

NeroBurner (approved):

PR #1259, opened 2022-08-01
Use Counter widget in SetDate

Riksu9000:
Counter was expanded with month mode, variable digit count and a minimum width.

The counter will always show at least two digits with leading zeroes if necessary. This doesn't impact anything currently, but it's worth considering and I'll see if I can find a neat solution to this.

InfiniSim_2022-08-01_182556

### DIFF
--- a/src/displayapp/screens/settings/SettingSetDate.cpp
+++ b/src/displayapp/screens/settings/SettingSetDate.cpp
@@ -11,13 +11,36 @@ namespace {
   constexpr int16_t POS_X_DAY = -72;
   constexpr int16_t POS_X_MONTH = 0;
   constexpr int16_t POS_X_YEAR = 72;
-  constexpr int16_t POS_Y_PLUS = -50;
   constexpr int16_t POS_Y_TEXT = -6;
-  constexpr int16_t POS_Y_MINUS = 40;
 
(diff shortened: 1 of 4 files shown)


src/displayapp/screens/settings/SettingSetDate.cpp
```diff
@@ -11,13 +11,36 @@ namespace {
...
+      screen->HandleButtonPress();
+    }
+  }
+
+  void ValueChangedHandler(void* userData) {
+    auto* screen = static_cast<SettingSetDate*>(userData);
+    screen->CheckDay();
+  }
+
+  int MaximumDayOfMonth(uint8_t month, uint16_t year) {
```
JF002, line 28: What an attention to detail! :+1:

JF002 (approved):

NeroBurner (approved):
Loving the new counter style and how much code in the buttonhandler gets cleaned up by using the counter widget

Riksu9000:
I have found a neat solution to the leading zero issue, and I'll open a new PR for that later.

NeroBurner:
I think the day with leading 0 is no problem at all. I would propose to keep it that way if it keeps the code cleaner. I don't think a leading 0 in a day counter is very confusing. Just my two cents :)

Issue #183, opened 2021-01-26
Heart rate measurement is stops, when the screen is turned off


PetersOtto:
I installed v 0.11.0.
The heart rate sensor works great! 

I compare the measurement of the PineTime with the measurement of my Garmin Forerunner. The value are in the same range. Thats  good, I think.

However the measurement is stopping, when the screen is going off.
It would be great to have a button, that deactivate the power off-mode from the screen. Then it would be possible to have a longer view on the measurement.

It seems that the Heart Rate Sensor of the Garmin is connected to accelerator- / motion sensor.  The Heart Rate Sensor starts working, when movement is detected. Perhaps that could be a solution for future.

Thank you for working on this firmware!

JF002:
Thanks for the comparison with another device! Good to know the results seem correct!

The behavior you describe is expected : this is a first integration of the heart rate sensor, and the UI/UX can still be improved in further release. 

I cannot measure it for now, but I guess that leaving the heart rate sensor all the time might use a lot of power and reduce the battery life. As you said, using info from the motion sensor might be a good idea, but this sensor is not integrated yet in InfiniTime. There's still work to do ;)

> Thank you for working on this firmware!

Thanks, I appreciate! And thanks for using it and sharing your experience!

piggz:
Id like to see a couple of modes, each should save power in different ways:
1. Allow the user to turn on/off the HRM explicitly.  This would be on constantly until the user turns it off.  This would be useful for exercise, where for example the watch is supplying the heartrate to another device
2. Have the device turn on the HRM at a user specified interval, typically 60 seconds.  This would be useful for when the firmware has access to a filesystem for logging the heartrate.  A companion app can then download that data later.

JF002:
I also think of a 3rd case : in addition to turn the HRM ON at a specified interval, we could also use the motion sensor to turn the HRM ON on motion even.

... But first, we need to implement the motion sensor :)

jones139:
> ... But first, we need to implement the motion sensor :)

I know, I have been feeling bad about lack of progress.   I am encouraged though that this HRM implementation is using FreeRTOS tasks and a message queue as far as I can tell from a quick look so I will use that style to implement the accelerometer......but first I need to get the new version to compile so I can add in the accelerometer - it is failing at the moment saying it will not fit in the flash memory "ld: pinetime-app-0.12.0.out section `.ARM.exidx' will not fit in region `FLASH'", so I must have my configuration wrong.......

[...]

PR #1261, opened 2022-08-02
Add sleep mode

### LINKED ISSUE #689: Sleep mode
It would be handy to have a way to set a range of hours where wake on lift will not turn on the watch. For example while sleeping you'll be tossing and turning and it can drain the battery a lot quicker.

At least, this would be a decent workaround for it being so sensitive.

Riksu9000:
A sleep mode is needed so the watch can be worn during the night without causing the menus to activate.

Sleep mode disables notifications, touch wakeup, motion wakeup and chimes.

InfiniSim_2022-08-02_105252

Fixes #689

### DIFF
--- a/src/components/settings/Settings.h
+++ b/src/components/settings/Settings.h
@@ -9,7 +9,7 @@ namespace Pinetime {
     class Settings {
     public:
       enum class ClockType : uint8_t { H24, H12 };
-      enum class Notification : uint8_t { ON, OFF };
+      enum class Notification : uint8_t { On, Off, Sleep };
       enum class ChimesOption : uint8_t { None, Hours, HalfHours };
       enum class WakeUpMode : uint8_t {
         SingleTap = 0,
@@ -219,7 +219,7 @@ namespace Pinetime {
         uint32_t screenTimeOut = 15000;
 
         ClockType clockType = ClockType::H24;
-        Notification notificationStatus = Notification::ON;
+        Notification notificationStatus = Notification::On;
 
         uint8_t clockFace = 0;
         ChimesOption chimesOption = ChimesOption::None;
--- a/src/displayapp/screens/settings/QuickSettings.cpp
+++ b/src/displayapp/screens/settings/QuickSettings.cpp
@@ -9,13 +9,21 @@ using namespace Pinetime::Applications::Screens;
 namespace {
   void ButtonEventHandler(lv_obj_t* obj, lv_event_t event) {
     auto* screen = static_cast<QuickSettings*>(obj->user_data);
-    screen->OnButtonEvent(obj, event);
+    if (event == LV_EVENT_CLICKED) {
+      screen->OnButtonEvent(obj);
+    }
   }
 
   void lv_update_task(struct _lv_task_t* task) {
     auto* user_data = static_cast<QuickSettings*>(task->user_data);
     user_data->UpdateScreen();
   }
+
+  enum class ButtonState : lv_state_t {
+    NotificationsOn = LV_STATE_CHECKED,
+    NotificationsOff = LV_STATE_DEFAULT,
+    Sleep = 0x40,
+  };
 }
 
 QuickSettings::QuickSettings(Pinetime::Applications::DisplayApp* app,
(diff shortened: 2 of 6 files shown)


FintasticMan:
Nice work! I think it should also disable chimes so that if you enable sleep mode you don't get woken up every (half) hour.

trman:
i want to know , what is the difference with  #961 , execpt that is you that is doing the pull?

trman:
(please don't tell me that  , that is always better when it's you that is doing the pull....
people migh no longer to try to add functionnly if you will close  or draft their their pull only to make shine yours
i've seen project crumble from inside with this kind of behaviour)

FintasticMan:
@trman I created #961, it is pull request with a very different goal and use-case. This pull request disables things that could be a distraction for going to sleep, while my pull request is a fix (debatable) for the issue of the touch panel recognising touches if you didn't touch the screen.

There aren't many solutions to the issue of distractions when you're sleeping, so this pull request would be pretty easy to merge. There are many ways to solve the issue of ghost inputs, one of which is mine. The way to figure out which works best is by trying them, and so far there haven't been any other implementations of a fix to compare against.

I am planning on updating my pull request and trying to fix the issues it has, but I don't think these pull requests are at all related.

trman:
when i see two pull that remove touch with one in draft and the other nearly merged 
it's normal to me me that it is  explained why the previous is not merged first.

for the cynical , i've seen @Riksu9000  closing pull with no reason except  he don't like the pull purpose...

lman0:
hi @Riksu9000  , is clicking on the button is enough to restore the touch screen?

FintasticMan:
> the https://github.com/InfiniTimeOrg/InfiniTime/pull/961 would no longer serve any purpose , @FintasticMan since it only need to click on the "sleep" icon in order to disable/make sleep the touch screen

This pull request doesn't disable touch, it only disables touch wakeup.

trman:
the explanation was lacking , so i thinked it was "touch" aka  touch screen  and "motion wake up"

lman0:
even now , the only thing that wake up  the screen in my case is the touch screen because i don't use the wake up by wrist or screen .
and i disable the notification before sleeping.
is this pull will  stop the screen wake up because of touch @Riksu9000 ?

minacode:
@Riksu9000 please include chimes in this. It makes sense and would resolve #1083 I opened some time ago and never found the time to work on.

Riksu9000:
I've now updated the description to be more clear about what this PR does and also disabled Chimes. Thanks @FintasticMan for clearing up the confusion.

Let's stay on topic from this point on.

mashuptwice:
This would be a good point to resolve #183 and parts of #895 with an always-on mode, similar to caffeine on LineageOs/Android. 

infinitime-always-on

infinitime-always-on_2nd_logo

calhix:
I like this idea, not sure if it would warrant a separate issue or not but I would appreciate a Do Not Disturb functionality that is configurable in the quick settings, which will automatically turn on and off at the times/schedule you specify like on phones.

Tiggilyboo:
Great work, I was about to make the same changes myself, beat me to it!

jakethelizard99:
Used the feature last night works flawlessly!  Thank you so much for following through greatly improves the usability of the watch for me!

src/components/settings/Settings.h
```diff
@@ -9,7 +9,7 @@ namespace Pinetime {
     class Settings {
     public:
       enum class ClockType : uint8_t { H24, H12 };
-      enum class Notification : uint8_t { ON, OFF };
+      enum class Notification : uint8_t { On, Off, Sleep };
```
NeroBurner, line 12: I understand why you changed `ON` to `On`, as `SLEEP` somehow feels out of place. But for me `On` and `Off` in code as keywords/enums feels out of place as well. Would it be ok to keep `ON` and `OFF` and just add `Sleep` as new entry?

```suggestion
      enum class Notification : uint8_t { ON, OFF, Sleep };
```

src/displayapp/screens/settings/QuickSettings.cpp
```diff
@@ -77,19 +81,23 @@ QuickSettings::QuickSettings(Pinetime::Applications::DisplayApp* app,
   btn3 = lv_btn_create(lv_scr_act(), nullptr);
   btn3->user_data = this;
   lv_obj_set_event_cb(btn3, ButtonEventHandler);
-  lv_btn_set_checkable(btn3, true);
   lv_obj_add_style(btn3, LV_BTN_PART_MAIN, &btn_style);
+  lv_obj_set_style_local_bg_color(btn3, LV_BTN_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_RED);
+  lv_obj_set_style_local_bg_color(btn3, LV_BTN_PART_MAIN, specialState, LV_COLOR_MAKE(0x60, 0x00, 0xff));
```
NeroBurner, line 86: can you add a comment what color this is?

src/systemtask/SystemTask.cpp
```diff
@@ -483,12 +486,11 @@ void SystemTask::UpdateMotion() {
   motionController.IsSensorOk(motionSensor.IsOk());
   motionController.Update(motionValues.x, motionValues.y, motionValues.z, motionValues.steps);
 
-  if (settingsController.isWakeUpModeOn(Pinetime::Controllers::Settings::WakeUpMode::RaiseWrist) &&
-      motionController.Should_RaiseWake(state == SystemTaskState::Sleeping)) {
-    GoToRunning();
-  }
-  if (settingsController.isWakeUpModeOn(Pinetime::Controllers::Settings::WakeUpMode::Shake) &&
-      motionController.Should_ShakeWake(settingsController.GetShakeThreshold())) {
+  if (settingsController.GetNotificationStatus() != Controllers::Settings::Notification::Sleep &&
```
NeroBurner, line 489: this is a very long convoluted conditional making my head dizzy. I think it would be better for readability and future bug prevention to keep this conditional simpler (even though we'd have to use more lines of code for it)

NeroBurner (changes requested):

lman0:
@Riksu9000  more than sleeping , it' s more a 'silent mode' to me .
a true sleep mode ,  for me ,would  slow down battery usage as well , with performance reduced , screen dimmed, color filtered but clock and alarm functional.

no forgetting that , maybe one day , like on other swartwatch we may have on quick setting an icon/notifcation icon for regular app , 
and app for slepp do exist , and use the sleep icon as well.

src/systemtask/SystemTask.cpp
Riksu9000: Would it be enough to separate the notification status check, but keep the wakeup methods in one?

src/components/settings/Settings.h
Riksu9000: All uppercase seems more out of place to me, since we don't use that style anywhere else.

JF002: I agree with @Riksu9000, `Notification::On` and `Notification::Off` are more consistent with the naming convention used in the project.

src/systemtask/SystemTask.cpp
NeroBurner: I think so, yes

src/displayapp/screens/settings/QuickSettings.cpp
```diff
@@ -9,13 +9,17 @@ using namespace Pinetime::Applications::Screens;
...
+      screen->OnButtonEvent(obj);
+    }
   }
 
   void lv_update_task(struct _lv_task_t* task) {
     auto* user_data = static_cast<QuickSettings*>(task->user_data);
     user_data->UpdateScreen();
   }
+
+  constexpr uint8_t specialState = 0x40;
```
JF002, line 22: The name of this constant does not provide enough clue about its meaning. If I understand correctly, it's  used to define a 3rd state to the button (default, checked, "special").

To be more expressive, we can maybe try this : 

```
static constexpr uint8_t notificationOnState = LV_STATE_CHECKED;
static constexpr uint8_t notificationOffState = LV_STATE_DEFAULT;
static constexpr uint8_t notificationSleepState = 0x40; // define an additional state to the button to handle Sleep mode
```

And why not use a `enum class` ?
```
enum class ButtonState : uint8_t {
  On = LV_STATE_CHECKED,
  Off = LV_STATE_DEFAULT,
  Sleep = 0x40; // define an additional state to the button to handle Sleep mode
}
```

src/components/settings/Settings.h
NeroBurner: ok :)

NeroBurner (approved):

JF002 (approved):

wmertens:
Would it be possible to go in/out of this mode when sleep tracking via Sleep as Android?

Issue #1263, opened 2022-08-02
Improve visibility for people with color vision deficiency


lman0:
### Verification

- [X] I searched for similar feature request and found none was relevant.

### Pitch us your idea!

add list of color correction by type of color blindness so these people can use pinetime

### Description

some people are born or become with Color vision deficiency 
https://en.wikipedia.org/wiki/Color_blindness
it make them  unable to see some color (like blue , green red,)
it would be great if it was possible to add in setting a color correction menu that list the kind of color blindness
'none' , 'red-green'  ...
and for each color correction add layer to color so the final color became visible to them .
like adding colors the base color of #1258 

it not a theme by the way, since the original color are not modified (in none case) but only corrected for a 'medical' purpose

Riksu9000:
Accessibility is definitely something important to look into. This issue can also be solved by carefully designing the UI in a way which makes it understandable without even considering the colors. Do you know if some UI elements are especially problematic?

I took a look at InfiniTime with a color blindness simulator and found a few issues which I'll look into fixing. Otherwise I think InfiniTime is already quite good about this. Most objects have identifying elements in white, which is the easiest to see for everyone.

ITCactus:
it also could be a part of the Global Color Scheme #1030
i mean like in some IDEs

lman0:
@Riksu9000 color blindness make people see without the color 
and if color hold a meaning , then these people are unable the see such meaning.

explanation  (please see the website , since it show visually how each kind see the color)
https://www.colourblindawareness.org/colour-blindness/types-of-colour-blindness/

here some kind :
* Red–green color blindness :  mediacal name "deuteranopia" *
* blue–yellow color blindness :  mediacal name "Tritanomaly" *

lman0:
i found a color listing that allow to change color red to color red for deutonmaly ...
https://www.disabled-world.com/disability/accessibility/websitedesign/color-spectrum-chart.php


please create a select list that change base color to each kind like "deuteranopia"

lman0:
what is the purpose of the green color of the notification icon?
if it has a meanning  please change the icon if you plan to remove the color

Riksu9000:
To avoid having to add big features not used by the majority, let's apply [universal design](https://en.wikipedia.org/wiki/Universal_design) concepts to the UI design. I'll close this issue for now. Anyone, please open a new issue if you have trouble using InfiniTime for any reason and we'll see what we can do about it.

PR #1266, opened 2022-08-03
Improve checkbox visibility

Riksu9000:
One of the necessary changes for https://github.com/InfiniTimeOrg/InfiniTime/issues/1263.

InfiniSim_2022-08-03_164153


src/displayapp/lv_pinetime_theme.c
```diff
@@ -215,8 +215,8 @@ static void basic_init(void) {
   lv_style_set_transition_prop_6(&style_cb_bg, LV_STATE_DEFAULT, LV_STYLE_OUTLINE_OPA);
 
   lv_style_reset(&style_cb_bullet);
-  lv_style_set_outline_opa(&style_cb_bullet, LV_STATE_FOCUSED, LV_OPA_TRANSP);
   lv_style_set_radius(&style_cb_bullet, LV_STATE_DEFAULT, LV_DPX(4));
+  lv_style_set_pattern_image(&style_cb_bullet, LV_STATE_CHECKED, LV_SYMBOL_OK);
```
NeroBurner, line 219: is `LV_SYMBOL_OK` defined by lvgl? Other times when we add a new glyph it needs to be defined in multiple places. Here it only is added to `fonts.json` and can be directly used?

Riksu9000: Yes, it's defined by LVGL. In this case we couldn't include Symbols.h anyway because it's a C++ file and this is a C file, unless #1258 was merged.

NeroBurner: ah thanks for the clarification and extra info!

NeroBurner (approved):

JF002 (approved):

PR #1267, opened 2022-08-05
Tweaked issue forms

Riksu9000:
When people suggest a solution to a generic issue they encounter, they have to open a feature request, which is not optimal. Submitting an issue through this new form will require writing about the issue that led to the idea, which will make it easier to discuss different solutions.

I figured we can keep the bug form as it is working well. 

You can preview the form by viewing the file through GitHub.


JF002:
I'm wondering if the distinction between 'bug' and 'issue' will be understood by everyone. I understand we want to change the "feature request" form to encourage people thinking about why they request this feature or improvement. But will they understand they have to use this new form to suggest their idea or improvement?

On the other side... I cannot find any better word right now...

JF002:
I think I prefer these changes (#1267) over the other ones (#1286). I'm just a little bothered by the wording : in my mind, "issue", "problem", "bug" translate to more or less the same thing : something is not working as expected : the watch shouldn't crash when I do this, the doc said this should be red, but it is green,... 

Basically, I think we have 3 types of "issues" : 

- [1] real bugs ("it crashes when I do this"), 
- [2] new feature requests and improvements to existing ones (add something that does not exist yet, like support multiple alarm or add a qr code app),
- [3] "this thing does not work for my use-case, it should do this instead" (ex : I can't take a shower with my pinetime -> add a "shower mode")

For now, [2] and [3] as managed as "Feature Request". The downside of this is that users request "a rain mode" instead of describing the actual "issue" that led to this request. That's why we are trying to improve the templates : encourage the user to think about the "root cause" of their request and to explain it so we can find the best way to answer to their request.

I think this new `issue-report.yaml` is quite good with those 2 fields (introduce the issue - something lacking of suboptimal - then suggest a solution). As I said, it's just the word "issue" that bothers me. 

This is probably a (spoken/written) language issue, so I'll trust anyone with better English skills than me ;-)

Here's a suggestion: the name and description could be changed to 

name : feature or improvement request
description : Suggest a new feature or an improvement to an existing one. Use this form if you encounter an issue with InfiniTime (that is not a bug) and that could be fixed by adding new feature or by extending existing ones.

Now, if we cannot find a better wording, we can also just apply those changes and see if the overall quality of feature request improves with them.

JF002:
@Riksu9000 It looks good to me! I've just created #1495 so the users know they can file their feature requests in the Discussion board.
If you're OK with this, we can merge those 2 PRs.

JF002 (approved):

Riksu9000:
I've improved the description of these forms to make it easier to select the correct form. Let me know what you think and if the wording can be improved.

JF002:
I like the new wording, it makes a better distinction between a bug and an issue :+1:

PR #1271, opened 2022-08-10
Counter: Determine the number of digits from the max value.

Riksu9000:
Previously the minimum number of digits was fixed at two, padded with leading zeros if necessary. Now it will be determined from the max value. 

Current:
`00, 01, 02 ... 99, 100, 101 ... 999, 1000 1001`

With this change:
`0000, 0001, 0002 ... 0099, 0100, 0101 ... 0999, 1000, 1001`

This issue doesn't happen with the current uses of the Counter.

### DIFF
--- a/src/displayapp/widgets/Counter.h
+++ b/src/displayapp/widgets/Counter.h
@@ -41,6 +41,7 @@ namespace Pinetime {
         int min;
         int max;
         int value;
+        const int leadingZeroCount;
         bool twelveHourMode = false;
         bool monthMode = false;
         lv_font_t& font;
--- a/src/displayapp/widgets/Counter.cpp
+++ b/src/displayapp/widgets/Counter.cpp
@@ -17,9 +17,17 @@ namespace {
       widget->DownBtnPressed();
     }
   }
+  constexpr int digitCount(int number) {
+    int digitCount = 0;
+    while (number > 0) {
+      digitCount++;
+      number /= 10;
+    }
+    return digitCount;
+  }
 }
 
-Counter::Counter(int min, int max, lv_font_t& font) : min {min}, max {max}, value {min}, font {font} {
+Counter::Counter(int min, int max, lv_font_t& font) : min {min}, max {max}, value {min}, font {font}, leadingZeroCount {digitCount(max)} {
 }
 
 void Counter::UpBtnPressed() {
@@ -71,14 +79,14 @@ void Counter::UpdateLabel() {
     if (value == 0) {
       lv_label_set_text_static(number, "12");
     } else if (value <= 12) {
-      lv_label_set_text_fmt(number, "%.2i", value);
+      lv_label_set_text_fmt(number, "%.*i", leadingZeroCount, value);
     } else {
-      lv_label_set_text_fmt(number, "%.2i", value - 12);
+      lv_label_set_text_fmt(number, "%.*i", leadingZeroCount, value - 12);
     }
   } else if (monthMode) {
     lv_label_set_text(number, Controllers::DateTime::MonthShortToStringLow(static_cast<Controllers::DateTime::Months>(value)));
   } else {
-    lv_label_set_text_fmt(number, "%.2i", value);
+    lv_label_set_text_fmt(number, "%.*i", leadingZeroCount, value);
   }
 }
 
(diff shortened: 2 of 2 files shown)


lman0:
i want to know , i this modification mean that if the counter is left overnight , it will go over the screen ? or it will crash?

Riksu9000:
@lman0 I think you're misunderstanding what this PR is about. Please familiarize yourself with the code.

lman0:
i just read your explanation , and questioned because i thought it was accurate...
i'am wrong ? does that mean that what you write is not what you "code"?

i understood :
"Previously the minimum number of digits was fixed at two"
-> the counter was limited to 2 digit max
"Now it will be determined from the max value."
-> now the number of digit can be  1 , 2 , 3....

if the max value is 999 , then 3 digit ?
what is the max value to begin with ? 99?  999 ? no limit?

lman0:
i just read your explanation , and questioned because i thought it was accurate...
if i understand your last comment right , Does people need to be dev to read comment because the pr explanation might  don't meant what is coded?

for instance ,here ,  what i understood :
"Previously the minimum number of digits was fixed at two"
-> the counter was limited to 2 digit max
"Now it will be determined from the max value."
-> now the number of digit can be  1 , 2 , 3....
am I wrong ? 

so if the max value is 999 , then 3 digit ?
what is the max value to begin with ? 99?  999 ? no limit?

i  do remember a "calculator" app , that was above the limit of digit and crashed... because the screen app  not "coded" to contain the number of digit and created a "overflow" , so it's the reason of  my worry

for instance , i do also  remember as well  , when people searched for the explanation of a functionality that have been added 
but not explained in the doc and it was thank to the proper pr explanation that is was resolved  , 
so a proper explanation , readable by _not dev_   help _everyone_

Riksu9000:
@lman0 I've updated the description, but as this is an entirely technical change and essentially a bugfix, I would rather have feedback from devs.

JF002:
@lman0 Github **is** a developer tool, and more especially the pull request functionality : developers write changes or additions and request core-developers/maintainers to pull those changes into the main code base. 
We do this in the public space, anyone is free to read and interact in the comment, but we do **not** expect everyone to understand everything that is being done or discussed in those comments.

As a developer, I can easily see that this PR only changes the formatting of the leading 0s based on the maximum value the widget is supposed to display. The changes are quite simple and documented enough in the description. I do not need more information.

You are of course free to ask questions and speak your opinion on any pull-requests or issues in the project, but remember that we are under **no** obligation to answer all of them. Answering those questions and debating endlessly on insignificant details or on topics that you seem to not understand is very time consuming for us and counter productive, and we hide comments when we feel like we are loosing our time.

If you want to contribute to the project (and I think you really want to), please try to be a bit more organized : comment on topics where you really think you can provide additional value, do not question each and every change done by developers, think a bit more before publishing your comments to avoid multiple edits afterwards, and, if you think you have a good/better idea, try to contribute it yourself.

lman0:
@JF002  where a user like me have right to take question on functional work of a pr  , without it put off topic?
 when it's precisly  about the pr understanding?
 only on issue?

lman0:
> @lman0 I've updated the description, but as this is an entirely technical change and essentially a bugfix, I would rather have feedback from devs.

thank ! 
if only there where some tag added to pr to say it's a technical  ( not visual or app ) change so user like me not pry

src/displayapp/widgets/Counter.h
```diff
@@ -41,6 +41,7 @@ namespace Pinetime {
         int min;
         int max;
         int value;
+        int leadingZeroCount;
```
JF002, line 44: If I'm not wrong, this variable could be declared as `const`?

src/displayapp/widgets/Counter.h
```diff
@@ -41,6 +41,7 @@ namespace Pinetime {
         int min;
         int max;
         int value;
+        const int leadingZeroCount;
```
NeroBurner, line 44: I think `const` must be removed as the `leadingZeroCount` may change when `SetMax()` is called (this update is not yet implemented in `SetMax()`

NeroBurner (changes requested):

src/displayapp/widgets/Counter.h
Riksu9000: The Counter would break anyway if the number of digits was changed with SetMax(), since resizing is not implemented, so I added a comment discouraging changing the number of digits. SetMax() is currently only used to set the number of days in a month.

NeroBurner (approved):

JF002 (approved):

PR #1273, opened 2022-08-11
Disable LVGL value_str

Riksu9000:
value_str is a way to add text on a button without a separate label.
This saves having an extra label object, but I believe it's removed in LVGL8.
Disabling this saves some memory, but using value_str everywhere would probably also save some memory. value_str can make slightly cleaner code, but isn't good for future proofing. Thoughts?


JF002:
Well if `value_str` is removed in LVGL8, it might make sense to remove it in case we want to update LVGL in the future.
Regarding memory usage, it could be interesting to check how `value_str` is implemented in lv_button : they might instanciate a `lv_label` under the hood.

JF002 (approved):

PR #1279, opened 2022-08-13
Merge CI to deduplicate events

Riksu9000:
We can have multiple checks in one file as different jobs that get triggered with the same events. The main workflow should run when there are changes that may affect the builds.

Do you think the code formatting check should also be moved here? It has slightly different triggering events, but can be tweaked to work. (Maybe there isn't a good reason to do so)


JF002:
I think it makes sense that building the firmware, the simulator and checking the code formatting should be triggered at the same time, yes. So merging those 3 workflows in 1 totally make sense. I'll ask  @Avamander and @NeroBurner opinion as (I think) they wrote those 2 additional workflows ;-)

NeroBurner (approved):
as long as we see with the checks which of the jobs failed I'm OK with the change

Riksu9000:
I'll merge this as I think we always want to build the firmware and the simulator at the same time.

PR #1282, opened 2022-08-14
Move all images to doc. Remove old images.

Riksu9000:
Concentrate doc related things in the doc folder. Remove old photos. If the old photos were to be kept, a history document should be written that displays them.


Avamander (approved):

NeroBurner (approved):

NeroBurner:
As the images are added by @JF002 I'd like to ask him directly if those pictures should be kept

JF002 (approved):

JF002:
> As the images are added by @JF002 I'd like to ask him directly if those pictures should be kept

Yes, I added those pictures at some point, but they are not needed anymore, and the code repository is certainly not the best place to store them ;)

PR #1284, opened 2022-08-15
Fix markdown format with autoformatters

Riksu9000:
I'm testing some autoformatters. They all seem to be opinionated and quirky. Here I've compiled the less opinionated changes to begin with.

Remove whitespace at the end of lines.
Remove leading space from unordered lists.
Enforce two newlines between sections.
Always use `-` for unordered lists.
Slightly tweaked tables.
Indentation fixes.
Escape special characters where necessary.


Itai-Nelken:
This doesn't seem to be a problem in the files changed, but 2 spaces at the end of a line mean a newline, so they might have been put intentionally.

JF002:
I'm OK with those changes, they'll make our docs more consistent. We'll probably want to specify those conventions in the documentation

Did you run a tool to make those changes automatically? If yes, I think it could be a good idea to document how to use it, provide a configuration file (like we did for clang-format/tidy), and maybe add it in the format workflow for Github Actions.

Riksu9000:
@JF002 Absolutely. I was initially working on adding a markdown formatter check, but none of the formatters I tested produced the exact format I was expecting so I need to experiment some more. These are cherry picked changes which guidelines and autoformatters agree with, so we can focus on the small details later.

bootloader/README.md
```diff
@@ -50,72 +54,67 @@ New firmware does not run properly but sets the valid bit and refreshes the watc
...
+`program image.bin 0x8000`
 
 ## OTA and DFU
+
 Pack the image into a .zip file for the NRF DFU protocol:
 
-`
-adafruit-nrfutil dfu genpkg --dev-type 0x0052 --application image.bin dfu.zip
-`
+`adafruit-nrfutil dfu genpkg --dev-type 0x0052 --application image.bin dfu.zip`
```
NeroBurner, line 108: I'd use a codeblock here with `sh` style to get syntax highlighting (can't do the suggestion properly because the \`\`\` are used for the codeblock and for the github suggestion)

```suggestion
```sh
adafruit-nrfutil dfu genpkg --dev-type 0x0052 --application image.bin dfu.zip
```
```

File: bootloader/README.md [hunk 2]
```diff
@@ -50,72 +54,67 @@ New firmware does not run properly but sets the valid bit and refreshes the watc
...
 
 Use NRFConnect or dfu.py (in <project root>/bootloader/ota-dfu-python) to upload the zip file to the device:
 
-`
-sudo dfu.py -z /home/jf/nrf52/bootloader/dfu.zip -a <pinetime MAC address> --legacy
-`
+`sudo dfu.py -z /home/jf/nrf52/bootloader/dfu.zip -a <pinetime MAC address> --legacy`
```
NeroBurner, line 112: I'd use a codeblock here with `sh` style to get syntax highlighting (can't do the suggestion properly because the \`\`\` are used for the codeblock and for the github suggestion)

```suggestion
```sh
sudo dfu.py -z /home/jf/nrf52/bootloader/dfu.zip -a <pinetime MAC address> --legacy
```
```

File: bootloader/ota-dfu-python/README.md [hunk 3]
```diff
@@ -1,13 +1,13 @@
 # Python nRF5 OTA DFU Controller
 
 So... this is my fork of dingara's fork of astronomer80's fork of
-foldedtoad's Python OTA DFU utility. 
+foldedtoad's Python OTA DFU utility.
 
-My own contribution is little more than a brute force conversion to 
-python3. It is sparsely tested so there are likely to be a few 
+My own contribution is little more than a brute force conversion to
```
NeroBurner, line 6: I've always kept to the "one sentence one line" rule for markdown. Makes diffs MUCH more readable and formatting won't mess with a small change

NeroBurner (changes requested):
While changing the docs, can we also use the "one sentence one line" rule for markdown. It makes the diffs much more readable, as a single word change won't lead to a whole paragraph diff as the text flows down to the next line

For GitHub we'd then need two spaces at the end of the line to keep the sentences in one paragraph (on GitLab one would need to put a blank line in between sentences to create a new paragraph)

bootloader/ota-dfu-python/README.md
Riksu9000: I'm inclined to agree, but none of the formatters I tested so far support this. We'll have to discuss this later. This PR isn't intended to fix everything, but only the most obvious issues.

doc/ble.md
```diff
@@ -70,34 +74,37 @@ When the service does not exist in the BLE specification, InfiniTime implements
...
+  - Call characteristic (extension to the Alert Notification Service): 00020001-78fc-48fe-8e23-433b3a1942d0
 
 - Since InfiniTime 1.7:
-    * [Motion Service](MotionService.md): 00030000-78fc-48fe-8e23-433b3a1942d0
 
+  - [Motion Service](MotionService.md): 00030000-78fc-48fe-8e23-433b3a1942d0
 
 - Since InfiniTime 1.8:
-    * [Weather Service](/src/components/ble/weather/WeatherService.h): 00040000-78fc-48fe-8e23-433b3a1942d0
+  - [Weather Service](/src/components/ble/weather/WeatherService.h): 00040000-78fc-48fe-8e23-433b3a1942d0
```
NeroBurner, line 94: are the added newlines necessary? I think keeping the enumerations together would be rendered just fine (less space between bullet points in the final render). For example:

```md
- Since IT 0.1.2
  - some service: `000-111`
- Since IT 0.1.2
  - some service: `000-111`
```
produces:

- Since IT 0.1.2
  - some service: `000-111`
- Since IT 0.1.2
  - some service: `000-111`

Whereas an example with spaces between them

```md
- Since IT 0.1.2

  - some service: `000-111`

- Since IT 0.1.2

  - some service: `000-111`
```
produces:

- Since IT 0.1.2

  - some service: `000-111`

- Since IT 0.1.2

  - some service: `000-111`

---

and while we're at it I recommend adding backticks to the BLE Service IDs as done in my examples above

doc/gettingStarted/about-software.md
```diff
@@ -6,9 +6,10 @@ A **firmware** is software running on the embedded hardware of a device.
 
 InfiniTime has three distinct firmwares:
 
- - **[InfiniTime](https://github.com/InfiniTimeOrg/InfiniTime)** is the operating system.
- - **[The bootloader](https://github.com/JF002/pinetime-mcuboot-bootloader)** is responsible for safely applying firmware updates and runs before booting into InfiniTime.
- - **The recovery firmware** is a special *application firmware* than can be loaded by the bootloader on user request. This firmware can be useful in case of serious issue, when the main application firmware cannot perform an OTA update correctly.
+- **[InfiniTime](https://github.com/InfiniTimeOrg/InfiniTime)** is the operating system.
+- **[The bootloader](https://github.com/JF002/pinetime-mcuboot-bootloader)** is responsible for safely applying firmware updates and runs before booting into InfiniTime.
+
```
NeroBurner, line 11: why is this empty line added? I think it can be removed

```suggestion
```

NeroBurner (approved):
two minor suggestions, after that I'm very happy with this PR. Thank you for making the docs much more coherent!

doc/ble.md
Riksu9000: Adding the newlines didn't change the rendering, because there were empty lines in between. I'd rather leave it rendering the same way for now.

```
- Since IT 0.1.2
  - some service: `000-111`

- Since IT 0.1.2
  - some service: `000-111`
```

JF002 (approved):
I'm OK with those changes. I think it would be a good idea to list those formatting rules in `coding-convientions.md` because I'll probably be the first one to forget about them ;-)

PR #1287, opened 2022-08-16
Clang tidy check

Riksu9000:
Added a new check that displays clang-tidy warnings for changed files. This check isn't blocking, because a lot of warnings exist in the codebase and the rules may need to be tweaked, but can be used for reference. I've added some fixes that trigger the check for demonstration.

I think we should work to fix these warnings or disable them if they're not important. These warnings aren't necessarily about fixing bad code, but rather consistent practices, so silencing or switching some of them isn't an issue. Let me know what you think.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -17,9 +17,8 @@ jobs:
         with:
           fetch-depth: 1000
 
-      - name: Configure git
-        run: |
-          git fetch origin "$GITHUB_BASE_REF":"$GITHUB_BASE_REF" --depth=1000
+      - name: Fetch base branch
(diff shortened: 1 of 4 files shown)


Avamander:
This has always been the end intention, that ideally a new PR would see if any additional warnings are generated and the author has a chance to fix them.

Riksu9000:
@JF002 Yes. They fix a few warnings and trigger the check on those files, so we can inspect it.

JF002:
Ok, great! And I've just noticed that the PR also adds a new github workflow to run clang-tidy :+1: 
This could be extended in the future to export the warning messages in a build artifacts, a bit like it's done in the clang-format workflow!

JF002 (approved):

PR #1288, opened 2022-08-16
Update and improve the bootloader documentation 

JF002:
Improve the documentation of the bootloader, add background information, info on design choices, remove outdated content.

Move the doc from /bootloader to doc/bootloader and remove outdated/unmaintained files from /bootloader.


Riksu9000 (changes requested):
Great cleanup.

Maybe this document should be in the bootloader repo and we simply link to it from here?

~~I propose we split paragraphs using semantic linefeeds https://rhodesmill.org/brandon/2012/one-sentence-per-line/. This will make diffs much simpler.~~ Rather this is what I maybe would've done, but maybe we can discuss this later when we decide on a markdown format.

doc/bootloader/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+- The application firmware will be flashed at **adress 0x8000 of the internal flash**. It must then be **linked** with a start adress of 0x8000 (instead of 0x00 by default).
+- The bootloader starts the **watchdog** just before loading the application firmware. The application firmware must **refresh the watchdog regularly**. If the watchdog is not refreshed in time, the microcontroller will reset, and the bootloader will consider the firmware as unfunctionnal and will revert to the previous one.
+- The application firmware must implement an **OTA procedure which copies the new firmware at offset 0x040000 of the external SPI flash memory**.
+- The application firmware must set the valid bit in flash memory (write `1` at offset 0x7BFE8 of the internal flash memory) to validate the installation. If this bit is not set, the bootloader will revert to the previous version at next reset.
+
+Those are basically the only constraints imposed on developers. They are free to use any language, framework, build system, BLE stack and communication protocol, as long as they implement those functionalities.
+
+### The OTA implementation
```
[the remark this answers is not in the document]
JF002: You are right! At the beginning, I started this document as a blog post to explain the design choices and the implementation of the OTA in InfiniTime. I changed my mind later on and decided to include that doc in the project documentation.
I'll rework this part to make it more generic.

JF002:
> Great cleanup.

Thanks :)

> Maybe this document should be in the bootloader repo and we simply link to it from here?

You're probably right. There was already some documentation in InfiniTime regarding the bootloader, but it makes more sense to move it to the bootloader project. I suggest we continue the review in this PR, and more it later on?

> ~I propose we split paragraphs using semantic linefeeds https://rhodesmill.org/brandon/2012/one-sentence-per-line/. This will make diffs much simpler.~ Rather this is what I maybe would've done, but maybe we can discuss this later when we decide on a markdown format.

Mhm this seems like a good idea. I'll give it a shot and see how it looks like ;)

doc/bootloader/README.md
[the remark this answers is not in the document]
JF002: I re-read the document : couldn't we just remove those sections (`The OTA implementation` and `Firmware validation`) as the section just above explains what needs to be done in the application firmware to integrate with the bootloader?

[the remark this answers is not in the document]
Riksu9000: You're right. I think these sections can be removed.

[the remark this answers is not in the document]
JF002: I removed `The OTA implementation` and moved `Firmware validation` to the getting started user guide :)

doc/bootloader/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+Here is basically how it works:
+
+- the application firmware exposes a BLE "firmware update" service. In InfiniTime we decided to implement the **nRF DFU** protocol.
+- this update service stores the new firmware in a temporary location in the *external flash memory* and checks the integrity of the data (CRC).
+- when the new firmware is completely transferred and checked, the watch is rebooted and the bootloader is ran again. The bootloader detects that a new firmware is available in the *external* flash memory and safely swaps the current firmware in the *internal* memory with the new one from the *external* memory.
+- When the swapping is done and checked, the new firmware is executed from the *internal* memory.
+- If the new firmware does not properly run, the bootloader will automatically revert to the older firmware by swapping again firmwares from the external to the internal memory.
```
NeroBurner, line 44: reorder for readability flow

```suggestion
- If the new firmware does not properly run, the bootloader will automatically revert to the older firmware by again swapping firmwares from the external to the internal memory.
```

doc/bootloader/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+This procedure involves multiple software modules and rely on some hardware features. Let's go into more details!
```
NeroBurner, line 46: I suggest one line per sentence, less messy diffs

```suggestion
This procedure involves multiple software modules and rely on some hardware features.
Let's go into more details!
```

I love the enthusiasm by the way!

doc/bootloader/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Memory map
+
+Multiple memories are available in the PineTime : the internal RAM (64KB) and flash (512KB) memories from the nRF52832 and an external flash memory (4MB) connected to the SPI bus.
```
NeroBurner, line 50:
```suggestion
Multiple memories are available in the PineTime:
the internal RAM (64KB) and flash (512KB) memories from the nRF52832 and an external flash memory (4MB) connected to the SPI bus.
```

doc/bootloader/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+RAM memory is not relevant here. Here is how we organized both flash memories:
```
NeroBurner, line 52: I suggest one line per sentence, less messy diffs

```suggestion
RAM memory is not relevant here.
Here is how we organized both flash memories:
```

doc/bootloader/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+![memory map](memoryMap.jpg)
+
+The internal flash memory contains the bootloader, a log segment (not used), the application firmware and a scratch area. This area is used by the bootloader when *swapping* firmware from the external memory.
```
NeroBurner, line 56: Maybe we could use "swap area" instead of "scratch area", as I don't know what a "scratch area" is

```suggestion
The internal flash memory contains the bootloader, a log segment (not used), the application firmware and a swap area.
This area is used by the bootloader when *swapping* firmware from the external memory.
```

doc/bootloader/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+The SPI flash memory contains a recovery firmware and an area big enough to store a whole firmware (the new firmware will be first written to this area before being swapped into the internal memory). The remaining space is available for the application firmware needs.
```
NeroBurner, line 58: I suggest one line per sentence, less messy diffs

```suggestion
The SPI flash memory contains a recovery firmware and an area big enough to store a whole firmware (the new firmware will be first written to this area before being swapped into the internal memory).
The remaining space is available for the application firmware needs.
```

doc/bootloader/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+### The bootloader
+[The bootloader](https://github.com/InfiniTimeOrg/pinetime-mcuboot-bootloader) is based on [MCUBoot, a secure bootloader for 32 bits microcontrollers](https://github.com/mcu-tools/mcuboot). It was originally written by [Lup Yuen](https://github.com/lupyuen).
```
NeroBurner, line 61: I suggest one line per sentence, less messy diffs

```suggestion
[The bootloader](https://github.com/InfiniTimeOrg/pinetime-mcuboot-bootloader) is based on [MCUBoot, a secure bootloader for 32 bits microcontrollers](https://github.com/mcu-tools/mcuboot).
It was originally written by [Lup Yuen](https://github.com/lupyuen).
```

doc/bootloader/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+MCUBoot is run at boot time. In normal operation, it just jumps to the reset handler of the application firmware to run it. Once the application firmware is running, MCUBoot does not run at all.
```
NeroBurner, line 63: I suggest one line per sentence, less messy diffs

```suggestion
MCUBoot is run at boot time.
In normal operation, it just jumps to the reset handler of the application firmware to run it.
Once the application firmware is running, MCUBoot does not run at all.
```

doc/bootloader/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+![MCUBoot boot sequence diagram](boot.png)
+
+But MCUBoot does much more than that : it can upgrade the firmware that is currently running by a new one, and it is also able to revert to the previous version of the firmware in case the new one does not run properly.
```
NeroBurner, line 67: replace `by a new one` with `with a new one`
```suggestion
But MCUBoot does much more than that:
it can upgrade the firmware that is currently running with a new one, and it is also able to revert to the previous version of the firmware in case the new one does not run properly.
```

doc/bootloader/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+To do this, it uses 2 memory 'slots' :
+- **The primary slot** (located in the internal flash memory of the nRF52832) : it contains the current firmware, the one that will be executed by MCUBoot
+- **The secondary slot** (located in the external SPI flash memory) : it is used to store the upgraded version of the firmware, when available.
+
+At boot time, MCUBoot detects that a new version of the firmware is available in the secondary slot and swaps them : the current version of the firmware is copied from the primary to the secondary slot and vice-versa.
```
NeroBurner, line 73:
```suggestion
At boot time, MCUBoot detects that a new version of the firmware is available in the secondary slot and swaps them.
The current version of the firmware is copied from the primary to the secondary slot and vice-versa.
```

doc/bootloader/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+When the swapping is done, it runs the new version of the firmware from the primary slot.
+
+![MCUBoot update sequence diagram](upgrade.png)
+
+The next time MCUBoot will be run (after an MCU reset), MCUBoot will check if the new firmware ran correctly (it must set a flag in memory). If it is not the case, it'll revert to the previous version of the firmware by copying it from the secondary to the primary.
```
NeroBurner, line 79: Too many `it` one after another, be more explicit

```suggestion
The next time MCUBoot will be run (after an MCU reset), MCUBoot will check if the new firmware ran correctly (it must set a flag in memory).
If it is not the case, the MCUBoot will revert to the previous version of the firmware by copying it from the secondary to the primary.
```

doc/bootloader/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+![MCUBoot update sequence diagram](recover.png)
+
+You'll find more information about MCUBoot, its design and how it works in [the documentation of the project](https://docs.mcuboot.com/).
+
+The [pinetime-mcuboot-bootloader](https://github.com/InfiniTimeOrg/pinetime-mcuboot-bootloader) integrates MCUBoot, provides low-level drivers for the display and SPI flash memory and implements a very basic UI. See [the documentation of the project for more details about this](https://github.com/InfiniTimeOrg/pinetime-mcuboot-bootloader#boot-flow).
```
NeroBurner, line 85:
```suggestion
The [pinetime-mcuboot-bootloader](https://github.com/InfiniTimeOrg/pinetime-mcuboot-bootloader) integrates MCUBoot, provides low-level drivers for the display and SPI flash memory and implements a very basic UI.
See [the documentation of the project for more details about this](https://github.com/InfiniTimeOrg/pinetime-mcuboot-bootloader#boot-flow).
```

doc/bootloader/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+An application firmware must implement the following functionalities to be compatible with our bootloader:
+
+- The application firmware will be flashed at **address 0x8000 of the internal flash**. It must then be **linked** with a start address of 0x8000 (instead of 0x00 by default).
```
NeroBurner, line 89:
```suggestion
- The application firmware will be flashed at **address 0x8000 of the internal flash**.
  It must then be **linked** with a start address of 0x8000 (instead of 0x00 by default).
```

doc/bootloader/README.md
```diff
@@ -0,0 +1,113 @@
...
+- The bootloader starts the **watchdog** just before loading the application firmware. The application firmware must **refresh the watchdog regularly**. If the watchdog is not refreshed in time, the microcontroller will reset, and the bootloader will consider the firmware as unfunctional and will revert to the previous one.
```
NeroBurner, line 90: suggesting `nonfunctional` from my computer dictionary, and `micro-controller`

```suggestion
- The bootloader starts the **watchdog** just before loading the application firmware.
  The application firmware must **refresh the watchdog regularly**.
  If the watchdog is not refreshed in time, the micro-controller will reset, and the bootloader will consider the firmware as nonfunctional and will revert to the previous one.
```

doc/bootloader/README.md
```diff
@@ -0,0 +1,113 @@
...
+- The application firmware must implement an **OTA procedure which copies the new firmware at offset 0x040000 of the external SPI flash memory**.
+- The application firmware must set the valid bit in flash memory (write `1` at offset 0x7BFE8 of the internal flash memory) to validate the installation. If this bit is not set, the bootloader will revert to the previous version at next reset.
```
NeroBurner, line 92:
```suggestion
- The application firmware must set the valid bit in flash memory (write `1` at offset 0x7BFE8 of the internal flash memory) to validate the installation.
  If this bit is not set, the bootloader will revert to the previous version at next reset.
```

doc/bootloader/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+Those are basically the only constraints imposed on developers. They are free to use any language, framework, build system, BLE stack and communication protocol, as long as they implement those functionalities.
```
NeroBurner, line 94:
```suggestion
Those are basically the only constraints imposed on developers.
They are free to use any language, framework, build system, BLE stack and communication protocol, as long as they implement those functionalities.
```

doc/bootloader/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Recovery firmware
+
+Together with the bootloader, we also built a [recovery firmware](https://github.com/InfiniTimeOrg/pinetime-mcuboot-bootloader#recovery-firmware). This firmware, which implements the bare minimum to support the OTA procedure (basic UI, BLE stack and DFU protocol) can be used in last resort, in case the application firmware does not run at all, or does not support the OTA procedure.
```
NeroBurner, line 98: `as a last resort` sounds better in my ears

```suggestion
Together with the bootloader, we also built a [recovery firmware](https://github.com/InfiniTimeOrg/pinetime-mcuboot-bootloader#recovery-firmware).
This firmware, which implements the bare minimum to support the OTA procedure (basic UI, BLE stack and DFU protocol) can be used as a last resort, in case the application firmware does not run at all, or does not support the OTA procedure.
```

doc/bootloader/README.md
```diff
@@ -0,0 +1,113 @@
...
+Data got corrupted during file transfer | Application firmware does a CRC check before applying the update, and does not proceed if it fails.
+New firmware does not run at all (bad file) | MCU executes random instructions and will most likely end up in an infinite loop or freeze in an error handler. The watchdog will reset the MCU after ~7s because the firmware does not refresh it. Bootloader reverts to the previous version of the firmware after reset.
+New firmware runs, does not set the valid bit and does not refresh the watchdog | The watchdog reset the MCU after ~7s and the bootloader reverts to the previous firmware.
+New firmware does not run properly, does not set the valid bit but refreshes the watchdog | Reset the watch if the application firmware provides any way to trigger a softreset. If not, wait for the battery to drain completely before charging the watch again. The bootloader will revert the firmware as soon as the watch is powered again.
+New firmware does not run properly but sets the valid bit and refreshes the watchdog | Reset the watch, or drain the battery and load the recovery firmware when the bootloader is running again.
+
+
+
```
NeroBurner, line 113: one newline at the end of the file is enough :)

```suggestion
```

doc/bootloader/README.md
FintasticMan: I think it should be `microcontroller`. See [this Wikipedia article](https://en.wikipedia.org/wiki/Microcontroller).

doc/bootloader/README.md
agittins: Perhaps to make the structure of the list easier to digest:
```suggestion
Multiple memories are available in the PineTime :
- 64KB of internal RAM on the nRF52832 microcontroller
- 512KB of internal flash memory on the nRF52832
- 4MB of external flash memory, which is connected to the SPI bus.
```

doc/bootloader/README.md
agittins: Perhaps starting with "We divide the internal flash memory into three areas..." would make it clear that this part of the layout is not inherent to the device, and reflects a design decision made by the project. Up to now the things being described are external constraints, and continuing with "this contains that" makes it sound like it wasn't up to [us].

doc/bootloader/README.md
agittins: Since this is a list of requirements for the application firmware, perhaps list the requirement itself first, then state the rationale. eg:
```suggestion
- It must be **linked** with a start address of 0x8000 (instead of the default of 0x00).
  This is because the first 32KB of internal flash is reserved for the bootloader, so the application firmware image will be copied to the internal flash at an **offset address** of 0x8000.
```

doc/bootloader/README.md
agittins: Switch to requirement first, rationale second. Explicitly state what will be reverted.

Question: Does the bootloader take into account the reboot reason, or is it just because the validation bit isn't set that it would revert the image? Would the firmware reversion still happen if the WDT fired some time after the validation bit was set?

Assuming it's just about the WDT firing before the validation is set:

```suggestion
- The application firmware must **refresh the watchdog regularly**.
  The bootloader starts the watchdog just before loading the application firmware.
  If the watchdog is not refreshed in time, the microcontroller will reset.
  After the reset the bootloader will notice that the validation bit has not been set, and will revert to the previous application firmware image.
```

doc/bootloader/README.md
agittins: Would it be better to call it "the validation bit", or perhaps "validation flag"? Saying "the valid bit" is confusing because a zero could be "the valid bit". Also using the term "bit" rather than "byte" suggests we should be more explicit about what needs to be written - would a byte equal to `0x01` be suitable, or is that byte really a bitmask of multiple flags, and they need to `OR` that location with `0x01`?

Replace version with application firmware for clarity.

Assuming a dedicated byte to contain 0x01:
```suggestion
- The application firmware must set the validation bit by **writing `0x01` at offset 0x7BFE8 of the internal flash memory** to validate the installation.
  If this bit is not set, the bootloader will revert to the previous application firmware at next reset.
```

doc/bootloader/README.md
agittins: Agreed re `as a last resort` and also suggest removing the comma to make the meaning a bit clearer.

```suggestion
This firmware, which implements the bare minimum to support the OTA procedure (basic UI, BLE stack and DFU protocol) can be used as a last resort in case the application firmware does not run at all, or does not support the OTA procedure.
```

doc/gettingStarted/updating-software.md
```diff
@@ -33,13 +33,25 @@ We have prepared instructions for flashing InfiniTime with Gadgetbridge and NRFC
 
 Firmware updates must be manually validated. If the firmware isn't validated and the watch resets, the watch will revert to the previous firmware. This is a safety feature to prevent bricking your device with faulty firmware.
 
-You can validate your updated firmware on InfiniTime >= 1.0 by following this simple procedure:
+If the new firmware is working correctly, open the settings and select the "Firmware" entry. This setting page displays the version of the firmware that is currently running, and allows you to validate the firmware, or reset the device to rollback to the previous version.
```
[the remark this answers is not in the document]
agittins: I'm more used to seeing it expressed as "roll back" in this context, being the verb form might be less confusing than the noun-form.

```suggestion
If the new firmware is working correctly, open the settings and select the "Firmware" entry.
This settings page displays the version of the firmware that is currently running, and allows you to validate the firmware, or reset the device to roll back to the previous version.
```

Issue #787, opened 2021-10-27
Step counter doesn't start on new day


coneng96:
### 

- [X] I searched for similar bug reports and found none was relevant.

### What happened?

Step counter doesn't start counting after 12AM unless watch is awaken

### What should happen instead?

Even if you don't wake the watch it should start counting.

### Reproduction steps

- Setup your watch to wake up with single or double tap
- Wait for time to pass 12:00AM
- Without waking up the watch (either by single, double or pressing the button), start walking
- After few steps wake the watch and see if the steps counter is still zero.

### More details?

The counter will work if you have the wake up setting to Raise Wrist.

### Version

v1.6.0

### Companion app

n/a

pswilde:
I experience this issue as well.
It works with raise wrist wake as it's likely the action of putting the watch on is triggering this motion. I have that feature disabled and can be 3000 steps into a walk before I check the time.
Not a vital issue, but one that would be good to fix.

drgnomage:
Experiencing the exact same issue, had to turn off raise to wake so my battery life was more than 2 days and began noticing this.

ncartron:
hum i'm not using any of "single tap/double tap/raise wrist" wake up methods, and my steps are properly counted (e.g. if i wake up  at 2 am to drink water, then go back to bed, when i wake up and check time, i can see like 100 or 200 steps.

are you saying that the problem is with single/double tap wake up methods?

drgnomage:
@ncartron I've found that if the bluetooth connection is up then it works as normal but if it's disconnected then it won't count steps until you wake the screen.

pshem:
I'm impacted too, on version 1.8.0. The same code that resets the step count should probably initialize it, but I don't have an unsealed version of InfiniTime to experiment with.

efuentes25:
I will be working on a fix for this.

ljr55555:
I have the same problem -- InfiniTime version 1.13.0 with a build date of Jun 24, 2023. I guess I'll try turning on the wrist raise to wake because sometimes it's midday before I notice I've got zero steps.

PR #1291, opened 2022-08-18
Solution to Step Counter not counting upon new day if wake up mode is Single Tap or Double Tap

aretinoco:
Addresses #787
Worked on with @efuentes25 

Creates a new function that wakes up the watch at midnight without lighting the screen so that it can immediately start counting steps on a new day.


efuentes25 (approved):

Riksu9000:
https://github.com/InfiniTimeOrg/InfiniTime/blob/ef6ed5d26e9c75017d51fdf889dfe991e63cca4e/src/systemtask/SystemTask.cpp#L471 It is not a good idea to implement a hack to work around this line.

PR #1293, opened 2022-08-21
Digital watchface seconds instead of colon

mashuptwice:
Shows seconds instead of colon seperator.

InfiniSim_2022-08-21_123405

Please be gentle with my code, this is the first bigger project I am working on.

Currently the label objects are split into hh, mm, s1, and s2.
The fine-positioning of the clocks elements and fixing the position of the 12H digits is done with pixel offsets.


FintasticMan:
I personally quite like this change, but I think it should definitely not be the default. I think that it could be behind a settings menu like PTS has.

mashuptwice:
@FintasticMan good point, that would make it dependent on #1028 or similar.

PR #1295, opened 2022-08-22
Change uknown incoming call from null to unknown

tomechio:
When using Gadgetbridge as a companion app, Infinitime shows that unknown call is coming from "null". 
This modification changes it to "Unknown"

image

image


Riksu9000:
I think this is a Gadgetbridge issue. Why not fix it there?

tomechio:
Hi,

I opened an issue six months ago, but no reaction from their side.
https://codeberg.org/Freeyourgadget/Gadgetbridge/issues/2604

Riksu9000:
They haven't closed it. There just hasn't been a contributor willing to fix it. Unless there's a good reason not to, this should be fixed in Gadgetbridge.

Issue #1114, opened 2022-05-02
Build Firmware with Optional Apps


boredom101:
### Verification

- [X] I searched for similar feature request and found none was relevant.

### Pitch us your idea!

Ability to choose which apps are included in the firmware

### Description

I think it would be useful if there was a way to choose which apps are included in your build.
I might try this myself if there is interest.

Jasperki:
I would vote for this.
Especialy the possibility to remove some, like pong or paint.

Riksu9000:
It's already possible to disable apps in a build simply by editing `ApplicationList.cpp` and replacing `Apps::(appname)` with `Apps::None` and reorganizing the arrays.

Currently InfiniTime includes all apps by default, so I'm guessing the suggestion is that we maintain more apps which wouldn't be in the official build, but can be swapped in during a build @boredom101?

Riksu9000:
If you're going to be tinkering with it, then I'm sure you can figure out a way to remove those apps from the build to free up some space, even without this being a supported feature. InfiniTime also isn't limited to 12 app slots. More screens can be added. Just to be clear.

GIJack:
this would be great. de-clutter the UI, just use the apps you need/want

mark9064:
This has been implemented with -DENABLE_USERAPPS :) See #1894

PR #1296, opened 2022-08-23
Build firmware with optional apps

minacode:
Please see #1408. It is a smaller implementation of this.

This PR implements #1114 .
It adds a config file `apps.txt` that contains a list of apps that should be included in the firmware. Their order is respected in the UI.
It also adds `tools/apps_prebuild.py`.

All you need to do is to build the docker container, change `apps.txt` and then build the firmware.
I hope, that this makes it possible to accept more apps, even if they are not included in the default build.

### Other implementation strategies
In this version there is a dict in the python script that contains the paths to all relevant source files for each app. We could automate this, by moving each app to its own directory like `src/displayapp/apps/alarm`. Each of those directories may also contain a meta-data file (that contains the ui-symbol for example).
The script can then just build all files within an apps directory when the app appears in `apps.txt`.

The same thing as for apps could be done for watch faces, but this is too much for this PR.

### Support
I propose that we test and maintain a default sample of apps similar to the ones that are present right now as well as a build with all apps (just for testing) and fix all eventual bugs for custom builds when they are reported.

How do we accept new apps when they are not part of the default apps, but still need to be tested on the hardware at least once?

### Todo
- [x] save .h/.cpp paths and class names in generation script
- [x] insert paths in `CMakeLists.txt`
- [x] insert case in `displayapp/DisplayApp.cpp`
- [x] insert value in `displayapp/Apps.h`
- [x] runs well on hardware


Riksu9000:
This isn't actually enough to remove the apps from the build, so the build size doesn't decrease and there isn't more space for other apps. This basically extracts the array from ApplicationList.h to a text file, so it's supposedly less intimidating to edit for non developers, but I don't think it's necessary and editing ApplicationList.h directly is hardly more difficult. We can add a document that instructs people how to hide apps, if that's really what they want.

minacode:
My thought was, that something really user friendly like itd or siglo could build a UI on top of that file interface. Do you know what else must be modified to cut apps out?

minacode:
It is very hacky, but my last commit builds a smaller image sucessfully (after running the Python script, so the automatic builds here fail). 
I know that @JF002 is working on a more convenient solution, but maybe this PR is a good itermediate step. What are your thoughts on this approach?

minacode:
After the last update, I got this image size when disabling all apps:
```
[ 88%] Linking CXX executable pinetime-mcuboot-app-1.10.0.out
           FLASH:      342564 B       480 KB     69.69%
             RAM:       54224 B        64 KB     82.74%
```

minacode:
I got the last version to build with zero apps and run without a problem on my watch.

minacode:
I am closing this one in favor of #1408.

PR #1297, opened 2022-08-25
Allow manual trigger of CI workflow

mashuptwice:
#1229 introduced a manual trigger for the workflows.
With #1279 the functionality  was removed again.

As this can be a useful feature for running workflows for other branches than develop, this PR reintroduces the workflow_dispatch trigger.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,9 @@ on:
     branches: [ develop ]
     paths-ignore:
       - 'doc/**'
-
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+    
 jobs:
   build-firmware:
     runs-on: ubuntu-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,6 +11,7 @@ on:
     paths-ignore:
       - 'doc/**'
   # Allows you to run this workflow manually from the Actions tab
+  # this can be useful for running workflows for other branches than develop
   workflow_dispatch:
     
 jobs:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,8 +10,8 @@ on:
     branches: [ develop ]
     paths-ignore:
       - 'doc/**'
-  # Allows you to run this workflow manually from the Actions tab
-  # this can be useful for running workflows for other branches than develop
+  # Allows to manually run the workflow from the Github Actions tab, optionally on another branch
+  #More information: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/
   workflow_dispatch:
     
 jobs:


.github/workflows/main.yml
```diff
@@ -10,7 +10,9 @@ on:
     branches: [ develop ]
     paths-ignore:
       - 'doc/**'
-
+  # Allows you to run this workflow manually from the Actions tab
```
Riksu9000, line 13: The comment should include the reasoning behind adding this rather than explaining what it means.

Riksu9000 (changes requested):

mashuptwice:
I've just kept the original two lines from the following commit. I don't understand why there are changes necessary if the same two lines were approved and merged only 7 weeks ago, with the exact same comment and similar reasoning:


> commit 6303bd22a577a78699247f3cbaad53bf26275399
> Author: Finlay Davidson <finlay.davidson@coderclass.nl>
> Date:   Sun Jul 10 16:46:24 2022 +0200
> 
> Fix Docker workflow not pushing on manual trigger
>     
>     Also fix issue where workflow doesn't run on pull requests.
>     Also allow other workflows to be triggered manually.
>     Also don't trigger any workflows on pull requests to master, as
>     there shouldn't be any pull requests against master.

 @Riksu9000 could you elaborate why the functionality needed to be removed in the first place?

Riksu9000:
I never approved the change and I didn't know what the purpose of this line was because there was no reasoning. These workflows are supposed to be triggered by events, so I didn't think it was necessary to manually trigger it.

mashuptwice:
@Riksu9000 understandable. Do you think [aee88e2](https://github.com/InfiniTimeOrg/InfiniTime/pull/1297/commits/aee88e2fcdad97196d2776ea52b66d85a155d81c) adds enough reasoning to prevent removal in the future?

JF002:
I used the manual trigger functionality when I was testing the new Docker image the workflow that builds and pushes the images in Dockerhub. It allows to manually run a build without having to do actual changes in the code. It should not be used most of the time, though.

NeroBurner:
@JF002 could you clarify your statement in respect to the current PR. Is it a "won't merge" or "we can merge it but we won't use it"?

Riksu9000 (approved):
The way this works is that when the default branch has a workflow containing `workflow_dispatch:`, a button appears in the Actions tab that allows running the workflow manually, but using the workflow file from another branch.

As far as I know, it isn't enough for just the feature branch to contain this. I added this to the default branch in my fork to be able to test workflows.

.github/workflows/main.yml
```diff
@@ -10,7 +10,10 @@ on:
...
+  # this can be useful for running workflows for other branches than develop
```
JF002, line 14: My understanding is that the main goal of `worklow_dispatch` is to allow manually triggering the workflow again. I even didn't know that we could use it to run the workflow on another branch.

See https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/.

I would rephrase this : "Allows to manually run the workflow, optionally on another branch" and maybe add the URL to the documentation.

JF002 (commented):
I don't remember exactly why we removed this feature. As I said, it was useful when I was testing the docker image : I wanted to restart the same build of InfiniTime (no changes needed) with a new build of the Docker image.

It shouldn't be used that often, but I see no reason why we couldn't re-enable it again.

Issue #1285, opened 2022-08-15
Resources (fonts & pictures) packaging


JF002:
### Verification

- [X] I searched for similar feature request and found none was relevant.

### Pitch us your idea!

Let's specify how resources will be packaged and handled by companion apps

### Description

In #321 I did a lot of experiments with the external flash storage and the file system. The goal is to use this external flash memory to store big resources (pictures and fonts) so that we can free quite a lot of space in the internal flash memory.

In [this comment](https://github.com/InfiniTimeOrg/InfiniTime/issues/321#issuecomment-1177289430), I listed the action points that must be done before we can release this feature. This discussion is about the resources packaging and management by companion app : **how will we packaged the resource files so companion app can upload them to the watch using the [BLE FS API](https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/doc/BLEFS.md)** ?

**Please note that this is a technical discussion mostly targeted to InfiniTime and companion apps developers.**

During my experiments, I used [ITD](https://gitea.arsenm.dev/Arsen6331/itd) to manually upload previously generated binary files to the watch. I sometimes needed to retry because the transfer would fail for some reason. I also had no way to check if the data were corrupted during the transfer. The upload of a full picture (240*240*16b = 115200B) took ~2 minutes, iirc.

Sending random binary files to the watch does not fit well with the end-users. I would like all the file to be automatically generated at build time (this is **not** the scope of this discussion), packaged into a single file with all the info needed for companion app to automatically upload the files to the watch.

Here's my proposal : 
 - package all the files in a compressed file format (.tar.gz or .zip). Maybe change the file extension to something else (InfiniTime Resource Format, IRF)
 - add a small 'manifest' file that companion app will be able to use to identify this archive as a resource pack for InfiniTime
   - Could be as simple as a JSON file with the target infinitime version : 
```
{ 
  "infinitime" : "1.11"
}
```
   - it could also contain the file mapping (path and name of each file from the archive -> path and name in the in the PineTime FS)
```
{
  "files": [
    {
      "file1.bin": "/resources/watchface1/background.bin"
    },
    {
      "file2.bin": "/resources/watchface2/font.bin"
    }
  ]
}
```
   - this file could also contain other meta data in the future
 - I don't think our BLE API supports any kind of CRC functionality. The companion app will be able to check the data validity by reading back the file it has just transferred.
[...]

Issue #321, opened 2021-05-01
External flash usage


Avamander:
It would be good to map out where LittleFS would potentially be used in InfiniTime, right now and in the future. What would have to be migrated and what can't be migrated.

Current usages of flash contain:
- [x] Bootloader scratch sections
- [x] Settings storage
- [x] Datetime backup (https://github.com/JF002/InfiniTime/issues/87)

Future usages might contain:
- [x] Resource storage (for LVGL)
- [ ] Buffer for HR/step data (https://github.com/JF002/InfiniTime/issues/493)
- [ ] ??? Please leave your comment below

joaquimorg:
I managed to do a first performance test between RAWFS versus LITTLEFS.
The use of LITTLEFS adds another 20Kb in the firmware.
As we can see in the video the performance between the two is evident.

[![Pinetime Lite - RAWFS vs LITTLEFS](https://res.cloudinary.com/marcomontalbano/image/upload/v1620165278/video_to_markdown/images/youtube--pqAwvxe-y-I-c05b58ac6eb4c4700831b2b3070cd403.jpg)](https://youtu.be/pqAwvxe-y-I "Pinetime Lite - RAWFS vs LITTLEFS")

The LITTLEFS introduces some advantages in the management of the files, but I think that the loss of performance and increase of the firmware does not justify the gain of functionalities.

In my case, I will not invest more time in the littlefs solution, I will try to improve RAWFS and add some features to allow partial updates or subdivide between resources and watch faces, so that I can have customizable watch faces without having to always send everything.

Avamander:
What's the LittleFS configuration you used?

joaquimorg:
> What's the LittleFS configuration you used?

https://github.com/joaquimorg/PinetimeLite/blob/1db8712ecd49ecb2ee37147eebe2e5b6dac521ac/src/components/fs/FS.cpp#L81

Avamander:
I will try to optimize the settings, because I think better read performance is doable and worth the investment. 

The advantages we'd get from LittleFS, such as reduced maintenance and development burden, full-flash wear leveling, better failure tolerance (CoW), thought-out and documented design, higher interoperability and alternative implementations are more than some.

joaquimorg:
In my tests I had already identified that LVGL is always opening files to read, and did not keep the files open while refreshing the screen.
[...]

PR #1299, opened 2022-08-26
Package resources + a metadata file into a single .zip file.

JF002:
Files (fonts and pictures) are converted and packed into a single .zip file using a CMake `custom_target`. This target

- Converts the fonts into .bin files (using a slightly modified version of `src/displayapp/fonts/generate.py`)
- Converts the images into .bin files (using a new script inspired from `generate.py` above)
- Generates a .json file with metadata (see below)
- Zip everything in a single `infinitime-resources-1.11.0.zip`

The content of the zip file looks like this:

```
infinitime-resources-1.11.0.zip
|
|---- 7segments_40.bin
|---- 7segments_115.bin
|---- infineat-1bin
...
|---- resources.json
```

resources.json:

```
{
    "resources": [
        {
            "filename": "7segments_40.bin",
            "path": "/fonts/7segments_40.bin"
        },
        {
            "filename": "7segments_115.bin",
            "path": "/fonts/7segments_115.bin"
        },i
        {
            "filename": "infineat-1.bin",
            "path": "/fonts/infineat-1.bin"
        }
    ],
    "obsolete_files": [
        {
            "path": "/background_v1.bin",
            "since": "1.11.0"
        },
        {
            "path": "/font_v2.bin",
            "since": "1.11.0"
        }
    ]
}
```

- The array `resources` contains the list of files from the .zip file that must be uploaded to the watch. It contains the name of the file in the .zip file and the path (filename) where it must be written to the watch.
- The array `obsolete_files` contains a list of files that can optionally remove from the watch. This is a hint to the companion app which lists the files that are not used anymore in this version of infinitime and that can be removed. It's up to the companion app to remove them to free some memory space. Each item contain the filename in the watch FS and the version of infinitime that made that file obsolete.

EDIT : here is an example package: [infinitime-resources-1.10.0.zip](https://github.com/InfiniTimeOrg/InfiniTime/files/9500512/infinitime-resources-1.10.0.zip)

And this is the expected result once the files are uploaded (screenshots using [ITD](https://gitea.arsenm.dev/Arsen6331/itd):
image

```
./itctl fs ls                            
d 0.0 B  .
d 0.0 B  ..
- 4.9 kB 7segments_115.bin
- 760 B  7segments_40.bin
- 4.4 kB bebas.bin
- 1.4 kB infineat-1.bin
- 1.8 kB lv_font_dots_40.bin
- 115 kB matrix.bin
-  40 B  settings.dat
- 440 B  teko.bin

```

To support this, companion apps will have to implement our [BLE FS API](https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/doc/BLEFS.md).

See [this comment](https://github.com/InfiniTimeOrg/InfiniTime/issues/1285#issuecomment-1221605579)


src/resources/generate-fonts.py
```diff
@@ -0,0 +1,80 @@
...
+    def __init__(self, d):
+        self.file = d['file']
+        if not os.path.exists(self.file):
+            self.file = os.path.join(os.path.dirname(sys.argv[0]), self.file)
+        self.range = d.get('range')
+        self.symbols = d.get('symbols')
+
+
+def gen_lvconv_line(lv_font_conv: str, dest: str, size: int, bpp: int, format: str, sources: typing.List[Source], compress:bool=False):
+    if format != "lvgl" and format != "bin":
```
yehoshuapw, line 22:
```suggestion
    if format != "lvgl" and format != "bin":
        format = "bin" if dest.lower().endswith(".bin") else "lvgl"
```

src/resources/generate-fonts.py
```diff
@@ -0,0 +1,80 @@
...
+        format = "lvgl"
```
yehoshuapw, line 23:
```suggestion
    if format != "lvgl" and format != "bin":
        format = "bin" if dest.lower().endswith(".bin") else "lvgl"
```

Arsen6331:
I'm currently implementing this for ITD. Should be ready soon if everything goes well. I'll make a branch with it when I'm done and then merge when this is released.

Arsen6331:
One question that I have is: should I load the resources (delete obsolete, upload new, etc.) before or after the DFU itself?

JF002:
> should I load the resources (delete obsolete, upload new, etc.) before or after the DFU itself?

That's a good question! Loading the new resources (and potentially removing older ones) before the firmware update might break the firmware that is currently running, if it's still using the resources that will be deleted. Loading them after the firmware update might break the new firmware if it tries to load resources that do not exist yet... 
This is actually one of to TODO task that need to be done before releasing the functionality : "*InfiniTime must handle the case where an app/watchface needs a resource from the FS and the corresponding file does not exist*" https://github.com/InfiniTimeOrg/InfiniTime/issues/321#issuecomment-1177289430

In the end, I think we need to ensure that the firmware (app & watchface) will never crash or misbehave if the resource does not exist. In this case, the update sequence (firmware then resource or resource then firmware) will be the choice of the user or the companion app.

JF002:
Here is a new resource package that maps files into subdirectories (`/fonts` and `/images`). It might be useful to ensure that implementations in companion apps handle the subdirectories correctly

[infinitime-resources-1.10.0.zip](https://github.com/InfiniTimeOrg/InfiniTime/files/9500512/infinitime-resources-1.10.0.zip)

src/resources/generate-fonts.py
JF002: Done, thanks!

JF002:
So my plan for this feature is:
  - Merge this branch to allow generation of resources
  - Extend Infineat to support external resource and merge it. It also brings Infineat watchface and the possibility to select more than 4 watchfaces in the settings.
  - Extend and merge the G7710 watchface
  - Talk with companion app developers and contributors hoping they'll add support for this feature in companion apps. ITD already supports it, and I'm currently working on the integration in Amazfish.
  - If anything goes wrong, we'll be able to temporarily disable the new watchfaces that need additional resources in the settings.
  - Release and enjoy!

PR #1306, opened 2022-08-28
Watchface accurate words

### LINKED ISSUE #1305: New WatchFace to give description of current time accurate to a minute
### Verification

### Pitch us your idea!

New Words WatchFace - PR to follow

### Description

PR to follow see screen shots

bmsleight:
Closed #1305

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -428,6 +428,7 @@ list(APPEND SOURCE_FILES
         displayapp/screens/WatchFaceDigital.cpp
         displayapp/screens/WatchFaceTerminal.cpp
         displayapp/screens/WatchFacePineTimeStyle.cpp
+        displayapp/screens/WatchFaceAccurateWords.cpp
 
         ##
 
--- a/src/displayapp/screens/Clock.cpp
+++ b/src/displayapp/screens/Clock.cpp
@@ -14,7 +14,6 @@
 #include "displayapp/screens/WatchFacePineTimeStyle.h"
 #include "displayapp/screens/WatchFaceAccurateWords.h"
 
-
 using namespace Pinetime::Applications::Screens;
 
 Clock::Clock(DisplayApp* app,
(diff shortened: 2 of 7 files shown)


Itai-Nelken:
This is is very similar to https://github.com/InfiniTimeOrg/InfiniTime/pull/660

bmsleight:
I also like #660
Similar, however - using existing fonts - to save space and accurate to a minute rather than fuzzy

bmsleight:
InfiniSim_2022-08-28_131613
InfiniSim_2022-08-28_132955

bmsleight:
Date in words (excluding year)
InfiniSim_2022-08-28_200900

bmsleight:
Specific dates have specific names
InfiniSim_2022-12-26_205312

bmsleight:
Happy to address issues.
Grateful for some guidance on what is needed or missing ?

I am starting to explore some improvements/hacks on my own branch of this branch. If current pull is not fit for inclusion in project it would help.

minacode:
It seems unlikely that a new watch face is added immediately because the memory and ram on the watch are running out. Work is done currently to remove things from the firmware to make space, but it may need some time. 

I wrote something about the arrays of strings in your implementation because I think that they are not memory-efficient (unless the compiler does magic).

bmsleight:
@minacode "I wrote something about the arrays of strings in your implementation because I think that they are not memory-efficient (unless the compiler does magic)."

Where do you write this - I can find it ?

minacode:
When you scroll up in this feed, it says that is a pending comment. 

What I meant is that you have those arrays of strings of an explicit size, but the strings you put in there are most often shorter. And some strings appear more than once. I am not sure if the compiler can figure this out, but if not then maybe you can define constant strings and use arrays of pointers to those strings. Then you don't need a maximum size for all and can just have the same pointer twice instead of the same string twice.

bmsleight:
@minacode My C skills are very rusty, thanks for the pointers and patience. [Pun intended]

bmsleight:
I think there is an error in my code. 
My watch is now in boot loop. After I validated my previous build and tried to upload a new build. 
Time to get the heat gun.
Closing until I work out what went wrong.

minacode:
Holding the button might bring you to a screen with "InfiniTime". You can upload a new firmware there. I had the same problem, when I tried something with a watch face lately. For me, it was a uninitialized pointer for an lvgl object.

minacode:
Oh. Thank you!

minacode:
I just saw this line in SettingsWatchFace.h

```
static constexpr std::array<const char*, 4> options = {"Digital face", "Analog face", "PineTimeStyle", "Terminal"};
```

This seems to be a good way to handle an array of strings.

PR #1307, opened 2022-08-30
Add support for BMA425's wrist-tilt interrupt (raise-to-wake)

Zoarial94:
This pull request is for adding support for the BMA425's wrist-tilt or raise-to-wake functionality. I have tested this on a PineTime Dev kit and have found no issues to far. The code will automatically use the interrupt if the device has a BMA425, and have no change for a BMA421 (untested). This behavior can be changed pretty easily with a checkbox/toggle. The software fallback for the raise-to-wake and shake-to-wake are still working with the fix mentioned above. 

I have not tested this code on a sealed device, or even on my wrist, but from what I can tell should work fine. 
I understand this will probably interfere with other efforts to improve the BMA421's code (such as #826).

------

Code breaking change
> The axes were remapped in the BMA425, and driver code.
> I have updated the software raise-to-wake and shake-to-wake to use the proper axes.
> The quick fix for using the remapped axes, it just to invert the sign. (Use -x, -y, & -z instead of x, y, & z)

### Change log:
SystemTask.cpp
- Added motionSensor as an argument to motionController's Init()
- Call shouldRaiseWake instead of Should_RaiseWake when deciding to wake up

Bma421.cpp and .h
- Fixed clang-format and initialization issues
- Added wristTiltInterrupt flag for raise-to-wake functionality
- Refactored Bma421::Init() to better match Bosch's example code.
  - Removed redundant step_detector_enable call. Step counting functionality appears to still be working.
  - Refactored all bma423_feature_enable calls to 1 function call
- Remapped BMA425's axes for proper raise-to-wake functionality.
  - Also remapped the axes in driver-space
  - The remap can be done in runtime, which allows for remapping of the right and left hand.
- Enabled BMA425's wrist-tilt feature and interrupt for raise-to-wake functionality

MotionController.cpp and .h
- Fixed clang-format errors
- Added shouldRaiseWake which executes a callback function for the desired raise-to-wake algorithm.
  - This was done so that the call back can be changed transparently in the driver during runtime.
- Added a reference to Drivers::BMA421 in the constructor
  - I believe MotionController should completely encapsulate the BMA421 object. This can be discussed later.
- Created separate function for software raise-to-wake and the BMA425's interrupt
- Updated the software raise-to-wake and shake-to-wake to use the correct axes since they are now remapped.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -128,7 +128,7 @@ Standard: Latest
 StatementMacros:
   - Q_UNUSED
   - QT_REQUIRE_VERSION
-TabWidth: 8
+TabWidth: 2
 UseCRLF: false
 UseTab: Never
 ...
--- a/src/components/motion/MotionController.cpp
+++ b/src/components/motion/MotionController.cpp
@@ -14,34 +14,41 @@ void MotionController::Update(int16_t x, int16_t y, int16_t z, uint32_t nbSteps)
   this->x = x;
   this->y = y;
   this->z = z;
-  int32_t deltaSteps = nbSteps - this->nbSteps;
+  uint32_t deltaSteps = nbSteps - this->nbSteps;
   this->nbSteps = nbSteps;
   if (deltaSteps > 0) {
     currentTripSteps += deltaSteps;
   }
 }
 
-bool MotionController::Should_RaiseWake(bool isSleeping) {
-  if ((x + 335) <= 670 && z < 0) {
+bool MotionController::Should_RaiseWake() {
+  bool isSleeping = true;
+  if ((-x + 335) <= 670 && -z < 0) {
     if (not isSleeping) {
-      if (y <= 0) {
-        return false;
-      } else {
-        lastYForWakeUp = 0;
+      if (-y <= 0) {
         return false;
       }
-    }
-
-    if (y >= 0) {
       lastYForWakeUp = 0;
       return false;
     }
-    if (y + 230 < lastYForWakeUp) {
-      lastYForWakeUp = y;
+
+    if (-y >= 0) {
+      lastYForWakeUp = 0;
+      return false;
+    }
+    if (-y + 230 < lastYForWakeUp) {
+      lastYForWakeUp = -y;
       return true;
     }
   }
   return false;
+}
+
+// Use the hardware interrupt for raise-to-wake
+// This is just a call back, and will hopefully not be used on anything other than a BMA425.
+// See MotionController::Init()
+bool MotionController::BMA425ShouldRaiseWake() const {
+  return bmaDriver->getAndClearWristTiltInterrupt();
 }
 
 bool MotionController::Should_ShakeWake(uint16_t thresh) {
@@ -49,7 +56,7 @@ bool MotionController::Should_ShakeWake(uint16_t thresh) {
   auto diff = xTaskGetTickCount() - lastShakeTime;
   lastShakeTime = xTaskGetTickCount();
   /* Currently Polling at 10hz, If this ever goes faster scalar and EMA might need adjusting */
-  int32_t speed = std::abs(z + (y / 2) + (x / 4) - lastYForShake - lastZForShake) / diff * 100;
+  int32_t speed = std::abs(-z + (-y / 2) + (-x / 4) - lastYForShake - lastZForShake) / diff * 100;
   //(.2 * speed) + ((1 - .2) * accumulatedspeed);
   // implemented without floats as .25Alpha
   accumulatedspeed = (speed / 5) + ((accumulatedspeed / 5) * 4);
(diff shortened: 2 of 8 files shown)


Zoarial94:
I have been using this feature on my sealed PineTime for 4+, and I found that it doesn't work that well. Sometimes it will just not activate, and other times it will active in weird position. I still support remapping the axes in this PR, but I am not certain the interrupt is the way to go for a reliable experience. It a possibly I have not implemented it properly. I would love to see some more input on this.

kieranc:
I'm very interested in this feature, I haven't had a proper look yet but I've built it and will start testing. Given the problems you've listed seem similar to with the current software raise to wake implementation, is there any possibility both are active? I'm including a DFU here for others to test it as actions aren't approved yet for this PR, I've flashed it successfully on my own watch but obviously YMMV.

[pinetime-mcuboot-app-dfu-1.10.0.zip](https://github.com/InfiniTimeOrg/InfiniTime/files/9548344/pinetime-mcuboot-app-dfu-1.10.0.zip)

kieranc:
I've done a bit of testing, it's certainly functional but not perfect - I've noticed a couple of times this evening that it was awake just lying on a desk, and there were some instances of it not waking when I hoped that it would. Maybe some sort of debug app with indicators for shake/software wrist raise/hardware wrist raise activations could be useful? 
I'm afraid I can't offer much in terms of code quality but hopefully someone else can weigh in. I can say that there's a fair amount of cleanup in the code which is great, but unrelated to this PR and for ease of review each PR should aim to address a single issue. You might want to break out the clang/gitignore stuff to a separate branch/PR.

Zoarial94:
Thank you for the feedback. I have not contributed to projects other than my own and I am still trying to get a feel for it. 

I have definitely seen the same behavior you are seeing, and I think the debug app is a great idea. Until now, I was just adding lables into the motion app. 

> ...is there any possibility both are active?
No. The way I implemented this was that a callback is made to a single function. Only 1 algorithm is checked at any point. 

I will see if I can work on something today or tomorrow.

src/components/motion/MotionController.cpp
```diff
@@ -14,69 +14,87 @@ void MotionController::Update(int16_t x, int16_t y, int16_t z, uint32_t nbSteps)
   this->x = x;
   this->y = y;
   this->z = z;
-  int32_t deltaSteps = nbSteps - this->nbSteps;
+  uint32_t deltaSteps = nbSteps - this->nbSteps;
```
FintasticMan, line 17: This shouldn't be unsigned, otherwise the check below will always return true.

.clang-format
```diff
@@ -135,7 +135,7 @@ Standard: Latest
 StatementMacros:
   - Q_UNUSED
   - QT_REQUIRE_VERSION
-TabWidth: 8
+TabWidth: 2
```
FintasticMan, line 138: I don't think this change was intended to be in this PR.

src/components/motion/MotionController.cpp
```diff
@@ -14,69 +14,87 @@ void MotionController::Update(int16_t x, int16_t y, int16_t z, uint32_t nbSteps)
...
   this->nbSteps = nbSteps;
   if (deltaSteps > 0) {
     currentTripSteps += deltaSteps;
   }
 }
 
-bool MotionController::Should_RaiseWake(bool isSleeping) {
-  if ((x + 335) <= 670 && z < 0) {
+bool MotionController::Should_RaiseWake() {
+  bool isSleeping = true;
```
FintasticMan, line 25: `isSleeping` isn't always true. Maybe add the `isSleeping` parameter to the callback function as well?

src/drivers/Bma421.cpp
```diff
@@ -35,9 +37,12 @@ Bma421::Bma421(TwiMaster& twiMaster, uint8_t twiAddress) : twiMaster {twiMaster}
 }
 
 void Bma421::Init() {
-  if (not isResetOk)
+  if (not isResetOk) {
```
FintasticMan, line 40:
```suggestion
  if (!isResetOk) {
```

src/drivers/Bma421.cpp
```diff
@@ -100,23 +123,43 @@ void Bma421::Write(uint8_t registerAddress, const uint8_t* data, size_t size) {
 }
 
 Bma421::Values Bma421::Process() {
-  if (not isOk)
+  if (not isOk) {
```
FintasticMan, line 126:
```suggestion
  if (!isOk) {
```

src/drivers/Bma421.h
```diff
@@ -35,15 +36,28 @@ namespace Pinetime {
       bool IsOk() const;
       DeviceTypes DeviceType() const;
 
+      bool getAndClearWristTiltInterrupt() {
```
FintasticMan, line 39: [Functions should use PascalCase](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/doc/coding-convention.md).

FintasticMan (changes requested):

Zoarial94:
Haha. Didn't expect feedback already. I've wanted to work on this again for a while, and I still won't really have the time for another month or so.

FintasticMan:
I haven't had time to test it on hardware, so I don't know how good the tilt detection itself is. If it's good, it might be alright to just enable it by default, but if it isn't, maybe a toggle in the settings would be good.

Zoarial94:
From my small amount of testing (not mounted to my wrist), it isn't great. I don't know if it's a config issue or if it's just how the sensor is. False positives can be reduced by preventing it from waking while the screen is facing up, but not sure about false negatives yet.

I have a feeling software might be the way to go.

src/components/motion/MotionController.cpp
Zoarial94: I did that as a placeholder since I knew I would need to refactor it. I think it makes sense to pull the logic for setting the tilt threshold into MotionController::Update. In my opinion, MotionController::Should_RaiseWake should only be called when sleeping.

src/components/motion/MotionController.cpp
Zoarial94: I was attempting best practices since I was given a warning from clang-tidy, but I will revert it for now until I get a better understanding of what's going on.

Zoarial94:
Both the hardware and software raise-to-wake functions work. I have flashed my sealed watch with the latest commits on this branch and I will try it out for a while and report what I find. From my initial findings, it seems to work fine. There is sometimes a slight delay before the screen turns on, but still acceptable. If it doesn't turn on after a second, a small shake seems to wake it.

[pinetime-mcuboot-app-dfu-1.11.0.zip](https://github.com/InfiniTimeOrg/InfiniTime/files/10875991/pinetime-mcuboot-app-dfu-1.11.0.zip)

Zoarial94:
I am happy to report using the hardware interrupt works pretty well. I would like to see what others think of it especially compared to the original software solution. In combination with a screen lock and lower-to-sleep functions, I would be able to go about my day without having to worry about accidentally opening an app or wasting too much battery life.

PR #1313, opened 2022-09-02
Redesign chime settings, make chime buzz duration configurable

ght:
The new UI has two spinbox-like elements, one for the chime frequency,
the other for the buzz duration when a chime happens.

Also add more chime frequency options, since switching from checkboxes
to a spinbox removes the fixed screen space overhead per option.

Changing the duration setting emits a buzz of the selected duration, to
get a feel for it.

Used size before: 400680B, after: 401580B (+900B)

Preview:
```
      @  Chimes
 ___              ___
/   \            /   \
| - | Every      | + |
\___/ half hour  \___/
 ___              ___
/   \ buzz for   /   \
| - | 120 ms.    | + |
\___/            \___/
```


minacode:
Do we really need a setting for this?

ght:
That is up to you to decide. I'm just sharing improvements I made on my personal fork. I added this because I wanted chimes every quarter hour and because the 35ms buzz is easy to miss e.g. when you're working outdoors or doing sports, so I wanted a stronger chime buzz.

If you would prefer to keep the old UI but use the refactored chime logic, just drop the last commit from this PR.

---

Slightly off-topic, but: I didn't open this only because I want to see this in mainline InfiniTime, but because I thought it could be useful to others that remix their own forks out of the menagerie of pull requests available here :) Some I can highly recommend:
- #1157 Never had to fiddle with indents in any editor since I merged this locally
- #585 No longer from 40% to dead in under half an hour
- #923 Handy to quickly look up what weekday some upcoming date is, or vice versa

ght:
You didn't come across as rude, I didn't mean to sound offended :)

I agree that the chime vibrations, along with all others, could use a rework in the long run, decoupling the apps that cause vibration from the MotorController via an intermediate VibrationController or something, which would allow centrally managing do-not-disturb/sleep-mode, priorities between vibrations (alarm clock going off while an incoming-call-notification is ringing should not be dropped, but queued until after the call), and maybe handle vibrations differently when watch is not worn, etc.

minacode:
I have exactly this on my todo-list! 😉 Maybe I should find time for this..

ght:
I removed the new UI from this PR, so it is now only a refactoring to clean up the chimes code.

PR #1316, opened 2022-09-05
Fix carried over step count after midnight

### LINKED ISSUE #1270: InfiniTime sends Steps count from the previous day in some cases
### Verification

### What happened?

InfiniTime sends Steps count from the previous day in some cases

InfiniTime should count from 0 after midnight

### Reproduction steps

so far there are 2 cases discovered where InfiniTime sends wrong steps count to Step count characteristic subscriber (e.g. companion app Gadgetbridge):

**case 1**
1. make some steps (e.g. 1000 steps) on day 1.
2. unlock PineTime before midnight to make sure steps are synced.
3. after the midnight (e.g. in the morning) receive notification to the PineTime (e.g. incoming call, or use "fing lost device" in Gadgetbridge)
4. InfiniTime sends step count from the previous day (1000 steps in this case) before reset the steps counter. that won't happen if you just unlock the PineTime. it's specifically related to the case with notification.

**case 2** (probable the issue #787. see [this comment](https://github.com/InfiniTimeOrg/InfiniTime/issues/787#issuecomment-1209173778))
1. make some steps (e.g. 1000 steps) on day 1.
2. unlock PineTime before midnight to make sure steps are synced.
3. after the midnight (e.g. in the morning or overnight) make some steps (e.g. 50 steps) **without** unlocking PineTime.
4. unlock PineTime, and it will send 2 steps count in less that a second: 
   - 1st one - combined values from previous day plus outstanding steps from the new day (1000 + 50 -> 1050steps)
   - 2nd one - zeroed value (literally "0"). also InfiniTime  displays "0" steps. that means all steps from the day 2 before the unlock (50 steps) are gone/zeroed. 

### More details?

there is a discussion about that issues and workaround on Gadgetbridge side ([Gadgetbridge#2783](https://codeberg.org/Freeyourgadget/Gadgetbridge/issues/2783)).

### Version

1.10

### Companion app

_No response_

minacode:
This PR is an attempt to fix #1270 .

I moved `stepCounterMustBeReset` from the system task to the motion controller.
Calling `MotionController::NbSteps()` returns `0` if the reset is set.

Calling `MotionController::Update()` automatically sets `stepCounterMustBeReset` to `false`, because we trust the update value and the reset is therefore obsolete. This might be a thing we want to discuss.

Also, there will still be lost steps between midnight and the actual reset of the IC after its first wake up. I do not understand, why the whole reset build-around thing exists, so maybe it might just be easier (and definitively cleaner) to just wake up the sensor at midnight and reset it.


ght:
IMHO `SystemTask` should not know about internal implementation details of `MotionController`. An API like the second line would be much better (excuse me if this isn't possible due to some limitation of the current code base, I'm not that deep in yet):
```diff
-motionController.SetStepCounterMustBeReset(true);
+motionController.ResetStepCount();
```

minacode:
The last commit does the following:
1. Systemtask does not call the MotionSensor (aka. Bma421/5) for values. It only interacts with it on initialization (or whatever happens in the beginning. I am not touching this :grinning: One may be able move this to MotionController and remove MotionSensor from Systemtask completely but this is not the focus here).
2. Systemtask issues MotionController to update itself instead of passing values to it.
3. MotionController now holds a reference of MotionSensor and fetches the values by itself.
4. Systemtask calls `motionController.ResetSteps()` at midnight. The MotionController handles the rest internally.

This is all untested and I will have another look at it tomorrow. I would appreciate if you take a look at it.

PR #1317, opened 2022-09-06
Make PineTimeStyle step count display configurable

kieranc:
This PR adds the ability to choose between 3 different options for the PTS step count display, the original full gauge, a half gauge with seconds (#915) and a numerical readout like the digital watchface. It also includes #872, I'd like some feedback on this but it can easily be made configurable if it's unpopular.
I hope I've correctly integrated the commits from @ialokim and @aereaux, thanks to both of them.
It looks a bit like this:
InfiniSim_2022-09-06_102238


ialokim:
Nice! Looking forward to having this merged and available in official builds :tada:

aereaux:
Great work!  If it would improve the PR I can also share the code I had that added a gradient to the step counter:

src/displayapp/screens/WatchFacePineTimeStyle.cpp
```diff
@@ -431,16 +500,19 @@ void WatchFacePineTimeStyle::Refresh() {
   stepCount = motionController.NbSteps();
   motionSensorOk = motionController.IsSensorOk();
   if (stepCount.IsUpdated() || motionSensorOk.IsUpdated()) {
-    lv_gauge_set_value(stepGauge, 0, (stepCount.Get() / (settingsController.GetStepsGoal() / 100)));
+    lv_gauge_set_value(stepGauge, 0, (stepCount.Get() / (settingsController.GetStepsGoal() / 100)) % 100);
```
NeroBurner, line 503: I'd remove this modulo, keeping this PR to just the gauge settings

NeroBurner (changes requested):
I'd remove the gauge modulo change, the others look good for me

src/displayapp/screens/WatchFacePineTimeStyle.cpp
kieranc: The gauge modulo has been discussed previously in #872, I'm in favor of it for the reasons outlined there. If it is deeply unpopular I will add a setting toggle for it, but I feel that the gauge continuing to do _something_ once the step goal is reached is an improvement over the current version.

NeroBurner: thanks for the background info, resolved

NeroBurner (approved):
looking good!

JF002 (approved):

JF002:
Thanks @kieranc, great work!

Issue #1320, opened 2022-09-08
Forward or Rewind Media by 5 Seconds Option in the Music App


dyllan500:
### Verification

- [X] I searched for similar feature request and found none was relevant.

### Pitch us your idea!

Forward or Rewind Media By 5 Seconds

### Description

Hello Everyone,

I have been enjoying my PineTime for a while now. I especially love using the media controls in the Music app. I did have a great idea earlier about having the ability to fast forward or rewind my media by 5 seconds. I think it maybe fairly easy to implement. I haven't ever contributed to an open source project before, so I am not sure how to go about this. I hoping with some guidance I may be able implement this code myself. I have listed what I would like the music app to do and the research I did on how to implement this below.

# My vision for the Music App is:

Default Buttons:
Previous_Track_Button < Pause/Play_Button > Next_Track_Button
Volume Buttons:
Volume_Down_Button < Pause/Play_Button > Volume_Up_Button
Forward/Rewind Buttons:
Rewind_Button < Pause/Play_Button > Forward_Button

When the user opens the music app the default buttons are shown, when they swipe up the volume buttons show, if they swipe up again the Forward/Rewind Buttons show, if the swipe up again then the default buttons are shown. Swiping Down will do the reverse. So swiping up or down just cycles through the various buttons.

# My Research:

I looked through the infiniTime code and I think it will be fairly easy to implement this.

I think the only code that needs to be added to MusicService.h is 2 more events.

MusicService.h
```
static const char EVENT_MUSIC_OPEN = 0xe0;
static const char EVENT_MUSIC_PLAY = 0x00;
static const char EVENT_MUSIC_PAUSE = 0x01;
static const char EVENT_MUSIC_NEXT = 0x03;
static const char EVENT_MUSIC_PREV = 0x04;
static const char EVENT_MUSIC_VOLUP = 0x05;
static const char EVENT_MUSIC_VOLDOWN = 0x06;
static const char EVENT_MUSIC_FORWARD = 0x07;
static const char EVENT_MUSIC_REWIND = 0x08;
```
But for Music.cpp code for the Forward and Rewind buttons will needed to be added. And a way to tell which buttons need to be displayed when the user swipes up or down. I think this code is fairly easy to add.

[...]

PR #1323, opened 2022-09-10
Add fast forward rewind to music app

dyllan500:
This is my first ever pull request on any open source project. Please feel free to critique me, so I can improve.
This pull request adds the feature I requested here: https://github.com/InfiniTimeOrg/InfiniTime/issues/1320

TLDR:
I added new buttons to the Music App that allows for fast forwarding and rewinding of content by 10s. I have tested this on a PineTime Dev kit and have found no issues, so far. To test this you will need to make the changes to GadgetBridge I outlined in my issue.


minacode:
Disclosure: I am not a maintainer, just like the project and new people joining it.

I scrolled through this once and it looks good for me. Have you tested it on the hardware?
And you can take a look at the simulator (see the [InfiniSim repo](https://github.com/InfiniTimeOrg/InfiniSim) ) to test your Code if needed (probably not if you test on hardware) and to make a screenshot or gif of the new UI. Would be interesting how the new buttons look.

Congrats to your first open source PR 🎉

dyllan500:
Screenshot_20220911_174001

Here is a screenshot of the new buttons.


The reason why the simulator is failing to build this code needs to be added to it. Once that is added to the simulator branch everything will build fine.

- sim/components/ble/MusicService.h

```
static const char EVENT_MUSIC_OPEN = 0xe0;
static const char EVENT_MUSIC_PLAY = 0x00;
static const char EVENT_MUSIC_PAUSE = 0x01;
static const char EVENT_MUSIC_NEXT = 0x03;
static const char EVENT_MUSIC_PREV = 0x04;
static const char EVENT_MUSIC_VOLUP = 0x05;
static const char EVENT_MUSIC_VOLDOWN = 0x06;
// below needs adding
static const char EVENT_MUSIC_FORWARD = 0x07;
static const char EVENT_MUSIC_REWIND = 0x08;
```

minacode:
I like it. Now we have to wait for further reviews. This may take some time.

src/displayapp/screens/Music.cpp
```diff
@@ -223,7 +243,11 @@ void Music::UpdateLength() {
 
 void Music::OnObjectEvent(lv_obj_t* obj, lv_event_t event) {
   if (event == LV_EVENT_CLICKED) {
-    if (obj == btnVolDown) {
+    if (obj == btnForward) {
```
minacode, line 246: This looks like a switch:
```cpp
switch (obj) {
  case btnForward:
    ...
}

**Review by minacode — CHANGES_REQUESTED** [review]

File: src/displayapp/screens/Music.cpp [hunk 2]
```diff
@@ -248,25 +272,59 @@ void Music::OnObjectEvent(lv_obj_t* obj, lv_event_t event) {
...
       lv_obj_set_hidden(btnVolUp, false);
-
       lv_obj_set_hidden(btnNext, true);
       lv_obj_set_hidden(btnPrev, true);
+      break;
+    case BTN_FOR_REWIND:
+      lv_obj_set_hidden(btnForward, false);
+      lv_obj_set_hidden(btnRewind, false);
+      lv_obj_set_hidden(btnVolDown, true);
+     lv_obj_set_hidden(btnVolUp, true);
```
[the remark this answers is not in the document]
dyllan500: I fixed the following code formatting with my latest commit.

src/displayapp/screens/Music.cpp
dyllan500: I can not make this a switch since the obj variable type is an object and not an integral type. There is no way for the compile to convert the obj to an integer needed for the switch statement.

src/displayapp/screens/Music.cpp
```diff
@@ -23,6 +22,7 @@
 #include "displayapp/icons/music/disc.c"
 #include "displayapp/icons/music/disc_f_1.c"
 #include "displayapp/icons/music/disc_f_2.c"
+#include <libraries/log/nrf_log.h>
```
minacode, line 25: This seems not to be needed. I could compile your code without it.

src/displayapp/screens/Music.cpp
```diff
@@ -248,25 +272,59 @@ void Music::OnObjectEvent(lv_obj_t* obj, lv_event_t event) {
...
+      break;
+    case BTN_FOR_REWIND:
+      lv_obj_set_hidden(btnForward, false);
+      lv_obj_set_hidden(btnRewind, false);
+      lv_obj_set_hidden(btnVolDown, true);
+      lv_obj_set_hidden(btnVolUp, true);
+      lv_obj_set_hidden(btnNext, true);
+      lv_obj_set_hidden(btnPrev, true);
+      break;
+    default:
```
minacode, line 302: I don't know if this actually improves the code, but you could do something like
```cpp
  case BTN_NEXT_PREV:
  default:
    lv_obj_ ...
```

This would reduce code duplication and make clear what the default is.

minacode (changes requested):
I tagged two minor things from reading your code. They should be easy to resolve.
And then I have a bigger question that came to my mind: Do we need so many buttons?
It should be possible to just have one set of buttons and to change their labels based on the current mode. I get that you just followed the previous implementation.

Your code does not build for me with the simulator. Maybe you have to update your branch? I did not test it on hardware.

Avamander:
As much as I would like either rewind or seek functionality, unfortunately do not see it working properly for enough users. Due to that reason I don't think this functionality should be included in base InfiniTime. Tops, maybe behind a compile- or runtime compilation flag.

minacode:
> do not see it working properly for enough users

Honest question: why?

Avamander:
@minacode 
> Honest question: why?

Not supported by most music players. If enabled runtime when compatibility is detected, then it would be fine.

minacode:
I don't know how but guess can we find that out on connection to the device, right? Without the user doing anything. The companions would have to decide if they support it.

Avamander:
> The companions would have to decide if they support it.

Detect/decide, yes.

dyllan500:
I understand like I said in my feature request I had to manually code in support for fast forward and rewind to the GadgetBridge companion app. Even after doing that I only found that VLC works well with the buttons. I am probably a rare user who uses VLC to listen to podcasts, who benefits from this feature.

PR #1324, opened 2022-09-11
Infineat external resources

JF002:
This branch contains https://github.com/InfiniTimeOrg/InfiniTime/pull/1024 merged with develop. I've also added support for external resources.

@dmlls Would you like to have a look at those changes? If it's OK for you, I can add those commit to your PR and eventually merge it :)


dmlls:
@JF002 looks good!

Maybe a stupid question, but should the Actions build run as is on the PineTime? I've tried it but it crashes. There's probably something obvious I'm missing :)

Also, just as a note, the case when the fonts are not loaded and falls back to the system fonts will cause misalignments. But this is probably fine, since at that point there is already something not working properly.

JF002:
@dmlls Thanks for the review! 

> Maybe a stupid question, but should the Actions build run as is on the PineTime? I've tried it but it crashes. There's probably something obvious I'm missing :)

I don't think you're missing anything obvious... I've just tried the DFU from the CI build (https://github.com/InfiniTimeOrg/InfiniTime/actions/runs/3032845417) and it works fine on my setup, even if the resources are not loaded yet. What happens when "it crashes" ? It boots and resets a few seconds later?

> Also, just as a note, the case when the fonts are not loaded and falls back to the system fonts will cause misalignments. But this is probably fine, since at that point there is already something not working properly.

I implemented this fallback mode to ensure that the watch wouldn't crash if the resources are not installed when running this code. And I confirm, the alignment is not perfect with the system font. That's something we can improve if we want the fallback mode to look good, indeed!

dmlls:
@JF002 not sure what I was doing wrong, but installing the DFU via Gadgetbdrige worked _neatly_ ;) Thanks!

<br>

> I implemented this fallback mode to ensure that the watch wouldn't crash if the resources are not installed when running this code.

Makes total sense!

<br>

> And I confirm, the alignment is not perfect with the system font. That's something we can improve if we want the fallback mode to look good, indeed!

In the original PR I fixed the alignment (see [`2ad51e6`](https://github.com/InfiniTimeOrg/InfiniTime/pull/1024/commits/2ad51e6a98004f0f163c10231821ad4b2da70163)). Only the date and BLE icon needed to be adjusted, resulting in:

<p align="center">pinetime-classic</p>

<br>

However this fix cannot be applied directly here as there were some small naming and layout changes. Nothing too complicated to implement, but this and next week will be super busy, so I don't want to block the PR.

What if we leave it like this for now and I'll come back to it later? :)

src/displayapp/DisplayAppRecovery.cpp
```diff
@@ -2,6 +2,7 @@
 #include <FreeRTOS.h>
 #include <task.h>
 #include <libraries/log/nrf_log.h>
+#include <components/fs/FS.h>
```
NeroBurner, line 5: including our headers should be done with `""`, excluding other libs (like lvgl, nrf, std-lib) should be done with `<>`

```suggestion
#include "components/fs/FS.h"
```

src/displayapp/screens/CheckboxList.h
```diff
@@ -0,0 +1,49 @@
+#pragma once
+
+#include <lvgl/lvgl.h>
+#include <cstdint>
+#include <memory>
+#include "displayapp/screens/Screen.h"
+#include "displayapp/Apps.h"
+#include "components/settings/Settings.h"
+
+#define MAXLISTITEMS 4
```
NeroBurner, line 10: we could make this define into a constexpr size_t and move it from the global namespace into the private CheckboxList one

src/displayapp/screens/CheckboxList.h
```diff
@@ -0,0 +1,49 @@
...
+      private:
+        const uint8_t screenID;
+        Controllers::Settings& settingsController;
+        const char* optionsTitle;
+        const char* optionsSymbol;
+        void (Controllers::Settings::*SetOptionIndex)(uint8_t);
+        uint8_t (Controllers::Settings::*GetOptionIndex)() const;
+        std::array<const char*, MAXLISTITEMS> options;
+
+        lv_obj_t* cbOption[MAXLISTITEMS];
```
NeroBurner, line 40: we could use a `std::array` here as well

```suggestion
        std::array<lv_obj_t*, MAXLISTITEMS> cbOption;
```

src/displayapp/screens/CheckboxList.h
```diff
@@ -0,0 +1,49 @@
...
+
+        lv_point_t pageIndicatorBasePoints[2];
+        lv_point_t pageIndicatorPoints[2];
```
NeroBurner, line 43: although more to type I like `std::array` :grin: 

```suggestion
        std::array<lv_point_t, 2> pageIndicatorBasePoints;
        std::array<lv_point_t, 2> pageIndicatorPoints;
```

src/displayapp/screens/Clock.h
```diff
@@ -42,12 +43,14 @@ namespace Pinetime {
         Controllers::Settings& settingsController;
         Controllers::HeartRateController& heartRateController;
         Controllers::MotionController& motionController;
+        Controllers::FS& fs;
```
NeroBurner, line 46: optional suggestion: we could rename it to `filesystem` to be more explicit

see comment (far) above

src/displayapp/DisplayAppRecovery.cpp
```diff
@@ -24,7 +25,8 @@ DisplayApp::DisplayApp(Drivers::St7789& lcd,
                        Pinetime::Controllers::TimerController& timerController,
                        Pinetime::Controllers::AlarmController& alarmController,
                        Pinetime::Controllers::BrightnessController& brightnessController,
-                       Pinetime::Controllers::TouchHandler& touchHandler)
+                       Pinetime::Controllers::TouchHandler& touchHandler,
+                       Pinetime::Controllers::FS& filesystem)
```
NeroBurner, line 29: we're using `fs` for the filesystem controller everywhere else (I think). I like the explicitness of `filesystem`

would you suggest to use `filesystem` instead of `fs` in the future?

src/displayapp/screens/WatchFaceInfineat.h
```diff
@@ -0,0 +1,144 @@
+#pragma once
+
+#include <lvgl/src/lv_core/lv_obj.h>
```
NeroBurner, line 3: I think we're using lvgl.h directly, instead of including from the internals of the lvgl library

```suggestion
#include <lvgl/lvgl.h>
```

NeroBurner (changes requested):
not a complete review, I ran out of time

src/displayapp/DisplayAppRecovery.cpp
JF002: Done

src/displayapp/screens/CheckboxList.h
JF002: Done! However, since this constant is also used in the ctor(), I added it as a public field of the class.

src/displayapp/screens/CheckboxList.h
JF002: Done

src/displayapp/screens/Clock.h
JF002: Done

src/displayapp/DisplayAppRecovery.cpp
JF002: Yes, I agree with you, `filesystem` is more explicit!

src/displayapp/screens/WatchFaceInfineat.h
JF002: Done

JF002:
Merged in #1024

PR #1199, opened 2022-06-24
WIP: screens: add dice app

w4tsn:
Hey there!

An app to do simple multi-dice rolls for games like D&D. The app makes use of the motionController and "wake on shake" setting (multiplied by a fixed amount) to allow rolling by shake movement. It also uses the motion sensor as random seed source.

InfiniSim_2022-06-24_074042
InfiniSim_2022-06-24_074053
InfiniSim_2022-06-24_074057

Let's talk if this is useful for the upstream project ;)

Todos:

- [ ] don't re-roll through shake movement, if a roll is currently displayed
- [ ] show roll statistics, such as lowest / highest roll, no. of 1s and 20s, 3 lowest rolls

PR #1326, opened 2022-09-12
New dice-rolling app: InfiniDice!

yusufmte:
InfiniDice is a dice-rolling applet where you can roll random numbers from d2 to d99. The functionality is:
- Set the number of "sides" of the dice
- Roll to randomly generate a result number in the interval from 1 to the number of sides

It starts at d2 (ie coinflip), and when you first open the app, a roll is made, so you get a "free" coinflip. Then you can change the number of sides and use it as a d6, d20, etc. The color of the result alternates with each roll (so that you can be sure it was actually rolled even when the result is the same).

Tested on PineTime hardware with InfiniTime 1.10.0, and found to be in good working order. See video:

https://user-images.githubusercontent.com/45940010/189685541-4fb9899f-cb8c-4026-8d25-8971b0421565.mp4

I often find myself looking for dice and think this could come in handy for tabletop players :) Any testing, feedback, or suggestions on optimization would be appreciated!


joseph58tech:
I believe this has been implemented before, check #565

minacode:
I like the simple interface and think they solve different problems. Maybe we could have both apps in a world where we are able to choose which apps we want.

src/displayapp/screens/Dice.h
```diff
@@ -0,0 +1,31 @@
...
+namespace Pinetime::Applications::Screens {
+  class Dice : public Screen {
+    public:
+      Dice(DisplayApp* app, Controllers::DateTime& dateTime);
+      ~Dice() override;
+      void Roll();
+      void NextColor();
+
+    private:
+      Controllers::DateTime& dateTime;
```
NeroBurner, line 17: because you need the dateTime object only in the constructor you can remove the stored reference here

src/displayapp/screens/Dice.h
```diff
@@ -0,0 +1,31 @@
...
+
+      lv_obj_t* btnRoll;
+      lv_obj_t* btnRollLabel;
+      lv_obj_t* resultLabel;
+
+      lv_color_t resultColors[3] = {LV_COLOR_YELLOW, LV_COLOR_MAGENTA, LV_COLOR_AQUA};
```
NeroBurner, line 23: you could use a `std::array<lv_color_t, 3>`. Then you won't need the `resultColorsLength` variable in the next line

src/displayapp/screens/Dice.h
```diff
@@ -0,0 +1,31 @@
...
+      uint8_t resultColorsLength = sizeof(resultColors) / sizeof(resultColors[0]);
+      uint8_t currentColorIndex;
+
+      uint8_t rollResult;
```
NeroBurner, line 27: is the rollResult needed outside of the `Roll()` function? I think the member variable can be removed.

src/displayapp/screens/Dice.h
```diff
@@ -0,0 +1,31 @@
...
+#include "components/datetime/DateTimeController.h"
+#include <lvgl/lvgl.h>
+
+namespace Pinetime::Applications::Screens {
+  class Dice : public Screen {
+    public:
+      Dice(DisplayApp* app, Controllers::DateTime& dateTime);
+      ~Dice() override;
+      void Roll();
+      void NextColor();
```
NeroBurner, line 14: the `NextColor()` function is a private helper. I don't think it is needed in the public interface of the App. please move into the private namespace

NeroBurner (changes requested):
lovely app :heart:

FintasticMan:
One feature that I think is quite nice to have in a dice app, is being able to specify a number of dice, rather than just always rolling one die. I don't think that for this you would need to show the results of all the dice individually, just the final result.

For the rest I think this app looks good! 👍

minacode:
#565 is doing exactly that already.

FintasticMan:
You're right, it does do that. But it hasn't been updated in quite a while, and so doesn't merge nicely with some of the changes in InfiniTime. The author of that PR also doesn't seem to be very active, so it's uncertain as to whether it will be updated.

minacode:
This problem is fundamental for InfiniTime. We need to fix the app problem. 
Shameless plug: #1296

src/displayapp/screens/Dice.h
yusufmte: addressed in 6c7d8c2fcae8f649b94a3d23fe4b42c4ea1fd806

src/displayapp/screens/Dice.h
yusufmte: addressed in 3bd3d817a04134a50d11062de7c836e6a5251d2a

src/displayapp/screens/Dice.h
yusufmte: addressed in 21e5ec188447c357409e91ea5b905d3e161d02b0

src/displayapp/screens/Dice.h
yusufmte: addressed in b1e522fc42e8819e7d8ca90e7a8e2b877f6fd303

yusufmte:
Thank you @NeroBurner kindly for your improvements on the code :heart: Thank you everyone for your comments and feedback as well.

@FintasticMan :
> One feature that I think is quite nice to have in a dice app, is being able to specify a number of dice, rather than just always rolling one die.

I definitely appreciate your sentiment, rolling multiples is a frequently needed and useful function of dice! :+1:  However I tend to agree with @minocode , there's a value to keeping the interface simple and intuitive, and #565 already implements multi-roll in a lovely way. But I might be thinking about ways to implement that smoothly without crowding the UI too much.

yusufmte:
Implemented rolling multiple dice in 52e4af53332a4e6d8c6e54b4862c08df300aeef0

The number of dice can range from 1-9 (default 1), and the sides can range from d2-d99 (default d2). The individual rolls are shown as well. Here's the new look:

infinidice_on_infinisim
infinidice_on_pinetime

As before, any feedback or further suggestions are much appreciated. I'll try to keep it up to date with the official develop branch in the meantime. Thanks everyone :)

minacode:
Maybe sort the results from highest to lowest?

src/displayapp/screens/Dice.cpp
```diff
@@ -14,9 +14,8 @@ static void btnRollEventHandler(lv_obj_t* obj, lv_event_t event) {
   }
 }
 
-Dice::Dice(DisplayApp* app, Controllers::DateTime& dateTime) : Screen(app) {
-
-  srand(dateTime.Uptime().count() * time(nullptr));
+Dice::Dice(DisplayApp* app, System::SystemTask& systemTask) : Screen(app), systemTask {systemTask} {
```
NeroBurner, line 17: SystemTask is never used, is it?

yusufmte: You are correct; for some reason I thought at one point that I would need it to get the system tick count, which is now being used for the seed (working better than uptime and other things I tried-- no operations needed, and seems there's about 1 tick per ~0.1 ms so very low chance of seed collision!) -- I must have mistakenly left in the argument & reference after that. Thanks for catching that! Reference removed in adc97a03ba0b9555970692ddf01c7e9ce56e2e20

yusufmte:
Thanks for the suggestions @minacode , I really appreciate it!

> Maybe sort the results from highest to lowest?

This is a neat concept, however after thinking about it I'm hesitant to implement it, as I see two plausible use cases it would affect:

(1) User wants the rolls sorted (eg if they need to know the highest of N rolls).
(2) User wants the exact original sequence of rolls (eg if asked "roll three dice, the result of the 1st will determine A, result of the 2nd will determine B, result of the 3rd will determine C").

If the rolls were sorted, it would be convenient for use case 1 but make use case 2 impossible. If the rolls were left in original sequence, it would only be inconvenient for use case 1 (but still possible/pretty easy, given 9 dice max), but make use case 2 possible/easy.

> Also: how about showing only the sum or the single values and swapping between them with a tap on the label?

This is an interesting idea, but IDK, with the current implementation, one can fill the display maximally (3-digit sum and 9x 2-digit individual rolls, as pictured) and it still fits nicely within the provided area.

infinidice_filled_display

From that standpoint I feel it would be better to have more info displayed at a time on a single static screen, if it can be helped, rather than make the user flip between modes. Also, the tapping would represent a hidden feature/function, whereas currently the function and control of the applet is plain just by looking at a static image.

The other thing influencing my thinking here is that when I feel ambivalent about whether to include a feature, the approach I'm favoring is to err towards keeping things simple and doing as little as possible, given the current memory/resource constraints on the PineTime.

minacode:
Ok, you convinced me 🙂

src/displayapp/screens/Dice.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+  lv_obj_t* dLabel = lv_label_create(lv_scr_act(), nullptr);
+  lv_obj_set_style_local_text_font(dLabel, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, &jetbrains_mono_42);
+  lv_obj_set_style_local_text_color(dLabel, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_WHITE);
+  lv_label_set_text_static(dLabel, "d");
+  lv_obj_align(dLabel, lv_scr_act(), LV_ALIGN_IN_TOP_LEFT, 57, 80);
+
+  dCounter.Create();
+  lv_obj_align(dCounter.GetObject(), nullptr, LV_ALIGN_IN_TOP_LEFT, 83, 24);
+  dCounter.SetValue(2);
```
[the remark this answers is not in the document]
yusufmte: good point 🤔️ I guess d6 may be the most common, besides you can simulate d2 or d3 with d6 (pick odds/evens, etc.). Adjusted in 13758ad4ec83e04253fd4b88fbfe8d9a5f875d0d

src/displayapp/screens/Dice.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include "displayapp/screens/Screen.h"
+#include "displayapp/widgets/Counter.h"
+#include "systemtask/SystemTask.h"
+#include <components/motion/MotionController.h>
```
NeroBurner, line 6: use `<>` includes for includes outside of InfiniTime, for example system libraries or std-libraries. For InfiniTime header please use `""`

```suggestion
#include "components/motion/MotionController.h"
```

src/displayapp/screens/Dice.h
```diff
@@ -0,0 +1,33 @@
...
+#include <lvgl/lvgl.h>
+#include <random>
+#include <array>
```
NeroBurner, line 9: inside a group of includes (like std-lib includes) try to order them alphabetically (if possible)

```suggestion
#include <array>
#include <random>
```

src/displayapp/screens/Dice.cpp
```diff
@@ -0,0 +1,93 @@
...
+}
+
+Dice::~Dice() {
+  lv_obj_clean(lv_scr_act());
+}
+
+void Dice::Roll() {
+  uint8_t resultIndividual;
+  uint16_t resultTotal = 0;
+  std::uniform_int_distribution<> distrib(1, dCounter.GetValue());
```
NeroBurner, line 71: is the empty `<>` brackets needed? If not I'd remove them

```suggestion
  std::uniform_int_distribution distrib(1, dCounter.GetValue());
```

NeroBurner (changes requested):
please fix code formatting, otherwise very good use of the `<random>` header and thank you for the informative commit message https://github.com/InfiniTimeOrg/InfiniTime/pull/1326/commits/c1b435bcfcb3a242a725333efaab8cd4e00e3762 explaining why you introduce more code to be more correct :+1:

src/displayapp/screens/Dice.h
yusufmte: thank you for the explanation & improvements on the code! :heart: fixed in 39235224f476390ad2e5ffc3418ca96885a491d1

src/displayapp/screens/Dice.h
yusufmte: fixed in 39235224f476390ad2e5ffc3418ca96885a491d1

src/displayapp/screens/Dice.cpp
yusufmte: I think they are necessary to tell the compiler to use the default result type (int); without the `<>` it doesn't compile for me. Mirrors usage in the example in https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution

src/displayapp/screens/Dice.h
```diff
@@ -1,12 +1,13 @@
 #pragma once
 
+#include "components/motion/MotionController.h"
 #include "displayapp/screens/Screen.h"
 #include "displayapp/widgets/Counter.h"
+#include "lvgl/lvgl.h"
```
NeroBurner, line 6: sorry about the misunderstanding, but lvgl isn't "our" project, it is in our project dir through a submodule, but it is not ours. So I'd move the include down and use  `<>`

src/displayapp/screens/Dice.cpp
NeroBurner: ah thanks, learned something!

src/displayapp/screens/Dice.h
yusufmte: ah yes-- sorry about that, thanks for explaining! fixed in 4647f8aa3e3f464f93e3cdd877d23daecaace796

src/displayapp/screens/Dice.h
```diff
@@ -3,10 +3,10 @@
 #include "components/motion/MotionController.h"
 #include "displayapp/screens/Screen.h"
 #include "displayapp/widgets/Counter.h"
-#include "lvgl/lvgl.h"
 #include "systemtask/SystemTask.h"
 
 #include <array>
+#include <lvgl/lvgl.h>
```
NeroBurner, line 9: I'm probably starting to go on your nerves. Sorry about that :sweat_smile: 

first system header, then std-lib-header

```suggestion
#include <lvgl/lvgl.h>

#include <array>
```

yusufmte: Not in the least :smile: I appreciate your time & patience going thru the code & making it better!

yusufmte:
Btw, I've clang-formatted all files, but the error & warnings I get when I run clang-tidy appear to be false -- I suspect this is the cause for the code formatting check fail. I get similar errors when running clang-tidy on other pre-existing files in the repo as well. For example:

```
Error while processing /home/yusuf/PineTime/InfiniTime-InfiniDice/src/displayapp/screens/Dice.cpp.
/home/yusuf/PineTime/InfiniTime-InfiniDice/src/displayapp/screens/Dice.cpp:1:10: error: 'displayapp/screens/Dice.h' file not found [clang-diagnostic-error]
#include "displayapp/screens/Dice.h"
         ^
```
(While Dice.h exists and the include is working otherwise)

```
/home/yusuf/PineTime/InfiniTime-InfiniDice/src/displayapp/screens/Dice.cpp:11:34: warning: parameter 'font' is unused [misc-unused-parameters]
  lv_obj_t* MakeLabel(lv_font_t* font, lv_color_t color, lv_label_long_mode_t longMode, uint8_t width, lv_label_align_t labelAlignment, const char* text, lv_obj_t* reference, lv_align_t alignment, int8_t x, int8_t y) {
```
(This warning appears for all parameters of the functions in the local namespace in Dice.cpp, while all parameters are used)

Do you know why this might be happening? Have you run into this problem before?

Thanks!

NeroBurner:
how do you call clang-format? The script used in the CI can be found here https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/tests/test-format.sh

According to the script you should be able to do the following (not tested locally, and maybe you need to install packages for git clang-format)

```sh
git clang-format-12 -q --style file --diff "develop" src/displayapp/screens/Dice.cpp
```

JF002:
What are the real world use-case(s) for such an app? This is an honest question, as I've never felt the need for this kind of app ;)

yusufmte:
@JF002 

> What are the real world use-case(s) for such an app? This is an honest question, as I've never felt the need for this kind of app ;)

The question is appreciated! Same as the reasons one would want physical dice or a coin to flip :) For anyone who plays many tabletop or board games, looking for dice or not having enough dice handy is a common problem. Or for randomizing the starting player, picking which path to take on a walk, or which friend will go get the food (similar to drawing straws). I made the app because I thought it's something I would use fairly often, and it's just neat to have dice on my wrist!

Like with any app, for the PineTime or otherwise, some people may enjoy and use it while some others may not find a use for it. As mentioned previously and in the [InfiniTime Vision](https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/doc/InfiniTimeVision.md), in an ideal world, people could just pick which apps they want to include in their version of the firmware. But for now, for something that doesn't require many resources, I would argue towards including it and having that extra provided functionality available even though everyone may not use it. One piece of evidence favoring that at least some portion the PineTime community might appreciate some wrist-dice is that this is the third instance of such an app-- see #565 and #1199 .

Danimations:
I like the dice rolling app idea, and would be happy to make some visuals for it, if there is interest in the display of something resembling a traditional dice. It could also be nice to add vibration to accompany the "roll" given that the most common use case will be be tabletop gamers.

NeroBurner (approved):
looking good for me!

I think we should get the basic functionality in and open a new Issue/PR for possible styling to not block this PR any further

yusufmte:
Thanks! I like the vibration idea; I've added that in https://github.com/InfiniTimeOrg/InfiniTime/pull/1326/commits/b47b13aecb5db37ea0d592e224d7b3de3e72c58e since it is a small change. Agree with getting basic functionality in first.

Thanks for the offer @Danimations -- I love the idea of displaying something resembling the traditional dice down the line (may be more realistic to implement now, given our recent newfound access to external resource storage)! I think it would be tricky with certain rarer dice sizes, but could be done with the more common ones (eg d2, d6, d20). The individual result label could display a sequence of images of the dice with the result shown, with the sum still shown numerically. Or the images could be used only when 1 die is rolled, and show the image of that die result while hiding the total / individual result text.

Danimations:
> Thanks for the offer @Danimations -- I love the idea of displaying something resembling the traditional dice down the line (may be more realistic to implement now, given our recent newfound access to external resource storage)! I think it would be tricky with certain rarer dice sizes, but could be done with the more common ones (eg d2, d6, d20).

Cool, glad to see you like the vibration idea and that external storage has been unlocked with some wizardry! What would work best, vectors or raster images for dice faces? And if vector, what format should I supply these in?

yusufmte:
It'll be a little bit before I get a chance to look into using the new external resources in the code (and probably separate PR to let this one go through when it's deemed ready), but I'd say raster in png format would probably work best!

Danimations:
Cool. I will make a rudimentary set of .PNG for D2 and D6 when I get a chance or an idle moment. The coin flip visual will be nice with some spinning animation... I'm an animator too, so I easily give you spinning and landed sprites for the coin toss/spin. Thanks for bringing this idea to life.

Avamander:
I think you could draw dice programmatically as well, shouldn't be too difficult. Loading resources is not instant and some aspects might be easier in a hybrid-graphics way.

Danimations:
> It'll be a little bit before I get a chance to look into using the new external resources in the code (and probably separate PR to let this one go through when it's deemed ready), but I'd say raster in png format would probably work best!

I've made some D6 .png files for you to play with... how shall I send them to you?

yusufmte:
Hey there, just checking in-- what's the current status on this? What are the current obstacles to merging and anything I can do to get it to that point? Thanks! 👍

Danimations:
Hi Yusuf. I have art for you, but broke my leg on the weekend and haven't
had a computer in hospital. I also haven't worked out how to upload files
to GitHub yet.

Dan

yusufmte:
Sorry to hear that- hope you have a smooth recovery! No rush at all with the files, it'll take me a little while to get to figuring out how to integrate them and I would probably open a separate issue for it.

JF002:
> Hey there, just checking in-- what's the current status on this? It is ready from my standpoint. What are the current obstacles to merging and anything I can do to help get it closer to that? Thanks! +1

I think the only obstacle left mostly my free time :) We do plan to review and merge this app soon, I just want to have a closer look at the code ;-)

> Hi Yusuf. I have art for you, but broke my leg on the weekend and haven't had a computer in hospital. I also haven't worked out how to upload files to GitHub yet.

Sorry to read this, I hope you'll get well soon! You can upload file simply by dragging them into the text area when you're editing you comment, or by clicking on the 'link' button above the editor. Not sure how to do it on the phone, though.

yusufmte:
> I think the only obstacle left mostly my free time :) We do plan to review and merge this app soon, I just want to have a closer look at the code ;-)

Thanks! I know the maintainers work on this in their free time and have lots to review so no hurry :) I don't mean to rush/be pushy, just wanted to check if there was something I was supposed to be working on! Take your time. 👍

JF002:
That's fine, your question was totally relevant knowing that it was initially planned for 1.11. But there were already quite a lot of changes so I preferred to release while it was stable before adding new changes ;)

yusufmte:
Makes sense! 🙂

src/displayapp/screens/Dice.h
```diff
@@ -0,0 +1,40 @@
...
+    ~Dice() override;
+    void Roll();
+
+  private:
+    lv_obj_t* btnRoll;
+    lv_obj_t* btnRollLabel;
+    lv_obj_t* resultTotalLabel;
+    lv_obj_t* resultIndividualLabel;
+
+    std::mt19937 gen;
```
JF002, line 27: We already use `rand()` to generate random numbers in Paddle, Twos and PTS watch face. Is there any reason to use another random number generator here?

yusufmte: Sorry for the late reply! I noticed that, and so I did initially use `rand()` here as well in my earlier version. However, there are significant flaws with `rand()` and a few reasons that the functions used here instead (from the [C++11 `<random>` library](https://en.cppreference.com/w/cpp/header/random)) provide superior randomness. For the other apps mentioned, I think it doesn't matter too much to have great randomness, because the experience is fine as long as it _feels_ random enough for the user (and having fewer lines of code is better).

But for dice, where the whole _point_ of the app is randomness, it is important to have very fair randomness (especially if the dice were being used for a game or competition).

More details below:

(1) `std::rand()` is considered somewhat deprecated for nontrivial needs and does not guarantee quality of the randomness of the sequence produced. More details [here](https://en.cppreference.com/w/cpp/numeric/random/rand):

> There are no guarantees as to the quality of the random sequence produced. In the past, some implementations of rand() have had serious shortcomings in the randomness, distribution and period of the sequence produced (in one well-known example, the low-order bit simply alternated between 1 and 0 between calls). rand() is not recommended for serious random-number generation needs. It is recommended to use C++11's random number generation facilities to replace rand(). (since C++11)

(2) Using `rand() % x` opens the possibility for modulo bias: that is, if `RAND_MAX` is not divisible by `x`, then the last "segment" is shorter and generating smaller numbers is therefore more likely. With `RAND_MAX` being usually `2^32`, this bias is exceedingly slight, but again, it is best to correct for it when the randomness is the point of the app in question. While you can correct for this bias with a few extra lines of code with `rand()`, I think it is more elegant to use a [standard library function](https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution) that already does so by picking integers uniformly distributed on a specified closed interval.

(3) The `<random>` library here has the neat [`std::seed_seq` function](https://en.cppreference.com/w/cpp/numeric/random/seed_seq), which mixes multiple sources of randomness together (as opposed to `srand`, which takes in just 1 seed), and makes seeds distributed over the whole 32-bit range even when the inputs are close or poorly distributed. This app mixes the system tick count and the X/Y/Z components of the PineTime's sensitive motion controller (credit to @w4tsn #1199 for the motion controller idea!), making seed collision extremely unlikely.

medeyko:
First, I would like to thank for your attention to the randomness quality ( avoiding modulo bias and other problems with rand() )!
However, I have two questions and one proposal.
1. What does impose the colors choice? Why there are three colors? Why they are exactly yellow, magenta & aqua? Are these 3 colors are used is some set of the games?
2. What is the key difference between this all and those that were proposed earlier? ( #565, #919, #1199) Why you had to make a new app instead of reworking the #565?
3. I would like to propose the following: in case when there is a coin flip mode (i.e. 1 d2, which seems to be a default mode) not only display "1" or "2", but also "heads" or "tails" respectively. Of course, it's a subtle addition, but I think it would add some charm to the app.

yusufmte:
@medeyko:

> What does impose the colors choice? Why there are three colors? Why they are exactly yellow, magenta & aqua? Are these 3 colors are used is some set of the games?

(1) Good question! I started out with just yellow and magenta. I picked these colors because there is no form of colorblindness that confuses them (or at least extremely rare-- as opposed to doing something like R/G/B). Then I also added aqua to introduce more contrast/variety, and to complete the "set" (see below). 3 seemed like a good number to have a good amount of variety without going overboard in the number of colors. These particular three colors are at maximally contrasting parts of the color wheel (they are (R,G,B) (1,1,0), (1,0,1) and (0,1,1), respectively), and are maximally bright to contrast nicely with the black background making them easy to see. Beyond that, it just seemed like a fun palette. Beyond making it clear that the result changed, the "colorful" theme seemed to match the concept of dice and randomness.

image

> What is the key difference between this all and those that were proposed earlier? (Add dice rolling app #565, Create a coin flip app #919, WIP: screens: add dice app #1199) Why did you have to create a new app instead of reworking the Add dice rolling app #565?

(2) To be honest, I actually didn't notice those when I started working on this app, and was just excited to start developing for the PineTime, and wanted dice. I did glance through the existing PRs but just missed it. It was my mistake. So to answer your question, there is no grand difference, and I think the other apps are awesome! If I had noticed it, I probably would have helped rework one of the other ones, or tried of a different new app, or worked on other feature requests. I just kept the PR open after that because it seemed there was some interest and that this one was closest to the most current version of the code-base (eg by using the existing Counter class, which didn't exist before). But if there is interest instead in re-focusing efforts to re-working/re-basing one of the other apps, I would have no problem with shifting my support to that effort and closing this PR. I just want dice for the PineTime.

> I would like to propose the following: in case when there is a coin flip mode (i.e. 1 d2, which seems to be a default mode) not only display "1" or "2", but also "heads" or "tails" respectively. Of course, it's a subtle addition, but I think it would add some charm to the app.

(3) Thanks, I really like that idea 🙂️ Just implemented in https://github.com/InfiniTimeOrg/InfiniTime/pull/1326/commits/ee1f39bf4a55aa677d27669d619a26bd64fe0358 .

> Maybe it would be right to make the "roll" button unclickable for about 5 seconds in order to minimize cheating/arguments during games.

Much appreciate the suggestion! But I am not sure I fully understand where this would be helpful. What situation are you picturing where the current setup enables cheating? Someone quickly clicks the button again because they did not like the first result? I mean, even with physical dice, you can pick up and roll it again right after the result becomes visible-- you just have to assume some degree of good faith with the people you are playing with. Or is it a setting where only the wearer of the watch can see the screen and their finger tapping? Even then, it seems to make sense to hold it out more visibly when rolling. I just worry this would be extra code/complexity to remove a potential feature (quick consecutive rolls) or solve a non-problem.

Thank you for looking thru the app and commenting! 🙂️ Greatly appreciate the thoughts and feedback.

yusufmte:
Just fixed the build error caused after rebase -- Trying to keep it up to date so that if merged in the future it will not be too convoluted of a process. It's building well and working well on PineTime hardware for me now with no bugs.

I'm not sure what the "PR comment / comment (pull_request_target)" workflow that is failing is or how to address that -- anyone have any ideas?

Boteium:
what does the 0xf201 icon do? this chart-line icon doesn't seem to be used.

FintasticMan:
That used to be the icon for the motion sensor debug app, which was removed. Presumably when updating this PR that change wasn't applied properly.

yusufmte:
> what does the 0xf201 icon do? this chart-line icon doesn't seem to be used.

> That used to be the icon for the motion sensor debug app, which was removed. Presumably when updating this PR that change wasn't applied properly.

Thanks for catching that! That's exactly what happened. Fixed in https://github.com/InfiniTimeOrg/InfiniTime/pull/1326/commits/1415cf9bc332a9ccd01490709f1518bc8bccdc74

Boteium:
Would it be too late to suggest UI changes ?
I think the alignment of the result can be slightly better.
Using full name to label the counter instead a single letter d gives clearer meaning, and we can also save some space by removing letter d from jetbrain_mono_42.

https://github.com/InfiniTimeOrg/InfiniTime/assets/8587031/ce09728f-95c9-41a5-9a8f-149c79378804

yusufmte:
> Would it be too late to suggest UI changes ?
> I think the alignment of the result can be slightly better.
> Using full name to label the counter instead a single letter d gives clearer meaning, and we can also save some space by removing letter d from jetbrain_mono_42.

Certainly not too late! This PR has been added to & removed from the last three or so milestones, so I am not clear on what the plans for it are with regards to the InfiniTime project, but my current goal is just to keep it up to date with the main branch. Certainly appreciate any suggestions & feedback in the meantime.

The "d" was meant to resemble the traditional "NdF" notation, but I like this suggestion. It does look more clear that way, and saving space by removing the 'd' would be great. I'm thinking of using "count" and "sides" as the labels. Will work on this and re-align the result.

yusufmte:
Thanks for the suggestion @Boteium ! Changed counter labels to full names and removed letter 'd', and exacted the centering of the result.

image

yusufmte:
Thanks to @Poohl for making a shake-to-roll feature! This feature has been merged into this PR in https://github.com/InfiniTimeOrg/InfiniTime/pull/1326/commits/e29bd40d42ac8b05851784c18c77480ce37b2543 and is working quite nicely.

minacode:
#1894 implements optional apps and was merged

yusufmte:
> #1894 implements optional apps and was merged

I noticed that! Very excited about it. I will update this branch/PR to match shortly.

yusufmte:
Rebased to 1.14! (Very excited about the customizable user app selection implemented in this version of Infinitime.)

I also want to bump this PR in general. It has been very ready for a long time, and continuing to work great in 1.14. The shake-to-roll function implemented by [@Poohl](https://github.com/Poohl) has been working super well also. This has probably been my most frequently used of the PineTime user apps, and have used it in a few different circles with very good success. Sometimes I have taken the PineTime off the band and passed it around during a tabletop game for players to roll. There have been numerous other PRs for dice apps in the past, so there is definitely interest in such an app from the PineTime community.

I know there are a lot of PRs to review and a lot of more important work to do for Infinitime in general -- all on a volunteer basis, which I recognize and greatly appreciate! However, can the maintainers weigh in on the plans for this at some point? Are there plans to add it to the 1.15 milestone? Thanks!

src/displayapp/screens/Dice.cpp
```diff
@@ -0,0 +1,199 @@
...
+#include "displayapp/screens/Screen.h"
+#include "displayapp/screens/Symbols.h"
+#include "components/settings/Settings.h"
+#include "components/motor/MotorController.h"
+#include "components/motion/MotionController.h"
+
+using namespace Pinetime::Applications::Screens;
+
+namespace {
+  static lv_obj_t* MakeLabel(lv_font_t* font,
```
FintasticMan, line 11: Functions in anonymous namespaces don't need to be declared as static, as they already have static linkage by default.

```suggestion
  lv_obj_t* MakeLabel(lv_font_t* font,
```

src/displayapp/screens/Dice.cpp
```diff
@@ -0,0 +1,199 @@
...
+    if (width != 0) {
+      lv_obj_set_width(label, width);
+    }
+    lv_label_set_align(label, labelAlignment);
+    lv_label_set_text(label, text);
+    lv_obj_align(label, reference, alignment, x, y);
+    return label;
+  }
+
+  static void btnRollEventHandler(lv_obj_t* obj, lv_event_t event) {
```
FintasticMan, line 34: Same here.

```suggestion
  void btnRollEventHandler(lv_obj_t* obj, lv_event_t event) {
```

src/displayapp/screens/Dice.cpp
```diff
@@ -0,0 +1,199 @@
...
+                           LV_LABEL_ALIGN_CENTER,
+                           Symbols::dice,
+                           btnRoll,
+                           LV_ALIGN_CENTER,
+                           0,
+                           0);
+
+  // Spagetti code in motion controller: it only updates the shake speed when shake to wake is on...
+  enableShakeForDice = !settingsController.isWakeUpModeOn(Pinetime::Controllers::Settings::WakeUpMode::Shake);
+  if (enableShakeForDice)
```
FintasticMan, line 128: You should put braces around this if-statement.

src/displayapp/screens/Dice.h
```diff
@@ -0,0 +1,62 @@
...
+#include "displayapp/apps/Apps.h"
+#include "displayapp/screens/Screen.h"
+#include "displayapp/widgets/Counter.h"
+#include "displayapp/Controllers.h"
+#include "Symbols.h"
+
+#include <array>
+#include <random>
+
+#define ROLL_HYSTERESIS 10
```
FintasticMan, line 12: We try to avoid using the preprocessor as much as possible, so I'd replace this with a static constexpr value in the class.

src/displayapp/apps/CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ else ()
     set(DEFAULT_USER_APP_TYPES "${DEFAULT_USER_APP_TYPES}, Apps::Paint")
     set(DEFAULT_USER_APP_TYPES "${DEFAULT_USER_APP_TYPES}, Apps::Paddle")
     set(DEFAULT_USER_APP_TYPES "${DEFAULT_USER_APP_TYPES}, Apps::Twos")
+    set(DEFAULT_USER_APP_TYPES "${DEFAULT_USER_APP_TYPES}, Apps::Dice")
```
FintasticMan, line 13: I don't think we want this to be a default app when merging, but we can leave it enabled for now for seeing the memory and flash usage from the tests.

FintasticMan (changes requested):
I really like this app. A couple of concerns: do we really need the C++ random generator, is it that much better than C's rand? Could you try using rand with the bias elimination implemented in #881, and see what the difference in flash usage is? If the difference isn't too big, I'd be very happy to use the better C++ generator.

Another concern is how well the shaking works when lower to sleep is also enabled. Could you check if there are any issues there? (I'm also not a big fan of the hack of enabling shake to wake if it's disabled to get it to work, but I don't think that needs to be solved now)

Thank you so much for this app, it's very nice, and I apologise for how long it's taking.

src/displayapp/screens/Symbols.h
```diff
@@ -34,6 +34,8 @@ namespace Pinetime {
         static constexpr const char* hourGlass = "\xEF\x89\x92";
         static constexpr const char* lapsFlag = "\xEF\x80\xA4";
         static constexpr const char* drum = "\xEF\x95\xA9";
+        static constexpr const char* dice = "\xEF\x94\xA2";
+        static constexpr const char* chartLine = "\xEF\x88\x81";
```
FintasticMan, line 38: Don't think this chartLine symbol is related.

yusufmte: Thank you! Removed in https://github.com/InfiniTimeOrg/InfiniTime/pull/1326/commits/3c4027ab3cf6231f7de64822db51929ab696a8e7

src/displayapp/apps/CMakeLists.txt
yusufmte: Sounds good! :slightly_smiling_face: I just included it because it seemed like most of the user apps were default currently (and it makes it 12 apps to fill up both screens of the list). But I think both options make sense, am happy with what the maintainers feel is most appropriate there!

src/displayapp/screens/Dice.h
yusufmte: Thanks! Great point. Corrected in https://github.com/InfiniTimeOrg/InfiniTime/pull/1326/commits/d4c69757e4f8be9e3b5059cd38f2f8a90a4ba5fa

src/displayapp/screens/Dice.cpp
yusufmte: Great point, thank you -- corrected in https://github.com/InfiniTimeOrg/InfiniTime/pull/1326/commits/d4c69757e4f8be9e3b5059cd38f2f8a90a4ba5fa

src/displayapp/screens/Dice.h
```diff
@@ -0,0 +1,61 @@
...
+        std::array<lv_color_t, 3> resultColors = {LV_COLOR_YELLOW, LV_COLOR_MAGENTA, LV_COLOR_AQUA};
+        uint8_t currentColorIndex;
+        void NextColor();
+
+        Widgets::Counter nCounter = Widgets::Counter(1, 9, jetbrains_mono_42);
+        Widgets::Counter dCounter = Widgets::Counter(2, 99, jetbrains_mono_42);
+
+        bool openingRoll = true;
+        uint8_t currentRollHysteresis = 0;
+        static constexpr uint8_t ROLL_HYSTERESIS = 10;
```
FintasticMan, line 43: We use camelCase for constants as well.

.gitignore
```diff
@@ -50,3 +50,8 @@ src/arm-none-eabi
 
 # clangd
 .cache/
+
+# node
+/node_modules
+/package-lock.json
+/package.json
```
FintasticMan, line 57: These changes are unrelated. Maybe you could make a separate PR for it?

FintasticMan (changes requested):
Just a couple of things.

yusufmte:
@FintasticMan thank you for your kind words & thoughtful feedback/review, I really appreciate it!! :heart: No apology necessary, you folks have been doing amazing work with InfiniTime and I know you have a ton on your plate!

> I really like this app! A couple of concerns: do we really need the C++ random generator, is it that much better than C's rand? Could you try using rand with the bias elimination implemented in #881, and see what the difference in flash usage is?

I noticed that `rand()` is used for RNG in other apps like Paddle and Twos, and so I did initially use it in earlier versions of the Dice. However, there are a few significant flaws with `rand()` (detailed below) that make it inferior to the C++11 `<random>` lib. I think that doesn't matter for the other apps mentioned, where the experience is fine as long as it generally feels random-- but for the Dice, since the whole point of the app is randomness, it feels important for the randomness to be very fair / high quality.

(1) According to official docs, `rand()` is considered deprecated for most needs as it does not guarantee quality of the randomness of the sequence produced. More details [here](https://en.cppreference.com/w/cpp/numeric/random/rand):

>There are no guarantees as to the quality of the random sequence produced. In the past, some implementations of rand() have had serious shortcomings in the randomness, distribution and period of the sequence produced (in one well-known example, the low-order bit simply alternated between 1 and 0 between calls). rand() is not recommended for serious random-number generation needs. It is recommended to use C++11's random number generation facilities to replace rand(). (since C++11)

(2) As you mentioned, there is the possibility of modulo bias with `rand() % x` (where is `RAND_MAX` is not divisible by `x`, then generating smaller numbers is more likely). With `rand()`, it's possible to correct for it with a few extra lines of code, but feels more elegant to instead use the [intended standard library function](https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution), which picks integers uniformly distributed on a specified closed interval.

(3) The `<random>` library has the very neat `std::seed_seq` function used here, which mixes multiple sources of randomness together (as opposed to `srand()`, which takes in just 1 seed), and makes seeds that are distributed over the whole 32-bit range even when the inputs are close or poorly distributed. This is nice here particularly because the PineTime has a few possible sources of randomness that can work together in tandem; the dice app mixes the system tick count and the X/Y/Z components of the PineTime's sensitive motion controller (credit to @w4tsn #1199 for the motion controller idea!), making seed collision vanishingly unlikely.
 
> Another concern is how well the shaking works when lower to sleep is also enabled. Could you check if there are any issues there? (I'm also not a big fan of the hack of enabling shake to wake if it's disabled to get it to work, but I don't think that needs to be solved now)

That's a great point; I am not sure as I haven't previously used lower to sleep on my own PineTime -- I will test this shortly and report back here soon!

FintasticMan:
Thank you for the thorough explanation. I think it shows you've put a lot of thought into this and I agree that using rand() is not well-suited for this application. Let's stick with the C++ random classes you've already implemented then!

yusufmte:
> Another concern is how well the shaking works when lower to sleep is also enabled. Could you check if there are any issues there? (I'm also not a big fan of the hack of enabling shake to wake if it's disabled to get it to work, but I don't think that needs to be solved now)

Tested current version on PineTime hardware, it's working well! The lower-to-sleep & raise-to-wake continues to work inside & outside the Dice app when these are enabled, and the shake-to-roll works as well.

.gitignore
yusufmte: Good point, thank you! Didn't mean to lump these into here. Removed changes to .gitignore in https://github.com/InfiniTimeOrg/InfiniTime/pull/1326/commits/8880acc1f69b34b7e89212549b867d9fc2724fd8 .

src/displayapp/screens/Dice.h
yusufmte: Thanks, fixed in https://github.com/InfiniTimeOrg/InfiniTime/pull/1326/commits/01b241bbde1433acac58a6521debb63ef32ee772 !

FintasticMan (approved):
Awesome! This looks good to me

yusufmte:
Awesome, thanks again!!! Let me know if anything else needed prior to merge!

NeroBurner:
@yusufmte thank you for your hard work and endurance!

We can finally say: [Alea iacta est](https://en.wikipedia.org/wiki/Alea_iacta_est)

yusufmte:
> @yusufmte thank you for your hard work and endurance!
> 
> We can finally say: [Alea iacta est](https://en.wikipedia.org/wiki/Alea_iacta_est)

Thank you friend, and many thanks to all who contributed and reviewed along the way! 🎉 Alea iacta est, indeed! :)

Issue #1223, opened 2022-07-05
Alarm gets dismissed by Notification


mashuptwice:
### Verification

- [X] I searched for similar bug reports and found none was relevant.

### What happened?

While an alarm was playing, a incoming notifications vibration dismissed the alarm.

### What should happen instead?

The alarm should not be interrupted or even dismissed by the notification

### Reproduction steps

1. Set an alarm
2. wait until alarm is vibrating
3. get a notification on a paired device
4. the alarm gets deactivated and the notification is shown on the watch

### Version

1.10.0

SuperPrower:
I've tried looking into this, and here is how I see it going:

1. We get an alarm, this starts an alarm app.
2. We get a notification. Notification event starts a notification preview app, which causes current app (Alarm) to be destroyed:
https://github.com/InfiniTimeOrg/InfiniTime/blob/95ff285991d399498d9bd7f60a503ef7665822ce/src/displayapp/DisplayApp.cpp#L305
3. Alarm app destroyer stops the alarm:
https://github.com/InfiniTimeOrg/InfiniTime/blob/7f45538eb53235ab4015fcf13533796c8759c7bc/src/displayapp/screens/Alarm.cpp#L134

Doesn't look like a bug for me, rather just unintended interaction.

minacode:
The watch could have a global state that captures that the alarm is going. Going to the alarm app and turning off the alarm would reset that state. As long as it is on, the watch is vibrating.
The problem right now is, that the alarm is only allowed to ring as long as the app is open. 

We might need a clear hint to the alarm app though, or people would crazy about how to turn off the vibrations 😄

SuperPrower:
> Going to the alarm app and turning off the alarm would reset that state.
I think a simpler variant could be to have app loader check if alarm is playing and not start notification preview app if it does. I don't believe any other app could hijack the screen like it. However, both this and the variable and checks on it sound like a complication that could be somehow avoided with some smart engineering.

minacode:
Then the notification would go unnoticed. I don't like this either. 

I think, the general problem is bigger than this issue, because notifications, phone calls, alarm, timer and metronome all use vibrations and we need a general strategy to resolve conflicts. 

I think, we should decouple vibrations from the apps, save which patterns are active and run the one with the highest priority. Maybe, we could also jump to the according screen. 

Example:
- alarm is ringing
- got to the alarm screen
- we receive a notification
- the alarm has the higher priority, so nothing changes
[...]

PR #1328, opened 2022-09-14
Vibration scheduler

minacode:
This PR aims to solve #1223 .
It implements intermediate logic between the motor controller and the rest of the system.
The main goal is to prevent simutaneously activation of vibrations.
The side effects of the old behaviour seem to be weird strong vibrations and vibrations cancelling themselves out.

This implementations exposes the following API on the `MotorController`:
`ActivatePhoneCall`, `ActivateTimer`, `ActivateAlarm`, `ActivateNotification` to mark the different vibrations as something that has to happen.
`DeactivatePhoneCall`, `DeactivateTimer`, `DeactivateAlarm` to mark the vibrations as being over.
`SingleVibration` to trigger a vibration for in-app use (eg. metronome). The single vibration is skipped, when another vibration is already happening.

The activated vibrations are handles according to their priority. The priorities from high to low are:
1. phone call
2. timer
3. alarm
4. notification

The active one with the highest priority is run. When it is over, the next highest one is run afterwards. This ensures that none of them is lost.

In the `Update` function, we can also define custom durations/pattern for each of the vibrations, such that they can be easier distinguished.

TODO:
- [] Are chimes as single vibration ok?
- [] Test this code.

### DIFF
--- a/src/systemtask/SystemTask.cpp
+++ b/src/systemtask/SystemTask.cpp
@@ -294,18 +294,18 @@ void SystemTask::Work() {
           if (state == SystemTaskState::Sleeping) {
             GoToRunning();
           }
-          motorController.RunForDuration(35);
+          motorController.ActivateTimer();
           displayApp.PushMessage(Pinetime::Applications::Display::Messages::TimerDone);
           break;
         case Messages::SetOffAlarm:
           if (state == SystemTaskState::Sleeping) {
             GoToRunning();
           }
-          motorController.StartRinging();
+          motorController.ActivateAlarm();
           displayApp.PushMessage(Pinetime::Applications::Display::Messages::AlarmTriggered);
           break;
-        case Messages::StopRinging:
-          motorController.StopRinging();
+        case Messages::StopAlarm:
+          motorController.DeactivateAlarm();
           break;
         case Messages::BleConnected:
           ReloadIdleTimer();
@@ -396,7 +396,7 @@ void SystemTask::Work() {
               GoToRunning();
               displayApp.PushMessage(Pinetime::Applications::Display::Messages::Clock);
             }
-            motorController.RunForDuration(35);
+            motorController.SingleVibration(35);
           }
           break;
         case Messages::OnNewHalfHour:
@@ -408,12 +408,12 @@ void SystemTask::Work() {
               GoToRunning();
               displayApp.PushMessage(Pinetime::Applications::Display::Messages::Clock);
             }
-            motorController.RunForDuration(35);
+            motorController.SingleVibration(35);
           }
           break;
         case Messages::OnChargingEvent:
           batteryController.ReadPowerState();
-          motorController.RunForDuration(15);
+          motorController.SingleVibration(15);
           ReloadIdleTimer();
           if (state == SystemTaskState::Sleeping) {
             GoToRunning();
@@ -429,7 +429,7 @@ void SystemTask::Work() {
           if (state == SystemTaskState::Sleeping) {
             GoToRunning();
           }
-          motorController.RunForDuration(35);
+          motorController.SingleVibration(35);
           displayApp.PushMessage(Pinetime::Applications::Display::Messages::ShowPairingKey);
           break;
         case Messages::BleRadioEnableToggle:
--- a/src/components/alarm/AlarmController.cpp
+++ b/src/components/alarm/AlarmController.cpp
@@ -104,5 +104,5 @@ void AlarmController::StopAlerting() {
     // set next instance
     ScheduleAlarm();
   }
-  systemTask->PushMessage(System::Messages::StopRinging);
+  systemTask->PushMessage(System::Messages::StopAlarm);
 }
--- a/src/components/motor/MotorController.cpp
+++ b/src/components/motor/MotorController.cpp
@@ -42,11 +42,12 @@ void MotorController::StopRinging() {
 }
 
 void MotorController::StopMotor(TimerHandle_t xTimer) {
+  auto* motorController = static_cast<MotorController*>(pvTimerGetTimerID(xTimer));
   nrf_gpio_pin_set(PinMap::Motor);
-  isShortVibrating = false;
-  if (!IsVibrating()) {
+  motorController->isShortVibrating = false;
+  if (!motorController->IsVibrating()) {
     // current vibration pattern is over, so check for next one
-    Update();
+    motorController->Update();
   }
 }
 
--- a/src/components/motor/MotorController.cpp
+++ b/src/components/motor/MotorController.cpp
@@ -9,7 +9,7 @@ void MotorController::Init() {
   nrf_gpio_cfg_output(PinMap::Motor);
   nrf_gpio_pin_set(PinMap::Motor);
 
-  shortVib = xTimerCreate("shortVib", 1, pdFALSE, nullptr, StopMotor);
+  shortVib = xTimerCreate("shortVib", 1, pdFALSE, this, StopMotor);
   longVib = xTimerCreate("longVib", pdMS_TO_TICKS(1000), pdTRUE, this, Ring);
 }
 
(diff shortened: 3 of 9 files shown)


ght (commented):
I like that this adds priorities to different vibration causes and handles overlapping vibrations instead of just ending them all when any of them is ended.

I don't like the hard-coding of the available kinds of vibrations, their names, and their priorities in the vibration controller. That should be none of it's business IMHO. How about parameterising those, so the API is something like this:
```cpp
uint8_t StartVibration(uint8_t priority, uint8_t duration); // returns an ID
void StopVibration(uint8_t id);
void StopAllVibrations();
```

I'll think about the basic architecture for this some more and prototype something too in the coming days when I get time.

minacode:
I thought about something like this too for a little while, but then went for the simple implementation, because what are we gaining from the abstract one? 

In the long run, I also wanted to add that when a vibration starts, the corresponding app is opened. 

Based on your idea, I could see some kind of initial registration from an app where it submits a config of something like
- it's name
- priority
- duration
- recurrence
And then we (de)activate based on the name.

How do we store them? Making an array of n priorities? Or one with n entries and give out ids like you did?
This leads to cleaner code for the future, but uses more RAM. I like the cleaner API though.

Joheyy:
I think the MotorController should not handle general notifications. If someone wants to implement a visual indication for notifications in the future for example this would all need to be copied or moved somewhere else. 

I think it might make sense to put the handler for this into a separate Notification Controller. It might seem like bloat but I think it might get more important in the future since notifications are one of the main components of this watch.

minacode:
The idea is good and something I had in mind too. 
With this PR, I wanted to bring a needed feature with a reasonable amount of functioning code. We can iterate on this again if needed. I have to admit that I am afraid of overengineering a simple feature to the point where it never gets merged because something might be useful sometimes.

Abstracting away from the exact four apps we have right now and doing so in another class are both reasonable in the future, but both are not needed at the point we are now. So maybe we can merge this PR and solve a real problem and then make a new PR and refactor things that need more attention.

If you all don't agree with me on this then I will do the changes in this PR. I am fine with that too. No bad mood from my side.

minacode:
Thanks for the feedback. I will do this.

minacode:
I am stuck with getting
```
/sources/src/systemtask/SystemTask.cpp: In constructor 'Pinetime::System::SystemTask::SystemTask(Pinetime::Drivers::SpiMaster&, Pinetime::Drivers::St7789&, Pinetime::Drivers::SpiNorFlash&, Pinetime::Drivers::TwiMaster&, Pinetime::Drivers::Cst816S&, Pinetime::Components::LittleVgl&, Pinetime::Controllers::Battery&, Pinetime::Controllers::Ble&, Pinetime::Controllers::DateTime&, Pinetime::Controllers::TimerController&, Pinetime::Controllers::AlarmController&, Pinetime::Drivers::Watchdog&, Pinetime::Controllers::NotificationManager&, Pinetime::Controllers::MotorController&, Pinetime::Drivers::Hrs3300&, Pinetime::Controllers::MotionController&, Pinetime::Drivers::Bma421&, Pinetime::Controllers::Settings&, Pinetime::Controllers::HeartRateController&, Pinetime::Applications::DisplayApp&, Pinetime::Applications::HeartRateTask&, Pinetime::Controllers::FS&, Pinetime::Controllers::TouchHandler&, Pinetime::Controllers::ButtonHandler&)':
/sources/src/systemtask/SystemTask.cpp:84:21: error: cannot bind non-const lvalue reference of type  Pinetime::Controllers::AlertController&' to an rvalue of type 'Pinetime::Controllers::AlertControlle '
   84 |     alertController(motorController),
      |                     ^~~~~~~~~~~~~~~
In file included from /sources/src/systemtask/SystemTask.h:21,
                 from /sources/src/systemtask/SystemTask.cpp:1:
/sources/src/components/alert/AlertController.h:10:7: note:   after user-defined conversion: 'Pinetime::Controllers::AlertController::AlertController(Pinetime::Controllers::MotorController&)'
   10 |       AlertController(MotorController& motorController);
      |       ^~~~~~~~~~~~~~~
```

and can not figure out how to do this right. I would apprechiate some help with this.
All I want to do is adding the AlertController to the SystemTask.

src/systemtask/SystemTask.cpp
```diff
@@ -81,6 +81,7 @@ SystemTask::SystemTask(Drivers::SpiMaster& spi,
     dateTimeController {dateTimeController},
     timerController {timerController},
     alarmController {alarmController},
+    alertController(motorController),
```
Joheyy, line 84: It's not possible to initialize `alertController` of type `Pinetime::Controllers::AlertController` with `motorController` of type  `Pinetime::Controllers::MotorController` since those are two different types.

src/systemtask/SystemTask.cpp
```diff
@@ -56,7 +56,7 @@ SystemTask::SystemTask(Drivers::SpiMaster& spi,
                        Controllers::Ble& bleController,
                        Controllers::DateTime& dateTimeController,
                        Controllers::TimerController& timerController,
-                       Controllers::AlarmController& alarmController,
+                       Pinetime::Controllers::AlarmController& alarmController,
                        Drivers::Watchdog& watchdog,
                        Pinetime::Controllers::NotificationManager& notificationManager,
                        Pinetime::Controllers::MotorController& motorController,
```
Joheyy, line 62: You need to pass an instance of `Pinetime::Controllers::AlertController` around here. It should probably look like:
`Pinetime::Controllers::AlertController& alertController,`

Compare with other controllers like the MotorController.

Joheyy (commented):
Note that you also need to initialize an instance of `AlertController` in `src/main.cpp` and pass it to the initialisation of `SystemTask` starting in line 136 in `main.cpp` (Compare with `motorController`)

src/components/alert/AlertController.cpp
```diff
@@ -0,0 +1,71 @@
+#include "components/motor/AlertController.h"
+
+using namespace Pinetime::Controllers;
+
+AlertController::AlertController(MotorController& motorControllerController)
+  : motorController {motor} {
+}
```
Itai-Nelken, line 7:
```suggestion
AlertController::AlertController(MotorController& motorController)
  : motorController {motorController} {
}
```

Joheyy:
@minacode 
I got it to compile although I haven't tested it yet. But you can take some inspiration from what I did if you want. 

https://github.com/Joheyy/InfiniTime/commit/73b12c649504c8b8e1bed7619b3d29e65e341959

minacode:
Thank you very much! You can make a PR at my fork and I will merge your changes to this PR. 
I am still a little confused why your approach compiles though. For example your change in `SystemTask.cpp` seems to break the pattern of the other lines around.

minacode:
Ok, so the last commit does a few things we surely will discuss.
1. It moves the call of `Update` to the SystemTask. 
This ensures that all active alerts will be run in the case that `Update` will do nothing because a single vibration from an app is running. There is still the possibility of an app spamming single vibrations or doing a very long one, but this wont happen in practice. We could add a `StopAll` function to the MotorController like @ght suggested. This should then only be called by the AlertController to gain priority.

2. It implements support for the DisplayApp to restore the original functionality (apps opening when vibrations start).
`Update` returns whether it started a new vibration. If it did, the SystemTask gets the corresponding Message from `DisplayMessage` and then forwards it to the DisplayApp. This part seems to me like we might refactor this. One idea is to only ping the DisplayApp and letting it catch the value by itself (although that would only be more complicated). Right now, the reference to the AlertController in the DisplayApp is not needed.

mashuptwice:
Looks like this PR would also lay a base for solving problems like #1223 

@minacode What is missing for this to be implemented?

minacode:
Yes, that issue was the start of this work 😊

The missing part is to test it. The functionality is complete, I think. Sorry that I haven't done it yet. My focus went to other PRs.

Also I have to rebase it. This PR is behind by 100+ commits.

mashuptwice:
@minacode ah, it is right in the first sentence, don't know how I could miss that...

minacode:
This is my focus PR once I have time again in the next days. If you want to help you can take a look at the code or test it. It's behind develop but it seems to run stably. We need to find out if it works as intended.

minacode:
There is a minor problem with this PR: when more than one notification happens, the notification app triggers only once and will probably show only the last one. If more than five notifications happen, the first one will be deleted before it could be read by the user. What do you think about it?

mashuptwice:
@minacode I have gadgetbridge set to only allow a notification every 5 seconds anyways.

Issue #1327, opened 2022-09-12
Temperature app


minacode:
### Verification

- [X] I searched for similar feature request and found none was relevant.

### Pitch us your idea!

Make an app that shows the temperature.

### Description

The BMA421/5 can measure temperature. The driver already implements the interface, but the temperature is never used.
We could make an app that shows the temperature. 
I am interested, what values the chip measures. Is it the chips temperature, the body temperature or the outside temperature.
For the latter two, we could maybe approximate the real values from a good parameterized linear function.

joseph58tech:
I never realized the BMA421/5 had a temperature sensor built in. While this is a good discovery, I'm not sure how accurate the temperature displayed will be considering that the sensor is inside of a sealed inclosure.

minacode:
That's why I would like to have some values and wonder if the measurements can be corrected to obtain a "relevant" temperature. My favorite would be the outside temperature.

minacode:
I did #1329 . If you are interested, please take a look at it.

minacode:
I am closing this, because an actual app seems very unlikely, but would like to keep the draft with the experimental implementation open for interested people.

PR #1329, opened 2022-09-14
Experiment: temperature sensor

minacode:
This PR is a test-implementation for #1327 .
I added the temperature value from the driver to the digital watch face.
And the value is: (drum roll...) 1. Always.

If someone sees something I did not see to make this useful, please send help!

I am adding this here as a Draft, because it might be of interest for some people, but is obviously not something we want to merge.

### DIFF
--- a/src/displayapp/screens/WatchFaceDigital.h
+++ b/src/displayapp/screens/WatchFaceDigital.h
@@ -55,7 +55,7 @@ namespace Pinetime {
         // DirtyValue<uint8_t> heartbeat {};
         // DirtyValue<bool> heartbeatRunning {};
         DirtyValue<bool> notificationState {};
-        DirtyValue<bool> temperature {};
+        DirtyValue<uint32_t> temperature {};
 
         lv_obj_t* label_time;
         lv_obj_t* label_time_ampm;
--- a/src/displayapp/screens/WatchFaceDigital.h
+++ b/src/displayapp/screens/WatchFaceDigital.h
@@ -55,7 +55,7 @@ namespace Pinetime {
         // DirtyValue<uint8_t> heartbeat {};
         // DirtyValue<bool> heartbeatRunning {};
         DirtyValue<bool> notificationState {};
-        DirtyValue<uint32_t> temperature {};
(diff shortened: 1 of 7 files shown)


ght:
Thanks for investigating! I was wondering about this too since finding out about it in the datasheet. Kinda disappointing that it does not seem to work.

minacode:
[Here](https://github.com/BoschSensortec/BMA423-Sensor-API/blob/master/examples/generic/temperature/temperature.c) is an example that might be useful.

src/displayapp/screens/WatchFaceDigital.h
```diff
@@ -51,19 +51,21 @@ namespace Pinetime {
...
         DirtyValue<std::chrono::time_point<std::chrono::system_clock, std::chrono::nanoseconds>> currentDateTime {};
         DirtyValue<bool> motionSensorOk {};
-        DirtyValue<uint32_t> stepCount {};
-        DirtyValue<uint8_t> heartbeat {};
-        DirtyValue<bool> heartbeatRunning {};
+        // DirtyValue<uint32_t> stepCount {};
+        // DirtyValue<uint8_t> heartbeat {};
+        // DirtyValue<bool> heartbeatRunning {};
         DirtyValue<bool> notificationState {};
+        DirtyValue<bool> temperature {};
```
ght, line 58:
```suggestion
        DirtyValue<uint32_t> temperature {};
```

ght (changes requested):
I just tested this branch and it actually works fine if you fix the type!

minacode:
Good catch, thank you!

minacode:
Now its always 32000. Do you have useful values?

ght:
Yes, but it changes very slowly. Try leaving it in the fridge for 5 minutes.

minacode:
So I am getting values between 25 and 32 (times 1000). They match the surroundings a little bit, but they change very slowly. 
But I don't think that they are sensitive enough for a temperature app. I will continue the measurements for a while though.

minacode:
I am closing this one because I sadly don't see any future for this.

PR #1333, opened 2022-09-14
AlarmController: Add saving alarm time to file

### LINKED ISSUE #1330: Alarm doesn't persist through resets
### Description

Sometimes devices reboots (by watchdog or because of low battery), then alarm time and settings are lost. As result alarm is missed and device has to be setup again to be used as wakeup source. This feature would be useful for regular repeatable alarms especially.

ght:
Save the set alarm time to the SPI NOR flash, so it does not reset to the default value when the watch resets, e.g. due to watchdog timeout.

Fixes #1330

Size impact: +248 (400332 -> 400580)

- [x] Fix dismissing a ringing alarm
- [x] Test on PineTime

### DIFF
--- a/src/components/alarm/AlarmController.cpp
+++ b/src/components/alarm/AlarmController.cpp
@@ -136,33 +136,35 @@ void AlarmController::SetRecurrence(RecurType recurrence) {
 }
 
 void AlarmController::LoadAlarmFromFile() {
-  lfs_file_t file;
-  AlarmData buffer;
+  lfs_file_t alarmFile;
+  AlarmData alarmBuffer;
 
-  if (fs.FileOpen(&file, "/alarm.dat", LFS_O_RDONLY) != LFS_ERR_OK) {
+  if (fs.FileOpen(&alarmFile, "/alarm.dat", LFS_O_RDONLY) != LFS_ERR_OK) {
     NRF_LOG_WARNING("[AlarmController] Failed to open alarm data file");
     return;
   }
 
-  fs.FileRead(&file, reinterpret_cast<uint8_t*>(&buffer), sizeof buffer);
-  fs.FileClose(&file);
-  if (buffer.version != alarmFormatVersion) {
-    NRF_LOG_WARNING("[AlarmController] Loaded alarm data has version %u instead of %u, discarding", buffer.version, alarmFormatVersion);
+  fs.FileRead(&alarmFile, reinterpret_cast<uint8_t*>(&alarmBuffer), sizeof(alarmBuffer));
+  fs.FileClose(&alarmFile);
+  if (alarmBuffer.version != alarmFormatVersion) {
+    NRF_LOG_WARNING("[AlarmController] Loaded alarm data has version %u instead of %u, discarding",
+                    alarmBuffer.version,
+                    alarmFormatVersion);
     return;
   }
 
-  alarm = buffer;
+  alarm = alarmBuffer;
   NRF_LOG_INFO("[AlarmController] Loaded alarm data from file");
 }
 
 void AlarmController::SaveAlarmToFile() {
-  lfs_file_t file;
-  if (fs.FileOpen(&file, "/alarm.dat", LFS_O_WRONLY | LFS_O_CREAT) != LFS_ERR_OK) {
+  lfs_file_t alarmFile;
+  if (fs.FileOpen(&alarmFile, "/alarm.dat", LFS_O_WRONLY | LFS_O_CREAT) != LFS_ERR_OK) {
     NRF_LOG_WARNING("[AlarmController] Failed to open alarm data file for saving");
     return;
   }
 
-  fs.FileWrite(&file, reinterpret_cast<uint8_t*>(&alarm), sizeof alarm);
-  fs.FileClose(&file);
+  fs.FileWrite(&alarmFile, reinterpret_cast<uint8_t*>(&alarm), sizeof(alarm));
+  fs.FileClose(&alarmFile);
   NRF_LOG_INFO("[AlarmController] Saved alarm data with format version %u to file", alarm.version);
 }
--- a/src/components/alarm/AlarmController.cpp
+++ b/src/components/alarm/AlarmController.cpp
@@ -19,11 +19,13 @@
 #include "systemtask/SystemTask.h"
 #include "task.h"
 #include <chrono>
+#include <libraries/log/nrf_log.h>
 
 using namespace Pinetime::Controllers;
 using namespace std::chrono_literals;
 
-AlarmController::AlarmController(Controllers::DateTime& dateTimeController) : dateTimeController {dateTimeController} {
+AlarmController::AlarmController(Controllers::DateTime& dateTimeController, Controllers::FS& fs)
+  : dateTimeController {dateTimeController}, fs {fs} {
 }
 
 namespace {
@@ -35,12 +37,14 @@ namespace {
 
 void AlarmController::Init(System::SystemTask* systemTask) {
   this->systemTask = systemTask;
+  LoadAlarmFromFile();
   alarmTimer = xTimerCreate("Alarm", 1, pdFALSE, this, SetOffAlarm);
 }
 
 void AlarmController::SetAlarmTime(uint8_t alarmHr, uint8_t alarmMin) {
   hours = alarmHr;
   minutes = alarmMin;
+  SaveAlarmToFile();
 }
 
 void AlarmController::ScheduleAlarm() {
@@ -105,4 +109,39 @@ void AlarmController::StopAlerting() {
     ScheduleAlarm();
   }
   systemTask->PushMessage(System::Messages::StopRinging);
+}
+
+void AlarmController::LoadAlarmFromFile() {
+  lfs_file_t file;
+  AlarmData buffer;
+
+  if (fs.FileOpen(&file, "/alarm.dat", LFS_O_RDONLY) != LFS_ERR_OK) {
+    NRF_LOG_WARNING("[AlarmController] Failed to open alarm data file");
+    return;
+  }
+
+  fs.FileRead(&file, reinterpret_cast<uint8_t*>(&buffer), sizeof buffer);
+  fs.FileClose(&file);
+  if (buffer.version != alarmFormatVersion) {
+    NRF_LOG_WARNING("[AlarmController] Loaded alarm data has version %u instead of %u, discarding", buffer.version, alarmFormatVersion);
+    return;
+  }
+
+  hours = buffer.hours;
+  minutes = buffer.minutes;
+  NRF_LOG_INFO("[AlarmController] Loaded alarm data from file");
+}
+
+void AlarmController::SaveAlarmToFile() {
+  lfs_file_t file;
+  AlarmData buffer = {.version = alarmFormatVersion, .hours = hours, .minutes = minutes};
+
+  if (fs.FileOpen(&file, "/alarm.dat", LFS_O_WRONLY | LFS_O_CREAT) != LFS_ERR_OK) {
+    NRF_LOG_WARNING("[AlarmController] Failed to open alarm data file for saving");
+    return;
+  }
+
+  fs.FileWrite(&file, reinterpret_cast<uint8_t*>(&buffer), sizeof buffer);
+  fs.FileClose(&file);
+  NRF_LOG_INFO("[AlarmController] Saved alarm data with format version %u to file", buffer.version);
 }
--- a/src/components/alarm/AlarmController.cpp
+++ b/src/components/alarm/AlarmController.cpp
@@ -37,13 +37,20 @@ namespace {
 
 void AlarmController::Init(System::SystemTask* systemTask) {
   this->systemTask = systemTask;
-  LoadAlarmFromFile();
   alarmTimer = xTimerCreate("Alarm", 1, pdFALSE, this, SetOffAlarm);
+  LoadAlarmFromFile();
+  if (alarm.isEnabled) {
+    NRF_LOG_INFO("[AlarmController] Loaded alarm was enabled, scheduling");
+    ScheduleAlarm();
+  }
 }
 
 void AlarmController::SetAlarmTime(uint8_t alarmHr, uint8_t alarmMin) {
-  hours = alarmHr;
-  minutes = alarmMin;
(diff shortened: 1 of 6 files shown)


ght:
Yes, both in the simulator and on the PineTime (that's when I caught the typo fixed in the first force-push).

I assumed it was mandatory for a PR (that is not marked as draft) to be tested :)

If you want to test it too, do not forget to verify the firmware before you reboot. I might have had to flash twice…

minacode:
> If you want to test it too, do not forget to verify the firmware before you reboot. I might have had to flash twice…

Lol, this makes it quite dangerous to test those kind of features, right? At least, you are not tinkering with Bluetooth.

minacode:
Since this was an issue in another issue: did you see my comments specific lines?

If not, tldr: you can use the data array for everything and omit the attributes. Also, pack activation and recurrence status also in the data struct because the are important to recover an active alarm.

minacode:
Also, #1278 gets rid of log-functions. I assume that you should follow that.

ght:
> Lol, this makes it quite dangerous to test those kind of features, right? At least, you are not tinkering with Bluetooth.

Not sure what you mean by dangerous. All that happened was that I first forgot to verify before I rebooted the watch, so the bootloader dutifully reverted to the previous firmware and I had to wait another 10 minutes while flashing again (via BLE DFU) for the next try.

> Since this was an issue in another issue: did you see my comments specific lines?

Did you comment on the commit somewhere? I didn't get any notifications besides the ones for your comments in this thread and can't find any comments elsewhere, sorry.

> If not, tldr: you can use the data array for everything and omit the attributes. Also, pack activation and recurrence status also in the data struct because the are important to recover an active alarm.

Good point, not sure how I missed that on the first version. I now added saving the recurrence and the enabled state along with the alarm time. The saved alarm is properly scheduled again after being loaded after boot if it was enabled

> Also, #1278 gets rid of log-functions. I assume that you should follow that.

IMHO that is the wrong way to go and I commented there. Should there be consensus to merge that PR however, I will remove the logging in this PR.

minacode:
> Not sure what you mean by dangerous.

Rebooting to the last verified image is a nice fallback, when something goes wrong. But when testing features requires a reboot one has to verify the test firmware and loses that fallback. 

> Did you comment on the commit somewhere?

Apparently I used that feature wrong. Sorry for the confusion.

toastom:
@minacode Verifying a firmware you're currently testing isn't as dangerous as it sounds at first. This is because if there is some issue with it, you can press and hold the button to reboot until you see the pinecone turn blue. This will rollback to the good now-secondary firmware even after you have validated the bad firmware. Source: [about-software.md](https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/doc/gettingStarted/about-software.md)

ght:
Adding the enabled state to persistence has made dismissing an alarm glitchy, I'll investigate more and fix later when I'm at my PC. Persistence works, but when the alarm rings, you have to exit the alarm app by long pressing the button as a workaround, as swiping down hangs the app.

ght:
The problems dismissing the ringing alarm should be fixed now; I'll mark the tasks in the first post as done once the fix is tested and confirmed working.

src/components/alarm/AlarmController.cpp
```diff
@@ -88,21 +103,66 @@ uint32_t AlarmController::SecondsToAlarm() {
...
 
 void AlarmController::StopAlerting() {
-  // Alarm state is off unless this is a recurring alarm
-  if (recurrence == RecurType::None) {
-    state = AlarmState::Not_Set;
+  isAlerting = false;
+  // Disable alarm unless it is recurring
+  if (alarm.recurrence == RecurType::None) {
+    alarm.isEnabled = false;
+    SaveAlarmToFile();
```
NeroBurner, line 123: instead of saving the file on every little change I recommend doing it like the settings file. It has a `bool settingsChanged = false;` flag. This flag can be set on modification and only if changed write out the new alarm-settings file

src/components/alarm/AlarmController.cpp
```diff
@@ -88,21 +103,66 @@ uint32_t AlarmController::SecondsToAlarm() {
...
+}
+
+void AlarmController::SaveAlarmToFile() {
+  lfs_file_t file;
+  if (fs.FileOpen(&file, "/alarm.dat", LFS_O_WRONLY | LFS_O_CREAT) != LFS_ERR_OK) {
+    NRF_LOG_WARNING("[AlarmController] Failed to open alarm data file for saving");
+    return;
+  }
+
+  fs.FileWrite(&file, reinterpret_cast<uint8_t*>(&alarm), sizeof alarm);
```
NeroBurner, line 165: never have seen `sizeof` as just a keyword without the braces
```suggestion
  fs.FileWrite(&file, reinterpret_cast<uint8_t*>(&alarm), sizeof(alarm));
```

src/components/alarm/AlarmController.cpp
```diff
@@ -88,21 +103,66 @@ uint32_t AlarmController::SecondsToAlarm() {
...
+    NRF_LOG_WARNING("[AlarmController] Loaded alarm data has version %u instead of %u, discarding", buffer.version, alarmFormatVersion);
+    return;
+  }
+
+  alarm = buffer;
+  NRF_LOG_INFO("[AlarmController] Loaded alarm data from file");
+}
+
+void AlarmController::SaveAlarmToFile() {
+  lfs_file_t file;
```
NeroBurner, line 159: nitpick: rename to `alarmFile` to be more explicit which file is meant (to be closer to the `Settings::SaveSettingsFromFile()` function)

```suggestion
  lfs_file_t alarmFile;
```

src/components/alarm/AlarmController.cpp
```diff
@@ -88,21 +103,66 @@ uint32_t AlarmController::SecondsToAlarm() {
...
+
+void AlarmController::SetRecurrence(RecurType recurrence) {
+  if (alarm.recurrence != recurrence) {
+    alarm.recurrence = recurrence;
+    SaveAlarmToFile();
+  }
+}
+
+void AlarmController::LoadAlarmFromFile() {
+  lfs_file_t file;
```
NeroBurner, line 139: nitpick: rename to `alarmFile` to be more explicit which file is meant (to be closer to the `Settings::LoadSettingsFromFile()` function)

```suggestion
  lfs_file_t alarmFile;
```

src/components/alarm/AlarmController.h
```diff
@@ -38,33 +38,44 @@ namespace Pinetime {
...
       }
-      void SetRecurrence(RecurType recurType) {
-        recurrence = recurType;
+      RecurType Recurrence() const {
+        return alarm.recurrence;
       }
+      void SetRecurrence(RecurType recurrence);
 
     private:
+      static constexpr uint8_t alarmFormatVersion = 1;
```
NeroBurner, line 60: I think limiting us to 128 format versions could bite us in the future. I don't expect to have more than 128 format changes, but I think an additional byte would give us a bit more breathing room

```suggestion
      static constexpr uint16_t alarmFormatVersion = 1;
```

src/components/alarm/AlarmController.h
```diff
@@ -38,33 +38,44 @@ namespace Pinetime {
...
+      struct AlarmData {
+        uint8_t version = alarmFormatVersion;
```
NeroBurner, line 62: see above

```suggestion
        uint16_t version = alarmFormatVersion;
```

NeroBurner (changes requested):
I like the persistent Alarm implementation, thanks for the PR

src/components/alarm/AlarmController.cpp
Itai-Nelken: It is valid without the parentheses, The only place where they are needed is for types (`sizeof(int)` for example).
That said I personally do prefer having parentheses.

src/displayapp/screens/Alarm.cpp
```diff
@@ -63,6 +63,7 @@ Alarm::Alarm(DisplayApp* app,
   minuteCounter.Create();
   lv_obj_align(minuteCounter.GetObject(), nullptr, LV_ALIGN_IN_TOP_RIGHT, 0, 0);
   minuteCounter.SetValue(alarmController.Minutes());
+
```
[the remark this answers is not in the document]
ght: Not sure where that came from, wasn't intentional :)

src/components/alarm/AlarmController.cpp
ght: Fixed. I checked and the [Linux kernel coding style](https://www.kernel.org/doc/html/v4.18/process/coding-style.html#spaces) and [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#sizeof) both recommend having parenthesis.

src/components/alarm/AlarmController.cpp
ght: I got derailed trying to pull the save/load functions out to avoid duplicating them and that was a left-over :) IMHO worth doing at some point, but one thing at a time.

src/components/alarm/AlarmController.h
ght: I was about to make it 16 bit originally, but don't think it makes sense. Even if the alarm format gets changed in every future release and we assume there will be a new release every month, 8 bit is still enough for over 20 years.

I added a comment to reserve the last value (`255`), so if – for whatever reason – the 254 versions should be exhausted quicker than expected, format version 255 can include a 64 bit extended version field.

NeroBurner: that sounds reasonable :+1:

src/components/alarm/AlarmController.cpp
NeroBurner: I think everywhere where you call `SaveAlarmToFile` you could instead set `settingsChanges = true;` and in the destructor of the Alarm call `SaveAlarmToFile` if `settingsChanges == true`

Am I missing something?

ght: I actually had that first, but the destructor does not get called if you don't close the alarm app, so we fail the user in the following scenario:
1. User opens the alarm app to set an alarm
2. User does not close the alarm app, watch turns off via idle timeout
3. Watch resets for some reason (maybe User was testing some fancy unstable feature)
The watch will load the last saved alarm when rebooting, either waking the user up at an unwanted time, or worse, not wake the user at the alarm set before the watch reset.

NeroBurner: another idea: let's save on exiting, and additionally immediately always when the on/off state is toggled. Then the enabled or disabled alarm is always stored. And loosing a disabled time because of not exiting the alarm app is tolerable

ght: See [`ce296ec` (#1333)](https://github.com/InfiniTimeOrg/InfiniTime/pull/1333/commits/ce296ece8335b31fe014d9ec0098689647172fea)

ght: One remaining inefficiency is that enabling saves twice: once for `SetAlarmTime` and again for `ScheduleAlarm`, but I think we can live with that. Avoiding that would cost more in binary size than it's worth IMHO.

NeroBurner:
The idea about saving only on toggle and on changes didn't let me go. So I implemented the behavior and tested the firmware sizes to see the impact of my changes. It surprised me, but my desired behavior didn't increase the size anymore
```
   text    data     bss     dec     hex filename
 399524     940   53400  453864   6ece8 pinetime-app-1.10.0.out - c853681d Reduce duplication in Twos (#1274)
 399756     940   53408  454104   6edd8 pinetime-app-1.10.0.out - 9ece9e3c AlarmController: Code style fixes
 399788     940   53408  454136   6edf8 pinetime-app-1.10.0.out - ce296ece Alarm: Only set time when enabling alarm and when exiting  ce296ece8335b31fe014d9ec0098689647172fea
 399788     940   53408  454136   6edf8 pinetime-app-1.10.0.out - neroburner-change, save only on enable toggle and change
```

Opened a PR to make it easy to get those changes into your branch: https://github.com/ght/InfiniTime/pull/1

@ght please review

ght:
@NeroBurner Thanks for trying it out! Cool to see that it does not necessarily increase binary size. This approach however has an API change I don't like: now the AlarmController no longer saves the alarm when data could be lost otherwise, but has to be told to do so explicitly by the caller. And not even consistently: in case of `SetAlarmTime` and `SetRecurrence` the caller has to remember to manually `SaveAlarm`, but in case of `ScheduleAlarm`, `DisableAlarm`, and `StopAlerting` it is saved automatically.

This gives me hope that it's doable without a size increase while keeping a clean API though. I'll look into it again too.

I really do not want the caller to have to remember to save. But the alternative is that the caller has to remember *not* to save unnecessarily, to avoid wearing out the flash.

---

What do you think of adding a timer to the alarm controller? Instead of committing every change to flash immediately, the timer is set to count down e.g. 10s or even just 5s and writing to flash only happens when the timer fires. That way it doesn't matter if the alarm app calls `SetAlarmTime` for every change, e.g. when `+` is held down to spin the minute up by 30.

NeroBurner:
The `SaveAlarm` function is just like the `SaveSettings` function in `Settings.h`. With the modification, that important changes (like alarm on/off) are saved immediately (as you've deemed important earlier https://github.com/InfiniTimeOrg/InfiniTime/pull/1333#discussion_r975739665 ). Everything else (like time and if it is repeating) is saved once the Alarm Screen is closed

`SaveAlarm` can't be overused, because it only does work if one of the settings did change (again just like done in `Settings.h`). This is a simple and consistent API change in my honest opinion.

minacode:
Ok, I like this behavior. We just have to keep in mind that every value change has to disable the alarm. Maybe this should be documented somewhere.

NeroBurner:
Opened a rebased PR https://github.com/InfiniTimeOrg/InfiniTime/pull/1367 with simplified `SaveAlarm()` functionality.

No need to work around bugs in the same firmware with saving the alarm. Instead such bugs should be fixed. This lets us keep the remaining code more simple

NeroBurner:
@ght I hope it is ok I hijack your PR, trying to champion your idea to get it into the firmware as soon as possible

JF002:
@NeroBurner Ok to merge this? I think those changes require changes in InfiniSim too ?

NeroBurner:
closing in favor of https://github.com/InfiniTimeOrg/InfiniTime/pull/1367

PR #1331, opened 2022-09-14
Add bluetooth icon for the analog watchface

tgc-dk:
Fixes the bluetooth part of #424.

PR #1336, opened 2022-09-19
WatchfaceAnalog: use StatusIcons widget

Itai-Nelken:
image

If merged, this pr replaces #1331.

### DIFF
--- a/src/displayapp/screens/WatchFaceAnalog.cpp
+++ b/src/displayapp/screens/WatchFaceAnalog.cpp
@@ -52,10 +52,9 @@ WatchFaceAnalog::WatchFaceAnalog(Pinetime::Applications::DisplayApp* app,
   : Screen(app),
     currentDateTime {{}},
     dateTimeController {dateTimeController},
-    batteryController {batteryController},
-    bleController {bleController},
     notificationManager {notificationManager},
-    settingsController {settingsController} {
+    settingsController {settingsController},
+    statusIcons(batteryController, bleController) {
 
   sHour = 99;
   sMinute = 99;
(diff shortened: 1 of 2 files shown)


NeroBurner (approved):

Riksu9000:
There will be more icons, after which the icons will start overlapping the watch face. I think in the case of this watchface, it's probably better to create custom indicators than use StatusIcons, which is intended to be used on status bars.

FintasticMan:
I personally think that the StatusIcons widget should be changed so that it can display in multiple layouts, for example for the PTS watch face. That would prevent code duplication when adding an icon that should be displayed on the watch face.

NeroBurner:
what if we merge this PR and improve the widget afterwards?

Alternatively a pretty fast improvement would be a "max-items" setting and set it to 3 on the Analog Watch Face. If there are more items then as 4th symbol a dot is drawn to indicate more items. Would this be an acceptable change to prevent overflow on the analog watch face?

JF002:
I don't think extending the StatusIcon widget to fit in the analog watch face is a good idea : the intent of this widget is to implement a status bar that covers the whole width of the display, on the top (hence the `LV_LAYOUT_ROW_TOP` layout). In the future, we might also want to add more icons or info on this widget, which probably won't fit with the analog watch face layout.
We implemented this widget to avoid code duplication in multiple places of the code (digital watch face, settings, app list).

Extending the widget to support another layout is of course possible, but I don't think that would be a good idea : it would make the code of `StatusIcons` more complicated, only to use this alternative layout at only 1 place. In this situation, I think using a custom "status widget" for the analog watch face would be a better idea (which is exactly how it's done right now).

NeroBurner (commented):
sounds valid, revoking approval

Itai-Nelken:
@tgc-dk #1331 can be reopened.

PR #1338, opened 2022-09-22
fix typo: notificatioManager -> notificationManager

minacode:
The title says it all. There is a typo in Clock that made its way into the other watch faces (but not WatchFaceAnalog!).


NeroBurner (approved):
Good catch! This shows the best way to write a new watch face is to copy an existing one 😁

JF002 (approved):

PR #1339, opened 2022-09-22
Refactor watch face to enum

### LINKED ISSUE #1334: Develop: watch face depends on notification status
### Verification

### What happened?

In recent test builds based on develop, the watch face changes not with its setting, but the notification setting from the quick settings. 

The watch face setting changes the watch face. 

### Reproduction steps

1. Build develop. 
2. Flash. 
3. Change the watch face settings and see the watch face doesn't change.
4. Change the notification settings and see the watch face change. 

### More details?

In the setting data, the notification status and the clock face are stored adjacently. Maybe this causes a wrong lookup. The uint8 enum for notification values holds the same values that correspond to the clock face values. I could not find the place where this goes wrong yet. 

Maybe refactoring the clock face values into an enum (like the one for apps) gives a type hint somewhere.

### Version

develop

### Companion app

None.

minacode:
This PR introduces the same enum-implementation for watch faces that already exists for apps. I think it is cleaner than using `uint8_t`.
The implementation of `CheckboxList uses `uint32_t` internally, so there is a conversion in between.

- [x] test on hardware
- [x] set enum values explicitly


src/displayapp/screens/settings/SettingWatchFace.cpp
```diff
@@ -65,7 +66,7 @@ void SettingWatchFace::UpdateSelected(lv_obj_t* object, lv_event_t event) {
     for (unsigned int i = 0; i < options.size(); i++) {
       if (object == cbOption[i]) {
         lv_checkbox_set_checked(cbOption[i], true);
-        settingsController.SetClockFace(i);
+        settingsController.SetClockFace(static_cast<WatchFace>(i));
```
Itai-Nelken, line 69: Maybe SetClockFace() should recieve a WatchFace and do the cast internally?

minacode: The cast is from an uint8_t to a WatchFace, right? SetClockFace already takes a WatchFace.

Itai-Nelken: Oh, I missed that. Sorry.

minacode:
I think we need to add another array to the CheckboxList with the values of each option. Right now, the values default to `uint8_t` indices, but we want to select watch faces. This also adds a generic type to the CheckboxList.

src/displayapp/screens/settings/SettingWatchFace.cpp
```diff
@@ -31,6 +33,7 @@ SettingWatchFace::~SettingWatchFace() {
   settingsController.SaveSettings();
 }
 
+
```
NeroBurner, line 36: I think this newline slipped in

```suggestion
```

src/displayapp/screens/settings/SettingWatchFace.cpp
```diff
@@ -5,7 +5,9 @@
 #include "displayapp/screens/Screen.h"
 #include "displayapp/screens/Styles.h"
 #include "displayapp/screens/Symbols.h"
+#include "displayapp/WatchFaces.h"
```
NeroBurner, line 8: two times the same include, remove one of them please

NeroBurner (changes requested):

minacode:
I guess I will put this PR on hold until #1351 is figured out.

src/displayapp/screens/settings/SettingWatchFace.cpp
minacode: Thank you!

src/displayapp/screens/WatchFaceTerminal.cpp
```diff
@@ -30,7 +30,6 @@ WatchFaceTerminal::WatchFaceTerminal(DisplayApp* app,
     settingsController {settingsController},
     heartRateController {heartRateController},
     motionController {motionController} {
-  settingsController.SetClockFace(3);
```
NeroBurner, line 33: could you make a small PR with just this cleanup line, for it to get merged faster

src/components/settings/Settings.h
```diff
@@ -3,6 +3,7 @@
 #include <bitset>
 #include "components/brightness/BrightnessController.h"
 #include "components/fs/FS.h"
+#include "displayapp/WatchFaces.h"
```
NeroBurner, line 6: other setting enums are defined in `Settings.h`. Do you think it would be a good idea to move the `WatchFace` enum into `Settings.h` as well? Just a comment. I'd like to hear your thoughts on this

minacode: No, because I see watch faces more like apps.

src/displayapp/screens/WatchFaceTerminal.cpp
minacode: I did it in #1423. Thank you! Should be a lot cleaner this way :)

minacode:
Testing results right now: pressing the settings button in the quick settings opens the quick settings again with an animation. The settings do not show at all.

minacode:
New testing result: it runs all well, except the background image of the Infineat watch face is missing. The Casio watch face looks as expected.

src/displayapp/WatchFaces.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+namespace Pinetime {
+  namespace Applications {
+    enum class WatchFace : uint32_t {
```
Riksu9000, line 5: By using uint8_t we should be able to avoid updating the settings version and avoid an unnecessary settings reset.

Riksu9000 (changes requested):
The commits should be fixed according to the commit conventions https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/CONTRIBUTING.md#commit-conventions

minacode:
Can I just squash them?

src/displayapp/WatchFaces.h
FintasticMan: Yep, and to do that we would also need to explicitly give the enumerations values like with WakeUpMode, because otherwise the values are undefined.

minacode: `CheckboxList` uses `uint32_t`. I still changed the enum to `uint8_t` since it should be fine.

FintasticMan:
I think you forgot to update the submodules in your local repo before committing.

Riksu9000:
All the commits seem to be fixing issues from previous commits, so they can all be squashed.

minacode:
Given the current ideas regarding getting rid of the apps enum, I am not sure if this PR makes sense anymore.

Riksu9000:
I'm sure this PR would be merged if the issues were fixed.

minacode:
Alright, here we go 😄

minacode:
I tried it on my watch and there is a bug with the font on the casio watch face that doesn't happen when I'm on 1.12. 
image

Riksu9000 (approved):
I don't think the issue relates to this PR. https://github.com/InfiniTimeOrg/InfiniTime/issues/1376

This can be squash merged.

minacode:
Ah, thank you. 😊 I wasn't aware of this issue anymore.

JF002 (approved):

Issue #1315, opened 2022-09-03
Levels of customization


minacode:
I am opening this issue as a general discussion about customization.
We have the problem that InfiniTime wants to keep unnecessary customization out in favor of good defaults, but we also have a user base that seems to like to tinker on a lot of little details. Both are completely valid. I do not think that "maintain your own fork" is a good default solution for this.

I propose the following three tiers of customization:
1. Settings that can be changed via the UI, because people actually change them on a stable setup (like turning Bluetooth off).
2. Settings that can be changed from the settings file via BLE FS without a UI, so not everything needs a UI. This would need changes on the file and maybe the companion apps. People could share those config with each other. An example could be custom vibration times/patterns.
3. Custom builds, for everything that needs no runtime configuration, like which apps, watch face or general features to include. I am also thinking on building the firmware completely without heart rate monitoring for people who are not interested in this.

ght:
For 2 there is also another way: instead of requiring the companion app, there could be a "secret" config screen like Firefox's `about:config` with very spartan design, basically just a long list of toggles or radio-groups, or something even more primitive.

This under-the-hood config screen should not be easily reachable by mistake by novice users, e.g. hidden behind long pressing on the last page of the firmware infos, or some other konami-code-like swipe/button-press combo.

Joheyy:
@ght I like your idea a lot.
Reminds me of the developer settings in android. So maybe a hidden setting that can be enabled by clicking multiple times on the about screen.

minacode:
I did a successful test in #1296 and if anyone from the project is actually interested in this, yours sounds cleaner regarding the cmake part and we should have a look at it.

nische:
I will take a look at this. The biggest work will be to change the main structure to use all component and screens as own libs and link them only they are used. 

I will do a POC in the next weeks

minacode:
The main thing that needed my code generation script was the order of apps in the app menu.
I followed the order of the apps as given in the config file. 
I don't know if this is possible with CMake.

nische:
There your right. CMake can't insert code snippets to add the apps. 
But the KConfig things can be use as defines in the source code. Then we don't need an additional script. (I think so :-D)

If we reorganize the components/Apps/Screens as small libs with own cmake files we can force the compiler to don't build this, if they don't needed. This will be the cmake part. But I'm not Sure if the work worth it

minacode:
[...]

PR #1341, opened 2022-09-23
Setting to disable step counter recording

### LINKED ISSUE #1308: Disable Step Counter/Goal
### Verification

### Pitch us your idea!

I'd like to be able to disable the step counter or its goal.

### Description

Simply setting the goal count to 0 and never getting notified about it would be fine. I'm just disabled, in a wheelchair most of the time, and it's not a useful metric for me the way heart rate monitoring or getting notifications on my watch are, and it's really not a feature I bought a PineTime for and I'd like to opt out.

taukakao:
Fixes #1308 

I think many people want to use InfiniTime although they are not interested in a step counter. 
This is not possible since you're being reminded about your steps on most watch faces and inside of your companion app. 

This PR adds a setting that makes it possible to almost completely disable step counter functionality.
It aims to only have light changes and not add bloat. (Most of the changed lines is white space)

### DIFF
--- a/src/components/settings/Settings.h
+++ b/src/components/settings/Settings.h
@@ -202,6 +202,15 @@ namespace Pinetime {
         return settings.stepsGoal;
       };
 
+      void SetStepCountEnabled(bool enabled) {
+        settingsChanged = settings.stepCountEnabled != enabled;
+        settings.stepCountEnabled = enabled;
+      };
+
+      bool IsStepCountEnabled() {
+        return settings.stepCountEnabled;
+      }
+
       void SetBleRadioEnabled(bool enabled) {
         bleRadioEnabled = enabled;
       };
@@ -230,6 +239,8 @@ namespace Pinetime {
         std::bitset<4> wakeUpMode {0};
         uint16_t shakeWakeThreshold = 150;
         Controllers::BrightnessController::Levels brightLevel = Controllers::BrightnessController::Levels::Medium;
+
+        bool stepCountEnabled = true;
       };
 
       SettingsData settings;
--- a/src/displayapp/screens/WatchFaceDigital.cpp
+++ b/src/displayapp/screens/WatchFaceDigital.cpp
@@ -61,15 +61,17 @@ WatchFaceDigital::WatchFaceDigital(DisplayApp* app,
   lv_label_set_text_static(heartbeatValue, "");
   lv_obj_align(heartbeatValue, heartbeatIcon, LV_ALIGN_OUT_RIGHT_MID, 5, 0);
 
-  stepValue = lv_label_create(lv_scr_act(), nullptr);
-  lv_obj_set_style_local_text_color(stepValue, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(0x00FFE7));
-  lv_label_set_text_static(stepValue, "0");
-  lv_obj_align(stepValue, lv_scr_act(), LV_ALIGN_IN_BOTTOM_RIGHT, 0, 0);
+  if (settingsController.IsStepCountEnabled()) {
+    stepValue = lv_label_create(lv_scr_act(), nullptr);
+    lv_obj_set_style_local_text_color(stepValue, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(0x00FFE7));
+    lv_label_set_text_static(stepValue, "0");
+    lv_obj_align(stepValue, lv_scr_act(), LV_ALIGN_IN_BOTTOM_RIGHT, 0, 0);
 
-  stepIcon = lv_label_create(lv_scr_act(), nullptr);
-  lv_obj_set_style_local_text_color(stepIcon, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(0x00FFE7));
-  lv_label_set_text_static(stepIcon, Symbols::shoe);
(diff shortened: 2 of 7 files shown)


Joheyy:
Here are screenshots of the affected Screens:

Setting Steps:
InfiniSim_2022-09-23_230112
Digital Watch Face:
InfiniSim_2022-09-23_230213
Watch Face Pine Time Style:
InfiniSim_2022-09-23_230203
Watch Face Terminal:
InfiniSim_2022-09-23_230221

minacode:
What do you think about solving this with some kind of conditional compiling and therefore cutting it from the image instead of adding even more to have less? Refer to my proposal in #1315. I think that the decision to disable step counting is a permanent one.

Joheyy:
@minacode That would be a good idea but not in the current state, in my opinion. 
As long as InfiniTime doesn't have a user interface to easily configure and compile this yourself this functionality would be limited to people who are comfortable in setting up and building on their own which is developers. 
I don't see this as a developer-only feature.

minacode:
I agree with the state right now and the problem that we need to take care of the standard user. 
It would be nice if we get to a good solution for this at some point.

Your implementation seems fine to me.

minacode:
Can you disable the measuring in the background?

Joheyy:
Theoretically, yes.
You'd need to call feature_disable BMA423_STEP_CNTR in bma423.c
But I think that would require quite a lot of changes and since step counting is a hardware feature anyway I don't think that would really have any meaningful impact.

JF002:
According to our [project vision](https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/doc/InfiniTimeVision.md), we tend to favor a solid default experience over customization and settings. 

I personally don't think that the step indicator is intrusive or annoying, even if you don't need that information. Moreover, the impact on battery life is yet to be measured, but I guess it's negligible.

Well this is my own opinion though... and I'm ready to hear your opinion about this ;)

minacode:
Not the author, but I have an opinion on that 😄
I believe in the vision of good defaults. But I also think that the best way to get there is through a community of developers. And they come here because they want to tinker on their watch. I also think that they will never agree on a single version for all people (if they have to use it themselves). Thats why I proposed #1315 . That way we could provide good defaults and easy enough customization for advanced users.

For this particular feature: I don't need it and already stated that I would rather have this implemented in a "less" manner rather than with more code (and more UI!).

Joheyy:
@JF002  
I know the project vision, but it doesn't say anything on the reasoning behind the single points. 
I have no idea why someone would want less customization if not for reducing either UI or code complexity, both of which are almost not affected by this. If you give me your reasoning behind this I might be able to argue against it. 

> I personally don't think that the step indicator is intrusive or annoying

Having unnecessary information on the watch face can be distracting or even irritating. ( #1308 )

JF002:
> I know the project vision, but it doesn't say anything on the reasoning behind the single points.

One of the reasons is that we want InfiniTime to be easy to use and to understand, it should *just work*. Adding many settings and options could make the whole system less predictive, more confusing and more complex to use.

I totally understand this description is very vague, and we have to find a good balance between "this is how it'll work for everyone" and "everything is configurable".

Another reason is that we have to keep the UI "simple" because we are constrained by the CPU power and the memory available on the PineTime. Adding settings app with many switches and settings does use code and RAM space. Not to mention that the display is quite small and low-resolution.

Last but no least : project maintenance. New settings mean that the code becomes more complex and more bug prone, that we have to take care of more use-cases when adding or modifying features and that we have to do more tests before the releases. Adding a simple new setting might seems like a small task, but maintaining it on the long term is not insignificant, and definitely not something I and the Core developer team take lightly.

I can totally understand the frustration when we don't want to add this or that small feature, but we also have to make sure that the project is sustainable in the long term ;-)

Joheyy:
Well, only thing I can argue is that this PR doesn't impact either of these points in any meaningful way.

--------------------------------------------

> it should just work

It still does, if you don't change this setting then nothing changes for the user.

> keep the UI "simple" & the display is quite small and low-resolution

I believe it is still quite simple. Only thing this adds is a single toggle on a page that virtually empty. 

> switches and settings does use code and RAM space

A tiny amount in this case. I can make an analysis tho if that's needed. 

> project maintenance

The step counter display is very straight forward. It's an icon with a label that is updated with a value. I think it's unlikely this part of the code will be edited a lot in the future. 

--------------------------------------------

This is all I have to say about it so I don't see a point in arguing more about it. If this doesn't justify this change please just close this PR and maybe #1308

JF002:
> Well, only thing I can argue is that this PR doesn't impact either of these points in any meaningful way.

You are absolutely right... only if you take this specific PR in consideration. There are many feature requests and PR that want to add a setting which default value won't change anything for the user, that only needs a single setting page with a simple UI and small memory footprint and that is not too intrusive in the code. 

If we added all those settings right away, we would end up with a huge settings menu, a combinatorial explosion of code paths and test cases, and with a code base full of `if(settingA) {doThis();} else {doThat();` everywhere.

I'm not saying we won't add any new settings, and I've not set my mind on this one yet, but I definitely want to keep the code base clean and maintainable. Maybe we need to think out of the box : why do we want to hide the step counter? Maybe we could change the way it's displayed on the UI so that it's less intrusive or disturbing? Or display it only when we detect that new steps are counted? Or maybe there are other UI elements we want to hide/show according to specific events or settings, so that we could handle them with a single setting?

PR #1346, opened 2022-09-30
Alarm: Display the time until alarm, even if it is off

maksalees:
This is version of #698 with a cleaner commit history and rebased.

Previously, if the alarm was turned off, when you pressed the information button, a message was displayed that the alarm was not set. The amount of time until the alarm is now displayed as well as if it is on.
What for? Allows you to save two clicks on a button if you just need to see how much time is left before the alarm

## Before
Before

## After
After


NeroBurner:
please rebase on current `develop` branch to fix the CI-jobs

Riksu9000:
In what situation would this be useful?

maksalees:
> In what situation would this be useful?

For example, when choosing a time for the alarm clock, so that there is enough time to sleep. In this case you should press fewer buttons.

maksalees:
> please rebase on current `develop` branch to fix the CI-jobs

I will rebase soon.

maksalees:
> > > In what situation would this be useful?
> > 
> > 
> > For example, when choosing a time for the alarm clock, so that there is enough time to sleep. In this case you should press fewer buttons.
> 
> I don't see it, because if I needed to be somewhere, I'd have to set a time regardless of how much sleep I'll get. If I don't have somewhere to be, or something to do at a certain time, I wouldn't set an alarm.

But it might be useful for someone.

Riksu9000:
That's an interesting idea, however because it's not an intended use case, there's no guarantee it will continue to be supported in a convenient way at least. For example if we had multiple alarms and the time setting was changed in some ways. Also setting the time can take a lot of time, so the reduction probably wouldn't be very significant in comparison.

On the other hand, the current behavior may help explain the state of the alarm for someone new to using it and give them confidence that it's set correctly.

PR #1348, opened 2022-09-30
feat(Notifications): Adjust alert layout to show more text

FrazerClews:
To make better use of the space available in the header, it makes sense to increase the width of the `alert_type`.

To avoid confusion with the `alert_counter`, a `｜` will separate the `alert_counter` from the `alert_type`.

I do have a branch ready which furthers this PR by making better use of the height of the header space as there is a lot of space. This allows more text in the subject content, but this may cause accessibility issues.

before
after


Commenter25 (approved):
Changes seem fully functional. For anyone wondering, here's a picture of the branch with a shortened header. I personally like it, but I can easily see it feeling cramped.

Screenshot of the branch increasing the notification height

RageGamerBoi:
Is there a way to slightly speed up the text as well?
For long notifications this would be very helpful.

NeroBurner (approved):
nice improvement with minimal change, I like it

Improving the scroll speed should be done in a new PR to keep this one minimal

The firmware build is failing because of the branch name containing a `/` character, the PR itself seems fine

FrazerClews:
Sorry, accidentally closed this PR trying to change the branch name, made another PR here https://github.com/InfiniTimeOrg/InfiniTime/pull/1422

PR #1350, opened 2022-10-02
Implement a return app stack

Riksu9000:
Each opened app (screen) is pushed on a stack, which is then popped from when returning instead of hard coded "previous apps". Return swipe and refresh directions are automatically determined from the app transition.

InfiniSim_2022-10-02_190230


FintasticMan:
Thanks for making this! It has been something that has been on my mind for a while now, and I was thinking I might try my hand at it myself, but you beat me to it :laughing:! I really like the way you've implemented a generic stack, so that it can be reused for other things as well.

It might be out of scope for this PR, but I was also thinking of having the transition direction determined by the "calling" app, instead of having a big switch statement that globally defines the transition direction for each app. This would allow for an app to be available from multiple screens, with a different transition for each.

FintasticMan:
I've been running this on my main watch for quite a while now, and the only issue I've encountered is that if a notification comes in while another one is still being previewed, it will return to the same notification preview screen, and buzz again.

Riksu9000:
I've fixed the issue by making it so the same app won't be added twice on the return stack. Properly fixing it to display the other notification would require a lot more changes.

JF002 (approved):

PR #1351, opened 2022-10-02
Refactoring of CheckBoxList : remove dependency to SettingController

JF002:
I'm not completely satisfied of the design of CheckBoxList. It currently depends on SettingsController and receives pointers to function to get and set the setting.

CheckBoxList is a UI component that we should be able to use anywhere. But the current implementation limit the usage of this class to settings that are handled by the `Settings` controller.

I tried to refactor this class : it now receives the `originalValue` (which is used to select the corresponding checkbox) and a `std::function` that is called when the value changes. This allows to completely de-couple `CheckBoxList` from `Settings`.

Now, I know that lambda with capture and std::function can have a big overhead (heap allocation, binary size and execution time). I couldn't find any in this specific case, but I wouldn't mind someone cross-checking this!

Any feedback about this refactoring?

### DIFF
--- a/src/displayapp/screens/CheckboxList.h
+++ b/src/displayapp/screens/CheckboxList.h
@@ -1,12 +1,12 @@
 #pragma once
 
-#include <lvgl/lvgl.h>
-#include <cstdint>
-#include <memory>
-#include <array>
-#include "displayapp/screens/Screen.h"
 #include "displayapp/Apps.h"
-#include "components/settings/Settings.h"
+#include "displayapp/screens/Screen.h"
+#include <array>
+#include <cstdint>
+#include <functional>
+#include <lvgl/lvgl.h>
+#include <memory>
 
 namespace Pinetime {
   namespace Applications {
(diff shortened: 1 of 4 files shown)


minacode:
For #1339 it would be good to allow any values behind the options (in this case WatchFaces). We could add this with a generic type. Instead of calculating the selected value directly from the index of the option, we would also need an array of values of the generic type corresponding to the options.

Avamander (approved):
Looks good to me. I trust it has been tested on physical hw.

src/displayapp/screens/CheckboxList.h
```diff
@@ -1,47 +1,40 @@
...
-        void (Controllers::Settings::*SetOptionIndex)(uint8_t);
-        uint8_t (Controllers::Settings::*GetOptionIndex)() const;
+        std::function<void(uint32_t)>OnValueChanged;
         std::array<const char*, MaxItems> options;
         std::array<lv_obj_t*, MaxItems> cbOption;
         std::array<lv_point_t, 2> pageIndicatorBasePoints;
         std::array<lv_point_t, 2> pageIndicatorPoints;
         lv_obj_t* pageIndicatorBase;
         lv_obj_t* pageIndicator;
+        uint32_t newValue;
```
NeroBurner, line 37: I'd rename the `newValue` to `currentValue` or even simpler `value`. For something to be new, there must be something old. But for the checkbox the `value` is the internal representation

JF002: Done!

Issue #1342, opened 2022-09-25
Notify on low battery


minacode:
### Verification

- [X] I searched for similar feature request and found none was relevant.

### Pitch us your idea!

Send a notification on low battery

### Description

The watch alerts the user once with a notification when the battery level goes under a threshold of maybe 50% (see #804) and the watch is not loading.

Companion apps can do this, but the watch should be able to do this by itself. I claim that no setting is needed.

I can do this in the next days, but if someone wants to do this instead, just comment below and I am fine with that.

RageGamerBoi:
This goes along perfectly with my issue #1321 
If somebody is working on this in the future, consider looking into that as well.

TheAwesome98-Real:
yes I've wanted this feature since I've had the watch

minacode:
If #1294 gets accepted, we could implement a low power mode that dims the screen, shortens the display time and maybe turns off bluetooth or touch wake up. The SystemTask could turn it on along with the issued notification.

minacode:
@RageGamerBoi: I completed the PR. The way I implemented it should also work for your shutdown feature in another PR. Just test for a threshold and send a message to the SystemTask which should be able to do what you wanted.

PR #1352, opened 2022-10-03
Low battery warning

minacode:
This PR implements a low battery warning (issue #1342) when the battery percentage reaches 15%. 
I added the threshold to the BatteryController that can be checked via `BatteryController::BatteryIsLow()`.
The BatteryController sends a `LowBattery` message to the SystemTask that reacts by pushing a notification.

I also updated the BatteryInfo settings screen to use the threshold to color its charging bar differently.


minacode:
@NeroBurner, since the battery percentage is now fixed and you already accepted my PR for InfiniSim I would appreciate if you could also take a look at this PR. But no pressure, I know you have a lot to do.

src/components/battery/BatteryController.h
```diff
@@ -39,6 +42,7 @@ namespace Pinetime {
       static constexpr nrf_saadc_input_t batteryVoltageAdcInput = NRF_SAADC_INPUT_AIN7;
       uint16_t voltage = 0;
       uint8_t percentRemaining = 0;
+      uint8_t lastPercentRemaining = 0;
```
NeroBurner, line 45: I'd replace this with a `bool batteryLowNotified = false;` to make the variable and conditions more explicit

minacode: Thank you! I will do this.

minacode: Resetting the flag if the watch is loading is enough, right? :grin:

NeroBurner: I think so, yes :+1:

src/systemtask/SystemTask.cpp
```diff
@@ -425,6 +425,15 @@ void SystemTask::Work() {
         case Messages::BatteryPercentageUpdated:
           nimbleController.NotifyBatteryLevel(batteryController.PercentRemaining());
           break;
+        case Messages::LowBattery: {
+          Pinetime::Controllers::NotificationManager::Notification notif;
+          std::array<char, 101> message {"Low Battery\0Charge your watch to prevent data loss.\0"};
```
Boteium, line 430: Let's change <char, 101> to <char, Pinetime::Controllers::NotificationManager::MessageSize + 1>

minacode:
Thank you for taking the time you did! I will keep this updated.

JF002:
Has anyone already had the opportunity to test this PR since we merged #1444 ? Does the low battery warning feels accurate?

minacode:
I tested it for stability before, but can do it again for the feeling. You are asking if 15% are a good threshold value, right?

JF002:
> You are asking if 15% are a good threshold value, right?

Yes, and also if the battery level reported by the new algorithm works good enough for this use-case.

minacode:
Ok, I will give it another try and report back in a few days.

src/systemtask/SystemTask.cpp
```diff
@@ -425,6 +425,16 @@ void SystemTask::Work() {
         case Messages::BatteryPercentageUpdated:
           nimbleController.NotifyBatteryLevel(batteryController.PercentRemaining());
           break;
+        case Messages::LowBattery: {
+          Pinetime::Controllers::NotificationManager::Notification notif;
+          std::array<char, Pinetime::Controllers::NotificationManager::MessageSize + 1> message {
+            "Low Battery\0Charge your watch to prevent data loss.\0"};
+          notif.message = message;
```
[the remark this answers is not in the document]
minacode: The `{}` init syntax is needed to write the C string to the `std::array`. A simple assignment does not work as far as I know. My hope is that the compiler figures that out.
Cleaning up this interface would be nice though. The easiest two methods would be
- moving the default fields `id` and `valid` to the end of the struct to allow unnamed struct initialization
- C++20 allows C style init of structs with `{ .name = value }`.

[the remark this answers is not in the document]
minacode: Lets leave it here as it is and I suggest that I make a PR in which I do the reordering.

NeroBurner:
15% gave me about 22 hours to charge the PineTime (although I think the current watch is at the upper limit of battery life with nearly 10 days from 100% to 0%)

no notification, just the linear interpolated battery curve

two full discharge curves:
Gadgetbridge two full discharge curves

Gadgetbridge 15% mark Gadgetbridge 12% mark Gadgetbridge 0% mark

minacode:
Sounds good.

minacode:
It just crashed at 15% battery and ended in a boot loop. Some of the last changes must have broken something :thinking: Good that we tested it again.

minacode:
My battery lasted approximately another day after reaching 15%. I saw 0%, so the new values are good on my watch. 
We can set the threshold to 20% if you really want to be sure though. 

Independently I fixed the crashes with a `std::move` but tested this with a threshold of 100% because I didn't want to wait so long for debugging 😉

src/systemtask/SystemTask.cpp
```diff
@@ -424,6 +424,16 @@ void SystemTask::Work() {
         case Messages::BatteryPercentageUpdated:
           nimbleController.NotifyBatteryLevel(batteryController.PercentRemaining());
           break;
+        case Messages::LowBattery: {
+          Pinetime::Controllers::NotificationManager::Notification notif;
+          std::array<char, Pinetime::Controllers::NotificationManager::MessageSize + 1> message {
+            "Low Battery\0Charge your watch to prevent data loss.\0"};
+          notif.message = std::move(message);
+          notif.size = 53;
```
JF002, line 432:
```suggestion
          constexpr char message[] = "Low Battery\0Charge your watch to prevent data loss.\0";
          constexpr size_t messageSize = std::min(sizeof(message), Pinetime::Controllers::NotificationManager::MaximumMessageSize());
          std::memcpy(notif.message.data(), message, messageSize);
          notif.size = messageSize;
```

With this, you avoid hard-coding the size of the string (we'll most probably forget to update it if we change the message). It also checks that the content of the message fits into the notification buffer.

JF002 (changes requested):

src/systemtask/SystemTask.cpp
minacode: Thank you! This is much better 😊

JF002 (approved):

PR #1358, opened 2022-10-05
Fix PTS steps style button label

kieranc:
Building current develop results in the new PTS 'Steps style' button having a blank/missing label. I think this is due to a conflict with #1273 and is fixed by this change.

edit: this problem may also affect Infineat watchface buttons


NeroBurner (approved):
Thanks for the quick fix!

NeroBurner:
Tested on my pinetime. Settings menu works as expected now. Thanks!

Riksu9000 (approved):

PR #1359, opened 2022-10-05
Introduce a restricted clock-only-mode when not woken with the button

tgc-dk:
To avoid random touches ends up changing settings or starting a timer, this introduces a restricted mode where no touch input is accepted unless the hardware button is used to turn on the watch. ~~A settings is introduces to enable this.~~
~~The restricted mode is only active on watchfaces.~~

This is not a perfect solution since it will require the button to be pressed several times to leave the restricted mode once in it.

---

edit 2025-05-24: updated description by @NeroBurner :

When the device is woken through pressing the button everything is the same.

But when woken through other means like single-tap or raise-to-wake,
then the screen is locked (and showing a lock screen on touch input)
until the button is pressed.
    
Only exception is when the alarm wakes the screen, then touch input is
still valid for the user to be able to press the red "stop alarm"
button.

### DIFF
--- a/src/displayapp/widgets/PopupMessage.cpp
+++ b/src/displayapp/widgets/PopupMessage.cpp
@@ -1,0 +1,75 @@
+#include "displayapp/widgets/PopupMessage.h"
+#include "displayapp/InfiniTimeTheme.h"
+#include <libraries/log/nrf_log.h>
+
+using namespace Pinetime::Applications::Widgets;
+
+namespace {
+  void event_handler(lv_obj_t* obj, lv_event_t event) {
+    auto* popupMessage = static_cast<PopupMessage*>(obj->user_data);
+    if (event == LV_EVENT_DELETE) {
+      popupMessage->HandleDelete(obj, event);
+    }
+  }
+}
+
+void PopupMessage::HandleDelete(lv_obj_t* object, lv_event_t event) {
+  if (object == popup && event == LV_EVENT_DELETE) {
+    // make sure to update the state of the popup pointer when someone
+    // deletes the popup object
+    popup = nullptr;
+    isHidden = true;
+  }
+}
+
+void PopupMessage::Create() {
+  popup = lv_obj_create(lv_scr_act(), nullptr);
+  lv_obj_set_size(popup, 90, 90);
+  lv_obj_align(popup, lv_scr_act(), LV_ALIGN_CENTER, 0, 0);
+  lv_obj_set_style_local_bg_color(popup, LV_OBJ_PART_MAIN, LV_STATE_DEFAULT, Colors::bg);
+  lv_obj_set_style_local_bg_opa(popup, LV_OBJ_PART_MAIN, LV_STATE_DEFAULT, LV_OPA_60);
+  lv_obj_t* lockBody = lv_obj_create(popup, nullptr);
+  lv_obj_set_size(lockBody, 55, 50);
+  lv_obj_align(lockBody, popup, LV_ALIGN_CENTER, 0, 10);
+
+  lv_obj_set_style_local_bg_color(lockBody, LV_OBJ_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_WHITE);
+  lv_obj_set_style_local_bg_opa(lockBody, LV_OBJ_PART_MAIN, LV_STATE_DEFAULT, LV_OPA_0);
+  lv_obj_set_style_local_border_color(lockBody, LV_OBJ_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_WHITE);
+  lv_obj_set_style_local_border_width(lockBody, LV_OBJ_PART_MAIN, LV_STATE_DEFAULT, 22);
+  lv_obj_set_style_local_border_side(lockBody, LV_OBJ_PART_MAIN, LV_STATE_DEFAULT, LV_BORDER_SIDE_FULL);
+  lv_obj_set_style_local_border_opa(lockBody, LV_OBJ_PART_MAIN, LV_STATE_DEFAULT, LV_OPA_100);
+
+  lv_obj_t* lockTop = lv_obj_create(popup, nullptr);
+  lv_obj_set_size(lockTop, 30, 35);
+  lv_obj_align(lockTop, popup, LV_ALIGN_CENTER, 0, -20);
+  lv_obj_set_style_local_bg_color(lockTop, LV_OBJ_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_WHITE);
+  lv_obj_set_style_local_bg_opa(lockTop, LV_OBJ_PART_MAIN, LV_STATE_DEFAULT, LV_OPA_0);
+  lv_obj_set_style_local_border_color(lockTop, LV_OBJ_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_WHITE);
+  lv_obj_set_style_local_border_width(lockTop, LV_OBJ_PART_MAIN, LV_STATE_DEFAULT, 6);
+  lv_obj_set_style_local_border_side(lockTop, LV_OBJ_PART_MAIN, LV_STATE_DEFAULT, LV_BORDER_SIDE_FULL);
+  lv_obj_set_style_local_border_opa(lockTop, LV_OBJ_PART_MAIN, LV_STATE_DEFAULT, LV_OPA_100);
+
+  // handler to keep popup pointer in sync
+  popup->user_data = this;
+  lv_obj_set_event_cb(popup, event_handler);
+
+  lv_obj_set_hidden(popup, isHidden);
+}
+
+void PopupMessage::SetHidden(const bool hidden) {
+  if (isHidden == hidden) {
+    return;
+  }
+  isHidden = hidden;
+  // create/delete on demand
+  if (popup == nullptr && !isHidden) {
+    Create();
+  } else if (popup != nullptr) {
+    lv_obj_del(popup);
+    popup = nullptr;
+  }
+}
+
+bool PopupMessage::IsHidden() const {
+  return isHidden;
+}
--- a/src/displayapp/widgets/PopupMessage.h
+++ b/src/displayapp/widgets/PopupMessage.h
@@ -1,0 +1,18 @@
+#pragma once
+#include <lvgl/lvgl.h>
+
+namespace Pinetime::Applications::Widgets {
+  class PopupMessage {
+  public:
+    PopupMessage() = default;
+    void Create();
+    void SetHidden(bool hidden);
+    bool IsHidden() const;
+    // public function only to be used by lvgl handler
+    void HandleDelete(lv_obj_t* object, lv_event_t event);
+
+  private:
+    lv_obj_t* popup = nullptr;
+    bool isHidden = true;
+  };
+}
--- a/src/systemtask/SystemTask.cpp
+++ b/src/systemtask/SystemTask.cpp
@@ -232,6 +232,7 @@ void SystemTask::Work() {
           }
           break;
         case Messages::SetOffAlarm:
+          unlockedByButton = true; // unlock so it is possible to press red stop button
           GoToRunning();
           displayApp.PushMessage(Pinetime::Applications::Display::Messages::AlarmTriggered);
           break;
@@ -241,6 +242,7 @@ void SystemTask::Work() {
           bleDiscoveryTimer = 5;
           break;
         case Messages::BleFirmwareUpdateStarted:
+          unlockedByButton = true; // prevent no screen-locked popup on firmware update
           GoToRunning();
           wakeLocksHeld++;
           displayApp.PushMessage(Pinetime::Applications::Display::Messages::BleFirmwareUpdateStarted);
@@ -268,7 +270,14 @@ void SystemTask::Work() {
             break;
           }
           if (state == SystemTaskState::Running) {
-            displayApp.PushMessage(Pinetime::Applications::Display::Messages::TouchEvent);
+            if (unlockedByButton) {
+              displayApp.PushMessage(Pinetime::Applications::Display::Messages::TouchEvent);
+            } else {
+              auto gesture = touchHandler.GestureGet();
+              if (gesture != Pinetime::Applications::TouchEvents::None) {
+                displayApp.PushMessage(Pinetime::Applications::Display::Messages::ShowIgnoreTouchPopup);
+              }
+            }
           } else {
             // If asleep, check for touch panel wake triggers
             auto gesture = touchHandler.GestureGet();
@@ -290,6 +299,7 @@ void SystemTask::Work() {
             action = buttonHandler.HandleEvent(Controllers::ButtonHandler::Events::Press);
             // This is for faster wakeup, sacrificing special longpress and doubleclick handling while sleeping
             if (IsSleeping()) {
+              unlockedByButton = true;
               fastWakeUpDone = true;
               GoToRunning();
               break;
@@ -331,6 +341,8 @@ void SystemTask::Work() {
           } else {
             state = SystemTaskState::AODSleeping;
           }
+          // lock when going to sleep
+          unlockedByButton = false;
           break;
         case Messages::OnNewDay:
           motionSensor.ResetStepCounter();
@@ -437,12 +449,14 @@ void SystemTask::GoToSleep() {
     return;
   }
   NRF_LOG_INFO("[systemtask] Going to sleep");
+  displayApp.PushMessage(Pinetime::Applications::Display::Messages::HideIgnoreTouchPopup);
   if (settingsController.GetAlwaysOnDisplay()) {
     displayApp.PushMessage(Pinetime::Applications::Display::Messages::GoToAOD);
   } else {
     displayApp.PushMessage(Pinetime::Applications::Display::Messages::GoToSleep);
   }
   heartRateApp.PushMessage(Pinetime::Applications::HeartRateTask::Messages::GoToSleep);
+  unlockedByButton = false;
 
   state = SystemTaskState::GoingToSleep;
 };
@@ -481,10 +495,21 @@ void SystemTask::HandleButtonAction(Controllers::ButtonActions action) {
     case Actions::Click:
       // If the first action after fast wakeup is a click, it should be ignored.
       if (!fastWakeUpDone) {
-        displayApp.PushMessage(Applications::Display::Messages::ButtonPushed);
+        if (!unlockedByButton) {
+          // the first button event unlocks the touch input
+          unlockedByButton = true;
+          displayApp.PushMessage(Pinetime::Applications::Display::Messages::HideIgnoreTouchPopup);
+        } else {
+          displayApp.PushMessage(Applications::Display::Messages::ButtonPushed);
+        }
       }
       break;
     case Actions::DoubleClick:
+      if (!unlockedByButton) {
+        // the first button event unlocks the touch input
+        unlockedByButton = true;
+        displayApp.PushMessage(Pinetime::Applications::Display::Messages::HideIgnoreTouchPopup);
+      }
       displayApp.PushMessage(Applications::Display::Messages::ButtonDoubleClicked);
       break;
     case Actions::LongPress:
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -403,6 +403,7 @@ list(APPEND SOURCE_FILES
         displayapp/widgets/PageIndicator.cpp
         displayapp/widgets/DotIndicator.cpp
         displayapp/widgets/StatusIcons.cpp
+        displayapp/widgets/PopupMessage.cpp
 
         ## Settings
         displayapp/screens/settings/QuickSettings.cpp
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -485,6 +485,12 @@ void DisplayApp::Refresh() {
         LoadNewScreen(Apps::Clock, DisplayApp::FullRefreshDirections::None);
         motorController.RunForDuration(35);
         break;
+      case Messages::ShowIgnoreTouchPopup:
+        popupMessage.SetHidden(false);
+        break;
+      case Messages::HideIgnoreTouchPopup:
+        popupMessage.SetHidden(true);
+        break;
     }
   }
 
@@ -508,6 +514,11 @@ void DisplayApp::LoadNewScreen(Apps app, DisplayApp::FullRefreshDirections direc
   // This is mainly to fix an issue with receiving two notifications at the same time
   // and shouldn't happen otherwise.
   if (app != currentApp) {
+    // We need to remove the popup
+    // If we keep the popup linked to the previous view, and this view is deleted, a bug will occur if we try to re-remove the popup.
+    // Not removing the popup will also prevent the popup to be raised on top of
+    // the new app
+    popupMessage.SetHidden(true);
     returnAppStack.Push(currentApp);
     appStackDirections.Push(direction);
   }
--- a/src/displayapp/DisplayApp.h
+++ b/src/displayapp/DisplayApp.h
@@ -16,6 +16,7 @@
 #include "components/stopwatch/StopWatchController.h"
 #include "components/alarm/AlarmController.h"
 #include "touchhandler/TouchHandler.h"
+#include "displayapp/widgets/PopupMessage.h"
 
 #include "displayapp/Messages.h"
 #include "BootErrors.h"
@@ -105,6 +106,7 @@ namespace Pinetime {
       Pinetime::Controllers::FirmwareValidator validator;
       Pinetime::Components::LittleVgl lvgl;
       Pinetime::Controllers::Timer timer;
+      Pinetime::Applications::Widgets::PopupMessage popupMessage;
 
       AppControllers controllers;
       TaskHandle_t taskHandle;
--- a/src/displayapp/Messages.h
+++ b/src/displayapp/Messages.h
@@ -24,6 +24,8 @@ namespace Pinetime {
         AlarmTriggered,
         Chime,
         BleRadioEnableToggle,
+        ShowIgnoreTouchPopup,
+        HideIgnoreTouchPopup
       };
     }
   }
--- a/src/systemtask/SystemTask.h
+++ b/src/systemtask/SystemTask.h
@@ -144,6 +144,8 @@ namespace Pinetime {
       void UpdateMotion();
       static constexpr TickType_t batteryMeasurementPeriod = pdMS_TO_TICKS(10 * 60 * 1000);
 
+      bool unlockedByButton = true;
+
       SystemMonitor monitor;
     };
   }


minacode:
> The restricted mode is only active on watchfaces.

Why is this necessary? Wouldn't this be useful regardless of the active app?

tgc-dk:
> > The restricted mode is only active on watchfaces.
> 
> Why is this necessary? Wouldn't this feature be useful regardless of the active app?

The reasoning is that if you (for example) are playing the "2" game and the screen goes to sleep due to no activity, if you then wake up the screen by shaking your arm, touch input will be ignored and if you push the hardware button you will exit the game, loosing the progress.
One way to solve this is of course to make the first hardware button push unlock the restricted mode, but IMHO this would probably be a bit confusing for users since it would change how the button-push is handled depending on how the wakeup was done...

minacode:
I think this is exactly how this should be done. The watch goes into a lock mode where the first button press does nothing but exiting that mode. Maybe we could have a popup with a lock symbol and something like "press the button to unlock" on touch events so users can better understand what is going on.

Another thing is a lock screen. We could always show the watch face on non-button wakeup (with a clue that this is a lock screen) and after pressing the button go back to the current screen. This is closer to what other devices like phones do.

Joheyy:
I agree with @minacode although I don't think a real "lock screen" makes sense. Maybe a visual indicator but a lock screen for a watch that only has a couple of modes is a bit much. 

One reason I prefer it is because it enables you to have for example the navigation or timer app open on the watch and look at it from time to time. Right now this would always randomly switch to different watch faces on it's own.

Joheyy:
And because I know it will be mentioned:
No, I don't think this is a weird workaround for the issue of a too sensitive touch panel. 
This is for when you currently don't want to interact with your watch but still look at it.

tgc-dk:
Thanks for the into @minacode and @Joheyy! I'll look into the possibility of implementing a "pop up" notifying about the lock and that the hardware button should be used to unlock.

minacode:
Keep in mind that the watch can wake itself (for example via notifications).

tgc-dk:
I've just pushed a new commit that  adds a 'Popup' notification and display it when a touch input is received during 'locked' wakeup. The lock is now active on all screens, though not when an app as caused the wakeup.
I have only tested this in the simulator so far...
Let me know what you think!
I've attached a small animation that shows the lock in action.

InfiniSim_2022-10-16_195522

Avamander:
That approach seems good, also very similar to that of say Apple Watch.

minacode:
I like it.

Can you explain the intuition behind the three WokenBy values? Button explains itself, but what exactly are WakeUpAction and Other? What cases do and don't they cover?

tgc-dk:
> Can you explain the intuition behind the three WokenBy values? Button explains itself, but what exactly are WakeUpAction and Other? What cases do and don't they cover?

The WokenBy values are:
* Button: Pretty self explanatory as you say)
* WakeUpAction: The actions selected in the wakeup settings, single/double tap, raise, shake
* Other: Other things that can wake the watch up, basically apps and notifications.

In reality it might be possible to replace the values with a simple boolean `wokenByAction`, but I'm not yet sure...

minacode:
I would like that in a comment, in the code. Or better names, I don't know. UserButtonWakeup, UserNonButtonWakeup, SystemAutoWakeup? Sounds weird as well..

src/systemtask/SystemTask.cpp
```diff
@@ -102,7 +102,9 @@ SystemTask::SystemTask(Drivers::SpiMaster& spi,
                      spiNorFlash,
                      heartRateController,
                      motionController,
-                     fs) {
+                     fs),
+    wokenBy {WokenBy::Other},
```
NeroBurner, line 106: please move the default values to the definition in the header, just my personal preference to keep the constructor for non-default values

src/displayapp/widgets/PopupMessage.cpp
```diff
@@ -0,0 +1,39 @@
+#include "displayapp/widgets/PopupMessage.h"
+#include <libraries/log/nrf_log.h>
+
+using namespace Pinetime::Applications::Widgets;
+
+PopupMessage::PopupMessage(const char* msg)
+  : message {msg},
+    isHidden {true},
+    btnPopup {nullptr} {
```
NeroBurner, line 9: please move the default values to the definition in the header, just my personal preference to keep the constructor for non-default values

src/displayapp/screens/settings/SettingWakeUp.cpp
```diff
@@ -73,6 +73,17 @@ SettingWakeUp::SettingWakeUp(Pinetime::Applications::DisplayApp* app, Pinetime::
...
+  cbOption[optionsTotal] = lv_checkbox_create(container1, nullptr);
+  lv_checkbox_set_text_static(cbOption[optionsTotal], "Button Unlock");
+  cbOption[optionsTotal]->user_data = this;
+  lv_obj_set_event_cb(cbOption[optionsTotal], event_handler);
+  if (settingsController.isWakeUpModeOn(Pinetime::Controllers::Settings::WakeUpMode::ButtonUnlocks)) {
+    lv_checkbox_set_checked(cbOption[optionsTotal], true);
+  }
+  optionsTotal++;
+
+
```
NeroBurner, line 86: useless newlines, please remove

```suggestion
```

src/displayapp/widgets/PopupMessage.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+#include <lvgl/lvgl.h>
+
+namespace Pinetime {
+  namespace Applications {
+    namespace Widgets {
+      class PopupMessage {
+      public:
+        PopupMessage(const char* msg);
```
NeroBurner, line 9: please add a note explaining that the string isn't copied, meaning the pointed to memory must be alive for the popup message to still work

NeroBurner (changes requested):
I think it would be a good idea to move the popup to displayapp and use `displayapp/Messages.h` to communicate between the systemtask and DisplayApp

src/systemtask/SystemTask.cpp
tgc-dk: done

src/displayapp/screens/settings/SettingWakeUp.cpp
tgc-dk: done

src/displayapp/widgets/PopupMessage.cpp
tgc-dk: done

tgc-dk:
> I think it would be a good idea to move the popup to displayapp and use `displayapp/Messages.h` to communicate between the systemtask and DisplayApp

Changed as suggested

pipe01:
Any updates on this?

tgc-dk:
I've introduced a second page to the WakeUp settings since there was no room left for new checkbox.
Note that the CI simulator build fails due to changes needed in InfiniSim to build with the changes in this PR, see InfiniTimeOrg/InfiniSim#131

@NeroBurner Are you up for a re-review?

riban-bw:
> I've just pushed a new commit that adds a 'Popup' notification and display it when a touch input is received during 'locked' wakeup. The lock is now active on all screens, though not when an app as caused the wakeup. I have only tested this in the simulator so far... Let me know what you think! I've attached a small animation that shows the lock in action.
> 
> InfiniSim_2022-10-16_195522 [ InfiniSim_2022-10-16_195522 ](https://user-images.githubusercontent.com/1773050/196055351-ff2126d1-f4eb-465b-87ea-b4338d8cd4ff.gif) [ ](https://user-images.githubusercontent.com/1773050/196055351-ff2126d1-f4eb-465b-87ea-b4338d8cd4ff.gif)

The popup seems rather intrusive. If we are trying to avoid activity that occurs fairly frequently then it is possible that the popup will obscure the display when the user wants to view the watchface. Could we just use a lock icon instead? This is a user configured feature that would probably be disabled by default so the user must enable it and therefore shoudl understand the purpose of the icon. (It could be reinforced by showing the same icon in the settings screen.) The lock icon could show in a notification space, e.g. top of screen. It could either be always visible when locked, shown on activity or visible and highlight on activity. This may also reduce the memory increase. (One icon may be less than lots of text?)

I really like the idea of a lock mechanism (as I too suffer the issue of unexpected activity) but don't want something obscuring the screen.

NeroBurner:
I'm in favor of the lock as it doesn't need to be translated and should be easily understood. I like the visual hint to use the lock in the setting screen as well

jgonyea:
I also would love to see this feature added.  I don't know how, but I constantly am waking the screen and accidentally tapping on screen elements.  This would solve this issue for me!

tgc-dk:
@riban-bw 
> The popup seems rather intrusive. If we are trying to avoid activity that occurs fairly frequently then it is possible that the popup will obscure the display when the user wants to view the watchface.

Please note that the lock screen first shows up after the screen has been woken up AND a touch gesture has been identified, meaning that the lock screen does not automatically show when the display is on.

@riban-bw @NeroBurner 
> I'm in favor of the lock as it doesn't need to be translated and should be easily understood. I like the visual hint to use the lock in the setting screen as well

This is a valid point, I'll change the lock screen to use an actual icon. Regarding using the lock in the in the settings, how do you envision this to be done? See the screenshot below for the current settings screen, shown is the (new) second settings page under "Wake Up". Where should the lock icon be placed?

InfiniSim_2024-01-27_181545


> The lock icon could show in a notification space, e.g. top of screen. It could either be always visible when locked, shown on activity or visible and highlight on activity.

If I understand you correctly, this would be nice, but it would require that InfiniTime has a global notification bar for it to be consisted across screens, which it don't, so I think I'll stick with the pop-up for now.

tgc-dk:
@NeroBurner @riban-bw 
Using the popup with a lock icon could look like this:
InfiniSim_2024-01-27_220733

jgonyea:
Could it also have an arrow pointing to the crown button to indicate how to
unlock?

NeroBurner:
This looks perfect ( in my commandline-progrmmers opinion 😁)

I wouldn't require a pointer as this would make it more complicated (with regards to screen orientation on other supported devices)

NeroBurner:
Regarding the setting and how to put a lock icon in there. No idea. Not enough space in that line. So I'd say keep it like it is now

tgc-dk:
@NeroBurner 
Turns out the new lock icon takes up to much spaces so linking fails. See https://github.com/InfiniTimeOrg/InfiniTime/actions/runs/7688311260/job/20949380052

So I've reverted the icon changes...

Suggestions for how to proceed?

minacode:
Could you use an external font for the icon? 
Or a smaller icon? 😄
Or draw the lock manually with LVGL somehow?

mark9064:
If you check the build logs, the difference in RAM usage is just 64B. It's just that infinitime is very close to the memory limit right now - keep the lock chance but decrease the heap size a little

tgc-dk:
@mark9064 
> If you check the build logs, the difference in RAM usage is just 64B. It's just that infinitime is very close to the memory limit right now - keep the lock change but decrease the heap size a little

Just to be sure... Would I be right in assuming this can be done by in `src/FreeRTOSConfig.h` changing `configTOTAL_HEAP_SIZE` from `(1024 * 40)` to `(1024 * 39)` ?

minacode:
`(1024 * 35)` was recommended before by JF to me for the calculator app.

tgc-dk:
Pushed the new changes. The simulator build still fails due to dependent on changes in InfiniTimeOrg/InfiniSim#131

src/touchhandler/TouchHandler.h
```diff
@@ -6,13 +6,25 @@ namespace Pinetime {
...
+      // Enum describes how the watch was woken:
+      //  * WakeUpAction: The actions selected in the wakeup settings, single/double tap, raise, shake
+      //  * Button: The hardware button
+      //  * Other: Other things that can wake the watch up, eg. apps and notifications.
+      enum class WokenBy { WakeUpAction, Button, Other };
+
+      // Enum describes how the reply from ProcessTouchInfo should be interpreted:
+      //  * NoAction: Do nothing, ignore input.
+      //  * TouchEvent: The input should be treated as a normal touch event.
+      //  * IgnoreTouchPopup: Show the popup for when ignoring touvh input.
```
NeroBurner, line 18:
```suggestion
      //  * IgnoreTouchPopup: Show the popup when ignoring touch input.
```

NeroBurner:
Just for visuals a screenshot of how the lock overlay looks on the Analog Watch Face

Analog Watch Face showing lock overlay

src/touchhandler/TouchHandler.h
```diff
@@ -6,13 +6,25 @@ namespace Pinetime {
...
+      enum class TouchProcessReply { NoAction, TouchEvent, IgnoreTouchPopup };
+
       struct TouchPoint {
         int x;
         int y;
         bool touching;
       };
 
-      bool ProcessTouchInfo(Drivers::Cst816S::TouchInfos info);
+      TouchProcessReply ProcessTouchInfo(Drivers::Cst816S::TouchInfos info, bool buttonUnlocksOn);
```
NeroBurner, line 27: FYI: If you set a default value for the new parameter this PR builds even without changes to InfiniSim

```suggestion
      TouchProcessReply ProcessTouchInfo(Drivers::Cst816S::TouchInfos info, bool buttonUnlocksOn=false);
```

NeroBurner:
@tgc-dk I rebased your changes on current main, resolved merge conflicts, squashed commits down to 3 commits
- first commit all your changes rebased and squashed
- second commit reducing heap size
- third default buttonUnlocksOn to false to make it work with current InfiniSim

The rebased branch can be found here: https://github.com/InfiniTimeOrg/InfiniTime/tree/button-unlock_neroburner

mark9064:
Can I suggest lowering the heap size to 39 rather than 40? Ideally the heap should be as large as possible as long as it compiles

tgc-dk:
> @tgc-dk I rebased your changes on current main, resolved merge conflicts, squashed commits down to 3 commits
> 
>     * first commit all your changes rebased and squashed
> 
>     * second commit reducing heap size
> 
>     * third default buttonUnlocksOn to false to make it work with current InfiniSim
> 
> 
> The rebased branch can be found here: https://github.com/InfiniTimeOrg/InfiniTime/tree/button-unlock_neroburner

Hi @NeroBurner 
I've commited you're suggested change, but I currently do not have the "bandwidth" to do much more... Feel free to create a PR based on your branch!
I will probably get some more time later in the year, so I can probably do the merge and resolve conflicts then, but I don't see any reason to wait...

NeroBurner:
On GitHub the PR receiver can force push on the branch to be merged? Didn't expect that, but now I force pushed into your repo and overwrote the branch. I hope this is OK for you :sweat_smile:

mark9064:
Is this ready for review?

tgc-dk:
> On GitHub the PR receiver can force push on the branch to be merged? Didn't expect that, but now I force pushed into your repo and overwrote the branch. I hope this is OK for you 😅

No worries :)

tgc-dk:
> Is this ready for review?

I believe so :+1:

NeroBurner:
I am testing if the RAM increase can be lessened by drawing the clock symbol using LVGL rectangles instead of a font symbol. Just a matter of time ⌛ 😁

mark9064:
That would be great, as fonts seem to be a ram sucker 👍

NeroBurner:
Got a lock symbol drawn with lvgl. Only two `lv_obj` needed (in addition to the pop-up background `lv_obj`). Not pretty, but help welcome :D

InfiniSim_2024-10-06_button_unlock_lvgl

comparing f0211e5fcbab990da6b989b533c3e42ca63f5e18 (lock with font symbol) against 2a8e953cb6b61a857a3ba51b8fb5279e24d678d6 (lock with lvgl)


- lock with font: 
  - default config:
    - FLASH: 376024 B -- 78.22%
    - RAM: 58KB -- 90.62%  (auto formatted to KB, don't know how to get B)
  - WatchFace::Terminal
    - FLASH: 300028 B -- 63.21%
    - RAM: 59184 B -- 90.31%
  - full analysis: [build_firmware_memory_2024-10-06_button_unlock_font.txt](https://github.com/user-attachments/files/17270542/build_firmware_memory_2024-10-06_button_unlock_font.txt)
- lock with lvgl:
  - default config:
    - FLASH: 375700 B -- 79.16%
    - RAM: 59336 B -- 90.54%
  - WatchFace::Terminal
    - FLASH: 299688 B -- 63.14%
    - RAM: 59120 B -- 90.21%
  - full analysis:  [build_firmware_memory_2024-10-06_button_unlock_lvgl.txt](https://github.com/user-attachments/files/17270541/build_firmware_memory_2024-10-06_button_unlock_lvgl.txt)

On `WatchFace::Terminal` we're saving 340B FLASH and 64B of RAM when drawing with LVGL instead of using the font (this is a 0.1% reduction in FLASH)

For comparison the current main branch with the same HEAP reduction as this PR currently has `(1024 * 35)` we have the following numbers.
- main with HEAP size 35KB:
  - default config:
    - FLASH: 374548 B -- 78.91%
    - RAM: 59320 B -- 90.52 %
  - WatchFace::Terminal
    - FLASH: 300232 B -- 63.26%
    - RAM: 59104 B -- 90.19%

Which means with lvgl lock we still add 16B RAM and we reduce FLASH by 544 B ( :question: )

NeroBurner:
I can revert the HEAP-size reduction now and we still stay in RAM bounds :tada: 

```
all default Apps/WatchFaces
USERAPP_TYPES:STRING=Apps::StopWatch, Apps::Alarm, Apps::Timer, Apps::Steps, Apps::HeartRate, Apps::Music, Apps::Paint, Apps::Paddle, Apps::Twos, Apps::Dice, Apps::Metronome, Apps::Navigation, Apps::Weather
WATCHFACE_TYPES:STRING=WatchFace::Digital, WatchFace::Analog, WatchFace::PineTimeStyle, WatchFace::Terminal, WatchFace::Infineat, WatchFace::CasioStyleG7710
           FLASH:      375700 B     474632 B     79.16%
             RAM:       64456 B        64 KB     98.35%
WatchFace::Terminal
           FLASH:      299688 B     474632 B     63.14%
             RAM:       64240 B        64 KB     98.02%
```

mark9064:
@NeroBurner is this now ready for review? If so could you rebase onto main?

NeroBurner:
finally found the time to rebase onto `main` (after 1.15.0 release). Ready for review

JF002:
I've been using this branch these last few days and, to be honest, this is much less intrusive than I first thought it would be :+1: 

I was a bit confused a few times by the fact that the lock only applies when the watch is woken by the motion sensor and not when it's woken by the button : I got used to the lock mode, and, sometimes, when I wake the watch using the button, I simply assume that it's lock and that I need to unlock it. After a few days, things became more natural to me !

Before merging, I would like to ask for more feedback : 
 - do you think this option should be enabled by default once we merge it?
 - if yes, do you think we need a setting to enable / disable it, or should we simply enable it ?

NeroBurner:
Happy you find it useful (and that it is stable enough for use 😁 )

For code simplicity I would prefer it to be always enabled without a setting

src/touchhandler/TouchHandler.h
```diff
@@ -6,13 +6,25 @@ namespace Pinetime {
   namespace Controllers {
     class TouchHandler {
     public:
+      // Enum describes how the watch was woken:
```
JF002, line 9: I'm wondering if this new functionality allowing to ignore touch events should be in the TouchHandler, or if it should be done at the SystemTask level (which wouldn't call the TouchHandler if we decide to ignore the TouchEvent) or in another object that would only do that.

This would allow to maintain the "single responsibility principle" on TouchHandler.

Or perhaps there were other reasons to implement this management in TouchHandler?

JF002:
> For code simplicity I would prefer it to be always enabled without a setting

I agree for the code simplicity :+1: 
Some users might find this new feature annoying (since it adds a new setup to unlock the watch). But I know that many users are very frustrated by the watch that wakes up and opens app by itself so the net result might be positive ?

So I guess we have a few options? Just keep the setting from this PR, or enable it by default and hide the setting (so we can re-enable it if we have good reasons to do so) or just remove it?

NeroBurner:
I propose to implementiert it unconditionally and in the process see if I can move the responsibility into SystemTask

If after merge we see that people testing the main branch complain about the lock being unconditional we could reintroduce the setting afterwards

This would slightly limit the size of this PR making the review a bit easier

al1-ce:
Testing this branch rn and shouldn't lock also apply to when notifications appear? Haven't tested with alarms, chimes and stuff but I think it'd make more sense to have to always unlock it if it wakes up from sleep on its own

al1-ce:
Another thing I'd add is previously mentioned
> I wonder if we shouldn't introduce the concept of "unlock mode" rather than adding this to the "wake up modes". What do you think?
It's quite confusing having to press button two times to turn off screen and then turn it back on + unlock. Just doesn't really make sense in this situation

src/components/settings/Settings.h
```diff
@@ -13,7 +13,14 @@ namespace Pinetime {
       enum class WeatherFormat : uint8_t { Metric, Imperial };
       enum class Notification : uint8_t { On, Off, Sleep };
       enum class ChimesOption : uint8_t { None, Hours, HalfHours };
-      enum class WakeUpMode : uint8_t { SingleTap = 0, DoubleTap = 1, RaiseWrist = 2, Shake = 3, LowerWrist = 4 };
+      enum class WakeUpMode : uint8_t {
```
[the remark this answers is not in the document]
NeroBurner: I simplified it in the process of moving the locking/locked handling to SystemTask. And now there is no setting

src/displayapp/screens/settings/SettingWakeUp.h
```diff
@@ -13,27 +16,40 @@ namespace Pinetime {
 
       class SettingWakeUp : public Screen {
       public:
-        SettingWakeUp(Pinetime::Controllers::Settings& settingsController);
+        SettingWakeUp(Pinetime::Applications::DisplayApp* app, Pinetime::Controllers::Settings& settingsController);
```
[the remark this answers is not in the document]
NeroBurner: removed as we have no new setting, so no changes needed in the `SettingWakeUp.h` screen

src/touchhandler/TouchHandler.h
NeroBurner: moved responsibility from `TouchHandler` to `SystemTask`. And simplified a bit

NeroBurner:
Rebased and simplified to a minimal working version. Now the screen has to be unlocked by the hardware-button. Otherwise the `lock` will appear on the next interaction after wakeup.

No wakeupmodes, just two new boolean variables in `SystemTask.h`

NeroBurner:
found one bug on watch. Help in fixing would be appreciated :sweat_smile: 
- enable double tap to wake
- wake up with double tap
- wipe up to get lock
- unlock with button
- wipe up again and the screens are swiped up twice instead of just once as expected

al1-ce:
It does that on a single tap too apparently

minacode:
> Also unlock when the alarm went off to be able to push the red "stop alarm" button.

Just an note: we have #2222 implementing a long press to stop the alarm and with that it should be fine to release the lock automatically and still be safe.

NeroBurner:
Fixed a bug in the previous version such that one needed two button presses to actually wake up from normal sleep. Now fixed.

And also the double swipe issue is fixed

@minacode Interesting! I'll keep the "alarm unlocks screen" behavior for now

mark9064:
Why did it create a merge commit :neutral_face: not trusting that button again

NeroBurner:
no idea. I like the GitLab feature where I just type `/rebase` as command, and it does a rebase. I miss that here

I'm somehow allowed to force-push. So no worries. All easily fixable :) Thanks for chiming in! :)

NeroBurner (approved):
I'm aware this is basically my own PR now, but still approving to get rid of my "changes requested" review status

NeroBurner:
Now having 7 days of uptime. Seems stable 🎉

mattvchandler:
I've been using this branch in my own build for over a year now. It's been really nice to prevent accidental touches.
However, I've noticed that if you touch the screen while a notification is shown, return to the watch face without unlocking, and then press the button, the watch crashes.

Here's a detailed method that seems to be 100% reproducible for me.
1. Have the watch face open - this only seems to happen consistently with the watch face, not with other apps.
2. Turn off the screen, either with the HW button or let it time out
3. Receive a notification. The screen turns on and goes to the notification preview
4. Touch the screen before the preview times out
5. After the preview times out and returns to the watch face, but before the screen turns off, push the hardware button
6. The watch will crash and reboot

I've also gotten a crash by waking the screen with a timer or alarm, and then double-pressing the button to get to the notifications. It seems less consistent, and I've never accidentally triggered a crash this way, but I noticed the possibility of the same problem as I was trying to find a solution to this crash, and tested it out as well.

1. Have the screen be off / locked, again with the watch face as the current app
2. Wake the screen by tapping, shaking, raising, an alarm or timer, etc. Any method other than the HW button.
3. Double-press the HW button to bring up the notification app
4. Touch the screen
5. Long press the HW button to return to the watch face
7. Press the HW button
8. The watch will crash and reboot

I haven't been able to reproduce either these methods consistently in InfiniSim.

For my own build, I worked around the problem by setting unlockedByButton = true in SystemTask.cpp when getting a new notification, or when double-pressing the button. It prevents the crash, but at least in the case of receiving a notification unlocks the screen without user interaction and can lead (and for me has led) to accidental presses, so it's not an ideal fix.

I haven't been able to track down the root cause. I think it's got to be something to do with the active app changing whilst locked, but I haven't been able to get any further than that.

NeroBurner:
can reproduce (the second guide with double-press). Thanks a lot for finding it and the thorough steps to reproduce the issue

atomdmac:
I've been using this branch for about a week and it's made my PineTime _much_ more usable for me.  The accidental touches were constantly causing my watch to end up on non-clock screens.  Thank you!!

p1gp1g:
Thank you for the PR!

When I touch the screen while it is locked, I see the popup until the screen is off. I think it would be better to show the lock for ~ half a second: it is enough to know that the screen is locked and allow to see the watch face after that

Less important, but it may be a good idea to show the popup for long press and swipe only on the watch face

BTW, it is possible to make the lock more transparent, it is very white with the casio face ?

p1gp1g:
> can reproduce (the second guide with double-press). Thanks a lot for finding it and the thorough steps to reproduce the issue

There is a double free, I'm opening a PR on your repo

EDIT: https://github.com/tgc-dk/InfiniTime/pull/6 this can be fixed by calling `popupMessage.SetHidden(true);` whenever we change app

NeroBurner:
Can confirm the second crash reproduction is now fixed. Thanks @p1gp1g

NeroBurner:
@p1gp1g 

> When I touch the screen while it is locked, I see the popup until the screen is off. I think it would be better to show the lock for ~ half a second: it is enough to know that the screen is locked and allow to see the watch face after that

I agree, but I'd like this PR to be kept simple. I essentially hijacked it and would like it to be merged as soon as possible. So I try to keep the PR itself to be simple and easily reviewable. So I propose a new PR after this one is merged

> Less important, but it may be a good idea to show the popup for long press and swipe only on the watch face

Good idea. I'd need to see if this a simple change or if it makes the code more complicated

> BTW, it is possible to make the lock more transparent, it is very white with the casio face ?

I'm no designer, so I'd like to delegate this to someone else (i.e. please open a a new PR after this one is merged)

p1gp1g:
I've found another bug (tested before my patch too), I'll dig it later to get a patch, but I think I know what to do:
- Lock the watch
- Single tap to wake the watch
- Double click to go to notifications app
- Single tap to show the popup
- Wait until it goes to sleep

It will crash

I think it can be fixed by calling `setHiddent(true)` in `SystemTask::GoToSleep`

NeroBurner:
can reproduce 😅 thanks a lot!

WhyNotHugo:
Fixes https://github.com/InfiniTimeOrg/InfiniTime/issues/1839

mark9064:
Ping me when this is ready for review :)

WhyNotHugo:
I've been testing this for a couple of weeks. My main issue is that the lock icon is shown as an overlay covering the time, so I can't read the time when it's visible. I'd prefer the lock to be shown on top of the time (e.g.: near the batter/bluetooth icons maybe), so I can still read the time while it's displayed.

atomdmac:
@WhyNotHugo That's basically how Bangle JS handles this mechanic and it works really well, IMO.

NeroBurner:
> I've found another bug (tested before my patch too), I'll dig it later to get a patch, but I think I know what to do:
> 
>     * Lock the watch
>     * Single tap to wake the watch
>     * Double click to go to notifications app
>     * Single tap to show the popup
>     * Wait until it goes to sleep
> 
> It will crash
> 
> I think it can be fixed by calling `setHiddent(true)` in `SystemTask::GoToSleep`

I just wanted to implement you fix, but that is effectively already done.

https://github.com/InfiniTimeOrg/InfiniTime/blob/3282333632f61c20f40d80157122225b32cea3e0/src/systemtask/SystemTask.cpp#L440

Do you maybe have other ideas, or did I misunderstand your fix?

NeroBurner:
found another workaround. Double press button also unlocks the watch

kieranc:
I just had a crash I think related to this PR, only changes are this plus font bpp changes the the bpp stuff has been stable for quite a while.

Notification arrived, I tried to swipe and got the lock icon, so I pressed the button to unlock, tried to swipe but the screen was unresponsive, then the watch rebooted.

I just made it happen again, I pressed the button just after/at the same time as the notification disappeared, perhaps related?

NeroBurner:
thanks for trying it out and finding a bug. I think I'll have to call hide overlay before changing the screen to be sure it is always properly unloaded from each frame. otherwise lvgl will delete it as part of the screens decunstructor

NeroBurner:
we already do that... so why does it segfault/reboot :thinking: 

https://github.com/InfiniTimeOrg/InfiniTime/blob/10344e5e68898a80a8701acb860b699b296e60ce/src/displayapp/DisplayApp.cpp#L513

p1gp1g:
> > I've found another bug (tested before my patch too), I'll dig it later to get a patch, but I think I know what to do:
> > ```
> > * Lock the watch
> > * Single tap to wake the watch
> > * Double click to go to notifications app
> > * Single tap to show the popup
> > * Wait until it goes to sleep
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It will crash
> > I think it can be fixed by calling `setHiddent(true)` in `SystemTask::GoToSleep`
> 
> I just wanted to implement you fix, but that is effectively already done.
> 
> https://github.com/InfiniTimeOrg/InfiniTime/blob/3282333632f61c20f40d80157122225b32cea3e0/src/systemtask/SystemTask.cpp#L440
> 
> Do you maybe have other ideas, or did I misunderstand your fix?

It may be because `displayApp.PushMessage(Pinetime::Applications::Display::Messages::HideIgnoreTouchPopup);` is called after `displayApp.PushMessage(Pinetime::Applications::Display::Messages::GoToSleep);`

Try:

```diff
 void SystemTask::GoToSleep() {
   if (IsSleeping()) {
     return;
   }
   if (IsSleepDisabled()) {
     return;
   }
   NRF_LOG_INFO("[systemtask] Going to sleep");
+  displayApp.PushMessage(Pinetime::Applications::Display::Messages::HideIgnoreTouchPopup);
   if (settingsController.GetAlwaysOnDisplay()) {
     displayApp.PushMessage(Pinetime::Applications::Display::Messages::GoToAOD);
   } else {
     displayApp.PushMessage(Pinetime::Applications::Display::Messages::GoToSleep);
   }
   heartRateApp.PushMessage(Pinetime::Applications::HeartRateTask::Messages::GoToSleep);
   unlockedByButton = false;
-  displayApp.PushMessage(Pinetime::Applications::Display::Messages::HideIgnoreTouchPopup);
 
   state = SystemTaskState::GoingToSleep;
 };

NeroBurner:
Thanks @p1gp1g committed your change. That could be it.

@kieranc could you test again?

tituscmd:
I think it would be better if this feature wasn’t enabled by default and could be turned off in the settings. I personally don’t run into the problem this PR fixes, so I likely wouldn’t use the feature and would appreciate the option to disable it.

Maybe the setting for this could be placed under an "Accessibility" page in the settings menu? This would be descriptive enough to be intuitive and allow future features to be included as well, unlike a "Unlock Watch" setting.

NeroBurner:
rebased onto 1.16.0 release

I noticed some power cycles indicating errors. Once happened after I was called with the call notification. Wasn't able to reproduce yet, just mentioning it

edit: just got a power cycle again after interacting with the "OTA disabled" message. Lock was shown, message screen got switched after time out, watch rebooted

NeroBurner:
fixed bug

Add a lvlg event handler when the `popup` object is deleted. This helps
to prevent double free memory errors due to `PopupMessage` class not
recognizing that the `popup` object was deleted by someone else (for
example when switching away from a notification)

tituscmd:
> I personally don’t run into the problem this PR fixes, so I likely wouldn’t use the feature and would appreciate the option to disable it.

Since this is now queued for the next version and might get addressed again soon, I would like to circle back to this comment of mine of restoring the setting for this feature.

PR #1367, opened 2022-10-12
Alarm persist to flash

NeroBurner:
rebase of https://github.com/InfiniTimeOrg/InfiniTime/pull/1333 with additional `SaveAlarm()` function as described below

---

Save the set alarm time to the SPI NOR flash, so it does not reset to
the default value when the watch resets, e.g. due to watchdog timeout
or a firmware update

Introduce `SaveAlarm()` member function which saves to file only if the
alarm setting is changed.
The `alarmChanged` boolean flag mimics the behavior of `settingsChanged`
flag from `Settings.h`

Also mark the ~~`SaveAlarmToFile()`~~ ~~`Save()`~~ `SaveSettingsToFile` function const, as it doesn't change
anything in the `AlarmController` object.

---

This PR needs https://github.com/InfiniTimeOrg/InfiniSim/pull/67 to work on `InfiniSim`

--- 

Fixes: https://github.com/InfiniTimeOrg/InfiniTime/issues/1330

### DIFF
--- a/src/components/alarm/AlarmController.cpp
+++ b/src/components/alarm/AlarmController.cpp
@@ -169,6 +169,7 @@ void AlarmController::SaveSettingsToFile() const {
   if (fs.DirOpen("/.system", &systemDir) != LFS_ERR_OK) {
     fs.DirCreate("/.system");
   }
+  fs.DirClose(&systemDir);
   lfs_file_t alarmFile;
   if (fs.FileOpen(&alarmFile, "/.system/alarm.dat", LFS_O_WRONLY | LFS_O_CREAT) != LFS_ERR_OK) {
     NRF_LOG_WARNING("[AlarmController] Failed to open alarm data file for saving");
--- a/src/components/alarm/AlarmController.cpp
+++ b/src/components/alarm/AlarmController.cpp
@@ -19,11 +19,13 @@
 #include "systemtask/SystemTask.h"
 #include "task.h"
 #include <chrono>
+#include <libraries/log/nrf_log.h>
 
 using namespace Pinetime::Controllers;
 using namespace std::chrono_literals;
 
-AlarmController::AlarmController(Controllers::DateTime& dateTimeController) : dateTimeController {dateTimeController} {
+AlarmController::AlarmController(Controllers::DateTime& dateTimeController, Controllers::FS& fs)
+  : dateTimeController {dateTimeController}, fs {fs} {
 }
 
 namespace {
@@ -36,11 +38,28 @@ namespace {
 void AlarmController::Init(System::SystemTask* systemTask) {
   this->systemTask = systemTask;
   alarmTimer = xTimerCreate("Alarm", 1, pdFALSE, this, SetOffAlarm);
+  LoadSettingsFromFile();
+  if (alarm.isEnabled) {
+    NRF_LOG_INFO("[AlarmController] Loaded alarm was enabled, scheduling");
+    ScheduleAlarm();
+  }
+}
+
+void AlarmController::SaveAlarm() {
+  // verify if it is necessary to save
+  if (alarmChanged) {
+    SaveSettingsToFile();
+  }
+  alarmChanged = false;
 }
 
 void AlarmController::SetAlarmTime(uint8_t alarmHr, uint8_t alarmMin) {
-  hours = alarmHr;
-  minutes = alarmMin;
+  if (alarm.hours == alarmHr && alarm.minutes == alarmMin) {
+    return;
+  }
+  alarm.hours = alarmHr;
+  alarm.minutes = alarmMin;
+  alarmChanged = true;
 }
 
 void AlarmController::ScheduleAlarm() {
@@ -53,18 +72,19 @@ void AlarmController::ScheduleAlarm() {
   tm* tmAlarmTime = std::localtime(&ttAlarmTime);
 
   // If the time being set has already passed today,the alarm should be set for tomorrow
-  if (hours < dateTimeController.Hours() || (hours == dateTimeController.Hours() && minutes <= dateTimeController.Minutes())) {
+  if (alarm.hours < dateTimeController.Hours() ||
+      (alarm.hours == dateTimeController.Hours() && alarm.minutes <= dateTimeController.Minutes())) {
     tmAlarmTime->tm_mday += 1;
     // tm_wday doesn't update automatically
     tmAlarmTime->tm_wday = (tmAlarmTime->tm_wday + 1) % 7;
   }
 
-  tmAlarmTime->tm_hour = hours;
-  tmAlarmTime->tm_min = minutes;
+  tmAlarmTime->tm_hour = alarm.hours;
+  tmAlarmTime->tm_min = alarm.minutes;
   tmAlarmTime->tm_sec = 0;
 
   // if alarm is in weekday-only mode, make sure it shifts to the next weekday
-  if (recurrence == RecurType::Weekdays) {
+  if (alarm.recurrence == RecurType::Weekdays) {
     if (tmAlarmTime->tm_wday == 0) { // Sunday, shift 1 day
       tmAlarmTime->tm_mday += 1;
     } else if (tmAlarmTime->tm_wday == 6) { // Saturday, shift 2 days
@@ -79,7 +99,10 @@ void AlarmController::ScheduleAlarm() {
   xTimerChangePeriod(alarmTimer, secondsToAlarm * configTICK_RATE_HZ, 0);
   xTimerStart(alarmTimer, 0);
 
-  state = AlarmState::Set;
+  if (!alarm.isEnabled) {
+    alarm.isEnabled = true;
+    alarmChanged = true;
+  }
 }
 
 uint32_t AlarmController::SecondsToAlarm() const {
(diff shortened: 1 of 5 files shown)


src/components/alarm/AlarmController.h
```diff
@@ -29,42 +29,57 @@ namespace Pinetime {
...
 
       void Init(System::SystemTask* systemTask);
+      void SaveAlarm();
       void SetAlarmTime(uint8_t alarmHr, uint8_t alarmMin);
       void ScheduleAlarm();
       void DisableAlarm();
       void SetOffAlarmNow();
       uint32_t SecondsToAlarm() const;
       void StopAlerting();
-      enum class AlarmState { Not_Set, Set, Alerting };
```
JF002, line 41: Why was this `enum` replaced by 2 bools?

src/components/alarm/AlarmController.h
```diff
@@ -29,42 +29,57 @@ namespace Pinetime {
...
       TimerHandle_t alarmTimer;
-      uint8_t hours = 7;
-      uint8_t minutes = 0;
+      AlarmData alarm;
       std::chrono::time_point<std::chrono::system_clock, std::chrono::nanoseconds> alarmTime;
-      AlarmState state = AlarmState::Not_Set;
-      RecurType recurrence = RecurType::None;
+
+      void LoadAlarmFromFile();
+      void SaveAlarmToFile() const;
```
JF002, line 82: The naming of those methods can probably be improved : `Alarm` is recurrent with the name of the class, and `File` is an implementation detail that brings no meaningful information to the interface of the class. What would you think about `SaveSettings()` or even `Save()` and `Restore()` or `RestoreSettings()`?

JF002 (changes requested):

src/components/alarm/AlarmController.h
NeroBurner: we've got the public function `SaveAlarm()`. The two mentioned are private functions.

I don't know how confusing it is to have `SaveSettings()` and `SaveAlarm()` :thinking: 

I think I like the `Save()/Restore()` pair the most and I think it is the least confusing alternative

src/components/alarm/AlarmController.h
NeroBurner: I think to have the split between `bool isEnabled` inside the stored settings and the temporary `bool isAlarming`, which shouldn't be stored

JF002: Fair enough, thanks!

src/components/alarm/AlarmController.h
JF002: Yep, it looks better to me !

I noticed that `Alarm` is redundant in other methods and fields that existed before this PR, so no need to fix that in this PR. We might want to clean that in the future :)

JF002 (approved):

src/components/alarm/AlarmController.cpp
```diff
@@ -88,21 +112,68 @@ uint32_t AlarmController::SecondsToAlarm() const {
...
+    alarm.recurrence = recurrence;
+    alarmChanged = true;
+  }
+}
+
+void AlarmController::Restore() {
+  lfs_file_t alarmFile;
+  AlarmData alarmBuffer;
+
+  if (fs.FileOpen(&alarmFile, "/alarm.dat", LFS_O_RDONLY) != LFS_ERR_OK) {
```
Riksu9000, line 151: When we add features that allow users to access the filesystem, we need to decide if the root of the filesystem will be generally accessible for the user, or if there'll be a "home directory". If the root will be accessible, system files must be placed in a directory.

I think the current structure used for the resources isn't really optimal if the root will be generally accessible. I think I'd prefer creating a safe "home directory" and limit general access to under that directory, which wouldn't require changes in this PR necessarily.

JF002: Currently, the filesystem is used to store
 - user settings
 - BLE bonding info
 - external resources since 1.11

The end users are not supposed to access those files directly : infinitime manages the settings/bonding data and the resources are uploaded following a specified procedure. But, they can, for example, customize the external resources if they want to (and if they know what they are doing).

What do you mean by "generally accessible" ? I don't think the users will ever need to create and manage their own file on the FS. This is at least not the case for now. Which files would you put in the "home directory"?

I'm OK to define a folder structure to keep the FS clean & tidy. For example:
 - system : settings + bonding info
 - resources : external resources (with subdir for fonts, pictures,...)
 - apps : config and data files specific for each application that need persistance

Riksu9000: I imagine users would need to be able to access the filesystem to install apps, translations, files for viewing with a gallery app, maybe customization patches, it could be anything.

JF002: But I guess those files will have to be stored at the exact location where InfiniTime and the apps will look for them, the users won't be able to upload those file anywhere they want in their home directory. For example, InfiniTime will look for translations in a specific folder and the gallery app will look for pictures in another specific folder. 

So how will we specify which file must be stored in `/home` rather than in another "system" folder?

Riksu9000: It's getting a bit off topic, but I think we should never place anything in the user directory, so only user uploaded files would be there. The thing I'm concerned with here is the user having to navigate among system files. Only giving access to a user directory would be my preferred way to resolve the issue. If we instead allow root access by default, then I think there should be no files directly at root.

JF002: I'm not sure I understand your concern here... Is this just a naming convention (file for certain type of access or use should be place in this or that folder, do not put files in `/`,...) or do you want to restrict the BLE FS API to only the home folder? Or something else?

Riksu9000: I don't want the user to encounter system files.

Let's imagine adding files to the filesystem. On my phone there are directories to put your music, pictures, etc.. On the PineTime with root access, you would see alarm.dat, settings.dat, etc., which doesn't seem right to me.

There are two solutions to the problem, and we can even implement both.

One is not placing any files in the root and making sure the user doesn't mistakingly enter system directories.

Second is not allowing access to the root in the first place. I suppose this would be done by restricting the BLE FS API, but I'm thinking about the issue more conceptually.

JF002: Yes, Ok, I understand. Currently, the users never interact directly with the FS. It's possible for them to replace a resource with a custom one if they want to, but that's not expected. But yes, in the future, with apps like the Gallery, they'll probably have more opportunities to use the filesystem.

Our current BLE FS API doesn't prevent anyone from accessing any file in the FS, and I don't think we should restrict the BLEFSAPI to a specific folder. Companion apps might want to provide an easy way to edit the settings, the alarm or even the bond info (or clear them). They also need to access the FS to upload the resources.

So yes, we can create folders for specific folders like I say in a previous comment (with an additional home folder for user customizable files, maybe), but I don't think we'll be able to prevent the users from accessing other folder than the home folder.
It is also possible for companion apps to implement UIs specific to some apps/files so that the user does not have to manually upload and update the file, this will be implemented by the companion app.

Riksu9000: What if we prepended all system directories with a dot and asked companion apps to hide directories beginning with a dot by default? I propose moving the alarm file to `.system/alarm.dat`.

JF002: That's actually a good idea! Yep, I approve :)

Riksu9000: What should we do about the existing files and directories? @JF002

NeroBurner:
rebased and squashed on current `main`

src/components/alarm/AlarmController.cpp
```diff
@@ -88,20 +111,67 @@ uint32_t AlarmController::SecondsToAlarm() const {
...
 }
+
+void AlarmController::SetRecurrence(RecurType recurrence) {
+  if (alarm.recurrence != recurrence) {
+    alarm.recurrence = recurrence;
+    alarmChanged = true;
+  }
+}
+
+void AlarmController::Restore() {
```
mark9064, line 145: Maybe `LoadAlarmFile` or similar could be more descriptive?

Reading Init
```c++
void AlarmController::Init(System::SystemTask* systemTask) {
  this->systemTask = systemTask;
  alarmTimer = xTimerCreate("Alarm", 1, pdFALSE, this, SetOffAlarm);
  Restore();
  ```
  The meaning of `Restore` isn't immediately clear to me
  If renaming, probably change `Save` too

**Review by mark9064 — COMMENTED** [review]
Haven't tested but looks sensible to me

File: src/systemtask/SystemTask.cpp [hunk 5]
```diff
@@ -344,8 +344,7 @@ void SystemTask::Work() {
         case Messages::OnNewHour:
           using Pinetime::Controllers::AlarmController;
           if (settingsController.GetNotificationStatus() != Controllers::Settings::Notification::Sleep &&
-              settingsController.GetChimeOption() == Controllers::Settings::ChimesOption::Hours &&
-              alarmController.State() != AlarmController::AlarmState::Alerting) {
+              settingsController.GetChimeOption() == Controllers::Settings::ChimesOption::Hours && alarmController.IsAlerting()) {
```
ght, line 347:
```diff
- settingsController.GetChimeOption() == Controllers::Settings::ChimesOption::Hours && alarmController.IsAlerting()) {
+ settingsController.GetChimeOption() == Controllers::Settings::ChimesOption::Hours && !alarmController.IsAlerting()) {
```
and same for `HalfHours` below, or chimes won't work. LGTM apart from this

Related: For anyone that wants more chime frequency options / configurable chime duration, see my [chimes-ui-redesign](https://github.com/ght/InfiniTime/tree/chimes-ui-redesign) branch

---
@NeroBurner Thanks for adopting my abandoned PR! I mostly stopped coding, and am a farrier's apprentice  instead for almost two years now and much happier :)

NeroBurner: Indeed, no chimes currently on my watch :sweat_smile: Thanks. Fixed in rebase

You're welcome! And thanks for still coming back to us reviewing this rebase :)

src/components/alarm/AlarmController.cpp
NeroBurner: the original function was changed from `LoadAlarmFromFile`/`SaveAlarmToFile` to `Restore`/`Save` because of feedback from JF. But the discussed part was the redundancy of the `Alarm` work in the function. So I think a rename to `LoadSettingsFromFile()`/`SaveSettingsToFile()` should be OK

https://github.com/InfiniTimeOrg/InfiniTime/pull/1367#discussion_r1007262187

src/components/alarm/AlarmController.cpp
NeroBurner: after a long time, moved `alarm.dat` into subfolder `.system`

src/components/alarm/AlarmController.cpp
NeroBurner: done!

mark9064 (approved):
Untested but all LGTM

mark9064:
I considered opening a comment on that text you just changed but decided it was too minor LOL

NeroBurner:
There is a segfault or freeze leading to a reboot on the second time when changing the alarm setting and exit the Alarm setting screen. Need to investigate

NeroBurner:
found it, need to close the `lfs_dir` after checking if the `.system` dir already exists or if we need to create it ourselfs

mark9064:
Did I miss that or did you only just add .system?

NeroBurner:
Second. I die add it based in an old Diskussion about where to put pibetime system files and that they shouldn't be in the root for the user to be directly available. JF and Riksu9000 agreed that a subfolder .system is a goog place to put our configs. And the user can put their files at the roor

https://github.com/InfiniTimeOrg/InfiniTime/pull/1367#discussion_r1061722304

PR #1386, opened 2022-10-23
Combine SettingSetTime and SettingSetDate

Elements6007:
This PR combines SettingSetTime and SettingSetDate into one setting app(SettingSetDateTime). I thought it would be nice to combine the two settings because it would put both in the same place and there would be one less setting in the setting menu. I hope you guys like it!

Note: I know that it does not work with InfiniSim because of the name change.

https://user-images.githubusercontent.com/88860704/197371890-4d6f5958-62b3-413c-aa6c-1323f0c64b9b.mp4

Thanks!


Elements6007:
> Very nice idea and good implementation. The only problem with this is mostly just that the user might not even realize it exists. A good case study of this is the music app, did you know it has volume control? It took me a few months to figure that out, and many people still don’t realize it exists…

Thanks for your input, but in my opinion the side indicators(i think my fingers were in the way in the video) hint to the second page adequately where as the music app has no such indicators. However, if others agree this is a problem then i will close the PR.



20221024_140720 (Small)

^better pic of indicator

RageGamerBoi:
Ah, ok, sorry I didn’t see that. Along with the title “Date & Time” it actually makes more sense to group them in a way… Very nice :)

Elements6007:
Sorry about that, I was trying to rename the branch because of the "/" character in the branch name which caused the firmware build to fail, I did not know that changing renaming a branch would do that. So looks like i have to make a new PR? ill only do it if people are interested in this though.

Elements6007:
Closing #1465

PR #1387, opened 2022-10-23
Better hardware abstraction using C++20 concepts

JF002:
This PR improves the hardware abstraction by integrating the Concepts features of C++20 for Spi, SpiMaster and SpiNorFlash.

My goal is to provide a design that abstracts the hardware nicely and clearly. I want to avoid preprocessor (`#ifdef`), CMake black magic and code generation as much as possible. I also want the abstraction to have the lowest possible overhead : most if not all of the abstraction work should be done at compile time. And I think that new features from C++20 will make this possible.

The 'interface' of the drivers are defined in `Pinetime::Drivers::Interface::Spi`, `Pinetime::Drivers::Interface::SpiMaster` and `Pinetime::Drivers::Interface::SpiNorFlash`. The actual implementation will be injected via the template parameter `T`. T is required to conform to the corresponding concepts `IsSpi`, `IsSpiMaster` and `IsFlashMemory`.

This implementation should be practically free in terms of memory and cpu resource usage, but it provides way to implement multiple variations of the same driver. It also describes very clearly the interface of the drivers thanks to the concepts.

This will allow us to support more easily other hardware : different SoC, different sensors and memory chips. It'll also make the implementation of InfiniSim much easier.

Note that the C++ standard was upgraded from 14 to 20.

This is still a work in progress, other drivers must still be ported to this new design.

But, as this is the first time I'm using concepts, this PR is also a request for comments : what do you think of this "hardware abstraction" design?

### DIFF
--- a/src/components/ble/DfuService.h
+++ b/src/components/ble/DfuService.h
@@ -9,12 +9,12 @@
 #undef max
 #undef min
 
+#include <drivers/SpiNorFlash.h>
+#include "infinitime.h"
+
 namespace Pinetime {
   namespace System {
     class SystemTask;
-  }
-  namespace Drivers {
-    class SpiNorFlash;
   }
   namespace Controllers {
     class Ble;
--- a/src/drivers/Spi.h
+++ b/src/drivers/Spi.h
@@ -1,28 +1,56 @@
 #pragma once
 #include <cstdint>
 #include <cstddef>
-#include "drivers/SpiMaster.h"
+#include <utility>
+#include <concepts>
 
 namespace Pinetime {
   namespace Drivers {
-    class Spi {
-    public:
-      Spi(SpiMaster& spiMaster, uint8_t pinCsn);
-      Spi(const Spi&) = delete;
-      Spi& operator=(const Spi&) = delete;
-      Spi(Spi&&) = delete;
-      Spi& operator=(Spi&&) = delete;
+    template <typename T>
+    concept IsSpi = requires(T s, const uint8_t* constData, uint8_t* data, const uint8_t* constCommand, uint8_t* command, size_t size) {
+                      { s.Write(constData, size) } -> std::same_as<bool>;
+                      { s.Read(command, size, data, size) } -> std::same_as<bool>;
+                    };
 
-      bool Init();
-      bool Write(const uint8_t* data, size_t size);
-      bool Read(uint8_t* cmd, size_t cmdSize, uint8_t* data, size_t dataSize);
-      bool WriteCmdAndBuffer(const uint8_t* cmd, size_t cmdSize, const uint8_t* data, size_t dataSize);
-      void Sleep();
-      void Wakeup();
+    namespace Interface {
+      template <class T>
+        requires IsSpi<T>
+      class Spi {
+      public:
+        Spi(T& spi) : impl {spi} {
+        }
+        Spi(const Spi&) = delete;
+        Spi& operator=(const Spi&) = delete;
+        Spi(Spi&&) = delete;
+        Spi& operator=(Spi&&) = delete;
 
-    private:
-      SpiMaster& spiMaster;
-      uint8_t pinCsn;
-    };
+        bool Init() {
+          return impl.Init();
+        }
+
+        bool Write(const uint8_t* data, size_t size) {
+          return impl.Write(data, size);
+        }
+
+        bool Read(uint8_t* cmd, size_t cmdSize, uint8_t* data, size_t dataSize) {
+          return impl.Read(cmd, cmdSize, data, dataSize);
+        }
+
+        bool WriteCmdAndBuffer(const uint8_t* cmd, size_t cmdSize, const uint8_t* data, size_t dataSize) {
+          return impl.WriteCmdAndBuffer(cmd, cmdSize, data, dataSize);
+        }
+
+        void Sleep() {
+          impl.Sleep();
+        }
+
+        void Wakeup() {
+          impl.Wakeup();
+        }
+
+      private:
+        T& impl;
+      };
+    }
   }
 }
--- a/src/drivers/SpiMaster.h
+++ b/src/drivers/SpiMaster.h
@@ -1,65 +1,70 @@
 #pragma once
 #include <cstddef>
 #include <cstdint>
-
-#include <FreeRTOS.h>
-#include <semphr.h>
-#include <task.h>
+#include <algorithm>
 
 namespace Pinetime {
   namespace Drivers {
-    class SpiMaster {
-    public:
-      enum class SpiModule : uint8_t { SPI0, SPI1 };
-      enum class BitOrder : uint8_t { Msb_Lsb, Lsb_Msb };
-      enum class Modes : uint8_t { Mode0, Mode1, Mode2, Mode3 };
-      enum class Frequencies : uint8_t { Freq8Mhz };
-      struct Parameters {
-        BitOrder bitOrder;
-        Modes mode;
-        Frequencies Frequency;
-        uint8_t pinSCK;
-        uint8_t pinMOSI;
-        uint8_t pinMISO;
+    template <typename T>
+    concept IsSpiMaster =
+      requires(T spi, uint8_t pin, const uint8_t* constData, uint8_t* data, const uint8_t* constCommand, uint8_t* command, size_t size) {
+        { spi.Init() } -> std::same_as<bool>;
+        { spi.Write(pin, constData, size) } -> std::same_as<bool>;
+        { spi.Read(pin, command, size, data, size) } -> std::same_as<bool>;
+        { spi.WriteCmdAndBuffer(pin, constCommand, size, constData, size) } -> std::same_as<bool>;
+        { spi.OnStartedEvent() };
+        { spi.OnEndEvent() };
+        { spi.Sleep() };
+        { spi.Wakeup() };
       };
 
-      SpiMaster(const SpiModule spi, const Parameters& params);
-      SpiMaster(const SpiMaster&) = delete;
-      SpiMaster& operator=(const SpiMaster&) = delete;
-      SpiMaster(SpiMaster&&) = delete;
-      SpiMaster& operator=(SpiMaster&&) = delete;
+    namespace Interface {
+      template <class T>
+        requires IsSpiMaster<T>
+      class SpiMaster {
+      public:
+        SpiMaster(T& spiMaster) : impl {spiMaster} {
+        }
+        SpiMaster(const SpiMaster&) = delete;
+        SpiMaster& operator=(const SpiMaster&) = delete;
+        SpiMaster(SpiMaster&&) = delete;
+        SpiMaster& operator=(SpiMaster&&) = delete;
 
-      bool Init();
-      bool Write(uint8_t pinCsn, const uint8_t* data, size_t size);
-      bool Read(uint8_t pinCsn, uint8_t* cmd, size_t cmdSize, uint8_t* data, size_t dataSize);
+        bool Init() {
+          return impl.Init();
+        }
 
-      bool WriteCmdAndBuffer(uint8_t pinCsn, const uint8_t* cmd, size_t cmdSize, const uint8_t* data, size_t dataSize);
+        bool Write(uint8_t pinCsn, const uint8_t* data, size_t size) {
+          return impl.Write(pinCsn, data, size);
+        }
 
-      void OnStartedEvent();
-      void OnEndEvent();
+        bool Read(uint8_t pinCsn, uint8_t* cmd, size_t cmdSize, uint8_t* data, size_t dataSize) {
+          return impl.Read(pinCsn, cmd, cmdSize, data, dataSize);
+        }
 
-      void Sleep();
-      void Wakeup();
+        bool WriteCmdAndBuffer(uint8_t pinCsn, const uint8_t* cmd, size_t cmdSize, const uint8_t* data, size_t dataSize) {
+          return impl.WriteCmdAndBuffer(pinCsn, cmd, cmdSize, data, dataSize);
+        }
 
-    private:
-      void SetupWorkaroundForFtpan58(NRF_SPIM_Type* spim, uint32_t ppi_channel, uint32_t gpiote_channel);
-      void DisableWorkaroundForFtpan58(NRF_SPIM_Type* spim, uint32_t ppi_channel, uint32_t gpiote_channel);
-      void PrepareTx(const volatile uint32_t bufferAddress, const volatile size_t size);
-      void PrepareRx(const volatile uint32_t cmdAddress,
-                     const volatile size_t cmdSize,
-                     const volatile uint32_t bufferAddress,
-                     const volatile size_t size);
+        void OnStartedEvent() {
+          impl.OnStartedEvent();
+        }
 
-      NRF_SPIM_Type* spiBaseAddress;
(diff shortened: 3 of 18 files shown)


src/components/ble/DfuService.h
```diff
@@ -9,13 +9,13 @@
 #undef max
 #undef min
 
+#include <drivers/SpiNorFlash.h>
+#include "infinitime.h"
```
NeroBurner, line 13: later this header is included as `port/infinitime.h` I think this was unintentionally left out here

```suggestion
#include "port/nfinitime.h"
```

src/components/ble/DfuService.h
```diff
@@ -9,13 +9,13 @@
 #undef max
 #undef min
 
+#include <drivers/SpiNorFlash.h>
```
NeroBurner, line 12: our own headers should be included with `""` instead of `<>`
```suggestion
#include "drivers/SpiNorFlash.h"
```

src/drivers/Spi.h
```diff
@@ -1,28 +1,56 @@
...
+          impl.Wakeup();
+        }
+
+      private:
+        T& impl;
+      };
+    }
   }
-}
+}
```
NeroBurner, line 56: adding newline at the end of the file
```suggestion
}

```

src/drivers/Spi.h
```diff
@@ -1,28 +1,56 @@
...
-      bool Read(uint8_t* cmd, size_t cmdSize, uint8_t* data, size_t dataSize);
-      bool WriteCmdAndBuffer(const uint8_t* cmd, size_t cmdSize, const uint8_t* data, size_t dataSize);
-      void Sleep();
-      void Wakeup();
-
-    private:
-      SpiMaster& spiMaster;
-      uint8_t pinCsn;
-    };
+    template <typename T>
```
NeroBurner, line 9: I'd rename the generic `T` to `SpiImpl` to be more specific about what this type represents
```suggestion
    template <typename SpiImpl>
```

src/drivers/SpiMaster.h
```diff
@@ -1,65 +1,70 @@
...
-      enum class Modes : uint8_t { Mode0, Mode1, Mode2, Mode3 };
-      enum class Frequencies : uint8_t { Freq8Mhz };
-      struct Parameters {
-        BitOrder bitOrder;
-        Modes mode;
-        Frequencies Frequency;
-        uint8_t pinSCK;
-        uint8_t pinMOSI;
-        uint8_t pinMISO;
+    template <typename T>
```
NeroBurner, line 8: I'd rename the generic `T` to `SpiImpl` to be more specific about what this type represents

```suggestion
    template <typename SpiImpl>
```

src/drivers/SpiMaster.h
```diff
@@ -1,65 +1,70 @@
...
-      volatile TaskHandle_t taskToNotify;
-      SemaphoreHandle_t mutex = nullptr;
-    };
+      private:
+        T& impl;
+      };
+    }
   }
-}
+}
```
NeroBurner, line 70:
```suggestion
}

```

src/drivers/SpiNorFlash.h
```diff
@@ -1,60 +1,116 @@
...
-      SpiNorFlash& operator=(const SpiNorFlash&) = delete;
-      SpiNorFlash(SpiNorFlash&&) = delete;
-      SpiNorFlash& operator=(SpiNorFlash&&) = delete;
-
-      struct __attribute__((packed)) Identification {
-        uint8_t manufacturer = 0;
-        uint8_t type = 0;
-        uint8_t density = 0;
-      };
+    template <typename T>
```
NeroBurner, line 9: suggest using more specific name for the type

```suggestion
    template <typename MemoryImpl>
```

src/drivers/SpiNorFlash.h
```diff
@@ -1,60 +1,116 @@
...
+          impl.Wakeup();
+        }
+
+      private:
+        T& impl;
+      };
+    }
   }
-}
+}
```
NeroBurner, line 116:
```suggestion
}

```

src/drivers/nrf52/Spi.h
```diff
@@ -0,0 +1,30 @@
...
+        void Sleep();
+        void Wakeup();
+
+      private:
+        Pinetime::Drivers::Nrf52::SpiMaster& spiMaster;
+        uint8_t pinCsn;
+      };
+    }
+  }
+}
```
NeroBurner, line 30:
```suggestion
}

```

src/drivers/nrf52/SpiMaster.h
```diff
@@ -0,0 +1,67 @@
...
+        SpiMaster::Parameters params;
+
+        volatile uint32_t currentBufferAddr = 0;
+        volatile size_t currentBufferSize = 0;
+        volatile TaskHandle_t taskToNotify;
+        SemaphoreHandle_t mutex = nullptr;
+      };
+    }
+  }
+}
```
NeroBurner, line 67:
```suggestion
}

```

src/drivers/spiFlash/SpiNorFlash.h
```diff
@@ -0,0 +1,62 @@
...
+          DeepPowerDown = 0xB9
+        };
+        static constexpr uint16_t pageSize = 256;
+
+        Pinetime::Drivers::Nrf52::Spi& spi;
+        Identification device_id;
+      };
+    }
+  }
+}
```
NeroBurner, line 62:
```suggestion
}

```

src/port/infinitime.h
```diff
@@ -0,0 +1,24 @@
...
+  namespace Drivers {
+#ifdef TARGET_DEVICE_PINETIME
+  using SpiMaster = Interface::SpiMaster<Pinetime::Drivers::Nrf52::SpiMaster>;
+  using Spi = Interface::Spi<Pinetime::Drivers::Nrf52::Spi>;
+  using SpiNorFlash = Interface::SpiNorFlash<Pinetime::Drivers::SpiFlash::SpiNorFlash>;
+#else
+  #error "No target device specified!"
+#endif
+  }
+}
```
NeroBurner, line 24:
```suggestion
}

```

src/drivers/St7789.h
```diff
@@ -1,10 +1,12 @@
 #pragma once
+#include "Spi.h"
```
NeroBurner, line 2: absolute paths please

```suggestion
#include "drivers/Spi.h"
```

src/drivers/nrf52/Spi.cpp
```diff
@@ -1,8 +1,8 @@
-#include "drivers/Spi.h"
+#include "Spi.h"
```
NeroBurner, line 1: absolute include paths please
```suggestion
#include "drivers/nrf52/Spi.h"
```

NeroBurner (changes requested):
I like the usage of the `impl` reference, as I was wary of the compile time overhead if so much code must be in the headers as templates require. But with the `impl` reference this should be ok as the heavy lifting is in the impl.cpp file

src/drivers/SpiMaster.h
```diff
@@ -1,65 +1,70 @@
...
-      enum class Frequencies : uint8_t { Freq8Mhz };
-      struct Parameters {
-        BitOrder bitOrder;
-        Modes mode;
-        Frequencies Frequency;
-        uint8_t pinSCK;
-        uint8_t pinMOSI;
-        uint8_t pinMISO;
+    template <typename T>
+    concept IsSpiMaster =
```
ArsenArsen, line 9: (outsider suggestion) (ditto for other places) Concepts should be named like adjectives rather than predicates, i.e. `SpiMaster` rather than `IsSpiMaster`.  This convention is established in the standard ([`same_as`, `derived_from`, `destructible`, `regular`, ...](https://en.cppreference.com/w/cpp/concepts)) and is in line with how other languages treat traits and trait-like features

JF002:
@NeroBurner 

> I like the usage of the impl reference, as I was wary of the compile time overhead if so much code must be in the headers as templates require. But with the impl reference this should be ok as the heavy lifting is in the impl.cpp file

In theory (and according to my experiments with compiler explorer), those calls to impl.xxx() should be optimized by the compiler, since the type of impl is known at compile time.

src/drivers/SpiMaster.h
JF002: I agree with this. However... this rule is not easy to apply here : 
 - in drivers/SpiMaster.h, there's `Pinetime::Drivers::IsSpiMaster`, the concept definition
 - in drivers/SpiMaster.h, there's also `Pinetime::Drivers::Interface::SpiMaster`, the actual driver, that owns an `impl` of type `IsSpiMaster`.
 - in port/infinitime.h, we declare `PineTime::Drivers::SpiMaster`, the type that will be used in the rest of the code.

Renaming the concept from `IsSpiMaster` to ` SpiMaster` will conflict with `SpiMaster` in InfiniTime.h. I'm open to suggestions regarding the naming of the types and namespaces, though!

JF002:
I think this PR is ready for review.

In this PR, I added a new `concept` and `proxy object` for the following low-level drivers:
 - Display (St7789)
 - HeartRateSensor (HRS3300)
 - MotionSensor (BMA42x)
 - SpiMaster
 - FlashMemory
 - TouchPanel
 - TwiMaster
 - Watchdog

The goals of these changes are:
 - They allow to easily customize the implementation of a low-level driver. This will be useful to support variants of the PineTime (Colmi P8, for example). 
 - They also allow a cleaner implementation in [InfiniSim](https://github.com/InfiniTimeOrg/InfiniSim) ([corresponding PR](https://github.com/InfiniTimeOrg/InfiniSim/pull/73)).
 - This is a first step to a complete hardware abstraction that will allow to port InfiniTime on other devices and implement unit tests, for example.
 - This is a cleaner design that is more expressive and that explicitly describes the interfaces that must be implemented to support new hardware without making use of runtime polymorphism.
 - I personally think this is a great use of this new feature (concepts) of modern C++(20).

The current naming of the classes / namespaces are:
```c++
namespace Pinetime {
  namespace Drivers  {
    template <typename DeviceImpl>
    concept IsDevice = { /* ... */ };
    
    namespace Interface {
      template <class T>
        requires IsDevice<T>
      class Device {
      /* ... */
      private:
        T& impl;
      };
    }

    using Device = Interface::Device<ActualDevice>;
  }
}
```

Basically:
 - the `concept` that describes the API of the classes are defined in `Pinetime::Drivers`. They are prefixed with `Is` to avoid  name collision. This has already been discussed [here](https://github.com/InfiniTimeOrg/InfiniTime/pull/1387#discussion_r1002752053) and I haven't fond a better naming convention yet.
 - The 'proxy object' that will forward the calls from the application to the actual implementation of the driver (needed to avoid runtime polymorphism) are declared in `Pinetime::Drivers::Interface`. We could maybe renamed this namespace as `Proxy` since that's what they are : proxy objects.
 - The selection **at build time** of the actual interface is done in header files in `port/`. This is the only place where you'll see `#ifdef`s. The classes in `Pinetime::Drivers::Interface` are aliased in `Pinetime::Drivers` so that this new design causes minimal changes to the rest of the application.

TODO:

- [ ] Review the naming convention
- [ ] Write design documentation
- [ ] Merge :)

Riksu9000:
It seems to me like there's no framework to choose the drivers yet, unless I'm missing something. How would you make two drivers selectable with CMake?

Riksu9000:
What about main.cpp, where the touchPanel is hard coded to be constructed with the Cst816S implementation?

JF002:
> What about main.cpp, where the touchPanel is hard coded to be constructed with the Cst816S implementation?

Yes, right. I didn't encounter this issue in InfiniSim since it has its own main.cpp file, that is distinct from the one in InfiniTime.
To solve this in InfiniTime, we could move the initialization of the driver objects into another file, specific to the target, or even provide 1 main.cpp file per target. This would give more freedom to make custom initialization on a target basis.

JF002:
@Riksu9000 
May I suggest that we limit the scope of this PR to the re-design of the low-level hardware drivers while keeping exactly the same functionalities that the current project? I think we can work on actually choosing the drivers and implement support for other variants based on this design in a future PR (which I plan to work on, obviously) ?

Unless there's any doubt that this design is correct and that we'll be able to actually support multiple hardware with it ?

Riksu9000:
https://stackoverflow.com/questions/21152018/is-it-worth-it-to-avoid-polymorphism-in-order-to-gain-performance 

Reading the linked discussion it seems like avoiding runtime polymorphism might not make a big difference, but it makes the architecture more confusing. I don't really understand why it wouldn't get optimized without the proxy, but I think it could be worth the tradeoff to make the code significantly simpler.

qookei:
This abstraction scheme as-is is unable to handle multiple differing implementations of the same concept being needed at runtime, for example if trying to port to a device that has two different SPI flashes on board, or (if this is extended to GPIOs), if some things are wired up to GPIOs from the SoC and some are going through an I2C GPIO expander chip. Handling these cases would require either explicitly templating the user on the peripheral type, or introducing virtual functions.

JF002:
@Riksu9000 
> Reading the linked discussion it seems like avoiding runtime polymorphism might not make a big difference, but it makes the architecture more confusing

This is an interesting discussion! Let me (try to) explain the goals of this new design.

The first goal of this design is **not** optimizations. Instead, it's more about "semantics" : I want the design to be as close as the hardware as possible.

InfiniTime is a firmware for smartwatches. To the InfiniTime point of view, a smartwatch is a device with a display, a touchpanel, a motion sensor, a heart rate sensor, a button, an external memory,... This list of peripherals is know and does not change : every PineTime has the exact same hardware.
Also, the kind of smartwatch targeted by InfiniTime is based on a very memory and resource constraint hardware : low power cpu, low RAM, low FLASH. 

So the idea is to provide a design that explicitly builds one and only one instance of each driver : the one that will be needed at runtime for a specific hardware.

Also, the concept brings a nice way to describe the interface of a driver, in my opinion.

In the end, I wanted this design to be as close as what we currently have in InfiniTime : everything is statically resolved and allocated.

BUT... you said that this new implementation is confusion, and that is certainly something I would like to avoid so I guess I still need to work a bit on that.

Using dynamic polymorphism and trust the compiler to optimize everything nicely might actually work! This is something I need to think about and experiment (compare).

@qookei 
> This abstraction scheme as-is is unable to handle multiple differing implementations of the same concept being needed at runtime

Is this really an issue? Up to now, I've I've never seen a smartwatch with multiple memory cheap (let alone 2 different ones), or multiple displays, multiple HR sensors,... Do you think there are reasonable reason to need to instantiate 2 different implementations of the same "category" ?

qookei:
> Is this really an issue? Up to now, I've I've never seen a smartwatch with multiple memory cheap (let alone 2 different ones), or multiple displays, multiple HR sensors,... Do you think there are reasonable reason to need to instantiate 2 different implementations of the same "category" ?

To be fair, in most cases I can't think of any applicable to InfiniTime, apart from potentially the aforementioned GPIO expanders vs SoC GPIO banks. I thought it still might be worth mentioning.

Avamander:
> Is this really an issue? Up to [...]

I tend to agree that it's unlikely to be an issue for InfiniTime any time soon.

JF002:
Thanks everyone for your feedback on this PR. I'll think a bit more about this new hardware abstraction since the additional complexity brought by the new design might not be needed after all.

PR #1392, opened 2022-10-25
Almost working morse area.

programmin1:
Morse text in a screen... except it does not compile, 
error: too many initializers for ‘std::__array_traits

I do not know why Infinitime is limited on apps in this array in ApplicationList.h??


FintasticMan (changes requested):
You need to set nScreens to 3 on line 39. It is also better to leave an empty line between each screen in the definition of the array, so that it's more obvious when reading the code which apps are on which screen.

minacode:
You also need to insert
```cpp
{nullptr, Apps::None}
```
elements for each empty slot on the new app menu screen. Otherwise the firmware will crash when you tap on those slots.

PR #1395, opened 2022-10-27
Enable CMake unity build

tnixeu:
This PR enables building InfiniTime with a CMake unity build.

Compared to a normal build the CMake unity build compiles on my machine with Docker in 6s instead of 9s and the pinetime-app.bin is 180 bytes smaller.

Some files currently cause compile errors. Those files are skipped by adding the property `SKIP_UNITY_BUILD_INCLUSION` to the file that causes problems.


src/displayapp/fonts/CMakeLists.txt
```diff
@@ -32,6 +32,12 @@ foreach(FONT ${FONTS})
    add_custom_target(infinitime_fonts_${FONT}
       DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/${FONT}.c
    )
+
+# skip files, which currently cause problems when truning on CMake untiy build
```
NeroBurner, line 36: truning?

src/CMakeLists.txt
```diff
@@ -1143,3 +1143,27 @@ if(BUILD_RESOURCES)
   add_subdirectory(resources)
 endif()
 
+# skip files, which currently cause problems when truning on CMake untiy build 
+
```
NeroBurner, line 1147:
```suggestion
```

src/displayapp/fonts/CMakeLists.txt
tnixeu: thx for spotting the typo.

It is fixed in the forced push

docker/build.sh
```diff
@@ -63,6 +63,7 @@ CmakeGenerate() {
     -S "$SOURCES_DIR" \
     -B "$BUILD_DIR" \
     -DCMAKE_BUILD_TYPE=$BUILD_TYPE \
+    -DCMAKE_UNITY_BUILD="ON" \
```
NeroBurner, line 66: little nitpick, the `""` aren't needed in this script
```suggestion
    -DCMAKE_UNITY_BUILD=ON \
```

NeroBurner (approved):
basically free CI build time and firmware size improvements? I'll take that! Thank you very much!

lman0:
what are the file removed ? does ti remove apss?

tnixeu:
> what are the file removed ? does you remove apps?

No files are removed. I probably have phrased it badly. Some files are skipped for inclusion into the unity build. Those files are compiled individually and linked just like without the CMake unity build.

JF002:
I've just learnt about this feature. According to the  [documentation](https://cmake.org/cmake/help/latest/prop_tgt/UNITY_BUILD.html), it combines target sources files into batches for faster compiling.

The doc also said that it should not be enabled by default (the option should be under the control of the developer via CLI arguments) and that it is recommended to disable the unity property on files that are known to cause issues, which is exactly what this PR does.

However... I'm a bit worried about the size reduction of the final binary. Do you know why the binary is smaller when built with this feature?

Avamander:
> However... I'm a bit worried about the size reduction of the final binary. Do you know why the binary is smaller when built with this feature?

I imagine it can function like weak LTO. Which we should/could enable anyways. Full LTO caused an increased RAM usage, albeit a significant flash decrease.

JF002 (approved):

PR #1397, opened 2022-10-28
Feature/proper batery percentage

AlexXZero:
This PR adds linear approximation util and use it for making battery discharge curve more linear.

These changes were done independently on https://github.com/InfiniTimeOrg/InfiniTime/pull/585, I guess my solution is more generic. The main idea of my PR is adding linear approximation which might be used in any other feature including this old PR.
I used my own measurements of battery, not sure if they accurate enough, so you are welcome to update my digits:
```C
  static const Utility::LinearApproximation<uint16_t, uint8_t, 5> aprox {{{
    {3200, 0},  // minimum voltage of battery before shutdown ( depends on the battery )
    {3600, 10}, // keen point corresponded to 10% of battery
    {3700, 25},
    {3800, 50},
    {4180, 100} // maximum voltage of battery ( max charging voltage is 4.21 )
  }}};
``` 
My measurements:

image
blue - measured voltage
red - measured percentage (using 1.9.0)
orange - filtered voltage (I'm working on adding filter in another branch)
green - expected percentage using this algorithm (actually data is filtered, but it shows trade line of approximation)

I can share my data by request.


Avamander:
Kindly rebase on top of `origin/develop`.

AlexXZero:
@Avamander Thanks for feedback, I will rebase my changes to develop + rename branch name as it looks like current name is not allowed.

Issue #1382, opened 2022-10-20
Should I submit PR's for apps that are a little niche?


Derek52:
Hi, I am working on some apps for my Pinetime to help me when gaming. I have already built 2, a life point counter for Magic The Gathering, and a life point counter for Yu-Gi-Oh. Next, I'm thinking about making a dice roll simulator. 

Anyway, I like my life counter apps a lot, and want to know if I can get them merged into upstream. To get these apps working I also built a couple of Counter widgets. I took the Counter that is used in the timer app, and wrote a couple of new versions, one is a 2 digit, one is a 4 digit. They may have uses for someone down the line. I'm not sure.

I think my life point counter apps have use for people. But, I don't think they are going to be used enough to make them apps enabled by default. But, there may be users who want to go through the code and enable them for their own watches. Also, if you merge them into upstream, I'll have a much easier time updating my own personal watch's build.

Basically, I want to know if there is interest in me submitting a couple of PR's that add apps to the watch, without automatically enabling them.

FintasticMan:
At the moment, that kind of app will not be merged because of space constraints, but I think it is a good idea to make pull requests anyway. This way you can get feedback on your app and its code, and when we have a way to load apps from the external flash, it might be merged sooner.

yusufmte:
As an aside, dice rolling app PR here https://github.com/InfiniTimeOrg/InfiniTime/pull/1326 in case you are interested :) Basic functionality in it is complete from my end and I've been using it on my pinetime. I'd be curious if it suits your needs / if it's similar to what you were thinking of.

minacode:
Well.. #1296.

Derek52:
@yusufmte I like your dice rolling app a lot. I like that it appears to change the font color when rolling to help you visualize that you have gotten a new result. I was trying to come up with a simple way to show the result had changed, when say, you get a 3 two times in a row. I guess I don't have to make a dice rolling app after all.

yusufmte:
@Derek52 Thanks -- glad you liked it!

By the way, IMO I agree with @FintasticMan about submitting the PR. Who knows, there may be more people interested in those specific uses than anticipated, and it wouldn't really affect the user experience for people who don't use it. Besides, the improvements to the Counter class could be generally useful. And the MTG counter would just be two 2-digit counters with a + and -, right? So it could perhaps be used as a general scorekeeping app for MTG but also for anything like ping pong, rock paper scissors, etc.?

[...]

PR #1406, opened 2022-10-31
I made a Life Point counter app for Magic The Gathering

Derek52:
I couldn't figure out how to get the simulator working, so I don't have screenshots to share. But, as referenced in #1382 I made an MTG life counter app. I also implemented a new 2 digit counter widget that could be useful. It has 2 plus buttons on the top, and 2 minus buttons on the bottom. These add and subtract either 10 or 1 from the total. The number displayed changes color as your life goes down, as a warning(i.e. it goes red when it's really low).

Holding the top left plus button will reset the counter to '40' the default starting value I gave it, for Commander games. I probably should refactor the app to have a dedicated reset button.

One thing I could use some feedback on is, how to make the app persist my life points upon closing the app. Currently, I just made the value of the counter, a static variable, which was advice given to me in the matrix dev room, I'm not sure if this was the best way to do that though. The relevant line of code is [here.](https://github.com/Derek52/InfiniTime/blob/048d7f7594da355502554b56b9eabdb6efc55e78/src/displayapp/widgets/TwoDigitCounter.h#L48)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -407,6 +407,7 @@ list(APPEND SOURCE_FILES
         displayapp/screens/Styles.cpp
         displayapp/Colors.cpp
         displayapp/widgets/Counter.cpp
+        displayapp/widgets/2DigitCounter.cpp
         displayapp/widgets/PageIndicator.cpp
         displayapp/widgets/StatusIcons.cpp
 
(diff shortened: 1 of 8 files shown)


minacode:
> One thing I could use some feedback on is, how to make the app persist my life points upon closing the app.

You can take a look at `src/components/settings/Settings.cpp` to see how the persistence is implemented. Then you can write and load the value to the flash at closing or loading the app. Also see the two PRs for alarm persistence.

PR #783, opened 2021-10-25
Stopwatch Persistence

desttinghim:
I was using the stopwatch app while to track the cure time of epoxy (I didn't use the timer because I was working on many items in quick succession). I found a few annoyances with the stopwatch while doing so:

- The current time is not displayed on the stopwatch screen
- The stopwatch app could not be left without the stopwatch resetting

I was already planning to dig into the code, so I decided this was a great way to introduce myself to it. With this pull request, the following features are added/removed:

- The current time is displayed on the stopwatch screen
- The stopwatch will keep running while other screens are visited
- The physical button no longer acts as a stop button (this is a personal preference)

To make this work I added `StopWatchController` to hold the state of the stopwatch, the elapsed time and 2 laps. The stopwatch app then uses this data when it's opened to display the correct values.

https://user-images.githubusercontent.com/10042482/138707672-75a924d5-1100-4cc7-a077-293d06c5dfd1.mp4

PR #1410, opened 2022-11-01
Stopwatch: add persistence

pptime02:
Stopwatch now keeps running in background if screen is left.
Implemented this by introducing a StopWatchController.
In additon, now an infinite lap count is possible (but only two laps are stored).
Largely based on the work of Louis Pearson (@desttinghim@github.com), see https://github.com/InfiniTimeOrg/InfiniTime/pull/783. Nevertheless, I removed all changes not related to the stopwatch persistence feature.


desttinghim:
Cool! Glad to see this is getting completed. I started working on rebasing it but I didn't want to get a build environment set up - thank you for taking it on :slightly_smiling_face:

src/components/stopwatch/StopWatchController.cpp
```diff
@@ -0,0 +1,101 @@
+#include "StopWatchController.h"
+
+using namespace Pinetime::Controllers;
+
+namespace {
+  TickType_t calculateDelta(const TickType_t startTime, const TickType_t currentTime) {
```
minacode, line 6: CalculateDelta...

minacode (changes requested):
Nice work on this feature. I scrolled through and found some things regarding code style. They are mostly functions starting with a lowercase character. I may not have found all of them.

desttinghim:
Those are probably leftover from the work I did

src/components/stopwatch/StopWatchController.cpp
```diff
@@ -0,0 +1,101 @@
+#include "StopWatchController.h"
```
Elements6007, line 1:
```suggestion
#include "components/stopwatch/StopWatchController.h"
```

pptime02:
Thank you all for taking a look on it. I think I applied all proposals, and found no further lowercase-functions.

minacode:
What about those struct types that end with `_t`? Do they fit our code style?

pptime02:
> What about those struct types that end with `_t`? Do they fit our code style?

I checked other structs by running `rg 'struct .*_t \{' InfiniTime/src/` and `rg 'struct .* \{' InfiniTime/src/`. Seems like the stopwatch was the only place where `_t` was used, thus I removed the remaining `_t` it in the last commit.

Xylemon:
Any chance of this getting rebased against `main`? This would be very useful.

mark9064:
Good news: #2141 is merged now :)

PR #1431, opened 2022-11-12
Notifications are better displayed

jspkay:
This is my first pull request ever, the name of the branch was supposed to be "minorUXimprovements". I basically changed the behaviour of the notification screens a little bit

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -165,11 +165,18 @@ void DisplayApp::Refresh() {
           brightnessController.Lower();
           vTaskDelay(100);
         }
+
         PushMessageToSystemTask(Pinetime::System::Messages::OnDisplayTaskSleeping);
         state = States::Idle;
         break;
       case Messages::GoToRunning:
         ApplyBrightness();
+
+        // reload main display app if we are notifications or quicksettings or app menu
+        if(currentApp == Apps::Launcher || currentApp == Apps::Notifications || currentApp == Apps::QuickSettings){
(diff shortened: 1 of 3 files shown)


minacode:
> This is my first pull request ever

Congrats and welcome! 😊

> I basically changed the behaviour of the notification screens a little bit

Can you explain what you did and maybe also why?

jspkay:
Thank you 😄
So, I received my pinetime not long ago and I've been using it since. I've noticed some problems in the experience. This request contains the fix to two problems:
- the title notification (if long) is not displayed correctly, because as soon as the title placed, it starts to move. The problem here is that the first characters cannot be seen. I made the title stay fixed for the first second, and only then starts to move
- another problem was the scrolling of notifications. If I receive multiple notifications all togheter, I would like to scroll among them without the need to get to the watchface first, so the new behavior allows to scroll a preview fo going backwards. Also, I changed the directing of the animation, for the aforementioned scrolling to result more natural
- finally, there were times that I left my watch on a notification or on the quick settings menu. The problem is that whenever I wake it up, I see the last notification or the quick settings, while it would be better to see the time if it went to sleep witouth opening an app

minacode:
Ok, so I get that your PR contains three separate changes. I recommend to also submit them as three separate PRs. They may be small, but that is a good thing. Having a discussion about all of them in one PR is not ideal.

jspkay:
Oh, I get it now... It's my first time, I thought it worked the opposite way: the more changes the better. Anyways, I understand why this way is much simpler.
So, what do I do? Do I remove the pr and open three different? Or what?

FintasticMan:
I would say that is probably the best way to go about it. Make sure to create a different branch for each of the pull requests.

PR #1434, opened 2022-11-13
Tennis score counter

rdeininger:
Added a tennis score counter app to the app list.
- Capable of counting game points, advantage and 3 sets
- Capable of handling tie break
- Allows to add or subtract game points
- Will show a "Wi" when the game is won and with another click on the "+" button will add it to the set score
- Will not save the game score on app close


minacode:
Something is very broken here.

PR #1435, opened 2022-11-13
Notification preview interaction

jspkay:
Thi pr proposes to make the new notifications to feel more natural in two ways:
- new notification (i.e. previews) are interactive. It is possible to scroll down to see older notifications, right to delete it and up to get back to the clock,
- the animation is more natural with respect to the previous behaviour, indeed the notification comes from down to up, as if it was piling with the others.

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -165,11 +165,18 @@ void DisplayApp::Refresh() {
           brightnessController.Lower();
           vTaskDelay(100);
         }
+
         PushMessageToSystemTask(Pinetime::System::Messages::OnDisplayTaskSleeping);
         state = States::Idle;
         break;
       case Messages::GoToRunning:
         ApplyBrightness();
+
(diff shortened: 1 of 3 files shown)


Riksu9000:
@jspkay Previews are now interactive, so what's left is the animation direction. It's true that in InfiniTime things tend to pop up from the bottom, however the notifications lie above the watchface. This is the reason the notifications come from above. Ideally notifications would probably just be displayed above the app without closing it.

Can you explain how this change would feel more natural?

jspkay:
@Riksu9000 thank you for the answer, I appreciate.
I see your point and it make logical sense, I agree. What bother me is that, when a new notification comes in, with present animation, **it feels natural to swipe up** to see more notifications, **because the new notification comes from the top**. Furthermore, the notifications pile up from the bottom, so that the older notifications are on the top of the pile. The idea of the animation is to give this sense of "piling from the bottom", which makes more sense to me. When you want to see an older notification, you simply "push the new notification back down" (i.e. swipe down), which is the usual way to go through the notifications.

Issue #1390, opened 2022-10-24
Return to Watch Face a Few Seconds After Display Sleep


RageGamerBoi:
### Verification

- [X] I searched for similar feature request and found none was relevant.

### Pitch us your idea!

About 10-15 seconds the display goes to sleep, if the watch has a navigation screen up it should return to the watch face.

### Description

If a navigation screen is opened and the watch goes to sleep because it isn’t being used, it should return to the watch face. This shouldn’t apply inside apps or for a few seconds after the watch goes to sleep because the page might need to be used again.

Screens where this would apply:
Quick settings (but not the actual settings menu)
App list
Notifications shade

This is useful when a menu screen is left open because it wasn’t closed, and you go to quickly check the time later and you’re greeted with the screen you had open and have to fumble with the watch to check the time.

minacode:
I disagree (a little). Personally, I send the watch to sleep by pressing the button a few times until it sleeps. Whenever I let it sleep on a screen I intend to use it later again. It's mostly the timer that I want to check quickly. 

On a less personal level, that behavior would erase any unsaved in-app state. Think of Twos for example. 
Given the PR that implements the app stack, we should maybe aim for persistence implementations for every app 🤔

Riksu9000:
A similar idea was previously closed as wontfix #570, however considering the navigation screens only it might not be a bad idea. With this change, quick settings and the app list would sort of be a part of the same "app" as the watch face conceptually.

Is this a common enough problem though that it is worth implementing this feature, even if #618 and #1592 were fixed?

RageGamerBoi:
This issue was originally trying to fix my own problems relating to #618 , regardless of touch driver problems etc.
I believe this should probably still be implemented in some fashion as even accidental touches cause problems, and seeing as the PineTime is primarily a watch, it seems like a logical solution.
Having the time in every app is a different discussion imho as a small clock in the corner is easy to miss at a glance, but idk.

RageGamerBoi:
I have put this suggestion as a reply on #618

Issue #570, opened 2021-08-11
Go to main screen after timeout


drzacek:
It looks like when I enter settings or app list, or notification*, PineTime remembers that screen and stays there forever. 
While it is okay for it to stay and not reset after 5s timeout (I can imagine it being annoying when you check firmware info etc, display dims and you're back in the clock and have to navigate back all the time), it should eventually go back to clock - when I check the time one hour later, I rather don't expect it to show the settings menu. 
And the point of having "wrist raise to wake" is not having to touch the screen (sometimes it is impossible).

Also, sorry for opening lots of issues lately, I only recently got my PineTime and I'm actively using/testing it as daily driver and find lots of stuff that can be improved imho.
I'm also setting up the dev environment again and fiddling with the code a bit, so I might start doing so actual work soon myself (no promises).

* sometimes I think it goes back to Clock screen on its own when notification was open. Maybe there was new notification in between or I accidentally touched the screen, still trying to catch this.

hubmartin:
Hi, for me it makes sense that menus/config screen should close after some bigger timeout and go to clock face. Not sure if closing should have the same timeout like turning screen off. Maybe make another timeout (like 10 seconds longer?) so if your screen turn off automatically, then you have few more seconds to wake the clock and finish your intention in the menu. Or maybe with the new dimming before sleep functionality it actually might go to main screen together with timeout.

Also the current wrist raise to wake sometimes triggers when you don't need it, so I wouldnt base going to main screen on that, or unless its detection is rock solid.

Just please do not jump to main screen automatically when I'm in an app. I use music player with wrist-wakeup which stays open to control volume and pause when I listen podcasts and work on the garden.

Riksu9000:
#480 Will probably help with this because it makes returning to the watch face much easier. Just hold the button down for a bit when you're done navigating the settings and it returns straight to the watch face.

>Also, sorry for opening lots of issues lately, I only recently got my PineTime and I'm actively using/testing it as daily driver and find lots of stuff that can be improved imho.

Don't be sorry. It's important for making this firmware as good as it can be :)

JF002:
>>  Also, sorry for opening lots of issues lately, I only recently got my PineTime and I'm actively using/testing it as daily driver and find lots of stuff that can be improved imho.

> Don't be sorry. It's important for making this firmware as good as it can be :)

I agree, feedbacks are very valuable for us to continue improving the project! And many devs open new PR after they saw a bug report of feature request !

[...]

PR #1436, opened 2022-11-13
Auto return to clock [UX improvement]

jspkay:
When the os goes to sleep after loading the quick settings or the notifications or the application list, it is better to reload the clock. This allows the user not to think about it when leaving the watch in those states and instead see the clock directly when the screen goes on again without the need of user interaction

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -165,11 +165,18 @@ void DisplayApp::Refresh() {
           brightnessController.Lower();
           vTaskDelay(100);
         }
+
         PushMessageToSystemTask(Pinetime::System::Messages::OnDisplayTaskSleeping);
         state = States::Idle;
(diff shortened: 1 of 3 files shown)


Riksu9000:
There were a few issues about this idea, which have been closed.
https://github.com/InfiniTimeOrg/InfiniTime/issues/1390
https://github.com/InfiniTimeOrg/InfiniTime/issues/570

https://github.com/InfiniTimeOrg/InfiniTime/issues/1390#issuecomment-1438547830 This comment sums up my thoughts. There are other issues that are related, but this shouldn't be considered a fix for those. Unless leaving the screen open is a common mistake, it's probably not worth fixing.

I updated the branch, but there were breaking changes in DisplayApp.

tituscmd:
This feature has been merged via #1980 so this PR can be closed :)

PR #1437, opened 2022-11-13
Quick settings cancel tap

jspkay:
I am not sure what is going on with all these changes. The only actual change to consider is the one to QuickSettings.cpp.
I am sorry for the inconvenience, I'm learning git practicing.

Anywas, the pr proposes to cancel a tap when the finger is dragged onto other elements

### DIFF
--- a/src/displayapp/screens/Notifications.cpp
+++ b/src/displayapp/screens/Notifications.cpp
@@ -135,6 +135,7 @@ bool Notifications::OnTouchEvent(Pinetime::Applications::TouchEvents event) {
     }
     return false;
   }
+  
   switch (event) {
     case Pinetime::Applications::TouchEvents::SwipeRight:
       if (validDisplay) {
--- a/src/displayapp/screens/Notifications.cpp
+++ b/src/displayapp/screens/Notifications.cpp
@@ -79,6 +79,12 @@ void Notifications::Refresh() {
       timeoutLinePoints[1].x = pos;
       lv_line_set_points(timeoutLine, timeoutLinePoints, 2);
     }
+
+    if(!this->scrolling && tick >= timeoutTickCountStart + timeoutStartScrolling){
+        this->scrolling = true;
+        lv_label_set_long_mode(currentItem->alert_type, LV_LABEL_LONG_SROLL_CIRC);
+    }
+
   }
 
   if (dismissingNotification) {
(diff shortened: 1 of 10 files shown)


minacode:
This is sadly not mergeable in this state. You have to create a clean PR, because the maintainers can only accept the whole PR or nothing in the end.

src/displayapp/screens/Notifications.cpp
```diff
@@ -135,7 +135,7 @@ bool Notifications::OnTouchEvent(Pinetime::Applications::TouchEvents event) {
     }
     return false;
   }
-
+
```
minacode, line 138: This is not needed.

src/displayapp/screens/settings/QuickSettings.cpp
```diff
@@ -10,8 +10,12 @@ using namespace Pinetime::Applications::Screens;
 namespace {
   void ButtonEventHandler(lv_obj_t* obj, lv_event_t event) {
     auto* screen = static_cast<QuickSettings*>(obj->user_data);
-    if (event == LV_EVENT_CLICKED) {
-      screen->OnButtonEvent(obj);
+    static lv_obj_t* pressedObj = nullptr;
```
minacode, line 13: I have a bad feeling with introducing a local static variable, but that might just be me. Is there any other way to implement this? I am just asking and would like your opinion.

minacode (commented):
Please also fix the missing spaces in your code. You can run clang-format with the file found [here](https://docs.infinitime.io/en/latest/how-to-contribute.html?highlight=Format#tl-dr)

src/displayapp/screens/settings/QuickSettings.cpp
jspkay: I think it's not a problem. The compiler should simply allocate some space near the function itself. I am not 100% sure tough. We should ask some experienced programmer. 
Regarding the implementation, there surely are other ways to implement this. I can think of a global variable (although scoped) but I think it's just worse, otherwise it's possible to add another member to the class, but it would be unnecessary, as I see this.
In conclusion, I guess a static variable is the simplest and safest way of doing this, tbh

minacode: Ok, I agree with you.

Riksu9000:
I like the idea. I think the dragging behavior should be changed, however doing it this way would require fixing it in each screen individually. We're preparing to use our own fork of LVGL, so we can fix issues like this at the root of the problem. If you're interested, I encourage you to look into fixing this issue directly in LVGL here https://github.com/InfiniTimeOrg/lvgl

PR #1438, opened 2022-11-13
Readable notification title

jspkay:
This pr proposes to maintain the title of a notification preview fixed for the first second, then start scrolling. This was done because long title directly starts to scroll and that usually results in an unreadable notification

### DIFF
--- a/src/displayapp/screens/Notifications.cpp
+++ b/src/displayapp/screens/Notifications.cpp
@@ -79,6 +79,12 @@ void Notifications::Refresh() {
       timeoutLinePoints[1].x = pos;
       lv_line_set_points(timeoutLine, timeoutLinePoints, 2);
     }
+
+    if(!this->scrolling && tick >= timeoutTickCountStart + timeoutStartScrolling){
+        this->scrolling = true;
+        lv_label_set_long_mode(currentItem->alert_type, LV_LABEL_LONG_SROLL_CIRC);
+    }
+
   }
 
   if (dismissingNotification) {
@@ -127,13 +133,18 @@ void Notifications::OnPreviewInteraction() {
 }
 
 bool Notifications::OnTouchEvent(Pinetime::Applications::TouchEvents event) {
-  if (mode != Modes::Normal) {
+  /*if (mode != Modes::Normal) {
     if (!interacted && event == TouchEvents::Tap) {
       interacted = true;
       OnPreviewInteraction();
       return true;
     }
     return false;
+  }*/
+
+  if(mode == Modes::Preview){
+    interacted=true;
+    OnPreviewInteraction();
   }
 
   switch (event) {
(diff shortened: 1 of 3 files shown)


src/displayapp/screens/Notifications.h
```diff
@@ -57,6 +58,8 @@ namespace Pinetime {
           lv_obj_t* label_accept;
           lv_obj_t* label_mute;
           lv_obj_t* label_reject;
+          lv_obj_t *alert_type;
```
minacode, line 61: I think it would be good to align the `*` to the type like the other lines do.

src/displayapp/screens/Notifications.cpp
```diff
@@ -79,6 +79,12 @@ void Notifications::Refresh() {
       timeoutLinePoints[1].x = pos;
       lv_line_set_points(timeoutLine, timeoutLinePoints, 2);
     }
+
+    if(!this->isTitleScrolling && tick >= timeoutTickCountStart + timeoutStartTitleScrolling){
+    	currentItem->StartTitleScroll();
+        this->isTitleScrolling = true;
```
minacode, line 85: The structure is not clear to me. Line 84 is hidden behind a function call, but the bool is set outside of that function. I would dismiss the function and just run the `lv_...` from line 369 here directly.

minacode (changes requested):
This is a nice little feature that should improve the UX.
I found two little things from scrolling through.

src/displayapp/screens/Notifications.h
jspkay: Sure thing! I didn't notice... In this case I just do another commit and the pr automatically gets the last version?

src/displayapp/screens/Notifications.cpp
jspkay: That was the first that I did, but I realized that I needed to put `alert_type` as a public member of `notificationItem` so I thought it was better to use a function. The boolean prevents multiple calls to the same function.
As far as I understand, though, you'd prefer the first alternative, am I right?

src/displayapp/screens/Notifications.h
minacode: Yes, exactly.

src/displayapp/screens/Notifications.cpp
minacode: I didn't notice that they are different objects, sorry. That makes more sense. Then I have a new question: Do you think it would be good to reset the Boolean when `currentItem` gets changed, like in line 39?

jspkay: Yeah, you are right... I didn't test a second long notification. Actually, we should check what happens when a new notification arrives (because I already forgot 🤣) so yeah, I'll make a commit in the near future to solve this. Thank you! :)

minacode: Good catch! I think there is a separate notification preview screen that you need to change too.

jspkay:
To completely generalized the concept (so that each notification will behave in the same way) I modified only NotificationItem using a lvgl task. I noticed that each time a notification is displayed, a new NotificationItem object is created, so I added a refresh task in the constructor which is called after 1s and destroyed aftewerads. Otherwise, is destroyed with the object, to avoid segmentation faults. This way, each notification will have a scrollable title after 1s. 
I want to add some settings in the feature, to modify the notification vibration intensity, the initial scroll time and few little things like this.

PR #1441, opened 2022-11-17
Change settings to use CBOR (attempt 2)

Arsen6331:
This PR is mostly the same as my previous PR (#1070). I've added the new settings that have been added since then as well as updated constraints on old settings to reflect new options.

I've also managed to narrow down the crashing issue I was seeing earlier to the load function. If I revert `Settings::LoadSettingsFromFile()`, the settings are successfully migrated to CBOR and the contents are as expected. Otherwise, InfiniTime crashes on startup. Based on my debugging, the function does complete successfully, and the crash occurs some time after that. To corroborate this, I see half of the watch face drawn on the screen before the crash, and the watch face is correct (PineTimeStyle) with the correct settings (Green sidebar, White text).

My theory is that some value may be getting read or converted incorrectly, which is causing whatever uses that value to crash. I'm not sure where this may be happening, and my unsealed PineTime is broken, so I can't use a debugger, but that didn't really help much last time, so I don't think it would be too useful anyway.

### DIFF
--- a/src/components/settings/Settings.cpp
+++ b/src/components/settings/Settings.cpp
@@ -1,6 +1,11 @@
 #include "components/settings/Settings.h"
+#include "libs/QCBOR/inc/qcbor/qcbor.h"
+#include <qcbor/qcbor_spiffy_decode.h>
+#include <qcbor/qcbor_encode.h>
+#include <qcbor/UsefulBuf.h>
 #include <cstdlib>
 #include <cstring>
+#include <vector>
 
 using namespace Pinetime::Controllers;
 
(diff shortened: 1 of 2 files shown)


Arsen6331:
I have considered that the watch could be running out of memory, but this shouldn't be using more than 500B or so, and I believe there is more memory available than that.

Arsen6331:
@Avamander @NeroBurner I'd appreciate it if you could help me debug this. I haven't been able to get very far with figuring out what's actually causing the problem.

jspkay:
I'd be happy to help, but I need some information... Can we have chat somewhere (like telegram or discord)?

Arsen6331:
> I'd be happy to help, but I need some information... Can we have chat somewhere (like telegram or discord)?

I prefer matrix (`@heisenbug:matrix.arsenm.dev`), but I am also on Discord as `Arsen6331#0230`

Avamander:
> @Avamander @NeroBurner I'd appreciate it if you could help me debug this. I haven't been able to get very far with figuring out what's actually causing the problem.

You can try the approach I used in Weather, it seems fairly stable. I wrote it very defensively to compensate for the dynamic parts that aren't so safe.

I'd also try and allocate both all final values and temporaries statically. Pseudocode example:
```
class Something {
    uint8_t final = 0;
    uint8_t final2 = 0;
    uint8_t tmp = 0;
    
    decode(); 
}

Something::decode() {
    QCBOR_Init();
    self.tmp = QCBOR_Decode_Int("Value"); // Check success
    self.final = self.tmp; // Check sanity here
    self.tmp = QCBOR_Decode_int("Value2"); // Check success
    self.final2 = self.tmp; // Check sanity here
    
    QCBOR_Finalize(); // Recheck for errors, close all decoders
}
```

NeroBurner:
wouldn't a local tmp variable make more sense? why add a tmp variable to the data struct?

Avamander:
> wouldn't a local tmp variable make more sense? why add a tmp variable to the data struct?

Not the data struct, the controller etc. holding all the data, that way it gets allocated along with the rest.

PR #1450, opened 2022-11-22
Step goal reset button

Dheenhasty:
Another idea to complete the pull request #1429 and the issue #1427 : 

So changing increment by 500. adding a reset button to go back to 10 000

Behavior of the increment changed by Click who permit de click faster if needed. 
Reset button will operate only on long pressed to avoid bad manipulation.

### DIFF
--- a/src/displayapp/screens/settings/SettingSteps.cpp
+++ b/src/displayapp/screens/settings/SettingSteps.cpp
@@ -2,6 +2,7 @@
 #include <lvgl/lvgl.h>
 #include "displayapp/DisplayApp.h"
 #include "displayapp/screens/Symbols.h"
+#include <string> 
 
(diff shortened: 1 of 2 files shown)


Riksu9000:
What issue is adding the reset button addressing?

Dheenhasty:
Honestly, just to simplify the test i have done (set the step to 45500 ..... and go back to 10 000 :) ) and play with lvgl a little bit.

but i can imagine that for example you usualy use 7500 step daily, but on some exercice you want to do a long walk with really high step ( 30 or 40 000 step) ....

the perfect idea would be to be able to set the step thru companion app.

Riksu9000:
#1575 changes the step size to 500 and allows changing the value faster, so a reset button shouldn't be necessary.

PR #1451, opened 2022-11-23
BLE CTS: fixed signedness of timezone

cybuzuma:
This is a hotfix for  #1161.

I am afraid I got misled by the BLE spec, which specified the timezone fields as `uint8`. But, to be able to express negative timezones, they need to be signed.

### DIFF
--- a/src/components/ble/CurrentTimeService.h
+++ b/src/components/ble/CurrentTimeService.h
@@ -47,8 +47,8 @@ namespace Pinetime {
       } CtsCurrentTimeData;
 
       typedef struct __attribute__((packed)) {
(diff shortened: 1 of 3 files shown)


Avamander:
> e uint8 1 This field represents the offset from UTC in number of 15-minute increments.
> Valid range from -48 to +56.

Lol, that is kinda funny. You could report this to Bluetooth SIG.

Avamander (approved):

JF002 (approved):

JF002:
I agree, this is very confusing. Thanks for the patch!

Issue #1168, opened 2022-06-04
World Clock


calhix:
### Verification

- [X] I searched for similar feature request and found none was relevant.

### Pitch us your idea!

An app to to be able to select multiple time zones to track

### Description

I think an app for a world clock would be really useful. I have family that live in different countries that I contact on a regular basis and being able to see their times from the watch would be really useful. Perhaps you could have a list of major cities in each timezone or the UTC times you could pick from to track.

medeyko:
I missed the multiple timezones display a lot, but I guess that there are too few memory left on the device to include this feature...

callingmybluff:
Can we then please close tickets that can not be added?

FintasticMan:
They will be able to be added at some point, when it's possible to load apps from the external flash. There is some really good progress on that from JF, who's got images and fonts to load at a pretty good speed.

calhix:
Good work so far, something I was thinking of is akin to the style of the world clock on the apple watch. It would be nice for the first screen in the app to be the list of cities you have selected to track (similar to the apple watch image from the link below). You would then maybe swipe left to a second screen which would have a list of major cities within each time zone (perhaps this would be a good start but it's missing quite a few cities imo: https://i.redd.it/imqp9i619a421.png) with a box to check beside each (just like the Wake Up app).
Perhaps optimally we would have a world clock integrated with gadgetbridge, where you could select the cities you want to track from there to then display within the world clock app on the pinetime.

Here is the icon we can use for it: https://fontawesome.com/icons/globe?s=solid
Apple watch: https://help.apple.com/assets/620FD4265AC6AE3F090A11D6/620FD42A5AC6AE3F090A11EE/en_US/ed82705883009a57983264270a0b94d0.png

PR #1454, opened 2022-11-24
Added World Clock Service

cybuzuma:
## Motivation

#365 asked for the watch to display multiple timezones.

### BLE

* Added custom World Clock Characteristic to BLE CTS Service

### `DateTimeController`

* Added store for four UTC offsets and 8 character labels for additional timezones
* Added accessors for the timezone slots

### Watch Faces

* Added a world clock watch face as a demo of the functionality

worldClock

### Settings

* Added world clock to watch face selection

setting

## Tests

* Tested against a patched version of Gadgetbridge (PR forthcoming)
Gadgetbridge

* Used my sealed PineTime without any observed problems

## Notes

I added the characteristic into the Current Time Service to not need an additional service and think timezones are related enough to current time to justify this non-standard extension. I created a new UUID in the style of the other custom UUIDs. I kept the characteristic definition very simple by requiring to always set all four slots. This makes handling of updating and resetting slots easier on the watch and moves the work into the more powerful companion.

The world clock watch face is meant mostly as a demo, I am by no means a UI developer but wanted to ship a way to check this new feature out. Adding and removing world clocks is displayed immediately, while updating the label of a clock will only update on every full minute. This is to lower the amount of redrawing the display.

So far, I only implemented code for Gadgetbridge (which has this function readily available in its framework). I am afraid I have not much experience in UI design neither on PC/Mac nor on iOS, so probably someone else needs to take up the work in the other companion apps.


calhix:
Great work! I didn't know there was a similar issue compared to the one I opened (#1168, though slightly different since I was thinking of an app rather than a watch face).

PR #1455, opened 2022-11-24
WeatherService daily min/max temperature fixes

kieranc:
This PR fixes an issue with the WeatherService GetToday[Min:Max]Temp functions which miscalculated the timestamp of the current day end. I also made a change which makes it find all events between day start & day end (instead of only non-expired events before day end) which as far as I can see is reasonable for this function (find highest and lowest values for the current 24h period). Testing will reveal how different weather data providers set expiry time and what that means for us...

### DIFF
--- a/src/components/ble/weather/WeatherService.cpp
+++ b/src/components/ble/weather/WeatherService.cpp
@@ -594,7 +594,7 @@ namespace Pinetime {
           }
         }
       }
-      
+
       return result;
     }
 
--- a/src/components/ble/weather/WeatherService.cpp
+++ b/src/components/ble/weather/WeatherService.cpp
@@ -554,8 +554,7 @@ namespace Pinetime {
       uint64_t currentTimestamp = GetCurrentUnixTimestamp();
       uint64_t currentDayEnd = currentTimestamp + ((24 - dateTimeController.Hours()) * 60 * 60) +
                                ((60 - dateTimeController.Minutes()) * 60) + (60 - dateTimeController.Seconds());
-      uint64_t currentDayStart = currentTimestamp - ((dateTimeController.Hours() * 60 * 60) +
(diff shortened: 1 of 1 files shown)


src/components/ble/weather/WeatherService.cpp
```diff
@@ -552,11 +552,12 @@ namespace Pinetime {
...
       uint64_t currentTimestamp = GetCurrentUnixTimestamp();
-      uint64_t currentDayEnd = currentTimestamp - ((24 - dateTimeController.Hours()) * 60 * 60) -
-                               ((60 - dateTimeController.Minutes()) * 60) - (60 - dateTimeController.Seconds());
+      uint64_t currentDayEnd = currentTimestamp + ((24 - dateTimeController.Hours()) * 60 * 60) +
+                               ((60 - dateTimeController.Minutes()) * 60) + (60 - dateTimeController.Seconds());
+      uint64_t currentDayStart = currentDayEnd - 86400;
       int16_t result = -32768;
       for (auto&& header : this->timeline) {
-        if (header->eventType == WeatherData::eventtype::Temperature && IsEventStillValid(header, currentTimestamp) &&
+        if (header->eventType == WeatherData::eventtype::Temperature && header->timestamp > currentDayStart &&
```
Avamander, line 560: It should either be `header->timestamp >= currentDayStart && header->timestamp < currentDayEnd` or `header->timestamp > currentDayStart && header->timestamp <= currentDayEnd` otherwise when `header->timestamp == currentDayStart` or `header->timestamp == nextDayStart` events will be ignored.

Avamander (changes requested):

src/components/ble/weather/WeatherService.cpp
kieranc: Thanks

Avamander (approved):

NeroBurner (approved):

PR #417, opened 2021-06-06
Initial work on adding internationalization.

MFAshby:
From https://github.com/JF002/InfiniTime/issues/363

This commit adds [lv_i18n](https://github.com/lvgl/lv_i18n) tooling for translating user-facing strings. It also adds a new settings page to allow selecting between up to 4 languages.

Done:
[X] Language selector for the settings screen (caveat, this can only support up to 4 languages at the moment)
[X] Replace hard-coded english text with lookups (done most text I can find) 
[X] Update main README for contributing translations

Known issue:
Some characters aren't rendering properly, for example `Batería` is rendering as `Batera` in the settings menu.

Known limitation:
The settings screen will only support up to 4 languages.

TODO:
[ ] Fix rendering issues for characters outside the usual ASCII
[ ] Calculate firmware size increase due to translations
[ ] If unacceptably large, either: 
1. build multiple firmwares, one for each language, and remove the user selection at runtime. 
2. wait until external flash memory can be used, and store strings files in it https://github.com/JF002/InfiniTime/issues/321

Issue #363, opened 2021-05-17
Translation (i18n) support


christianfl:
As InfiniTime is only translated in English so far I thought it would be nice to get i18n support. This would improve UX and could make it easier to get new users in the future.

I found this but not sure wether this is suitable: https://github.com/lvgl/lv_i18n

Is it somehow realistic to use this or any other i18n support possibility keeping the limited resources of the PineTime in mind? Selecting languages could either be automatically by the Companian app (phone system language) or by a "Language" pane in the PineTime. I have no real C++ experience but after having some working code example I could implement the strings/translations and could provide English and German translations.

P.S. Thanks for InfiniTime to all contributors! Just got two sealed PineTimes for a friend and myself today and looking forward to wether or how I can contribute in the future.

Raupinger:
Lol, I thought of this yesterday and wanted to open an issue once I get to it. Anyways, here are my ideas and opinions:
I thought about this too and already looked at the lvgl implementation as well. I would prefer to have the language set at compile time to save space. My idea was to have a Header file that defines every string that needs to be localized and you just change the language by giving the compiler a different file. This approach would also fit nicely with needing to create different builds support umlaute or accents. It does have the drawback of making writing code a bit harder since you can't just put strings in there anymore. With the lvgl method you just call a method with the string and it handles everything. Ideally we would have that but with a compiler directive instead of a runtime function.

christianfl:
Hehe, nice coincidence 😃

I can totally understand the save spacing argument. However, couldn't we somehow get the best of both worlds? Just compiling with different flags means there would be independent builds for each and every language? This would sometime generate a long list of new builds and a user would have to flash it to change it...

I thought maybe it could be possible to use lv_i18n or some other gettext implementation and just push and apply the user language file via Companian app? Don't know how this could be done though. Using JSON or YAML for each translation instead of directly using C/header files has the advantage that its easier to translate by non-devs and these files are greatly supported by translation services such as Weblate or Transifex (if that's an option in the future). They get compiled to C anyway.

MFAshby:
I've made a start on this using lv_i18n for now. This was very easy to integrate into the InfiniTime build!

It should be possible to roughly calculate the application size increase from adding this additional translations. If it's too big, then it looks possible to tweak the build system to include only 1 set of translations in a firmware. I think this means we can use this existing tool without having to invent something new.

[...]

PR #1458, opened 2022-11-25
Internationalization and Localization

cybuzuma:
## Motivation

After #417 was closed, #363 is still open. I thought I would give it a try and see how much memory really is used for translation. As shown below, it is not that much so that I propose this PR.

### lv_i18n Markup

* Wrapped occurrences of strings with `_()` macro.
* Where this was not possible, I wrapped the usage of the strings with `_()` and added an `#if 0` block with the actual strings for the extractor to pick them up. See the doc or the code for examples.

### Languages

* Added English translation. This is basically a non-op. lv_i18n uses the passed string as fallback if no translation is found. As I keept the strings as is, there is no need for an explicit English translation, saving a little bit of space.
* Added German translation, the only other language I feel confident enough to provide a translation.

### BLE

* Added the Language characteristic to the Device Information Service. This characteristic is specified by BLE, but not used any more. As I was too lazy to create a new service for it, I added it to Device Information, which I feel is reasonable. With this characteristic, companion apps can change the language of the watch.

### Flash footprint

* Compared to develop, this PR with one additional language has approx. 2KB or 0.5% more flash usage. I think this is quite reasonable for this kind of feature.

### Documentation

* Wrote a short HowTo

## Notes

I checked most of the german translation, but might have missed some broken ones. Still, I want to get this PR out for people to interact with it.

I patched Gadgedbridge to set the language of the watch in this [PR](https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/2986).


JF002:
Thanks @cybuzuma ! I appreciate that you implemented your idea and provide actual numbers regarding the memory usage!

Regarding memory usage, how will it scale with additional languages? Also, I guess it works only with "standard ASCII"  languages ?

Also, in the PR, I noticed that the translation seems to be duplicated : once in `lv_i18n.c` and once in de.yml. Is this intentional?

cybuzuma:
I guess the scaling depends a bit on the language, but as the screen space is limited (which means additional languages simply can not fit much more characters), I would suppose that it scales quite linear with 1kB per language. A good guess is to take the size of the yml file minus 1.7KB for the English keys which are only required once as the compiler optimizes them away as far as I can tell. This gives 1KB for the German one I did.

Yes, InfiniTime is ASCII only at the moment. I also noted this in the documentation.
As far as I saw, lvgl supports utf-8, but I have not looked into what that would take. The biggest thing there will probably be the additional glyphs in the fonts.

`lv_i18n.c` is generated from all the `language.yml` files, that's why they duplicate. I think we could move generating it into the build but as I said I wanted to get this out first as soon as possible. I can look into that in a future PR.

cybuzuma:
re failed builds:

build-simulator:
InfiniSim fails as it has its own source listings which do not contain `lv_i18n.c`. I think it would be easiest to first merge this PR and then fix InfiniSim when updating InfiniSim's InfiniTime submodule. Otherwise it will be a mess with the submodule versions.

build-firmware:
I put the translation source file into the lvgl source array which is not used by all targets, whereas some sources used by these targets are i18ned. Now, I added `lv_i18n` to the main and recovery sources. I did not add it to the recovery loader to keep it small and stable. For this, I had to remove i18n from `BrightnessController` which is used by the recovery loader. I added a comment to remind further investigators of it.

cybuzuma:
rebased onto develop and squashed a bit for cleaner history.

cybuzuma:
I looked into Unicode a bit more with the following results:

lvgl is already ready for Unicode/UTF-8, the only missing part is to enable additional Glyphs in `src/displayapp/fonts/fonts.json`.

At the moment, InfiniTime has ASCII as well as Cyrillic Glyphs enabled: https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/src/displayapp/fonts/fonts.json#L6

I looked into the space requirements of enabling additional Unicode ranges (cf [here](https://jrgraphix.net/research/unicode.php)) with the following results:
* **Latin-1**: additional 2.4 KiB
* **Greek and Coptic**: additional 2 KiB
* **Cyrillic**: additional 2.9 KiB

As the used font, `JetBrainsMono-Bold.ttf` is not very complete in its Unicode support, I replaced it with [Noto Sans Bold](https://en.wikipedia.org/wiki/Noto_fonts). With this font, I get the following results:
* **Latin-1**: additional 3KiB
* **Latin-Extended A**: additional 4.2KiB
* **Latin-Extended B**: additional 7.3KiB
* **Greek and Coptic**: additional 5.2KiB
* **Cyrillic**: additional 11.6KiB

Infinitime 1.11.0 has enough space left to enable all of the above, then weighting in at 440 KiB or 91.8% full:

i18n

I have not found a nice overview which Unicode ranges are needed for certain languages, but I think **Latin-1** should allow German, Spanish and French.

In Summary, I'd like to propose to extend this PR:
* Enable Latin1, disable Cyrillic. Further Ranges can be enabled if translations are rolling in (I would add a note about this in the documentation)
* Switch to Noto, to be future proof regarding Unicode. It is licensed under a free license. Still, I do not know how the process on choosing JetBrainsMono was and how we could decide this. For now, JetBrainsMono supporst almost all of Latin1.

Boteium:
I would suggest we build one firmware per language. I maintain a Traditional Chinese fork and my firmware has already crammed with tons of CJK glyph. 5kb is about 100 CJK glyph, which is quite a lot.

Actually, I've remove lots of apps and watchface just to make room for my pending PRs and stuff such as calculator and calendar in my personal build. I think we would hit the size limit if we are to merge apps such as Gallery, Calculator, Calendar, even for a English-only build.

The size does scale as predicted. ASCII is around 1kb, Greek/Cyrillic 2kb-ish and CJK will be 3kb just like how utf-8 codepoint is encoded.

Boteium:
utf-8 definitely works.

P_20221129_192352

PR #1459, opened 2022-11-25
PineTimeStyle weather display

kieranc:
Here is an initial draft of weather support for PineTimeStyle. It has a new button to toggle the weather display, and once enabled it displays the current temperature and an icon which can show sun, cloud, sun & cloud, sun cloud & rain or cloud & rain.

The initial goal was to have daily minimum and maximum temperatures displayed on the watchface plus the icon, but while the min/max functions work, more investigation needs to be done about how companion apps provide weather data especially with relation to expiry times. If weather is sent to the watch with a 1h expiry time, it will be purged from the timeline after an hour and then can't be included in daily min/max temperature. I've tested using ITD which works well but currently sets the expiry time to 1 hour, and TinyWeatherForecastGermany which sets the expiry time to 24h. Using TWFG and GadgetBridge currently results in temperatures being displayed in Kelvin, this will be fixed in the next GB release. The temperature is displayed simply in degrees (not °C/°F) partly for space but also so we can display farenheit or celsius depending on what the companion app supplies without any settings required on the watch. Gadgetbridge has a setting to choose metric or imperial units which would allow it to send farenheit if desired, but this will require another small patch to GB to make it work.

I'm opening this as a draft to get some input, one bug I haven't dealt with yet is that on boot, with no weather data available, the temperature shows 124°, precipitation 24 and clouds 155%, or some combination like that. Not critical, but I'd like to have something non numerical displayed ideally if there's no weather data available. I'm also getting an overload warning on compilation so I might have missed something in displayapp. It's not totally clear if precipitation represents a percentage chance or a quantity in mm/hr, it might depend on the weather provider. For me it seems like MM.

The calculation I've used to set the icon seems crude and can probably be improved:

```
    if ((clouds <= 30) && (precip == 0)) {
      lv_label_set_text(weatherIcon, Symbols::sun);
    } else if ((clouds >= 70) && (clouds <= 90) && (precip == 1)) {
      lv_label_set_text(weatherIcon, Symbols::cloudSunRain);
    } else if ((clouds > 90) && (precip == 0)) {
      lv_label_set_text(weatherIcon, Symbols::cloud);
    } else if ((clouds > 70) && (precip >= 2)) {
      lv_label_set_text(weatherIcon, Symbols::cloudShowersHeavy);
    } else {
      lv_label_set_text(weatherIcon, Symbols::cloudSun);
    };
```
InfiniSim_2022-11-25_194741

The BLE icon has been moved next to the battery icon, the notification icon has been moved to the top left, and the calendar shifts down when weather is enabled to make space.

It could maybe do with moon icon for nighttime?


minacode:
In your code snippet you used something like `precip = 0` a few times, but I guess you mean `precip == 0`.
The first version sets `precip` to `0` and then evaluates it to false, because it is zero.
The second version checks `precip` and `0` to be equal.

ItzSwirlz:
Can you send a screenshot of what it would do if no weather information was available?

kieranc:
> Can you send a screenshot of what it would do if no weather information was available? Would the spacing/padding for the weatherIcons and whatnot still be shown?

Currently looks like this on boot with no weather data:

image

Clouds, precip and temperature should all be zero in this situation, but based on the icon precip is over 2, clouds is over 70 while temp is 0.

src/components/settings/Settings.h
```diff
@@ -38,12 +38,14 @@ namespace Pinetime {
         Pink
       };
       enum class PTSGaugeStyle : uint8_t { Full, Half, Numeric };
+      enum class PTSWeather : uint8_t { On, Off };
```
minacode, line 41: Is there a third possible value in the near future or could it just be a `bool`?

kieranc: it could be a bool currently, I'm not sure if/when more options will be available but since this only a first approach I do expect it to change in the future. What's the cost/benefit of bool vs enum?

minacode: `bool` should have the same size as `uint8_t` do. That's not the problem. But using `bool` would simplify the `...::Get() == ...::On` comparisons.

kieranc: I think I'd rather leave it as-is if there's no performance impact, for consistency with the other settings and future options.

src/displayapp/screens/Clock.cpp
```diff
@@ -34,6 +35,7 @@ Clock::Clock(DisplayApp* app,
     settingsController {settingsController},
     heartRateController {heartRateController},
     motionController {motionController},
+    weatherService(weather),
```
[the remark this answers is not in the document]
kieranc: Thanks, should be resolved.

JF002 (commented):
I've just tested this PR with Gadgetbridge and Tiny Weather Forecast Germany and it works great... once I figured that I need to enable the weather display in the settings of PTS :)

I think it would be great to add a bit of documentation about the weather feature : how to enable it on the companion app side (mostly link to https://codeberg.org/Freeyourgadget/Gadgetbridge/wiki/Weather#user-content-install-and-configure-tiny-weather-forecast-germany) and how to enable it in PTS. Maybe also mention the minimum version of gadgetbridge required (the one that sends °C/°F instead of Kelvin).
This doc could be added in the "New to InfiniTime?" section : https://github.com/InfiniTimeOrg/InfiniTime#new-to-infinitime.

kieranc:
> I've just tested this PR with Gadgetbridge and Tiny Weather Forecast Germany and it works great... once I figured that I need to enable the weather display in the settings of PTS :)

I think there's an as yet unidentified bug which means there's a difference between the temperature displayed on the watch and the current temperature displayed in TWFG, I haven't managed to pin it down yet. It's only a couple of degrees, and confirmation of what I'm seeing would be helpful, currently TWFG says 13° and the watch says 12°, but 12 would be correct for any time in the next 4 hours. It could be a timestamp issue since TWFG supplies data for the next several hours, but I tried setting my timezone to UTC and there was still a difference. Temperature data supplied by itd works as expected, but it only provides a single data point.
 
> I think it would be great to add a bit of documentation about the weather feature : how to enable it on the companion app side (mostly link to https://codeberg.org/Freeyourgadget/Gadgetbridge/wiki/Weather#user-content-install-and-configure-tiny-weather-forecast-germany) and how to enable it in PTS. Maybe also mention the minimum version of gadgetbridge required (the one that sends °C/°F instead of Kelvin). This doc could be added in the "New to InfiniTime?" section : https://github.com/InfiniTimeOrg/InfiniTime#new-to-infinitime.

I can absolutely take care of that but I have just started a new job so it might be next week :-)

JF002:
> I think there's an as yet unidentified bug which means there's a difference between the temperature displayed on the watch and the current temperature displayed in TWFG, I haven't managed to pin it down yet. It's only a couple of degrees, and confirmation of what I'm seeing would be helpful.

I tried with multiple city, and it seems like the temperature displayed in PTS is always the one circled in red in the next screenshot:

pt-weather

I guess those 3 values are min/max/current temperature?

> I can absolutely take care of that but I have just started a new job so it might be next week :-)

No worries, take your time :)

kieranc:
> I tried with multiple city, and it seems like the temperature displayed in PTS is always the one circled in red in the next screenshot:
> 
> I guess those 3 values are min/max/current temperature?

Looks that way, that corresponds with what I'm seeing too. Not horribly broken behavior, but not exactly perfect either. I need to do more debugging, it's raining outside and TWFG says moderate rain, but my icon shows cloud/sun.

kieranc:
Updated to check for valid data before trying to display it, which means on boot the watch shows a ban icon 🛇 and -- in place of the temperature. When weather data is available it is immediately displayed. 
I still seem to have a small offset compared to temperature as displayed in TWFG and it's no longer showing the hourly min temp, I have 19 cur/20 max/17 min on my phone and 21 showing on the watch. I still kind feel like it's a timestamp issue (although 21 degrees would need it to be 6 hours out which is excessive), but I'm running a build with debug labels enabled so I'll keep an eye on it.

JF002:
I also noticed inconsistencies between the data displayed in TWFG and the data displayed in InfiniTime. 

For example : TWFG displays temperature (min/max) 26 / 27 in Madrid right now, but InfiniTime displays 19°C. TWFG forecasts 19° in 6 hours.

I added logs in WeatherService : InfiniTime receives 2 temperature events. In this example, it received `2685` and `1985`. Since 1985 will be divided (integer division) by 100, the resulting value will be 19.

If I had to make a guess, I would say that Gadgetbridge sends MAX and MIN values for the next 6 hours. Since InfiniTime only handles 1 temperature, it takes the last one which is the minimum one. 

I had a look at the code : GB effectively sends 2 temperature info : 

 - "Current temperature" : https://codeberg.org/Freeyourgadget/Gadgetbridge/src/branch/master/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/pinetime/PineTimeJFSupport.java#L690
 - "24h temperature forecast" : https://codeberg.org/Freeyourgadget/Gadgetbridge/src/branch/master/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/pinetime/PineTimeJFSupport.java#L718 . Strangely enough, the 2nd temperature sent by GB does not seem to be the one that is forecast in 24h so... I'm not sure.

They differ by the "Expires" field.

However, InfiniTime does not use that `expires` field to store multiple values for the same field so only the last one is taken into account.

So it looks like there're some inconsistencies/incompatibilities between the implementations of the weather service in InfiniTime and the weather client in GB. We'll probably have to collaborate with the developers of Gadgetbridge to fix those issues.
@Avamander any thoughts about this? Do we have any documentation on how the Weather API is supposed to be implemented and used?

I also tried using ITD to check if it produces the same results but the command `./itctl update weather` just seem to do nothing. 
@kieranc have you already tried with ITD? 

All in all, if we are confident that InfiniTime and PTS displays the data that is being sent, I'm OK to merge this PR and fix those issues with GB in another PR.

Avamander:
> Since InfiniTime only handles 1 temperature, it takes the last one which is the minimum one.

It is handled, `GetTodayMinTemp`/`GetTodayMaxTemp` are separate functions in `WeatherService`.

> However, InfiniTime does not use that expires field to store multiple values for the same field so only the last one is taken into account.

Ideally the one with lower `expires` field is or should be used when asking for the current temperature, as short-term forecast is more accurate than a long-term one. It doesn't/shouldn't matter when asking for the daily extremes. `WeatherService`'s `GetCurrentTemperature` should probably get this improvement at some point.

> So it looks like there're some inconsistencies/incompatibilities between the implementations of the weather service in InfiniTime and the weather client in GB.

I think the primary inconsistency stems from the lack of true expiration information from Gadgetbridge's weather sources. However I don't think it's the cause of the current issues.

kieranc:
> > Since InfiniTime only handles 1 temperature, it takes the last one which is the minimum one.
> 
> It is handled, `GetTodayMinTemp`/`GetTodayMaxTemp` are separate functions in `WeatherService`.
If I've understood correctly these functions provide the minimum and maximum temperatures for a 24h period, but what we're talking about here is the ability to have more than one temperature value stored for a 1h time period which is currently not supported. Since the companion apps provide 3 temperature values per period/timeline entry (current, min, max) we should probably add the ability to handle them, or discard them and only accept whichever is provided as the current temperature. I've tested using quickweather using openweathermap data and it also doesn't seem to display the correct values but I don't have much debugging access to investigate further.

Itd works as expected but I think it only provides one temperature value and one period/entry so that makes sense.

Avamander:
> If I've understood correctly these functions provide the minimum and maximum temperatures for a 24h period, but what we're talking about here is the ability to have more than one temperature value stored for a 1h time period which is currently not supported

Yeah, no, there's no such limitation. There's just not a premade function to quickly return those values. Until it is added the easiest approach is to just search the timeline "manually"

JF002:
> I think the primary inconsistency stems from the lack of true expiration information from Gadgetbridge's weather sources.  

Gadgetbridge specifies an expiration info, but yeah, it's a hard-coded value.

> WeatherService's GetCurrentTemperature should probably get this improvement at some point.

If I understand correctly, there's no issue with the data coming from GB, but `WeatherService::GetCurrentTemperature()` should be improved to return the value with the lower expiration date?

JF002:
@kieranc Before I forget : it would be nice to add a bit of documentation about the weather functionality (which companion apps and weather providers are supported, how to enable it, maybe link to the companion apps documentation,...). Is this something you could do?

kieranc:
> @kieranc Before I forget : it would be nice to add a bit of documentation about the weather functionality (which companion apps and weather providers are supported, how to enable it, maybe link to the companion apps documentation,...). Is this something you could do?

Absolutely, I'll try get it crafted over the next couple of days.

Avamander:
> Gadgetbridge specifies an expiration info, but yeah, it's a hard-coded value.

Because its sources don't/can't provide that info.

> If I understand correctly, there's no issue with the data coming from GB, but `WeatherService::GetCurrentTemperature()` should be improved to return the value with the lower expiration date?

Not lower expiration date (the soonest to expire from now), the one that expires the quickest (the soonest to expire in general). A daily forecast going to expire at the end of the day would be the least correct, an hourly forecast going to expire in an hour (maybe after the daily one) would be the most correct.

JF002 (approved):

JF002:
Links to documentation by @kieranc on the wiki : [PTS](https://wiki.pine64.org/wiki/PineTimeStyle) and the [weather integration](https://wiki.pine64.org/wiki/Infinitime-Weather).

PR #1460, opened 2022-11-25
Fixing build-firmware not returning error at the correct step

### LINKED ISSUE #1292: Firmware build CI fails at the wrong step
### Verification

### What happened?

If the build fails, the check fails during the unzipping step.

The CI should fail at the build step

### Reproduction steps

PRs that fail the build check pass the build step, but fail at unzipping.

### More details?

I believe this happened after switching to building with docker. I think it's because the script ignores the exit code generated by the compiler and exits successfully regardless. The check fails at unzipping, because the build didn't complete and didn't generate the files.

https://github.com/Riksu9000/InfiniTime/runs/7930084225?check_suite_focus=true Here is an example of the issue. The build step incorrectly shows it being successful.

### Version

develop

### Companion app

_No response_

cybuzuma:
If I see it correctly, I can not verify this fix, as the PR build does not use the build.sh from this PR. But from looking at it, this should fix #1292

### DIFF
--- a/.devcontainer/build.sh
+++ b/.devcontainer/build.sh
@@ -35,6 +35,8 @@ main() {
   if [ "$DISABLE_POSTBUILD" != "true" -a "$BUILD_RESULT" == 0 ]; then
     source "$BUILD_DIR/post_build.sh"
   fi
+  # assuming post_build.sh will never fail on a successful build
+  return $BUILD_RESULT
 }
 
 GetGcc() {
@@ -74,4 +76,12 @@ CmakeBuild() {
   fi
 }
 
-[[ $SOURCED == "false" ]] && main "$@" || echo "Sourced!"
+if [[ $SOURCED == "false" ]]; then
+  # It is important to return exit code of main
+  # To be future-proof, this is handled explicitely
+  main "$@"
+  BUILD_RESULT=$?
+  exit $BUILD_RESULT
+else
+  echo "Sourced!"
+fi
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -41,6 +41,8 @@ main() {
   if [ "$DISABLE_POSTBUILD" != "true" -a "$BUILD_RESULT" == 0 ]; then
     source "$BUILD_DIR/post_build.sh"
   fi
+  # assuming post_build.sh will never fail on a successful build
+  return $BUILD_RESULT
 }
 
 GetGcc() {
@@ -77,4 +79,12 @@ CmakeBuild() {
   fi
 }
 
-[[ $SOURCED == "false" ]] && main "$@" || echo "Sourced!"
+if [[ $SOURCED == "false" ]]; then
+  # It is important to return exit code of main
+  # To be future-proof, this is handled explicitely
+  main "$@"
+  BUILD_RESULT=$?
+  exit $BUILD_RESULT
+else
+  echo "Sourced!"
+fi


.devcontainer/build.sh
```diff
@@ -74,4 +76,4 @@ CmakeBuild() {
   fi
 }
 
-[[ $SOURCED == "false" ]] && main "$@" || echo "Sourced!"
\ No newline at end of file
+[[ $SOURCED == "false" ]] && main "$@" || echo "Sourced!"
```
NeroBurner, line 79: for the build `$SOURCED` will be "false". Then when `main "$@"` returns a non 0 exit status (build failed), then the `||` part will be triggered and instead of an error the `echo "Sourced!"` will be executed. At least that's how I read this statement

I think we should split this statement into two lines:

```suggestion
[[ $SOURCED == "false" ]] && main "$@"
[[ $SOURCED != "false" ]] && echo "Sourced!"
```

same change should be done for `docker/build.sh`

NeroBurner (changes requested):

.devcontainer/build.sh
FintasticMan: The original line is correct. For some reason the shell interprets `command && command || command` as an if-else. You can test this by running `true && echo true || echo false` and `false && echo true || echo false` in your terminal.

NeroBurner: The statement `true && false || echo false` prints `false` on the command line
```

Comment (NeroBurner) [reply .devcontainer/build.sh:79]: I think we should replace it with an explicit if/else if that is the intent of this line. Just to prevent confusion in the future

```suggestion
if [[ $SOURCED == "false" ]]; then
  main "$@"
else
  echo "Sourced!"
fi
```

cybuzuma: I'm witth @NeroBurner on this one. `true && false || echo false` prints `false` in my bash as well.
An explicit if else seems to be more clear in any way.

NeroBurner (approved):
LGTM now. Thanks for the fix

.devcontainer/build.sh
```diff
@@ -35,6 +35,8 @@ main() {
   if [ "$DISABLE_POSTBUILD" != "true" -a "$BUILD_RESULT" == 0 ]; then
     source "$BUILD_DIR/post_build.sh"
   fi
+  # assuming post_build.sh will never fail on a successful build
+  return $BUILD_RESULT
```
Riksu9000, line 39: I think this should `exit` instead of `return`. `exit` will stop running the script and return the correct value, but `return` will continue executing the script, so running any command after main would override the value. I would change this to avoid issues in the future.

docker/build.sh
```diff
@@ -41,6 +41,8 @@ main() {
   if [ "$DISABLE_POSTBUILD" != "true" -a "$BUILD_RESULT" == 0 ]; then
     source "$BUILD_DIR/post_build.sh"
   fi
+  # assuming post_build.sh will never fail on a successful build
+  return $BUILD_RESULT
```
Riksu9000, line 45: Same here.

Riksu9000 (changes requested):

.devcontainer/build.sh
NeroBurner: the script seems like it can be sourced as well. If someone would source it in another script an then call the `main` function it will terminate its shell as well. Which isn't behavior I'd expect from a function

Riksu9000: Ok. Then `exit` should be placed after the call to `main`.

Riksu9000: Explicit `exit` is very important to make the true intent of the code apparent and avoid bugs. Without any indication, it's very easy to break the workflow again, for example just by adding an echo after main for more verbose output. In fact, I think a comment should also be added unless the return value of main is explicitly used when exiting.

cybuzuma: Agreed, pushed an update with explicit exit code handling and a comment.

Riksu9000 (approved):

docker/build.sh
```diff
@@ -77,4 +79,12 @@ CmakeBuild() {
   fi
 }
 
-[[ $SOURCED == "false" ]] && main "$@" || echo "Sourced!"
+if [[ $SOURCED == "false" ]]; then
+  # It is important to return exit code of main
+  # To be future-proof, this is handled explicitely
+  main "$@"
+  BUILD_RESULT=$?
+  exit BUILD_RESULT
```
NeroBurner, line 87: I think we need to access the variable explicitly
```suggestion
  exit $BUILD_RESULT
```

.devcontainer/build.sh
```diff
@@ -74,4 +76,12 @@ CmakeBuild() {
...
 }
 
-[[ $SOURCED == "false" ]] && main "$@" || echo "Sourced!"
\ No newline at end of file
+if [[ $SOURCED == "false" ]]; then
+  # It is important to return exit code of main
+  # To be future-proof, this is handled explicitely
+  main "$@"
+  BUILD_RESULT=$?
+  exit BUILD_RESULT
```
NeroBurner, line 84: I think we need to access the variable explicitly
```suggestion
  exit $BUILD_RESULT
```

docker/build.sh
cybuzuma: Argh, yes. I just can't bash...

NeroBurner (approved):

NeroBurner:
Thanks @cybuzuma for the fix and working through all the reviews

Issue #1440, opened 2022-11-14
Customize vibrations on hour and half hour (avoid nights/weekends)


famiclone6502:
### Verification

- [X] I searched for similar feature request and found none was relevant.

### Pitch us your idea!

Only chime on weekdays and during the day

### Description

Chimes help break my focus during work hours and check for meetings I need to attend. Can you add a menu option to start and stop these during custom hours? Also, a check box for whether or not they occur only weekdays (Monday through Friday)? Otherwise it will keep me awake at night, and I do not need them during the weekend. Right now I need to disable the chimes manually on my PineTime.

I've been using a Timex Data-Link with a custom Motorola 6805 application to perform chimes on a specific schedule. Right now it does hourly chimes during the day (9 AM to - 6 PM) and only during weekdays. I do appreciate the half-hour chimes that the PineTime does already, which I wanted on the Timex.

Ideally, customization could eventually go even further. There's a lot of history in different types of chimes clocks/watches have offered like different repeaters and sonneries. Personally, I would like an additional reminder a minute before the hour or half-hour. Customizing with a shorter vibration for that, plus maybe an audible chime on the hour would be helpful. Others may also only want specific days of the week. These are not crucial to me but would be nice to consider in the design of related features. 

This request builds upon issue #778, which added the option for vibration on the hour and half hour. Without this feature I would not have bought the PineTime, so thank you for fulfilling this request.

minacode:
> Otherwise it will keep me awake at night

Sleep mode disable chimes. Maybe that it worth a try for you.

famiclone6502:
> Sleep mode disable chimes. Maybe that it worth a try for you.

From what I understand, it would still be a manual process to enter sleep mode. I also still want other notifications outside of work hours (i.e. during the weekend), just not chimes.

Boteium:
Automatic sleep mode sound awesome. It can also save battery during night.

PR #1461, opened 2022-11-26
add auto sleep

Boteium:
This PR add the option to automatically toggle on/off sleep mode during the night
This should address issue https://github.com/InfiniTimeOrg/InfiniTime/issues/1440

### DIFF
--- a/src/displayapp/screens/settings/SettingAutoSleep.cpp
+++ b/src/displayapp/screens/settings/SettingAutoSleep.cpp
@@ -1,0 +1,145 @@
+#include "displayapp/screens/settings/SettingAutoSleep.h"
+#include <lvgl/lvgl.h>
+#include "displayapp/DisplayApp.h"
+#include "displayapp/screens/Screen.h"
+#include "displayapp/screens/Symbols.h"
+#include "components/settings/Settings.h"
+
+using namespace Pinetime::Applications::Screens;
+using AutoSleepOption = Pinetime::Controllers::Settings::AutoSleepOption;
+
+namespace {
+  void event_handler(lv_obj_t* obj, lv_event_t event) {
+    auto* screen = static_cast<SettingAutoSleep*>(obj->user_data);
+    screen->UpdateSelected(obj, event);
+  }
+}
+
+SettingAutoSleep::SettingAutoSleep(Pinetime::Applications::DisplayApp* app, Pinetime::Controllers::Settings& settingsController)
+  : Screen(app), settingsController {settingsController} {
+
+  lv_obj_t* title = lv_label_create(lv_scr_act(), nullptr);
+  lv_label_set_text_static(title, "Sleep Setting");
+  lv_label_set_align(title, LV_LABEL_ALIGN_CENTER);
+  lv_obj_align(title, lv_scr_act(), LV_ALIGN_IN_TOP_MID, 15, 15);
+
+  lv_obj_t* icon = lv_label_create(lv_scr_act(), nullptr);
+  lv_obj_set_style_local_text_color(icon, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_ORANGE);
+  lv_label_set_text_static(icon, Symbols::clock);
+  lv_label_set_align(icon, LV_LABEL_ALIGN_CENTER);
+  lv_obj_align(icon, title, LV_ALIGN_OUT_LEFT_MID, -10, 0);
+
+  lv_obj_t* container1 = lv_cont_create(lv_scr_act(), nullptr);
+  lv_obj_set_style_local_bg_opa(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, LV_OPA_TRANSP);
+  lv_obj_set_style_local_pad_all(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, 10);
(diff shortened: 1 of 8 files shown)


src/displayapp/screens/settings/SettingAutoSleep.cpp
```diff
@@ -0,0 +1,145 @@
...
+    auto* screen = static_cast<SettingAutoSleep*>(obj->user_data);
+    screen->UpdateSelected(obj, event);
+  }
+}
+
+SettingAutoSleep::SettingAutoSleep(Pinetime::Applications::DisplayApp* app, Pinetime::Controllers::Settings& settingsController)
+  : Screen(app), settingsController {settingsController} {
+
+  lv_obj_t* title = lv_label_create(lv_scr_act(), nullptr);
+  lv_label_set_text_static(title, "Sleep Setting");
```
Itai-Nelken, line 22:
```suggestion
  lv_label_set_text_static(title, "Auto Sleep");
```
I think it looks better when the same/extended description is used as in the settings entry.

src/displayapp/screens/settings/SettingAutoSleep.cpp
```diff
@@ -0,0 +1,145 @@
...
+  lv_obj_set_style_local_pad_inner(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, 5);
+  lv_obj_set_style_local_border_width(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, 0);
+
+  lv_obj_set_pos(container1, 10, 40); // orig: 10, 60
+  lv_obj_set_width(container1, LV_HOR_RES - 20);
+  lv_obj_set_height(container1, LV_VER_RES - 30); // res - 50
+  lv_cont_set_layout(container1, LV_LAYOUT_COLUMN_LEFT);
+
+  char optionName[2][11] = {
+    "Auto Sleep",
```
Itai-Nelken, line 44:
```suggestion
    "Start Sleep",
```
This makes more sense in my opinion.

Itai-Nelken (commented):
I have been using this for the last few days and it works very well though it is a bit annoying that it always returns to notifications on mode when stopping sleep. It would be better if it would return to the previous notification state in my opinion.

src/displayapp/screens/settings/SettingAutoSleep.cpp
Boteium: thanks. had to update array size though.

Boteium:
Restoring previous state should be something easy to implement.

We would have to modify "QuickSettings.cpp", because notification state can change after auto sleep is set.
we would also need to modify notification state variable.
Notification::Sleep should became an independent variable, or, a new variable for storing previous state should be introduced.

However, I have concern that too many changes might block this PR from being merged.
Do you think we should take one baby step at a time and make state-restoring a future PR ?

Itai-Nelken:
> However, I have concern that too many changes might block this PR from being merged.
> Do you think we should take one baby step at a time and make state-restoring a future PR ?

I think you're right. Baby steps are better.

FintasticMan:
I think that returning to notifications off if it was set to sleep before auto sleep kicked in is probably the best solution. I think it might also be good for sleep to turn off when the alarm rings.

escoand:
It would be nice to have even more checkboxes. Sleep Bluetooth comes to my mind.

Boteium:
I originally design it this way because I might manually enter sleep/off mode (go to bed early) but may forget to turn it on in the morning. That's why there are two checkboxs and why "stop sleep" is set to "on" instead of previous state. The UI also make sure users cannot set start/stop to the same time. (only one checkbox will be active in this case)

I agree "sleep mode" should be something simple and easy to use. I also like the idea of even fine grained customization. I do have auto brightness on my build so I can certainly see the reason of auto sleep bluetooth. But meragable PR is my focus for now. Maybe we can keep the default UI simple but let user set the advanced options on the companion app when PR 1441 is ready.

RageGamerBoi:
Is it in 24h time or 12h?
(i.e. in your example, is it 1:00 PM to 8:30 AM or something else?)

Boteium:
It's 24H.

RageGamerBoi:
Maybe reducing it to 30 minute increments would allow localization indicators like time zone to fit…
Seems less confusing than 8h which seems like it would leave people going “8 hours from what???”

Boteium:
update UI to reduce confusion
InfiniSim_2023-06-22_224919

src/displayapp/screens/settings/Settings.h
```diff
@@ -41,6 +41,7 @@ namespace Pinetime {
           {Symbols::clock, "Date&Time", Apps::SettingSetDateTime},
           {Symbols::batteryHalf, "Battery", Apps::BatteryInfo},
           {Symbols::clock, "Chimes", Apps::SettingChimes},
+          {Symbols::clock, "Quiet Hour", Apps::SettingQuietHour},
```
FintasticMan, line 44: You need to bump the `nScreens` variable to 4.

Boteium: thank you. fixed.

mark9064:
Works well :+1:
Only change I'd suggest is that the minutes spinner shouldn't change the hours. No other time picker I've ever used does this and I found it quite surprising. Definitely a personal preference thing though, not a deal breaker at all

PR #1462, opened 2022-11-26
Star Trek Watchface

ljahn:
## Description

I made a watch face inspired by Star Trek.

base

It supports all the icons and values of the standard digital face.  
The status icons for Bluetooth, battery, power and notifications are on the left (the empty spot between battery and Bluetooth in the picture below is there because I could not make the small notification indicator "i" appear in InfiniSim).  
Step count is under the main clock digits in brackets. The distance between the brackets leaves enough space for the default step goal of 10,000. It breaks down at 100,000 steps, where the number reaches over the right bracket.
If the heart rate measurement is enabled, it is displayed above the date section.
It also supports both 12h and 24h formats.

full

## Technical stuff

I used the new external resources for the large number font and the small brackets.  
I include the brackets as images, because I could not be bothered to draw them as rectangles and circles, like the rest of the display, because they are so detailed. They use up less space in total than the font.
I thought about saving only one and turning it to make the other, but I could not get lvgl to do this.

The font for the large time numbers is [Edge of the Galaxy](https://www.fontspace.com/edge-of-the-galaxy-font-f45748), which is public domain.  
The smaller text for dates and stuff is the InfiniTime standard font. 

## Testing

I tested this on InfiniSim (thoroughly, because I had a memory leak at first [from not freeing the font in the destructor :facepalm:]) and also run it at my personal PineTime without crash so far.

## Things that could be changed

One could change the small font to a font closer to the Star Trek displays, but since it is so tiny, I don't think it would enhance the watch face much (apart from maybe removing the dot from inside the zero).

I am no graphics designer and therefore not good at selecting a balanced color palette. This could use some doing by someone with that skill.

I am unsure if it is okay to call this watch face "Star Trek" in the end, because of licensing issues.


ItzSwirlz:
This is cool! 

In the future, I want to try to get a Stardate thing going - I've tried getting behind how the Star Trek Online Stardate works but perhaps as another time format that could be a nice future feature. https://www.stoacademy.com/tools/stardate.php

I don't think licensing will be a problem - I doubt for a FOSS project any bullets will be shot, but if it is a problem I suppose naming it after a character or a ship. TMI the only copyrighted thing is the actual show + logo

CCF100:
I love this watchface! I really hope it gets merged

infinitytec:
Nice work!

CCF100:
Hey, you might want to look at the watch face initialization code, I've been running this watch face on my watch for a few days, and it's started rebooting whenever the watch attempts to chime on the hour/30 minute mark

It will also briefly display incorrectly right before it crashes...

ljahn:
I cannot replicate the crash.
Not on my watch, which is 1.11.0 plus the commits from this PR (#1462) and #1479,
Nor in InfiniSim with the same as on my watch or even just a build of this branch as is.

@CCF100 Could you please provide some info on what exactly you are running on your watch so I can replicate?

CCF100:
> I cannot replicate the crash.
> Not on my watch, which is 1.11.0 plus the commits from this PR (#1462) and #1479,
> Nor in InfiniSim with the same as on my watch or even just a build of this branch as is.
> 
> @CCF100 Could you please provide some info on what exactly you are running on your watch so I can replicate?

I'm just running the changes in this PR. I haven't merged in any other branches.

CCF100:
PXL_20230309_153033709 MP
It will display like this right before it crashes

ljahn:
Aha! It is using the fallback font and the brackets are gone. This means that the external resources are missing somehow.
Is the time also wrong?

My watch has still not crashed since I enabled the chimes, but I am also not running the exact code in this branch.
You could try to build from this [branch](https://github.com/ljahn/InfiniTime/tree/starTrekOn1.11), which is just version 1.11 plus the watch face.
This  branch here is rebased onto the `develop` (now `main`) branch from the time I created the pull request, maybe the instability comes from there.
The missing resources also hint at that.

infinitytec:
TBH, I kinda prefer the fallback font. It's more readable.

CCF100:
> Aha! It is using the fallback font and the brackets are gone. This means that the external resources are missing somehow. Is the time also wrong?
> 
> My watch has still not crashed since I enabled the chimes, but I am also not running the exact code in this branch. You could try to build from this [branch](https://github.com/ljahn/InfiniTime/tree/starTrekOn1.11), which is just version 1.11 plus the watch face. This branch here is rebased onto the `develop` (now `main`) branch from the time I created the pull request, maybe the instability comes from there. The missing resources also hint at that.

I think the problem might be rather it's having issues loading the external resources, causing it to crash and reboot. It has the external resources installed.

CCF100:
@ljahn I'm flashing the new branch to my watch now, will report back on any instability with this branch ^.^

CCF100:
@ljahn I still experience crashing when I have chimes enabled with the rebased branch :/

FintasticMan:
This is also an issue with the Infinineat and Casio watch faces, but in those cases it doesn't crash, just doesn't load the resources.

ljahn:
I think I found the problem.
I had a fallback font, but no fallback images, so when the external resources wouldn't load, displaying the brackets around the step count (which are images) would lead to crashes.
I added fallbacks for those now, too.

@CCF100 please test when you have time.
I was still unable to replicate the crash, maybe my watch has a golden flash chip that always delivers :wink:.

CCF100:
> I think I found the problem. I had a fallback font, but no fallback images, so when the external resources wouldn't load, displaying the brackets around the step count (which are images) would lead to crashes. I added fallbacks for those now, too.
> 
> @CCF100 please test when you have time. I was still unable to replicate the crash, maybe my watch has a golden flash chip that always delivers wink.
Haven't checked this PR in awhile, I'll check it in a few

CCF100:
> > I think I found the problem. I had a fallback font, but no fallback images, so when the external resources wouldn't load, displaying the brackets around the step count (which are images) would lead to crashes. I added fallbacks for those now, too.
> > 
> > @CCF100 please test when you have time. I was still unable to replicate the crash, maybe my watch has a golden flash chip that always delivers wink.
> Haven't checked this PR in awhile, I'll check it in a few
> 

Alas, the problem still exists. For whatever reason, enabling the chimes causes the watch to crash every time the chimes occurs...

ljahn:
> TBH, I kinda prefer the fallback font. It's more readable for me.

I increased the small gaps in the digits. I thinks this makes it more readable while still keeping the style.
 
widergaps
WiderGapsWatchface

ljahn:
> > > I think I found the problem. I had a fallback font, but no fallback images, so when the external resources wouldn't load, displaying the brackets around the step count (which are images) would lead to crashes. I added fallbacks for those now, too.
> > > @CCF100 please test when you have time. I was still unable to replicate the crash, maybe my watch has a golden flash chip that always delivers wink.
> > > Haven't checked this PR in awhile, I'll check it in a few
> 
> Alas, the problem still exists. For whatever reason, enabling the chimes causes the watch to crash every time the chimes occurs...

I am currently out of ideas what the reason could be.
It still has never happened for me and I checked through my code again if I can see anything obvious and found nothing.
When I wrote this, I took the digital watch face as a base and mainly edited stylistic stuff, I have not touched the fundamental clock logic or the interaction with the OS in any way (especially nothing related to chimes, since I do not use them).

So my hypothesis is that it it still something to do with the external resources, but I don't know what.
Do I have to free the image buffers in some way?
Does chiming for some reason use a lot of memory and the clock goes OOM under unlucky conditions, because the watch face is heavy and leaves not much free space?
I don't know :shrug: 

I am thankful for any additional testers and information on this :smiley:

lman0:
> ￼
Not in subject but
Where did you the bracelet ?
It so cool!

CCF100:
> @CCF100
> Not in subject but
> Where did you the bracelet ?
> It so cool!

My girlfriend got it for me as an anniversary gift

ljahn:
I updated the watch face to 1.12.0
The old version on 1.11.0 remains at [starTrekOn1.11](https://github.com/ljahn/InfiniTime/tree/starTrekOn1.11)

ljahn:
Well there is something happening :wink: 

I have channeled my inner masochist and drawn the detailed brackets around the step count fully in lvgl.
Furthermore, I have made the font of the time digits configurable via a menu that opens on long press (like with the the configurable watchfaces).
This means you can now choose if you want readability or style :wink: @infinitytec.
This also means no external resources are needed when running the system font.

Lastly, I think I found the bug causing the crash. @CCF100  The destructor of the watchface tried to free the font when it was closed. But if the fallback font was loaded, this would result in attempt to free the system font, which lead to a crash. This is fixed now as well.

This branch is now at 1.13.0, not at the current development status, so if you build this, you get 1.13.0 plus the StarTrek watchface.

ljahn:
I continued to do more work on this and I now consider it ready for review @JF002 

This watch face now supports all the usual icons and information like steps and heart rate.
I also added animations, to make it more lively (they can be disabled).
Additionally, we have weather (which I copied from PTS, I think one could do a thing like the BatteryIcon for the weather as well), a config option to display seconds or not, and an option to use the system font for better legibility.
It works without external resources, but then the Star Trek font is unavailable.

Here is a gif of what it looks like:
InfiniSim_2023-11-08_233900
The flickering at the bottom is an artifact of the gif, it does not happen on the watch.

The one thing that reaches deeper into the system is the startup animation. I added `OnLCDWakeup()` and `OnLCDSleep()` functions to the `Screen` class to be able to react to those events and call them in `DisplayApp`.

FintasticMan:
Yeah, I can't see any reason why it should fail... I'll try re-running the workflow.

ljahn:
Well, it fixed itself :tada: :wink:

ljahn:
I updated the watchface to version 1.14.0.
It now includes the `WatchFaceTraits` and all the stuff necessary for the new watchface handling.
I also changed out the weather service for the new `SimpleWeatherService`.

I tested on my PineTime and so far there is no problem. Alas, I cannot get Gadgetbridge Nightly to send weather to the watch (validated with PTS face), so there could be a problem with that, but I do not expect it, because I mostly copied the PTS code for weather again.

mueller-ma:
I compiled this branch and also have issues with the weather. I'm using the latest stable version of Gadgetbridge. Weather is working on PineTimeStyle without any issues. On the Star Trek watchface the weather disappears quite often, e.g. after swiping to the right (to open the menu with brightness, flashlight, ...). Then I have to disable weather in the watch face settings and enable it again to make it work.

ljahn:
@mueller-ma By pure coincidence, I fixed this today :smile:.
I had an error in the check if the weather was updated, which lead to it never receiving updates apart from when the weather mode was toggled.
Now it should work as intended.

Do you have any other input on the watch face?
I'm thinking about aligning the weather and date information a little more to the right, its a bit close to the border as it is now.

mueller-ma:
Overall I really like the watch face. A few thoughts on it:
- I like months as numbers more than as string ("05" instead of "May"). 
- I'd prefer a "bluetooth not connected" icon instead of a missing bluetooth icon in that case.
- I really like that the steps are written as number and as progress bar at the same time.
- It's also good that the new font is optional as you cannot update external resources with Gadgetbridge right now and I haven't looked into other companion apps yet.
- Maybe move the temperature next to the weather icon, similar to heartbeat.
- Seconds can be displayed 👍 
- Maybe update the name, so it doesn't contain trademarks?

ljahn:
@w4tsn True, there is not much happening here, but I don't intend to push people to get this merged in any way.

I made the watchface mainly for myself and for the challenge. I regularly update it to every new release, because I run it on my pinetime and of course want the newest features :wink:.

I maintain this merge request to offer it to the infinitime community, but I am also perfectly fine with it never getting merged.

Handrail9:
@ljahn Silly question, but do I have to compile this from source to use it? I am trying to compile it in docker but I keep running into an error :( Was wondering if youd be able to share the file

ljahn:
I did the great rebase and squash that was long overdue. This is now all in one commit and back on top of current `main`.
I even incorporated color handling for the AOD mode.

ljahn:
Hi @Handrail9,
It should now build cleanly with the newest version of everything.
I compiled a pre release version that you can find [here](https://github.com/ljahn/InfiniTime/releases/tag/31.10.24prerelease). That is just this branch as it is now.

abrahamlnx:
man this is an amazing watchface, definitly the best I've seen

abrahamlnx:
I was hoping to see this watchface added to the main 1.15 branch, welp I guess I'll skip this update to keep using this one

ljahn:
@abrahamlnx 
Dont't worry :wink:. I just rebased this onto 1.15 and will run a few days of testing on my watch. Then I will push 1.15 as well :smile:.

ljahn:
This is now the Star Trek watch face at a clean 1.15.
If there is interest from the team to merge this, I will of course rebase to current HEAD. 

For all those who don't want to self compile, here's a built version: [pinetime-mcuboot-app-dfu-1.15.0.zip](https://github.com/user-attachments/files/18142295/pinetime-mcuboot-app-dfu-1.15.0.zip)
No changes to the resources are necessary compared to the previous version.

ljahn:
I created separate pull requests for all parts of the Star Trek watch face that change something in parts of InfiniTime that are not the watch face in itself.
#2207 #2208 #2209 
This should help getting this reviewed, by breaking it apart and separating the stuff that affects others and needs to be discussed.

ljahn:
Because this PR got so many and the watch face got so many new features through the years that reading this PR is quite hard now, I decided to start fresh in #2221.

PR #1464, opened 2022-11-27
[new watchface] watchface inspired by the Pebble Watch

faxe1008:
## Preview

InfiniSim_2022-11-27_180038

## New resources
SquareRegular_72 (src/resources/fonts/SquareRegular.ttf)
SquareRegular_20 (src/resources/fonts/SquareRegular.ttf)

The smaller font is rendered with 2bpp, since otherwise the non-numerical characters get smushed. 
Originally I intended to use "leco 1976" as it would be the closest match but after researching I think this is not a public domain font.

[https://www.fontspace.com/square-font-f19483](https://www.fontspace.com/square-font-f19483)

## Description
Hi! Weeks ago I stumbled upon this  Pebble watchface:
image

I really liked the idea of having the step count be indicated by a progress bar wrapping around the edge of the screen.
However other then that I took a quite minimalist approach in cutting stuff that I usually do not use or need:

- Heart rate: I normally have the sensor disabled, so there is not much point in displaying
- Current year: I think besides the beginning of the year where it might be a nice reminder, everyone knows what the current year is :^)
- Bluetooth status: I usually have the radio enabled and the watch connected. Applying the reverse would be interesting - having an indicator when the watch is somehow disconnected.

Reducing the amount of displayed things allowed for the battery percentage to be displayed more prominently, since this is a thing that I do regularly check upon. Since adding an icon for the charging state felt out of place in this "icon-less" design I added a animation of the battery bar when the watch is charging. 

The name of the watchface is Gravel since it is inspired by a Pebble watchface but other then that I changed quite a few things :^). If anyone has a better suggestion feel free to drop it.


ThePinkUnicorn6:
This looks great and I would love to use it, one thing I think would be nice to have is the ability to change the accent colour of the step counter the same way that you can in PineTimeStyle.

RageGamerBoi:
The heart rate counter could be rendered only if the sensor is enabled, that sounds like the best compromise.

PR #1465, opened 2022-11-27
Combine Date and Time Settings 

Elements6007:
Previous PR #1386
changes since previous PR:

- different branch name ("/" caused workflow to fail, see CI #1903 )
- retains git history (unlike #1386)

https://user-images.githubusercontent.com/88860704/204163428-870acd82-cd40-4c73-b7c2-b4cc7c4a3390.mp4

(please excuse my clumsy hands) 

Works with InfiniSim after a few small changes. ill open a PR over there later.
unnamed (1)unnamed unnamed (2)

Thanks :)

FINAL:
InfiniSim_2023-01-02_160904


Riksu9000:
These screens are related so condensing the options seems like a good idea, however I'm not sure about the scrollbar being the only indicator here.

How about pressing the option first brings up SetDate, and when you press Set, it moves to the SetTime screen, which then exits back to the settings menu?

Elements6007:
> These screens are related so condensing the options seems like a good idea, however I'm not sure about the scrollbar being the only indicator here.
> 
> How about pressing the option first brings up SetDate, and when you press Set, it moves to the SetTime screen, which then exits back to the settings menu?

Something like this?

https://user-images.githubusercontent.com/88860704/205189227-b14581ef-489d-44e0-934c-50b5e367c1ec.mp4

I could not figure out a way to get it to exit to the settings screen once the time is set(I'm still a noob).

There is, however, one possible issue(more like just a nuisance). What if a user wanted to just set the date? After pressing Set it would go to SetTime, then the user would have to do two down gestures to exit, instead of only one. 

In my opinion the scrollbar is sufficient, but im more than happy to change it if y'all think its not enough :)

Riksu9000:
Actually I think the issue is that swiping to move to another screen that is not a list seems a bit off to me. By making it proceed by pressing Set, you also wouldn't need the scrolling at all. I can implement this idea to try if you don't mind.

Elements6007:
> Actually I think the issue is that swiping to move to another screen that is not a list seems a bit off to me. By making it proceed by pressing Set, you also wouldn't need the scrolling at all. I can implement this idea to try if you don't mind.

Ok, so i was able to get it to work. what I have is `screens.OnTouchEvent(Pinetime::Applications::TouchEvents::SwipeUp);` in void SetDate to bring it to the SetTime screen, and in void SetTime I have `running = false;` to return it to the menu. I do not know if its the proper way of implementing this, but if you think its adequate ill commit it. Im also going to keep the swipe feature as i see it does no harm.

Elements6007:
@Riksu9000 If you could check out the last commit that would be great :)

Riksu9000:
It's close to what I had in mind. The disabling of the button should be removed and I personally wouldn't make these screens a screenlist. I think it adds to the clutter when you could just press the set button to proceed. Then this feature shouldn't require many lines changed either.

Elements6007:
true, if its not a screenlist then its unnecessary clutter. I made a kind of draft commit of another way of doing this [here](https://github.com/Elements6007/InfiniTime/commit/12366bf236f172982f765621689ea7b23bad2e54). This way there is minimal changes, no file renames, and very simple. However there may be a few issues with this approach. First, it might be confusing looking through to find two files for seemingly one application(at least it would be for me). Second, since there is no scrollbar there is a lack of indication that SetTime exists(besides the title "Date&Time").

I could combine SetDate and SetTime into one file like in this PR, but unfortunately i cant figure out how to. I could also add something like the page indicator but i think it would turn out pretty janky.

Riksu9000:
Now that I think about it, I think what felt weird about this was just the scrollbar. What if instead we had page indicator dots on the side? I think we should try and see how that would look.

Elements6007:
so keep it as a screenlist and add dots like this?
Screenshot 2022-12-08 131922new 
Screenshot 2022-12-08 131945

Riksu9000:
Yeah something like that. Maybe using white and gray for the colors.

Elements6007:
> Yeah something like that. Maybe using white and gray for the colors.
PXL_20221216_023943303_2


so heres it with the dot indicators. However, i put these indicators in the setdatetime file. I tried really hard to make it a widget but i cant seem to find how to do make it work(especially making multiple of the same lv objects), if you could work on it or show me what to do that would be great :) 

Im sorry i cant work on the dot indicators further and I fully understand if you dont have the time to work on this.

Riksu9000:
That's okay. I understand it may have been too much to ask. I can work on the widget at some point and I'll report back once it's ready.

Elements6007:
@Riksu9000 so i had kind of an epiphany and was able to figure it out after i took a fresh look at it, turns out it was pretty simple. It still has to be fine tuned to make it better centered. I dont know if it should be in a separate file(how it is in the commit) or in one file in DotLabel, it works either way. I hope you catch this before you start working on it!

src/displayapp/screens/settings/SettingSetDateTime.cpp
```diff
@@ -0,0 +1,240 @@
...
+  lv_obj_align(btnSetDate, lv_scr_act(), LV_ALIGN_IN_BOTTOM_MID, 0, 0);
+  lv_obj_set_style_local_bg_color(btnSetDate, LV_BTN_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_MAKE(0x38, 0x38, 0x38));
+  lblSetDate = lv_label_create(btnSetDate, nullptr);
+  lv_label_set_text_static(lblSetDate, "Set");
+  lv_obj_set_event_cb(btnSetDate, SetDateEventHandler);
+  lv_btn_set_state(btnSetDate, LV_BTN_STATE_DISABLED);
+  lv_obj_set_state(lblSetDate, LV_STATE_DISABLED);
+
+  return std::make_unique<Screens::DotLabel>(0, 2, app, title);
+}
```
Riksu9000, line 127: A tidier solution would probably be to keep the original SetDate and SetTime files/screens, and simply use them in the new screenlist like so.

```suggestion
std::unique_ptr<Screen> SettingSetDateTime::CreateScreen1() {
  return std::make_unique<Screens::SettingSetDate>(...);
}
```

Riksu9000 (changes requested):

src/displayapp/screens/settings/SettingSetDateTime.cpp
Elements6007: I disagree, I think it should all be in one file. Both SetDate and SetTime are shown as one screen, thus they should be found in one file. Also getting back setDate and setTime isnt going to be easy because of how I merged two branches to combine the files and retain the git history, if I did it get them back it would be really messy. Therefore this PR would need to be scrapped. Having three different files for this would just be too complex for something this simple in my opinion. 

additionally, I dont exactly how your suggestion would work...lol

Riksu9000: Files don't add complexity, they create organization. I don't see a good reason to combine two classes that are completely independent from each other into one, where half the functions and variables relate to one screen, and the other half to a different one.

Riksu9000: You can always clean up the history with git-rebase, force pushing, or we can just squash merge this PR.

Elements6007: Ok then.... could you elaborate on how that would work? especially what goes into the (...) part. I usually figure things out by  looking how other apps work but there isnt really anything similar to that :/

Riksu9000: The parameters of the constructor for that screen.

https://github.com/InfiniTimeOrg/InfiniTime/blob/afea7ca0d1d670bdee04cfe80a1d8c36efa4fca0/src/displayapp/DisplayApp.cpp#L413
https://github.com/InfiniTimeOrg/InfiniTime/blob/afea7ca0d1d670bdee04cfe80a1d8c36efa4fca0/src/displayapp/screens/settings/SettingSetDate.h#L14

Elements6007: Ok, that makes much more sense. Thanks!

Elements6007:
@Riksu9000 I reverted one of the commits and implemented suggestion, hopefully it doesn't dirty the pr too much. Anyway I had some trouble adding the dot indicators into it. There is a way if you make both setdate and settime a screenlist but i figured its kind of dumb to have a screenlist inside of a screenlist. I also couldnt figure out how to make it proceed to the next screen properly after pressing set. I did manage to get it to scroll, but it ends up in a black screen. If you could work those two that i would really appreciate it (this time i promise i wont work on it)

src/displayapp/widgets/DotIndicator.cpp
```diff
@@ -0,0 +1,25 @@
...
+  static constexpr uint8_t dotSize = 12;
+  lv_coord_t yPosition = LV_VER_RES / 2 - nScreens * 8;
+
+  for (int i = 0; i < nScreens; i++) {
+
+    dotIndicator[i] = lv_obj_create(lv_scr_act(), nullptr);
+    lv_obj_set_size(dotIndicator[i], dotSize, dotSize);
+    lv_obj_set_style_local_bg_color(dotIndicator[i], LV_BTN_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_GRAY);
+    lv_obj_set_pos(dotIndicator[i], LV_HOR_RES - 16, yPosition);
+    yPosition = yPosition + 32;
```
Riksu9000, line 21: I'm assuming 8 should be dotSize / 2, 16 should be dotSize and 32 should be dotSize * 2.

src/displayapp/screens/DotLabel.cpp
```diff
@@ -0,0 +1,13 @@
+#include "displayapp/screens/DotLabel.h"
```
Riksu9000, line 1: DotLabel isn't needed anymore and can be deleted.

src/displayapp/DisplayApp.cpp
```diff
@@ -409,16 +408,8 @@ void DisplayApp::LoadApp(Apps app, DisplayApp::FullRefreshDirections direction)
...
-    case Apps::SettingSetDate:
-      currentScreen = std::make_unique<Screens::SettingSetDate>(this, dateTimeController);
-      ReturnApp(Apps::Settings, FullRefreshDirections::Down, TouchEvents::SwipeDown);
-      break;
-    case Apps::SettingSetTime:
-      currentScreen = std::make_unique<Screens::SettingSetTime>(this, dateTimeController, settingsController);
-      ReturnApp(Apps::Settings, FullRefreshDirections::Down, TouchEvents::SwipeDown);
-      break;
-    case Apps::SettingChimes:
-      currentScreen = std::make_unique<Screens::SettingChimes>(this, settingsController);
```
Riksu9000, line 421: You've accidentally also removed SettingChimes.

src/displayapp/screens/settings/SettingSetDateTime.cpp
```diff
@@ -0,0 +1,57 @@
...
+              [this]() -> std::unique_ptr<Screen> {
+                return screenSetTime();
+              }},
+             Screens::ScreenListModes::UpDown} {
+}
+
+std::unique_ptr<Screen> SettingSetDateTime::screenSetDate() {
+
+  return std::make_unique<Screens::SettingSetDate>(0, 2, app, dateTimeController, *this);
+}
```
Riksu9000, line 40: I think you should be able to create the dotIndicator in here, which would be a more appropriate place for it I think. Then the screen numbers don't need to be passed to the screens.

Riksu9000 (changes requested):

src/displayapp/screens/settings/SettingSetDateTime.cpp
Elements6007: I tried it and it just crashes.

src/displayapp/widgets/DotIndicator.cpp
Elements6007: Kinda hard for me to explain but for the first dot its half the vertical resolution(120) minus the number of screens times eight, which is 104. so thats (224, 104) for the first dot. For the second its 32 minus the last dot which is (224, 136). They are both 16px away from the vertical center. Now if theres say three screens then its not perfectly centered but its barely noticeable. Im going to see if i can perfect it.

src/displayapp/screens/settings/SettingSetDateTime.cpp
Riksu9000: Works fine for me.

```suggestion
std::unique_ptr<Screen> SettingSetDateTime::screenSetDate() {
  Widgets::DotIndicator dotIndicator(0, 2);
  dotIndicator.Create();
  return std::make_unique<Screens::SettingSetDate>(app, dateTimeController, *this);
}
```

src/displayapp/widgets/DotIndicator.cpp
Riksu9000: Let's make LVGL do the hard work for us with containers.

```c++
void DotIndicator::Create() {
  lv_obj_t* dotIndicator[nScreens];
  static constexpr uint8_t dotSize = 12;

  lv_obj_t* container = lv_cont_create(lv_scr_act(), nullptr);
  lv_cont_set_layout(container, LV_LAYOUT_COLUMN_LEFT);
  lv_cont_set_fit(container, LV_FIT_TIGHT);
  lv_obj_set_style_local_pad_inner(container, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, dotSize);
  lv_obj_set_style_local_bg_opa(container, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, LV_OPA_TRANSP);

  for (int i = 0; i < nScreens; i++) {
    dotIndicator[i] = lv_obj_create(container, nullptr);
    lv_obj_set_size(dotIndicator[i], dotSize, dotSize);
    lv_obj_set_style_local_bg_color(dotIndicator[i], LV_BTN_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_GRAY);
  }

  lv_obj_set_style_local_bg_color(dotIndicator[nCurrentScreen], LV_BTN_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_WHITE);

  lv_obj_align(container, nullptr, LV_ALIGN_IN_RIGHT_MID, 0, 0);
}

```

Elements6007: Oh, that works much better, Thank you.

src/displayapp/screens/settings/SettingSetDateTime.cpp
Elements6007: looks like i did it wrong, oops

Elements6007:
@Riksu9000 is it good now?

Riksu9000 (approved):
Looks good, just one comment. I'll approve this ahead of time. Thanks for the PR.

JF002:
I've just tested this PR. Combine those 2 settings in a single one is a good idea, and it looks good!

However, the app did not jump automatically to "set time" when I tap the button "Set" under the date setting. I also cannot see the page indicator in those settings and the new widget `DotLabel` does not seem to be used anywhere in the code. Is that expected?

Riksu9000:
@JF002 Did you use an older version of this PR? There's no issue for me.

Elements6007:
> I've just tested this PR. Combine those 2 settings in a single one is a good idea, and it looks good!
> 
> However, the app did not jump automatically to "set time" when I tap the button "Set" under the date setting. I also cannot see the page indicator in those settings and the new widget `DotLabel` does not seem to be used anywhere in the code. Is that expected?

Works for me. Also DotLabel was removed in 992333ce211543e9c0e173b45dd3a7eeb04d0654

JF002:
Oh... I kinda forgot to update my local branch... sorry for the confusion! Everything seems to work as expected!

JF002 (approved):

Elements6007:
Thanks for all your help on this @Riksu9000, i really appreciate it! I learned alot :)

Issue #837, opened 2021-11-18
Quick Replies


calhix:
I think it would be cool to be able to quickly reply to certain messages on the watch. You scroll to the message and hit a Reply button. You would then have a list of options such as yes, thanks, sounds good, later, no, etc. that would reply to the sender. Perhaps in the future you could add your own custom ones through Gadgetbridge.

Kozova1:
Android already has a feature like this for some messaging apps, perhaps it is possible to somehow reuse the suggested responses there via Gadgetbridge?

geekbozu:
I think in order for us to support this the companion app need support for it. @Avamander might know more.

JF002:
The protocol we use for the notification does not allow this kind of feature (the notification app sends the notification message/title and waits nothing in return).
Of course, we could add something similar to the phone call control (accept, ignore, reject), but that's not in our scope right now.

Issue #227, opened 2021-03-12
Calculator App.


FlameKat53:
I would appreciate a calculator app very much. i would be making it myself but i'm still a mega noob. please help (anyone who wants to get involved) in the Discord/Matrix/Telegram

Raupinger:
Just to give an update on it. I've got a WIP calculator app. My current parsing + evaluation function runs on my desktop but crashes the pinetime. I've not yet found the time to properly debug it so the fix might require anything from changing a single line to a complete rewrite.

Raupinger:
I've got a WIP PR open on this now #375

rimugu:
Any chance we see this in a release?

joseph58tech:
https://github.com/InfiniTimeOrg/InfiniTime/pull/375

PR #1466, opened 2022-11-28
add note taking app with t9 keyboard

Boteium:
I've created a note taking app with T9 keyboard.
This work is heavliy inspired by the Calculator. https://github.com/InfiniTimeOrg/InfiniTime/issues/227
So, thank you all for your contribution and may your PRs be merge soon.

About Notepad
- T9 keyboard with prediction
- Note is saved in memory and will be lost on reboot
- To reduce memory footprint (dictionary size), only words less than or equal to 6 characters can be predicted. So try to type "some" + "time" + "s" instead of "sometimes"
- Special Key:
-- tap "<=" for backspace
-- long press "<=" to exit edit mode
-- "_" serves both as space and enter (to select T9 predicted word)
-- "<>" to switch character or next predicted word
-- '123' and "Aa" to switch between number/alphabet

Maybe the T9 keyboard can one day become an system component, so any app can use it to get user input ?
I am dreaming of https://github.com/InfiniTimeOrg/InfiniTime/issues/837

https://user-images.githubusercontent.com/8587031/204256601-38807dc2-7147-4f12-83a9-ecc5b3d306d4.mp4

P_20221128_182231

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -399,6 +399,7 @@ list(APPEND SOURCE_FILES
         displayapp/screens/CheckboxList.cpp
         displayapp/screens/BatteryInfo.cpp
         displayapp/screens/Steps.cpp
+        displayapp/screens/Notepad.cpp
         displayapp/screens/Timer.cpp
         displayapp/screens/PassKey.cpp
         displayapp/screens/Error.cpp
@@ -611,6 +612,7 @@ set(INCLUDE_FILES
         displayapp/screens/HeartRate.h
         displayapp/screens/Metronome.h
         displayapp/screens/Motion.h
+        displayapp/screens/Notepad.h
         displayapp/screens/Timer.h
         displayapp/screens/Alarm.h
         displayapp/Colors.h
--- a/src/components/settings/Settings.h
+++ b/src/components/settings/Settings.h
@@ -50,6 +50,8 @@ namespace Pinetime {
         int colorIndex = 0;
       };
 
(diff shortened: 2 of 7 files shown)


minacode:
Great work! I have two questions regarding the functionality. 
1. Can it handle more than one note? 
2. Could this play together with the gallery app (that aims to show text files) in any way?

Boteium:
>     1. Can it handle more than one note?

Not currently. No. Adding it should be easy though.

The reason I did not add more features into it is because I want to keep the code as simple as possible. It's more a keyboard than a note taking app now. Keeping it simple would allow us the option to split it into a keyboard and note taking app in the future.

The note taking app I envisioned should save the text to external storage. Otherwise, supporting multiple notes takes up too much space in the memory. The decision to keep it in the memory for now is because I want it to behave like a "clipboard". The keyboard saves user input into the clipboard. Then the notepad import the text from clipboard and save it to external storage.
 
>     2. Could this play together with the gallery app (that aims to show text files) in any way?

I would love this too.

minacode:
I agree with your answers. Let's see what cool things could be enabled by a keyboard 🙂
While scrolling through your code I found that you use snake case. InfiniTime uses PascalCase for function names for example.

NitroNils:
I found this app in [this blog post](https://wrily.foad.me.uk/fun-with-pinetime-smart-watch) and it's astounding!
I see the use of a generic keyboard available on the PineTime. Unfortunately it doesn't seem to be merged after two years. Lack of time?

PR #1052, opened 2022-03-25
Nimble: Reduce power usage, add compatibility to LFCLK calibration

StarGate01:
This PR has been broken out of https://github.com/InfiniTimeOrg/InfiniTime/pull/1050 .

- Reduce BLE power usage by only powering up radio peripherals and the high frequency clock when needed (`BLE_LL_RFMGMT_ENABLE_TIME` parameter)
- Use NRF SDK to access high frequency clock, in order to not interfere with potential LFRC calibration.

PR #1471, opened 2022-12-02
reduce power usage by lowering ble crystal startup time

Boteium:
Pull request https://github.com/InfiniTimeOrg/InfiniTime/pull/1052 reduces power usage by powering down ble crystal when in idle. The crystal startup time is set to 1500 (mynewt default)

The nrf52 spec says it should be 360 in section 19.4.2. https://infocenter.nordicsemi.com/pdf/nRF52832_PS_v1.8.pdf

I think we can safely lower it to 400.
I've been using this parameter for several days.
However, I only have a sealed device and cannot verify if this actually saves battery, or, if this parameter even works.

So, I will set this PR as draft for now.


Boteium:
accidentally close this because of renaming repo.

Issue #1463, opened 2022-11-26
Bricking pinewatch using large size firmware for testing


AlexXZero:
### Verification

- [X] I searched for similar bug reports and found none was relevant.

### What happened?

Brick pinewatch after testing feature (twice)

### What should happen instead?

Linker error if build is not fit in available flash memory.

### Reproduction steps

1. Update code with adding feature so image size should use more than 95% of available internal flash.
2. Flash image to the dev kit using DFU.
3. Reboot the device to back into previous firmware

Then pinewatch start to be in constant reboot.

### More details?

I tried to figure out what I'm doing wrong and found that InfinityTime uses magic internal flash addresses for saving some boot configuration, so I guess that rewriting this data is cause of this issue. I would suggest to do the following thing to avoid this issue for debug builds and for future InfinityTime releases which might start use more flash memory:
1. Update linker script to have addresses of each magic variable in it.
2. Update linker script with decreasing size of available internal flash memory size to avoid overlapping firmware and boot configuration flags.
3. Update bootloader (not sure if it is possible to do for all users) to use last page for storing boot flags.
4. Update documentation with flash memory map to let developers know which memory limitations they have.

I found at least one magic boot flags which is not present in the linker script and which might be overwritten by firmware:
```
FirmwareValidator::validBitAdress {0x7BFE8}; // Magic address!!!
```
So to avoid overlapping with this address available internal flash memory should be decreased to 0x5000 bytes!
```diff
 MEMORY
 {
-  FLASH (rx) : ORIGIN = 0x08020, LENGTH = 0x78000
+  FLASH (rx) : ORIGIN = 0x08000, LENGTH = 0x73000
   RAM (rwx) :  ORIGIN = 0x20000000, LENGTH = 64K
 }
```
Alternative option is somehow update bootloader to the latest page for this purposes, i.e. address `0x7FFFE8`.

Note: I'm not sure if any other magic addresses are used somewhere else.

### Version

https://github.com/AlexXZero/InfiniTime/releases/tag/1.11.0-mod3

AlexXZero:
In a few days I'm going to add an instruction with photo's how I unbrick my dev kit. I remember that documentation of this procedure wasn't full enough for me (I did it using nrf52832 dev board) so I needed to find third party resources to make it clear for me.

AlexXZero:
[...]

PR #1472, opened 2022-12-02
Fix linker script to prevent using not available flash memory

AlexXZero:
There is a fix for the issue found in the ticket https://github.com/InfiniTimeOrg/InfiniTime/issues/1463
It replace magic numbers with named constants in the linker script for mcuboot application.


gcc_nrf52-mcuboot.ld
```diff
@@ -1,11 +1,39 @@
...
+ */
 
 SEARCH_DIR(.)
 GROUP(-lgcc -lc -lnosys)
 
+MCUBOOT_SIZE = 0x8000;
+MCUBOOT_APP_IMAGE_HEADER_SIZE = 0x20;
+MCUBOOT_APP_IMAGE_TRAILER_SIZE = 432;
+APP_OFFSET = MCUBOOT_SIZE + MCUBOOT_APP_IMAGE_HEADER_SIZE;
+APP_SIZE = 464K - MCUBOOT_APP_IMAGE_HEADER_SIZE - MCUBOOT_APP_IMAGE_TRAILER_SIZE;
```
Riksu9000, line 27: 464K is still a magic number. It should be `SCRATCH_OFFSET  - MCUBOOT_SIZE` I think.

gcc_nrf52-mcuboot.ld
```diff
@@ -1,11 +1,39 @@
-/* Linker script to configure memory regions. */
+/****************************************************************
+ * Memory map configuration for using application with MCU-boot *
+ ****************************************************************/
+/*
+ * Internal nRF52 flash memory:
+ * 0x00000000: MCUBoot(28 kB)
+ * 0x00007000: Reboot logs (4 kB)
+ * 0x00008000: MCUBoot header (0x20 bytes)
```
Riksu9000, line 8:
```suggestion
 * 0x00008000: MCUBoot header (20 bytes)
```

gcc_nrf52-mcuboot.ld
```diff
@@ -1,11 +1,39 @@
...
 SEARCH_DIR(.)
 GROUP(-lgcc -lc -lnosys)
 
+MCUBOOT_SIZE = 0x8000;
+MCUBOOT_APP_IMAGE_HEADER_SIZE = 0x20;
+MCUBOOT_APP_IMAGE_TRAILER_SIZE = 432;
+APP_OFFSET = MCUBOOT_SIZE + MCUBOOT_APP_IMAGE_HEADER_SIZE;
+APP_SIZE = 464K - MCUBOOT_APP_IMAGE_HEADER_SIZE - MCUBOOT_APP_IMAGE_TRAILER_SIZE;
+SCRATCH_OFFSET = 0x7c000;
+SCRATCH_SIZE = 4K;
```
Riksu9000, line 29: Maybe all values should be in hex, instead of mixing decimal and kibibytes.

gcc_nrf52-mcuboot.ld
```diff
@@ -1,11 +1,39 @@
...
+ */
 
 SEARCH_DIR(.)
 GROUP(-lgcc -lc -lnosys)
 
+MCUBOOT_SIZE = 0x8000;
+SCRATCH_SIZE = 0x1000;
+SCRATCH_OFFSET = 0x7c000;
+MCUBOOT_APP_IMAGE_HEADER_SIZE = 32;
+MCUBOOT_APP_IMAGE_TRAILER_SIZE = 432;
```
Riksu9000, line 27: Mixing decimal and hex can be confusing. Maybe only use hex for these.

gcc_nrf52-mcuboot.ld
```diff
@@ -1,11 +1,39 @@
...
+APP_OFFSET = MCUBOOT_SIZE + MCUBOOT_APP_IMAGE_HEADER_SIZE;
+APP_SIZE = SCRATCH_OFFSET - MCUBOOT_SIZE - MCUBOOT_APP_IMAGE_HEADER_SIZE - MCUBOOT_APP_IMAGE_TRAILER_SIZE;
+
 MEMORY
 {
-  FLASH (rx) : ORIGIN = 0x08020, LENGTH = 0x78000
+  /* MCUBOOT (r) : ORIGIN = 0x0, LENGTH = MCUBOOT_SIZE */
+  FLASH (rx) : ORIGIN = APP_OFFSET, LENGTH = APP_SIZE
+  /* SCRATCH (r) : ORIGIN = SCRATCH_OFFSET, LENGTH = SCRATCH_SIZE */
+  SPARE_SPACE (r) : ORIGIN = SCRATCH_OFFSET + SCRATCH_SIZE, LENGTH = 12K
```
Riksu9000, line 36: Same here.

gcc_nrf52-mcuboot.ld
AlexXZero: Agree.

gcc_nrf52-mcuboot.ld
AlexXZero: For me `12K` here is more obvious, rather than `0x3000`. It is also consistent with RAM size (see line below). So I would suggest to leave this line as is.

JF002:
@AlexXZero Thank you **very much** for your in-depth analysis in #1463 and for this PR!

I had a look at the changes, and I can see that I did a few mistakes when I designed the memory map for the bootloader and the firmware! The image header and trailer were not taken into account!

Also, the last 12KB that are left unused are a bit intriguing... To be honest, I can't remember what was my intent : I either made a calculus error, and I intended to use them as scratch area, but forgot to extend the area size to 16KB...
In the short term, I guess we'll leave it as it is, but we can brainstorm about how to best use those 12KB :)

Have you already had the opportunity to check that those changes in the linker script work correctly, and that it prevents from generating a faulty (too big) firmware?

JF002:
I did a few additional checks : generate the biggest possible firmware file, and then the corresponding image. The goal is to check that everything work fine when flashing the biggest size allowed.

So, according to @AlexXZero's calculation, we should be allowed to build a firmware that is 474672 (0x73e30) bytes of size.

```
           FLASH:      473728 B     474672 B     99.80%
             RAM:       54376 B        64 KB     82.97%
post build steps for pinetime-mcuboot-app-1.11.0
 473728	    940	  53432	 528100	  80ee4	pinetime-mcuboot-app-1.11.0.out
Usage: imgtool.py create [OPTIONS] INFILE OUTFILE
Try 'imgtool.py create -h' for help.

Error: Image size (0x73e74) + trailer (0x630) exceeds requested size 0x74000
```

In this case, the firmware size is 473728B of code and 940B of data which is 474668B total. It's 4B less than the maximum, but I couldn't reach the exact number of 474672.

Notice that ``--print-memory-usage`` does not take constant data into account while arm-none-eabi-size does.

There are 2 issues here : 
 - why does imgtool think that the image size is 0x73e74 (474740B) while it's either 474668B or 474700B with the image header ? What do those 40B account for?
 - why does imgtool want to generate a trailer of 0x630 (1584B) instead of 0x1b0 (432B) calculated by @AlexXZero  ?

I do not have the answer to the 1st question. 

The 2nd question can be explained by `min-write-size` (see [here](https://github.com/InfiniTimeOrg/InfiniTime/issues/1463#issuecomment-1334572727) and the `--align` parameter of imgtool.
According to @AlexXZero, `min-write-size` is set to 1(*), while `--align` is set to 4. Let's try with a 1 byte align:

```
Error: Image size (0x73e74) + trailer (0x1b0) exceeds requested size 0x74000
```

Yep, trailer sizes match, but not the image size.

So, in theory, we should change the imgtool CLI to use an align value of 1, but I would like to understand why there's a difference in the image size between the build and the call to imgtool first.

---

(*) And I think they are right. 
In the bootloader, the internal flash is configured this way (see repos/apache-mynewt-core/hw/mcu/nordic/nrf52xxx/src/hal_flash.c:74):
image
See `hf_align=1`.

Still in the bootloader, the external spi flash is configured this way : (repos/apache-mynewt-core/hw/drivers/flash/spiflash/src/spiflash.c)
image
`hf_align` is also set to 1.

Since both memories are configured with an align value of 1, mcu boot should use a min-write-size equal to 1.

JF002:
Ok, I found where do those 40bytes come from : `imgtool.py` adds a few TLV (type-length-value records) at [https://github.com/mcu-tools/mcuboot/blob/v1.6.0/docs/design.md#image-format](the end of the image).

See https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/tools/mcuboot/imgtool/image.py#L374. I added a few `print`s and figured out that the TLV account for exactly 40Bytes.

So I guess we'll probably take that into account in the linker script by removing 40 bytes to the max size of the firmware.

gcc_nrf52-mcuboot.ld
```diff
@@ -1,11 +1,39 @@
...
+ * 0x00040000: Application 2 (including MCUBoot header) (464 kB)
+ * 0x000b4000: User files - littlefs (3376 kB)
+ */
 
 SEARCH_DIR(.)
 GROUP(-lgcc -lc -lnosys)
 
+MCUBOOT_SIZE = 0x8000;
+SCRATCH_SIZE = 0x1000;
+SCRATCH_OFFSET = 0x7c000;
```
JF002, line 25:
```suggestion
TLV_SIZE =  0x28; /* Metadata added by imgtool at the end of the image */
SCRATCH_OFFSET = 0x7c000;
```

gcc_nrf52-mcuboot.ld
```diff
@@ -1,11 +1,39 @@
...
+MCUBOOT_APP_IMAGE_HEADER_SIZE = 0x20;
+MCUBOOT_APP_IMAGE_TRAILER_SIZE = 0x1b0;
+APP_OFFSET = MCUBOOT_SIZE + MCUBOOT_APP_IMAGE_HEADER_SIZE;
+APP_SIZE = SCRATCH_OFFSET - MCUBOOT_SIZE - MCUBOOT_APP_IMAGE_HEADER_SIZE - MCUBOOT_APP_IMAGE_TRAILER_SIZE;
```
JF002, line 29:
```suggestion
APP_SIZE = SCRATCH_OFFSET - MCUBOOT_SIZE - MCUBOOT_APP_IMAGE_HEADER_SIZE - MCUBOOT_APP_IMAGE_TRAILER_SIZE - TLV_SIZE;

```

JF002: I would really appreciate if you could confirm my findings :)

JF002 (changes requested):

gcc_nrf52-mcuboot.ld
AlexXZero: Thanks for your investigation, I reviewed your finding and seems you are completely right, so I applied your suggestions.

AlexXZero:
@JF002 It seems we also need to change the imgtool arguments, I will review how to do it soon and also will test building and running an image with maximum possible size.

JF002:
@AlexXZero The new memory map looks good! Waiting for your test results :)

AlexXZero:
I have built and tested image with maximum possible size (actually I also couldn't reach the rest 4 bytes, but I guess it is caused by alignment between .text and .data sections)

Test firmware is based on [cfc86d7](https://github.com/InfiniTimeOrg/InfiniTime/pull/1472/commits/cfc86d7d85167b1340740bbf4ec319d62c1513cf) commit + the following changes:
```diff
diff --git a/gcc_nrf52-mcuboot.ld b/gcc_nrf52-mcuboot.ld
index 6c3a769b..0fb3e9e9 100644
--- a/gcc_nrf52-mcuboot.ld
+++ b/gcc_nrf52-mcuboot.ld
@@ -166,6 +166,12 @@ SECTIONS
     PROVIDE(__stop_nrf_balloc = .);
   } > FLASH
 
+  .test_dummy :
+  {
+    . = ALIGN(4);
+    . += 56212;
+  } > FLASH
+
 } INSERT AFTER .text
 ```
 
 Tested actions:
 1. Uploading DFU image onto pinewatch dev kit
 2. Tested if test version is working correctly
    - button is working
    - touch sensor is working
    - watch screen is working
    - menu is working
 3. Tested reboot
    - Approved current version first
    - Hold the button to produce software reset
 4. Tested if rolling back to previous version
    - Hold the button to produce software reset
    - Hold the button in bootloader mode until it start process of rolling back to previous firmware version
    
All test results are match expected behavior.
@JF002 I think this PR is ready and I will prepare updating documentation in another branch/PR to make it easier to review. Let me know if I need to do something else.

JF002:
@AlexXZero Thanks for your test results, it looks good to me! I'm OK to create a new PR for the documentation.

Just one minor detail : I think we should also update `gcc_nrf52.ld`. This linker script is used in development, when flashing the firmware directly at offset 0x00 so that you don't need to run the bootloader when flashing a new version of the firmware via SWD. The only difference with `gcc_nrf52-mcuboot.ld` is the flash offset set to 0x00 instead of 0x8000.

JF002:
I added the changes in gcc_nrf52-mcuboot.ld and also updated the documentation in the bootloader readme file (https://github.com/InfiniTimeOrg/pinetime-mcuboot-bootloader/commit/2bd9ead9a22605ade0c9770b6f41ab2f099d39b9) so that we can merge this branch.

Thanks for you work,  @AlexXZero !

PR #1475, opened 2022-12-06
typo in the description

tom-programming:
AMD64 and x86_64 are the same architecture - ARM64 should be specified instead


Avamander:
Thanks!

PR #1479, opened 2022-12-07
Metronome: Make the BPM estimate a moving average

ljahn:
## Moving Average BPM estimation

It is already possible to tap on the number in the metronome app to estimate the bpm of tapping.

However, I found this estimate to be a bit jumpy, so I made it a moving average over the last three measurements.  
This way, there is some smoothing applied by considering the history and not only the instantaneous measurement.
The displayed value is always the average of the last three measurements.
This  also reduces the measurement inaccuracy for tapping along constant tempo pieces.

It has the side effect, that a change in tapping speed will lead to a 3 tap ramp-up/down until the new estimation is reached.

### Possible changes

There is nothing special about the number three, one could make the history longer, which would further improve the measurement accuracy for constant tapping, but would also increase the ramp time for a change in tapping speed.  

The ramping could be counteracted by comparing the instantaneous measurement to the last estimation and dropping the history if it differs significantly (say 20%).


ljahn:
Closing as stale.
Haven't used it much myself.

PR #1482, opened 2022-12-08
Refactor notifications

minacode:
This PR does two things (I know, but I hope it is ok here):
1. Move the `NotificationManager` out of `src/components/ble` into its own component `src/components/notifications` because it does not directly have anything to do with BLE.
2. It reorders the fields of the ´Notification` struct such that one should now be able to use the following string literal syntax to create a new notification:
```cpp
NotificationManager::Notification notif {
  "String literal with the notification text",
  42,
  NotificationManager::Categories::SimpleAlert
};
```

---

I also noticed that `src/displayapp/screens/WatchFaceAnalog.h` is the only watch face where the include happens in the header. Maybe someone with better C++ knowledge than me has an opinion on this :grinning: 

The reason for this PR is @NeroBurner's comment on #1352 where this would simplify the notification creation.

Right now three files create notifications:
1. `ImmediateAlertService.cpp`
2. `AlertNotificationService.cpp`
3. `AlertNotificationClient.cpp`

but all of them use either `os_mbuf_copydata` or `std::memcpy` to write the message text.


Boteium:
Notification does need some refactors. For example, If the messages were stored in a circular ring buffer instead of five fixed sized char array, we can either store more notifications or increase the max message size without increasing memory usage. But that's probably asking too much.

minacode:
Sounds interesting, but maybe we should do that in another PR. This one already includes two things (own component and reordering). You want to solve that there is a lot of wasted space because of the text arrays?

Boteium:
Definitely in another PR.
My experience is that half of my notifications are very short. Yet each short message still occupy 100-char space regardless of its actual length. Maybe some ring buffer like solution can amend this ? Storing all message in one big char array and only record each message's start and end position, perhaps ?

FintasticMan:
> I also noticed that `src/displayapp/screens/WatchFaceAnalog.h` is the only watch face where the include happens in the header. Maybe someone with better C++ knowledge than me has an opinion on this :grinning:

If something from a header is used in any way, it should be included in that file. Because a reference to NotificationManager is stored in the watch face classes, I believe that it should be included in all the watch face headers.

minacode:
@Boteium, sounds reasonable. You should try this!

@FintasticMan, thank you. I will add it to all the headers then.

src/displayapp/screens/WatchFaceDigital.h
```diff
@@ -7,6 +7,7 @@
 #include "displayapp/screens/Screen.h"
 #include "components/datetime/DateTimeController.h"
 #include "components/ble/BleController.h"
+#include "components/notifications/NotificationManager.h"
```
NeroBurner, line 10: the `NotificationManager` class is forward declared below. So we need either the include or the forward declaration below, but not both

NeroBurner:
The changes here look good. What have you planned to do until you remove the draft status?

A PR over at InfiniSim fixing the simulator CI job linking to this issue would be very much appreciated :)

minacode:
Nice, I was just waiting for input. I will do the PR for InfiniSim.

NeroBurner (approved):

Riksu9000:
The PR needs to be rebased and the commits need to be fixed according to the contribution guide. https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/doc/contribute.md#commit-conventions

The fields should be reoredered in a separate commit with an explanation why.

Riksu9000:
The files really shouldn't be moved in the same PR either.

minacode:
So the best is to close this PR and open two new ones?

Riksu9000:
You don't *have* to close this PR. You could remove the file changes from this PR and have just a single commit with a good description, if you decide to go that route. The moving of the files is a bit subjective and not that important right now, but you could open another PR for that.

PR #1485, opened 2022-12-12
Change clang-format to block indent

minacode:
This PR changes the `AlignAfterOpenBracket` option to `BlockIndent`. See [here](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) for all options available. 

Instead of 
```
a_very_long_function_name_that_reaches_far_to_the_right(arg1,
                                                        arg2,
                                                        arg3
);
```
the new format would look like
```
a_very_long_function_name_that_reaches_far_to_the_right(
  arg1,
  arg2,
  arg3
);
```

I like the second option more because it removes the big blocks that are sometimes shifted far to the right side. On a smaller screen or a split-screen setup the code is not easily visible like that. Diffs do not look as good as well, because of the amount of whitespace.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,7 +1,7 @@
 ---
 Language: Cpp
 AccessModifierOffset: -2
-AlignAfterOpenBracket: Align
+AlignAfterOpenBracket: BlockIndent
 AlignConsecutiveMacros: true
 AlignConsecutiveAssignments: false
 AlignConsecutiveDeclarations: false


Avamander:
I don't think really narrow screens are an use-case worth reformatting the entire codebase for or worth the effort in general.

minacode:
I think that good looking and better usable code could be worth it and cannot see any advantage of the current format. 
Take a look at the second half of `src/displayapp/screens/Clock.cpp` for example.

NeroBurner:
the new format would be more inline with my personal coding style, and I think git diffs in the review process would also benefit greatly from this format change.

I have no strong opinion on the 120 vs 140 screen width. I'm mostly coding on a 1080p screen landscape with two files side by side (roughly a per file width of about 94 character). So slightly favoring 120, but no strong opinion

Avamander:
> I think that good looking and better usable code could be worth it and cannot see any advantage of the current format.

But it's not better on normally-sized screens. Even my 5:4 aspect ratio monitor doesn't struggle one bit.

> Take a look at the second half of [src/displayapp/screens/Clock.cpp](https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/src/displayapp/screens/Clock.cpp) for example.

There's a clear hierarchical connection between the segments of code, as there should be.

NeroBurner:
> There's a clear hierarchical connection between the segments of code, as there should be

For me the hierarchial connection is still available with the new block indenting. But with less spaces and thus better GitHub review diff-view

currently:

```cpp
std::unique_ptr<Screen> Clock::WatchFaceDigitalScreen() {
  return std::make_unique<Screens::WatchFaceDigital>(app,
                                                     dateTimeController,
                                                     batteryController,
                                                     bleController,
                                                     notificationManager,
                                                     settingsController,
                                                     heartRateController,
                                                     motionController);
}
```

after block indent:
```cpp
std::unique_ptr<Screen> Clock::WatchFaceDigitalScreen() {
  return std::make_unique<Screens::WatchFaceDigital>(
    app,
    dateTimeController,
    batteryController,
    bleController,
    notificationManager,
    settingsController,
    heartRateController,
    motionController);
}
```

minacode:
I understand that it is big pain to reformat everything, especially for all the open PRs. 

I would like the smaller width as well because of me working a lot in split screens.

I don't get why we indent everything with two spaces and this one syntax is aligned differently. From my experience (which might not be too big) it is also quite unusual to do so.

minacode:
Feel free to close this PR if you do not agree with the changes I proposed.

JF002:
I guess this is mostly a case of personal preferences. 
I'm not sure I'll be able to provide any other argument other than "I prefer the way the code is currently indented". It seems more readable to me. But I totally understand that other people think the other way around.

I don't think changing this will bring sufficient advantages that are worth reformatting the whole code base. So I suggest we close this suggestion as not planned.

Issue #1405, opened 2022-10-31
Heart rate updated too infrequently


stevensu1838:
### Verification

- [X] I searched for similar feature request and found none was relevant.

### Pitch us your idea!

Increase the heart rate data logging interval

### Description

I've noticed that the heart rate on Pine Time watch is updated every few seconds. How can I increase the heart rate data logging interval to collect more heart rate data in the same time period? Like how can I make the heart rate data update every second? Thanks a million. Cheers.

JF002:
Are you asking this question as a user or as a developer?

The update interval is caused by how the current algorithm is designed. As a user, you cannot increase or decrease it, InfiniTime does not provide this functionality (yet?). As a developer, you'll need to improve the HR algorithm so that it yields measurements more often. Your contribution would be more than welcome ;-)

stevensu1838:
@JF002 Hi, Thank you so much. Do you know what exactly the heart rate data logging interval is? Can you please show the heart rate interval part of the current algorithm? I am a newbie but would love to explore and contribute. Cheers

JF002:
@stevensu1838 According to the code, we run the heart rate task every 40ms when the data acquisition is enabled (https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/src/heartratetask/HeartRateTask.cpp#L32).

Then, the PPG algo returns an actual heart rate value when is has accumulated 200 samples (https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/src/components/heartrate/Ppg.cpp#L61), which means that a new value is available every ~8s.

Reducing those constants will probably decrease the accuracy of the measurements, though.

from-nibly:
Just a thought, the way you'd probably need to accomplish more frequent measurements is not through shorter samples. but rather a sliding window of samples. if you take 200 samples, you can update the "measurement" by adding a 201st sample and only including the most recent 200 samples in your math. That way you'd be able to have a heart rate measurement every 40ms (after the first 8 seconds that is)

JF002:
@from-nibly #1486 will improve the accuracy and speed of measurement :)

Issue #682, opened 2021-09-20
Feature request: zero out HRM when stopped


xan-m:
I'm working on graph data persistence for HRM and Battery for https://github.com/xan-m/Infini-iOS and I noticed last night that if the HRM is stopped on InfiniTime (i.e. if it was running, but you press the stop button), it retains the last heart rate value and continues to report it over the BLE HRM characteristic as the "current" heart rate on every connection/reconnection. 

In my case last night, I had the watch just barely out of range of my phone with my iOS app open with the HRM stopped on the watch, so it was disconnecting and reconnecting pretty frequently. Every time the watch and phone were able to establish a connection, the watch reported the HRM value that it had been stopped at, and it was saved to the graph. I ended up with close to 30 reports of the heart rate value that I had stopped the HRM at.

I've got a kludgy workaround in place in my app to keep these values from being reported, but it potentially sacrifices the occasional real data point to reduce saturation of false data points. If InfiniTime zeroed this value out instead, it would be much easier to filter out (hopefully nobody's heart rate is 0...)

I realize this is an edge case, but I don't see much utility in retaining the last heart rate value after stopping the HRM. If it is useful to others, perhaps a "pause" and a "stop" button could be implemented so that users have the option to retain or clear the value at their discretion.

Issue #594, opened 2021-08-17
Bug: Heart rate monitoring continues while charging


FrazerClews:
If the heart rate monitor is on before charging, it will continue to stay on during and after charging.

I have made a WIP branch [here](https://github.com/JF002/InfiniTime/compare/develop...FrazerClews:frazer/heartReadingStopp) for now before I go any further with the patch which needs testing, just in case something else is planned to fix this.

Riksu9000:
It also keeps on measuring if it's just placed on a table. The ambient light sensor should probably be used instead to detect when the watch is taken off and then the measurement should be stopped.

hatmajster:
Erm, its not true, monitoring is turned off as soon as the screen go black, exactly like on hand - so its not working all the time while charging.

PR #1486, opened 2022-12-12
Refactored Ppg for frequency based algorithm.

Ceimour:
Changed driver settings in HRS3300.cpp for better signal to noise at 10Hz. Changed measurementStarted delay to 100ms in HeartRateTask.cpp. Removed unused ppg methods from HeartRateTask.cpp. Added ambient light sensor acquisition and handling in HeartRateTask.cpp. Added FFT and interpolation Ppg support code from Arduino FOSS implementations. Removed currently unused time domain based code.
Changed all doubles in PPG/Arduino code to float type so that FPU is used.


minacode:
There are files under the Apache License 2.0. Are they compatible with this project?

minacode:
Since this PR contains a lot, any sort of documentation helps you to potentially get this through. 

For the start it think you should describe a little why you did what you did to get an overview of the different changes. The maintainers don't have much time so presenting your ideas and reasons well is very helpful.

Then we could discuss if we like the changes and if they could maybe separated into smaller pull requests.

Also please tell us what this PR improves.

Ceimour:
Thanks for taking a look at this!

According to Apache.org (first 2 paragraphs) we are good to use Apache V2.0 within GPLV3, just not the other way around:
https://www.apache.org/licenses/GPL-compatibility.html

minacode:
Ok, cool.

Are you aware that your replies contain some email stuff? No offense, just wanted to tell you in case you didn't notice.

Avamander:
If replying over email, it's better not to quote the previous message. I've also manually truncated it from people's comments so the thread would look nicer.

Ceimour:
Motivation:
After receiving my new watch I was very impressed with the look, feel and quality of InfiniTime. I expected a hobby project but this something different, something I actually want to use! I experimented with the various features and I noticed the heart rate monitor accuracy issue (like issue 532 below). I set up a dev environment and examined the incoming data. Much of the time it looked really good after Daniel Thompson’s filtering was applied, but then there would sporadic instances of high amplitude noise within the heart rate frequency range. So it’s the old garbage in, garbage out problem. Since I’ve had some experience working in the frequency domain I thought it might be easier to discriminate against the noise using frequency analysis.

Changes:
No fundamental architecture changes.
Previous code in the Ppg methods has been replaced to implement a sample buffer with an overlap window (currently at 8 samples), frequency analysis, a bandwidth filter that covers a wider heart rate range and various small methods for peak search, averaging, etc.
All floating point types are 32bit float to take advantage of the FPU.
No dynamic allocations.
Adds 1-2% to stack and flash. Not sure of the exact amount.

Improvements:

  *   First HR result in a little 6.4 seconds
  *   Updates every 800ms
  *   Good stable HR accuracy (when HR is stable compare with a predicate device and should be within +/-1bpm)
  *   Tracks HR fairly well when the rate is changing slowly (experiment by waiting for a stable HR, take a breath, hold and HR should rapidly decrease)
  *   Uses spectral averaging to increase signal to noise ratio
  *   Uses sample averaging to hang on to previous HR (about 9 second history and drops to zero when the history buffer is all zeros)
  *   Spectral and sample averaging allow the algorithm to ignore quite a few noise events without going to zero at each noise event

Cons:

  *   Changing HR will lag a bit behind some predicate devices
  *   It is still possible to get errant readings due to noise although these generally go away quickly as the HR signal takes over

Pull Request:
I think this needs to be a single pull request because all the code in the algorithm is integrated to produce a single output. Although it looks like it addresses 532 and coincidentally 1405 and 1092 below.

Applied Algorithm (Ppg::HeartRate):
The algorithm is basically similar to the same processing chains found in many papers for extracting information from PPG:

  1.  Acquire samples from the HRS3300
     *   If acquisition is reset or initial, acquire all 64 samples before processing (before going to step 2)
     *   If acquisition is not reset, acquire 8 (overlapWindow)  samples (before going to step 2)
  2.  Detrend the dataset
     *   Remove the slope
     *   Convert to the difference between each sample
     *   Turns the problematic large, DC shifts into single sample spikes
  3.  Bandpass filter the data 30 to 240Hz
     *   Removes low frequency data that can interfere with low HR detection
     *   Removes the high frequency spikes from the detrending step
  4.  Apply a window function
     *   Make the signal appear continuous to the FFT
  5.  Calculate the FFT (Fast Fourier Transform)
  6.  Calculate the power spectrum (ComplexToMagnitude())
  7.  Apply result to spectral average
     *   Helps to enhance the more frequently occurring heart rate frequency while diminishing random noise
  8.  Check signal to noise ratio (SNR)
     *   Nothing complex, just compare max to mean
  9.  Search for peaks in the spectrum only if SNR is good,
     *   Look for peaks above the peak detection threshold (currently 60% of max)
     *   If only one peak above threshold, return the peak position in peakLocation
  10. Check peak width
     *   Wide peaks can occur from noise or rapidly changing HR
     *   If peak is too wide, set peak location to 0.0
  11. Check if HR is within limits (currently 40-230bpm)
     *   If out of range set HR to 0.0
  12. If HR == 0.0, reset spectral averaging for next pass
  13. Set Ambient Light Sensor threshold
     *   Used in next pass in (Ppg::PreProcess and HeartRateTask.cpp)
     *   HeartRateTask will reset the acquisition if ALS threshold is exceeded (goto 1)
  14. Add current peak location (zero or otherwise) to Ppg::HeartRateAverage and assign the result back to peakLocation
  15. Return result
     *   > 0.0 is valid HR
     *   -1 tells HeartRateTask to reset acquisition
     *   0.0 is no valid peak found this pass
  16. Goto 1
Execution time ~20ms


Relevant feature requests:
Heart rate measurement accuracy
https://github.com/InfiniTimeOrg/InfiniTime/issues/532

Increase the heart rate data logging interval
https://github.com/InfiniTimeOrg/InfiniTime/issues/1405

Ring Buffer for Heart Rate Data
https://github.com/InfiniTimeOrg/InfiniTime/pull/1092


Potential feature requests that can be addressed in future pull requests:
Zero out HRM when stopped
https://github.com/InfiniTimeOrg/InfiniTime/issues/682

Bug: Heart rate monitoring continues while charging
https://github.com/InfiniTimeOrg/InfiniTime/issues/594

Sleep tracking
https://github.com/InfiniTimeOrg/InfiniTime/issues/307

Heart rate measurement is stops, when the screen is turned off
https://github.com/InfiniTimeOrg/InfiniTime/issues/183


Please let me know if you have any questions, need more data or if there are any issues with the code.
Thanks!

Ceimour:
Thanks. I didn’t even think of that! So used to just finishing and hitting the send button.

Ceimour:
Thanks, got it!

minacode:
Your description sounds good as far as I read it. More reliable heart rate data would be a very good thing! Can you restore the formatting that it seemed to have when you wrote it? ;) You can just edit the comment if that is possible for you.

Ceimour:
**Motivation:**
After receiving my new watch I was very impressed with the look, feel and quality of InfiniTime. I expected a hobby project but this is something different, something I actually want to use! I experimented with the various features and I noticed the heart rate monitor accuracy issue (like issue 532 below). I set up a dev environment and examined the incoming data. Much of the time it looked really good after Daniel Thompson’s filtering was applied, but then there would sporadic instances of high amplitude noise within the heart rate frequency range. So it’s the old garbage in, garbage out problem. Since I’ve had some experience working in the frequency domain I thought it might be easier to discriminate against the noise using frequency analysis.

**Changes:**
  *   No fundamental architecture changes.
  *   Previous code in the Ppg methods has been replaced to implement a sample buffer with an overlap window (currently at 8 samples), frequency analysis, a bandwidth filter that covers a wider heart rate range and various small methods for peak search, averaging, etc.
  *   All floating point types are 32bit float to take advantage of the FPU.
  *   No dynamic allocations.
  *   Adds 1-2% to stack and flash. Not sure of the exact amount.
  *   Changed driver settings in HRS3300.cpp for better signal to noise at 10Hz.


**Improvements:**
  *   First HR result in as little as 6.4 seconds
  *   Updates every 800ms
  *   Good stable HR accuracy (when HR is stable compare with a predicate device and should be within +/-1bpm)
  *   Tracks HR fairly well when the rate is changing slowly (experiment by waiting for a stable HR, take a breath, hold and HR should rapidly decrease)
  *   Uses spectral averaging to increase signal to noise ratio
  *   Uses sample averaging to hang on to previous HR (about 9 second history and drops to zero when the history buffer is all zeros)
  *   Spectral and sample averaging allow the algorithm to ignore quite a few noise events without going to zero at each noise event

**Cons:**
  *   Changing HR will lag a bit behind some predicate devices
  *   It is still possible to get errant readings due to noise although these generally go away quickly as the HR signal takes over

**Pull Request:**
I think this needs to be a single pull request because all the code in the algorithm is integrated to produce a single output. Although it looks like it addresses 532 and coincidentally 1405 and 1092 below.

Applied Algorithm (Ppg::HeartRate):
The algorithm is basically similar to the same processing chains found in many papers for extracting information from PPG:

  1.  Acquire samples from the HRS3300
     *   If acquisition is reset or initial, acquire all 64 samples before processing (before going to step 2)
     *   If acquisition is not reset, acquire 8 (overlapWindow)  samples (before going to step 2)
  2.  Detrend the dataset
     *   Remove the slope
     *   Convert to the difference between each sample
     *   Turns the problematic large, DC shifts into single sample spikes
  3.  Bandpass filter the data 30 to 240Hz
     *   Removes low frequency data that can interfere with low HR detection
     *   Removes the high frequency spikes from the detrending step
  4.  Apply a window function
     *   Make the signal appear continuous to the FFT
  5.  Calculate the FFT (Fast Fourier Transform)
  6.  Calculate the power spectrum (ComplexToMagnitude())
  7.  Apply result to spectral average
     *   Helps to enhance the more frequently occurring heart rate frequency while diminishing random noise
  8.  Check signal to noise ratio (SNR)
     *   Nothing complex, just compare max to mean
  9.  Search for peaks in the spectrum only if SNR is good,
     *   Look for peaks above the peak detection threshold (currently 60% of max)
     *   If only one peak above threshold, return the peak position in peakLocation
  10. Check peak width
     *   Wide peaks can occur from noise or rapidly changing HR
     *   If peak is too wide, set peak location to 0.0
  11. Check if HR is within limits (currently 40-230bpm)
     *   If out of range set HR to 0.0
  12. If HR == 0.0, reset spectral averaging for next pass
  13. Set Ambient Light Sensor threshold
     *   Used in next pass to determine if ambient light was detected
     *   HeartRateTask will reset the acquisition if ALS threshold is exceeded
  14. Add current peak location (zero or otherwise) to Ppg::HeartRateAverage and assign the result back to peakLocation
  15. Return result based on value of peakLocation
     *   > 0.0 is valid HR
     *   -1 tells HeartRateTask to reset acquisition
     *   0.0 is no valid peak found this pass
  16. Goto 1 via HeartRateTask
 
Execution time ~20ms


Relevant feature requests:
Heart rate measurement accuracy
https://github.com/InfiniTimeOrg/InfiniTime/issues/532

Increase the heart rate data logging interval
https://github.com/InfiniTimeOrg/InfiniTime/issues/1405

Ring Buffer for Heart Rate Data
https://github.com/InfiniTimeOrg/InfiniTime/pull/1092


Potential feature requests that can be addressed in future pull requests:
Zero out HRM when stopped
https://github.com/InfiniTimeOrg/InfiniTime/issues/682

Bug: Heart rate monitoring continues while charging
https://github.com/InfiniTimeOrg/InfiniTime/issues/594

Sleep tracking
https://github.com/InfiniTimeOrg/InfiniTime/issues/307

Heart rate measurement is stops, when the screen is turned off
https://github.com/InfiniTimeOrg/InfiniTime/issues/183

Ceimour:
Reformatted pull request description. Not sure what the SOP is here, should I delete the old one and this comment as well?

minacode:
I don't think it is necessary to delete the comments. Thank you for posting it again! 😊 This will help a lot.

yehoshuapw:
I didn't quite have as much time to test this as I wanted too, so this will be brief (making the hrs better was something I never quite got to, so thanks!)

it seems to have some issue - every about 10 seconds or so, it resets back to `Not enough data`.

Ceimour:
Thanks for testing it! I know the signal gets polluted with quite a lot of noise from movement, heaving breathing, fast heart rate, etc., basically anything that adds a lot of energy. I have also noticed that I have hard time getting a reading when it is cold out and there is a large temperature difference between the top of the watch and my wrist. Anyway, if it continues to be an issue the averaging buffer size can be increased to mitigate. I hope that you and others will have time to try it for a longer period. I've only tried it on a couple of other people, so really looking forward to the feedback. We might also want to think of different instructions to the user like, "Not enough data, please remain still..." or something like that.

src/components/heartrate/Ppg.cpp
```diff
@@ -2,98 +2,242 @@
...
+  float offset = signal[0];
+  float m = (signal[size - 1] - offset) / (size - 1);
+
+  for (int i = 0; i < size; i++) {
+    signal[i] -= (m * i + offset);
+  }
+  for (int i = 0; i < size - 1; i++) {
+    signal[i] = signal[i + 1] - signal[i];
+  }
 }
```
Riksu9000, line 230: Move functions that can be made static into an anonymous namespace. That way we know they don't cause side effects.

src/components/heartrate/Ppg.h
```diff
@@ -3,29 +3,80 @@
...
+      // Heart rate Region Of Interest end (bins)
+      const int hrROIend = static_cast<int>((240.0f / 60.0f) / freqResolution + 0.5f);
+      // Minimum HR (Hz)
+      const float minHR = 40.0f / 60.0f;
+      // Maximum HR (Hz)
+      const float maxHR = 230.0f / 60.0f;
+      // Threshold for high DC level after filtering
+      const float dcThreshold = 0.5f;
+      // ALS detection factor
+      const float alsFactor = 2.0f;
```
Riksu9000, line 48: Make constants static constexpr

src/drivers/Hrs3300.cpp
```diff
@@ -26,18 +28,21 @@ void Hrs3300::Init() {
...
-  WriteRegister(static_cast<uint8_t>(Registers::Enable), 0x60);
+  // Wait time changed to zero and current to 12.5mA - Ceimour
+  WriteRegister(static_cast<uint8_t>(Registers::Enable), 0x70);
 
   // (partly) 20mA drive, power on, "magic" (datasheet says both
   // "reserved" and "set low nibble to 8" but 0xe gives better results
   // and is used by at least two other HRS3300 drivers
-  WriteRegister(static_cast<uint8_t>(Registers::PDriver), 0x6E);
+  // Current 12.5mA and more magic (low nibble 0xF) - Ceimour
+  WriteRegister(static_cast<uint8_t>(Registers::PDriver), 0x2f);
```
Riksu9000, line 38: Old comments don't need to be retained. Clearly explain what the magic values do. Maybe make them binary.

src/heartratetask/HeartRateTask.cpp
```diff
@@ -65,8 +65,17 @@ void HeartRateTask::Work() {
     }
 
     if (measurementStarted) {
-      ppg.Preprocess(static_cast<float>(heartRateSensor.ReadHrs()));
-      auto bpm = ppg.HeartRate();
+      auto start = xTaskGetTickCount();
+      auto ambient = ppg.Preprocess(static_cast<float>(heartRateSensor.ReadHrs()), heartRateSensor.ReadAls());
+      int bpm = ppg.HeartRate();
+      auto end = xTaskGetTickCount();
+      total = end - start;
```
Riksu9000, line 72: These don't seem to be used.

src/heartratetask/HeartRateTask.cpp
```diff
@@ -65,8 +65,17 @@ void HeartRateTask::Work() {
     }
 
     if (measurementStarted) {
-      ppg.Preprocess(static_cast<float>(heartRateSensor.ReadHrs()));
-      auto bpm = ppg.HeartRate();
+      auto start = xTaskGetTickCount();
+      auto ambient = ppg.Preprocess(static_cast<float>(heartRateSensor.ReadHrs()), heartRateSensor.ReadAls());
```
Riksu9000, line 69: Don't use auto for simple types.

src/components/heartrate/Ppg.cpp
```diff
@@ -2,98 +2,242 @@
...
-Ppg::Ppg()
-  : hpf {0.87033078, -1.74066156, 0.87033078, -1.72377617, 0.75754694},
-    agc {20, 0.971, 2},
-    lpf {0.11595249, 0.23190498, 0.11595249, -0.72168143, 0.18549138} {
+float Ppg::PeakSearch(const float* xVals, const float* yVals, int length, float start, float end, float th, float& width) {
+  int peaks = 0;
+  bool enabled = false;
+  float v0, v1, v2, minBin = 0.0f, maxBin = 0.0f, peakCenter = 0.0f;
+  v0 = Interpolation::CatmullSpline(xVals, yVals, length, start - 0.01f);
+  v1 = Interpolation::CatmullSpline(xVals, yVals, length, start);
```
Riksu9000, line 105: Prefer initializing while declaring.
```suggestion
  float minBin = 0.0f, maxBin = 0.0f, peakCenter = 0.0f;
  float v0 = Interpolation::CatmullSpline(xVals, yVals, length, start - 0.01f);
  float v1 = Interpolation::CatmullSpline(xVals, yVals, length, start);
```

src/components/heartrate/Ppg.cpp
```diff
@@ -2,98 +2,242 @@
...
+                              specLen,
+                              static_cast<float>(hrROIbegin),
+                              static_cast<float>(hrROIend),
+                              threshold,
+                              peakWidth);
+    peakLocation *= freqResolution;
+  }
+  // Peak too wide? (broad spectrum noise or large, rapid HR change)
+  if (peakWidth > maxPeakWidth)
+    peakLocation = 0.0f;
```
Riksu9000, line 81: Always use braces.

Riksu9000 (changes requested):
From simple testing this seems to work very well. The fact that it resets to zero when I'm moving around and presumably can't get a clean signal is giving me confidence that the reported values are more trustworthy, and the faster reporting of the heartrate is a much better experience.

Can you explain how you can get better results with less frequent data? Is more data just not necessary? Will the code work as well if the frequency was increased, or would it require tweaking?

How is the average power consumption compared to the previous code?

I made some comments on the code. The suggestions should be applied in all applicable places.

Libs should be places in the libs directory. Can you add the libs in one commit, and make the changes in another, so we can keep track of the changes? Libs also should not be reformatted.

.vscode/launch.json
```diff
@@ -44,8 +44,8 @@
         {
             "cwd": "${workspaceRoot}",
             // TODO: find better way to get latest build filename
-            "executable": "./build/src/pinetime-app-1.3.0.out",
-            "name": "Debug OpenOCD ST-LINK pinetime-app-1.3.0.out",
+            "executable": "./build/src/pinetime-app-1.11.0.out",
```
JF002, line 47: If the vscode launch script must be improved, let's do this in another PR.

src/components/heartrate/Ppg.cpp
```diff
@@ -2,98 +2,242 @@
...
-namespace {
-  int Compare(int8_t* d1, int8_t* d2, size_t count) {
-    int e = 0;
-    for (size_t i = 0; i < count; i++) {
-      auto d = d1[i] - d2[i];
-      e += d * d;
-    }
-    return e;
-  }
+// Refactored for frequency based algorithm - Ceimour
```
JF002, line 15: You don't need to document the history of the changes in the code, git takes care of that :)

src/components/heartrate/Ppg.h
```diff
@@ -3,29 +3,80 @@
...
+
+      arduinoFFT FFT;
+      // Raw ADC data
+      std::array<float, dataLength> dataHRS;
```
JF002, line 52: Just FIY : the previous implementation was mostly based on a 200B array. This improved implementation uses 4 arrays of 64 floats, which is 4 *  64 * 4 = 1024B.

This memory is statically allocated on the global stack (since those arrays are members of Ppg that is a member of HeartRateTask which is declared in the global scope in main.cpp

src/drivers/Hrs3300.cpp
JF002: Agreed, and you don't need to document the changes in the code, git does that for you.

src/heartratetask/HeartRateTask.h
```diff
@@ -26,6 +26,7 @@ namespace Pinetime {
       static void Process(void* instance);
       void StartMeasurement();
       void StopMeasurement();
+      __ULong total;
```
JF002, line 29: Prefer standard types (`uint32_t` I guess ?)

JF002 (changes requested):
I tested this PR, and it looks **very** promising. See the following video (the sealed pinetime on the left is running InfiniTime 1.11, the wired devkit on the right is running this PR):

https://user-images.githubusercontent.com/2261652/211158485-6b14d703-b43b-4c1f-b1c8-147fd9ab53c2.mp4

The value is updated more often, it detects very quickly that I removed my finger, and it doesn't sense any heart beat in the breadboard, which is nice! :+1: 
I do not have the equipment needed to check the accuracy of the measurements, though.

I think there are a few things that must be improved in the code, but I really hope we'll complete the review, validate the results and approve those changes :)

I fully agree with @Riksu9000's review : 
 - Document how this signal processing work and why it's better than the previous one (do not go into much details, but give a bit of info for those like me who do not know much about signal processing)
 - Move the 3rd party libraries in the folder `libs`.

I also added a few comments in my review. I know there are quite a few things to change, but they are mostly minor changes needed to conform with our coding standards :)

We must keep in mind memory usage : as I wrote in a comment : this implementation uses 4x time amount of RAM memory than before (from 200B to 1024+B). This seems to be fine as the compiler does not complain about overlapping regions, but we know that we are already quite tight in RAM memory.

JF002:
Here is the current measurements while running this branch:
image

When the sensor is not running, with the medium brightness, the watch use ~10-20mA (15mA avg). 
When the sensor is running, the watch draws ~12-30mA (20mA avg). So the sensor roughly draws 5mA in average.

> How is the average power consumption compared to the previous code?

The original code draws 24mA avg (12 - 40mA) when the sensor is running, so this new implementation draws 4mA less than the original code, which is nice!

src/components/heartrate/Ppg.cpp
Ceimour: Thanks for reviewing the code. The libs needed to be refactored to use float types rather than the original double types. This is done in order to take advantage of the FPU. Shall I move them into libs as is? Under a directory named signalprocessing or ?

Since we are working in frequency space we are primarily concerned with Nyquist and frequency resolution. The highest analyzed frequency is 4Hz (240bpm) so we need to acquire data at a minimum 8Hz. Frequency resolution is 1/TotalTime(seconds). I wanted to keep the total acquisition time to somewhere around 5 seconds for good UX. So I ended up at 10Hz with a 64 sample buffer resulting in a frequency resolution of 1/6.4s or 0.15625Hz, multiplied by 60 it is 9.375bpm. Yes that's poor resolution for pulse rate results! Thankfully the pulse rate frequency data has a nice Gaussian distribution and we can interpolate fairly accurate values. That coupled with pulse rate result averaging gives values that agree nicely with my off the shelf pulse oximeter. So we can certainly run at higher rates keeping in mind that we still need to acquire for at least 6.4s to keep the existing frequency resolution. It looks like it's taking ~20ms to run the existing algorithm, so the hard limit would be ~50Hz. Keep in mind that to keep the existing frequency resolution we would need to acquire for 6.4s resulting in a larger data buffer.

Ceimour: Forgot to explain why it works. We can get better results using this method because the metrics for determining good versus bad pulse rate data are simpler. We can just look for a single peak above a defined threshold in a power spectrum that has low noise characteristics (simple max to average in this case) and assume that peak is the pulse. Then an output averaging ring buffer is used to store the good pulse rate values and also store failed pulse rates as zero. When the averaging buffer becomes all zeroes, reset the acquisition as if we had just pressed the "Start" button. So basically we just look at the values we think are good results. There are times one may experience incorrect readings but they should be short lived after physically remaining still and allowing good values to fill the buffer.

src/heartratetask/HeartRateTask.cpp
```diff
@@ -29,7 +29,7 @@ void HeartRateTask::Work() {
     uint32_t delay;
     if (state == States::Running) {
       if (measurementStarted)
-        delay = 40;
+        delay = 100;
       else
         delay = 100;
```
FintasticMan, line 34: This if-statement seems unnecessary...

src/heartratetask/HeartRateTask.h
Ceimour: Thanks for the feedback! Note that because of the undefined behavior of the sensor (mainly because of the incomplete and confusing HRS3300 datasheet) the "magic" values above are experimental adjustments modifying the values that you ported over from Daniel Thomson's code. I did wonder if I should retain old comments or not because of the sensor remaining in a state of perpetual reverse engineering! Makes sense that Git will track it all.
The current data you collected is of interest to me. Is it possible for you send me the data set with the acquisition frequency or timestamps? I'm trying to understand why and how the driver oscillates after changing the low nibble in PDriver (0xE or 0xF) with wait time set to zero. The current oscillations look like a good way to measure that!

Riksu9000: Please post a new comment instead of replying to the thread about using __ULong. Code review threads will be hidden once resolved.

src/components/heartrate/Ppg.cpp
Riksu9000: Each lib should be in it's own directory. I suggest moving the libs to directories named Arduino-Interpolation and arduinoFFT according to the repo names where they're from.

src/heartratetask/HeartRateTask.cpp
Ceimour: Thanks! Fixed the magic number by using a public constant in the Ppg class.

src/heartratetask/HeartRateTask.h
Ceimour: Thanks, missed this one. I will address in the next commit.

src/components/heartrate/Ppg.cpp
Ceimour: These should be fixed in recent commits.

src/components/heartrate/Ppg.cpp
Ceimour: Fixed in recent commits. Clang-tidy comes back clean except for some complaints about short variable names.

src/heartratetask/HeartRateTask.cpp
FintasticMan: Cool! My point was more that no matter whether `measurementStarted` is true or false, `delay` is always set to 100, so the if-statement could be removed.

src/heartratetask/HeartRateTask.cpp
Ceimour: This was used to time the algorithm execution time in milliseconds. Removed.

src/heartratetask/HeartRateTask.cpp
Ceimour: Understood. Since it is possible to change the acquisition frequency in the future I thought it best to make it so there is just a single value to change. Now there is one constexpr (deltaTms) to change in Ppg.h and everything else is calculated from that. I used to have to change two things in order to update the acquisition rate. Thanks again for pointing that out, I had forgot to address that issue!

lman0:
any news ?
it's the most precise heart rate counter  since infinitme creation.
it make the pinetime nearly on par with an spo2 meter (for the heart rate part).
incomparable to the previous version!
i use it for 1 week already with no concern.

Ceimour:
@lman0 , thanks for testing this!
You are correct that movement can trigger a reset in the data acquisition. For a future improvement I have implemented an adaptive filter that will mitigate this behavior somewhat. But since this pull request is already quite large I hesitate to add it at this point.
A reset is produced when the data is noisy (caused primarily from motion, but the sensor sometimes just outputs bad data) for a long period. By long period, I mean more than 10 seconds. The reset text should say "Acquiring, please remain still..." or something like that (another future pull request). The basic philosophy is to attempt to only display the truth. Although it is averaged so it could be the true pulse rate up to 10 seconds in the past. But I think to display values older than that when bad data has been present for over ~10 seconds is misleading.

I think that in the future with wording and display changes it will provide a better user experience. It's all about user expectations!

src/components/heartrate/Ppg.cpp
Ceimour: Libs moved.

src/components/heartrate/Ppg.h
Ceimour: All constants now constexpr.

src/drivers/Hrs3300.cpp
Ceimour: Old comments removed and current comments updated.
Left the driver constants as hex for now. Let me know if you feel strongly about binary constants and I'll change them. Historically they have been hex in most other available example code.

.vscode/launch.json
Ceimour: I reverted the name but had to keep the change for the executable in order to run the debugger.

src/components/heartrate/Ppg.cpp
Ceimour: Cleaned up comments.

src/heartratetask/HeartRateTask.h
Ceimour: Removed algorithm time measurement code so "_ULong total" no longer exists.

Ceimour:
@lman0 
Thanks for sharing the perception that the pulse rate measurement going into reset while in active measurement mode seems confusing. In this recent commit I have removed that bit of code that invokes the reset while measuring.
Please give a try and let me know if it feels like the correct behavior.
Thanks again!

lman0:
@Ceimour  can you check why the action crashed?
I can't test your new commit since there us no artifact...

Ceimour:
@lman0 
I see that this repeats until exit with failure:
Retrieving check runs named Compare build size on InfiniTimeOrg/InfiniTime@25fe200768eb46bae12beffc75664defccddd6e4...
Retrieved 0 check runs named Compare build size
No completed checks named Compare build size, waiting for 10 seconds...

.vscode/launch.json
Riksu9000: You don't need to commit the change.

Riksu9000:
The PR needs to be rebased and commits that fix mistakes squashed. The libs should be added in one commit, and modified in another, so we can see the changes made.

Riksu9000:
Mark reviews as resolved when you've resolved them.

src/components/heartrate/Ppg.h
Ceimour: Thanks! In an upcoming commit I was able to shave 128 bytes off the dataHRS buffer by changing to u16.

Ceimour:
@Riksu9000 
Thanks for the instruction! 
I have rebased and now I find that I am not currently on a branch. My git skills are minimal and I am stuck as to how to complete the following command recommended by git:
git push origin HEAD:<name-of-remote-branch>

Riksu9000:
@Ceimour After rebasing you should still be on the same local branch. Because the history of the local and remote branches are different, you have to use the `--force` flag when pushing.

Ceimour:
@Riksu9000 Thanks again for the help! I seem to be missing something (new?) related to build size.
I get 1 failing check related to this message:
"Retrieved 0 check runs named Compare build size"

Riksu9000:
You haven't fixed the conflicts. You need to rebase **on top of latest develop**. Also the commits aren't clean. Libs are added twice and removed once. The formatting of the libs should be left original.

Ceimour:
@Riksu9000 OK, I think I understand now. Sorry if my confusion is causing you additional and unnecessary work. I know you are quite busy and I really appreciate your feedback and instruction.

lman0:
why did you have closed the pull  @Ceimour ?
why not make it draft ?

Ceimour:
@Riksu9000  I did the update. I assumed I had to run Sync to update to the latest develop but that seemed to automatically close the PR. Please let me know if I need to do anything to fix this. Thanks.

Ceimour:
@lman0 Sorry about that, I'm clearly not getting git! Waiting for help.

lman0:
@Ceimour  https://gist.github.com/Farhan-Haseeb/fd4cbb476de4e2931c322d7bacb75e5f
how to reopen pr

lman0:
> @lman0 Sorry about that, I'm clearly not getting git! Waiting for help.
this inside this link is explained how reopen pr (scroll for picture)
https://gist.github.com/Farhan-Haseeb/fd4cbb476de4e2931c322d7bacb75e5f

Ceimour:
@lman0 I was instructed by @Riksu9000 to force push and that appears to have closed the PR. It looks like I could add commits and then it would enable the Reopen but I will wait for further instruction. If I did something wrong in the rebase force push cycle, I just want to make sure I don't do any more damage!

Avamander:
It closed because there are no new commits in this branch.

Ceimour:
@Avamander Thanks for pointing me in the right direction! When I rebased it somehow detached from the branch and the PR appeared to close after pushing. So I went ahead and cloned and was restored to the feature branch. I then made a minor change, committed and pushed. So far I still have the disabled "Reopen" button here. Is there another method of reopening this PR? I see an option in the main Pull requests tab that recognizes that I have a recent push, but it appears that the only option is to create a PR (nothing indicating reopening).

lman0:
I have stumbled on this , with git command https://gist.github.com/robertpainsi/2c42c15f1ce6dab03a0675348edd4e2c

Ceimour:
@lman0 Thanks for the link, looks like that will work! Still waiting to see if it is okay to proceed.
Thanks again!

JF002:
@Ceimour Thanks for your follow-up on this PR, and I'm really sorry you encounter those issues with git... rebasing can be very veeeery confusing sometime. 

As of now, the PR appears empty:
image

You might want to wait for @Riksu9000 instructions (he's great with those advanced git functionalities), but it looks like the last commit you pushed is 049fbba, which is the last commit on develop. Since your branch contains all your changes, it should still be possible to fix this PR:
 - checkout `develop` and pull all changes (`git checkout origin/develop` and `git pull origin develop`)
 - checkout your branch (`git checkout heartratefreq`)
 - rebase your branch on develop (`git rebase develop`)

When I tried this on my side, there were a few conflicts that needed to be merged (probably because we applied changes to some files after to created your branch). You'll need to fix those conflicts, and then commit and (force)push again your branch to update this PR.

Ceimour:
@JF002 Thanks! Now I think I see where I went wrong. I will follow your suggestions and try again.

Ceimour:
@JF002 Thanks again for the instruction. I did as you suggested and as the last step from within branch heartratefreq ran "git push --force". All completed without error. Is there anything else I need to do to get it to reopen?

Ceimour:
@JF002 I was able to reopen. The reopen button appeared to be greyed out but I tried it and it worked. How should I address the Merge prompt? Update with merge or update with rebase?

.vscode/launch.json
Ceimour: Change is reverted.

pentamassiv:
Very nice PR! I believe what Riksu9000 wants you to do is to clean up the commit history. Right now the PR contains 8 commits. Some of the commits revert accidental changes you made or other small changes. If this was merged as it is, all these commits would get merged as well.
If somebody would investigate a bug and has a look at the last commits, this can be confusing. Git allows you to change old commits (the history). You can [change the old commits](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) so that it looks like you created a perfect PR on the first try.

Here is what you could do:
Change the history of your last 8 commits:
`git rebase -i HEAD~8`
Running this command gives you a list of commits in your text editor. If you rearrange or change those lines, you are changing the history of your commits. If you change it to the following:

```
pick 4a30070a Refactored Ppg for frequency based algorithm. Changed driver settings in HRS3300.cpp fo>
fixup 4ed64f5d Reverted Interpolation and arduinoFFT libraries back to author's original in order to >
fixup ffc7ebfc Reverted .vscode to develop 361e381ac32.
fixup 03855831 Moved Arduino libraries to src/libs.
pick 1c8aefc6 Arduino libraries InterpolationLib and arduinoFFT ported to InfiniTime. Changed all dou>
pick bcb1bb50 Fixed arduinoFFT.cpp constants are double, should be float. Modified Ppg.cpp to use ano>
fixup 187df497 Changed magic number to use bpm variable for readability.
fixup 5b7afaa5 Fixed unintended merge from original.
```

The order of the commits will be changed. Additionally the commits where the line starts with "fixup" will be squashed with the commit from the line above.

You might have to rebase it to the newest "develop" branch by then. You can do that by running:
`git rebase origin/develop`
If there were any merge conflicts, you have to resolve them.

Because you changed the history, you will have to "force push" to the repo. That means you have to overwrite your old commits. You can do that by running:
`git push --force`

I hope that helps you a little bit

Ceimour:
@pentamassiv Thanks for the help! I was finally able to get the rebase done but there is some issue with a post build routine looking for "Compare build size" and it always fails. Hopefully I will work this out soon. Thanks again.

lman0:
hi 
i have see that the simulator is broken @Ceimour  with you pul
maybe you need to do a pr for the new lib you here as well ...
what do you think @NeroBurner  ?

JF002:
@Ceimour be reassured, I do not forget this PR!

I was a bit surprised by the additional flash memory usage of this PR : +10KB. The changes in this PR alone do not explain this big size increase.

So I had a look at the `.map` file : 
 - ArduinoFFT adds 1228KB
 - InterpolationLib add 498KB

So were do those ~8300B come from? I compared map files from `main` and from this PR and found new entries from libgcc.a and libm.a : 

- libm.a : 5188B
- libgcc.a : 2108B

which account for 7296B.

The symboles are

- in libgcc.a
  - _aeabi_ldivmod.o 160B
  - _udivmoddi4.o 720B
  - _arm_addsubdf3.o 888B
  - _arm_muldivdf3.o 1060B

- in libm.a
  - __kernel_rem_pio2f 1652B
  - __ieee754_powf 1576B
  - two_over_pi 792B
  - __ieee754_rem_pio2f 612B
  - __kernel_cosf 260B
  - scalbnf 228B
  - powf 188B
  - ...

So mostly functions and variables related to float arithmetic. And indeed, in `arduinoFFT.cpp` we can find calls to `sqrtf()`, `powf()`, `cosf()`,...

Do we have any alternatives here? Could those calculations be done in integer, for example, to avoid linking with those heavy symbols? 
I'm not saying that we'll refuse this PR because of this, we really want to keep the memory usage (both ram and flash) under control!

Ceimour:
@JF002 Thanks for the analysis. I was also shocked to see the amount of flash taken by this! I believe I can simplify things to get the size down.
I'll let you know when I push changes. Thanks again!

Ceimour:
@JF002 I was able to get rid of the trig and other functions that were causing the library to be linked. At least I assume that was the case since the size reduced by ~8K. Let me know if I need to fix the rebase. Wasn't sure what to do with the other commits so I left then in there as 'pick'.
Thanks!

Ceimour:
@JF002 Made a mistake on previous push, forgot to rebase on main. This one should be good.

JF002:
@Ceimour Amazing job :1st_place_medal: , thank you! 

I've just checked the flash memory usage : these changes add a bit less than 2KB, which seems much more reasonable than before !

Regarding RAM usage, I couldn't find any dynamic allocation (which is nice), and the whole HeartRateTask class uses less than 1KB. It's probably more than the previous implementation, but these improvements are probably worth it.

This PR looks good to me. I'll ask @Riksu9000 and @FintasticMan to check whether all their conversations (review) are now solved :)

src/drivers/Hrs3300.cpp
Riksu9000: Old comments are still retained, so there are contradictory comments.

src/libs/arduinoFFT/arduinoFFT.cpp
```diff
@@ -0,0 +1,510 @@
...
+// 		if (dir == FFT_FORWARD) {
+// 			this->_vReal[i] *= weighingFactor;
+// 			this->_vReal[this->_samples - (i + 1)] *= weighingFactor;
+// 		}
+// 		else {
+// 			this->_vReal[i] /= weighingFactor;
+// 			this->_vReal[this->_samples - (i + 1)] /= weighingFactor;
+// 		}
+// 	}
+// }
```
Riksu9000, line 286: We shouldn't merge big blocks of commented code. Are you sure they need to be commented out to reduce the binary size? Wouldn't they just be removed by compiler optimizations?

src/libs/arduinoFFT/arduinoFFT.cpp
Riksu9000, line 1: https://github.com/kosme/arduinoFFT/blob/develop/src/arduinoFFT.h There's a develop version of this library, which is templated to allow using float instead of double. Can we use that without modification?

src/components/heartrate/Ppg.cpp
```diff
@@ -2,106 +2,276 @@
...
-  auto spl_int = static_cast<int8_t>(spl);
+  // Hanning Coefficients
+  // Note: Harcoded and must be updated if dataLength changes.
+  // This data is symetrical so just using the first half (saves 128B when dataLength is 64).
+  static constexpr float hanning[Ppg::dataLength >> 1] {
+    0.0f,        0.00248461f, 0.00991376f, 0.0222136f,  0.03926189f, 0.06088921f, 0.08688061f, 0.11697778f,
+    0.15088159f, 0.1882551f,  0.22872687f, 0.27189467f, 0.31732949f, 0.36457977f, 0.41317591f, 0.46263495f,
+    0.51246535f, 0.56217185f, 0.61126047f, 0.65924333f, 0.70564355f, 0.75f,       0.79187184f, 0.83084292f,
+    0.86652594f, 0.89856625f, 0.92664544f, 0.95048443f, 0.96984631f, 0.98453864f, 0.99441541f, 0.99937846f};
+}
```
Riksu9000, line 125: Where do these come from? If they're precalculated results of some calculation, can it be created using constexpr?

src/components/heartrate/Ppg.cpp
```diff
@@ -2,106 +2,276 @@
   SPDX-License-Identifier: LGPL-3.0-or-later
   Original work Copyright (C) 2020 Daniel Thompson
   C++ port Copyright (C) 2021 Jean-François Milants
+  Modified to use frequency space pulse rate metrics and ALS sensor monitoring Copyright (C) 2022 Ceimour
```
Riksu9000, line 5: I wouldn't call this modified, when there's nothing original left. This makes me wonder if this should be a new file, so it's not confused with the old one.

Ceimour:
@JF002 Thanks!

Riksu9000 (commented):
The last commit contains fixes for the first commit, which should be applied to the first commit to meet the commit requirements, however the entire commit should not necessarily be squashed. Ideally each commit would contain much fewer changes, because the descriptions are *long*.

src/libs/arduinoFFT/arduinoFFT.cpp
JF002: I've just checked : un-commenting those functions do not change the final binary size. The compiler/optimizer are smart enough to remove those functions at build time. So, yes, no need to comment them.

src/components/heartrate/Ppg.cpp
JF002: Yep, this PR does not keep any of the original code of ppg.c/h. How should that be handled regarding the license and copyrights?

Riksu9000: There's not really any licensed code left, so maybe it isn't applicable? Unless we can come up with a new name for this file, we could remove this file in one commit and add a new file with the same name in another.

src/libs/arduinoFFT/arduinoFFT.cpp
JF002: Interesting. This would allow us to use a non-modified version of the library, right?

src/components/heartrate/Ppg.cpp
JF002: So we could either remove the copyright notice or remove the old file and add the new file with the same name, then.

src/libs/arduinoFFT/arduinoFFT.cpp
Ceimour: I like that! I'll try that out. Thanks!

src/components/heartrate/Ppg.cpp
FintasticMan: Given that none of the original code is retained, the copyright notice should be removed in the commit that removes the last bit of the original code.

FintasticMan:
Other than the points Riksu has raised, I have no issues with this code.

src/components/heartrate/Ppg.cpp
Ceimour: @JF002 Licensing and copyright removed.

src/components/heartrate/Ppg.cpp
Ceimour: @Riksu9000 I tried to use constexpr to calculate but it would not allow cos() which is needed for the calculation. Maybe it will be allowed in the future? If so, I will update to more elegant and versatile constexpr technique.

FintasticMan: Some mathematical functions are made `constexpr` with C++23, but that does not include `cos()` or `sin()`.

src/libs/arduinoFFT/arduinoFFT.cpp
Ceimour: @Riksu9000 Thanks! I wish I had seen that earlier. I replaced the old lib with the template version and it seems to work great! I did make on change within the arduinoFFT.h file changing "#define sqrt_internal sqrt" to "#define sqrt_internal sqrtf" which results in ~3KB less flash storage.

src/libs/arduinoFFT/arduinoFFT.cpp
Ceimour: @JF002 This library was removed in favor of the template version that @Riksu9000 found.

src/libs/arduinoFFT/arduinoFFT.cpp
FintasticMan: I think that we don't need a very high level of precision for these calculations, so I think it might be worth testing the speed and space impacts of using the `FFT_SPEED_OVER_PRECISION` and `FFT_SQRT_APPROXIMATION` options. That would eliminate the need for the sqrt_internal change and we could perhaps include the library as a submodule.

src/CMakeLists.txt
```diff
@@ -666,9 +663,10 @@ set(INCLUDE_FILES
         drivers/TwiMaster.h
         heartratetask/HeartRateTask.h
         components/heartrate/Ppg.h
-        components/heartrate/Biquad.h
-        components/heartrate/Ptagc.h
         components/heartrate/HeartRateController.h
+	libs/arduinoFFT-develop/src/arduinoFFT.h
+	libs/arduinoFFT-develop/src/defs.h
+	libs/arduinoFFT-develop/src/types.h
```
FintasticMan, line 669: The indentation here uses a tab, rather than spaces.

```suggestion
        libs/arduinoFFT-develop/src/arduinoFFT.h
        libs/arduinoFFT-develop/src/defs.h
        libs/arduinoFFT-develop/src/types.h
```

src/heartratetask/HeartRateTask.cpp
```diff
@@ -85,21 +101,23 @@ void HeartRateTask::Work() {
 }
 
 void HeartRateTask::PushMessage(HeartRateTask::Messages msg) {
-  BaseType_t xHigherPriorityTaskWoken = pdFALSE;
+  BaseType_t xHigherPriorityTaskWoken;
+  xHigherPriorityTaskWoken = pdFALSE;
```
FintasticMan, line 105: I don't think this change was intentional.

FintasticMan (changes requested):

src/heartratetask/HeartRateTask.cpp
Ceimour: @FintasticMan This was present when I started this PR. @JF002 was this done during the WaspOS port? What should be done with this code?

src/CMakeLists.txt
Ceimour: @FintasticMan Thanks! I will fix.

src/heartratetask/HeartRateTask.cpp
FintasticMan: I was talking about the change from `BaseType_t xHigherPriorityTaskWoken = pdFALSE;` on one line to these 2 lines.

src/libs/arduinoFFT/arduinoFFT.cpp
Ceimour: @FintasticMan I did test using FFT_SPEED_OVER_PRECISION and it increased the flash storage ~2KB. The most economical solution was to set sqrtf. I really wish there was an #ifndef for sqrt_internal! Let me know if the extra ~2KB is tolerable and I can switch over to FFT_SPEED_OVER_PRECISION. I observed no difference in accuracy when I tested it.

src/heartratetask/HeartRateTask.cpp
Ceimour: @FintasticMan Got it. Not sure why I changed it. Most likely a clang-tidy misunderstanding on my part.

src/libs/arduinoFFT/arduinoFFT.cpp
FintasticMan: Have you also tried `FFT_SQRT_APPROXIMATION`?

FintasticMan: I've just done some preliminary testing, only looking at the flash usage with the different options.

| options enabled | FLASH usage |
| --------------- | ----------- |
| neither (`sqrtf`) | 407624B |
| `FFT_SPEED_OVER_PRECISION` (`sqrtf`) | 407624B |
| neither (`sqrt`) | 410616B |
| `FFT_SPEED_OVER_PRECISION` (`sqrt`) | 410616B |
| `FFT_SQRT_APPROXIMATION` | 409352B |
| both | 409352B |

From this, I think that we should be using `FFT_SPEED_OVER_PRECISION` but not `FFT_SQRT_APPROXIMATION`, with the `sqrtf` function for square roots. The speed over precision option doesn't impact the flash usage at all, and uses some optimisations that increase the speed. The square root approximation probably would be faster, but because it needs to use a new function, it adds to the space usage.

Ceimour: @FintasticMan Applied your suggestion and tested. The accuracy seems unchanged.

src/components/heartrate/Ppg.cpp
Ceimour: Left it "as is" for now. Note that changing dataLength will break it in the compiler, so should be low risk of future developer error.

src/libs/arduinoFFT/arduinoFFT.cpp
FintasticMan: Nice one! I think this PR is good to be merged then.

I have opened a pull request to the FFT library for `sqrt_internal` to be wrapped in an `ifndef`, so that we don't have to modify the library: kosme/arduinoFFT#83. I think we shouldn't wait for that to get merged, and address putting the library into a git submodule in a future PR.

FintasticMan (approved):
This looks good to me! I've opened a PR to the FFT library that would allow us to specify the square root function to use without modifying the library: kosme/arduinoFFT#83. I don't think we should wait for that before merging this. We can update the library and put it in a git submodule in a later PR.

src/libs/arduinoFFT/arduinoFFT.cpp
Ceimour: Thanks for your time and advice!

JF002 (approved):

JF002:
@Ceimour Thank you very much for your work and also for you patience and courage while we reviewed this PR!
I'm really happy to finally merge those changes which greatly improve the measurements and usability of the HR sensor in InfiniTime!

Ceimour:
@JF002 Thanks! And thanks to everyone who tested and helped guide me through the process. You all made the code and implementation much better!

Anasdec:
I noticed the HR readings drops a lot to zero especially during motion. I was wondering if you have got the time to implement the adaptive filtering above? 

Otherwise, is changing some of the parameters such as peakDetectionThreshold, maxPeakWidth, and signalToNoiseThreshold would help? 

Thanks.

Ceimour:
@Anasdec  Thanks for your interest in in improving this. Adaptive filtering unfortunately would require the use of the math libraries mentioned above (adds ~8KB). Not sure it would help that much since the fundamental issue is the poor signal quality from the sensor. I came to the conclusion that the ONLY use case for this current implementation is stationary, spot measurements. I still hold out hope that someone out there will find a way to get better signal to noise from the sensor.

You may be able to get a better motion related experience by increasing the spectrum averaging and/or results averaging. The downside is more averaging, more lag in the reading. 

If you are really keen on playing with peak and signal metrics, I recommend using a dev unit wired to a debugger such that the watch heart rate sensor can be in contact with the wrist. Then make a small change and observe the effect. Iterate to get desired end result. Warning, it will take up a lot of your time!

Anasdec:
@Ceimour  Thanks for coming back. I'm keen to look into this, but I don't have much experience. 

To increase the spectrum averaging as you suggest. Does that mean I need to do the following :

Change: datalength from 64 to 128 and the Hanning Coefficients (via python - numpy.hanning(128) and only use the first 64 values)? 

am I correct?

Ceimour:
@Anasdec 
In Ppg.h:

To increase spectrum averaging adjust:
static constexpr uint16_t spectralAvgMax = 2;

To increase heart rate result history, adjust the array size as you mentioned above:
std::array<float, 20> dataAverage;

Note that as the spectrum average is increased, the peak may spread out (peak width) as the heart rate varies. If you find that you are suddenly unable to get a result, the max peak width may need to be increased:
static constexpr float maxPeakWidth = 2.5f;

Cheers!

PR #1494, opened 2022-12-26
Improve discharge curve values

FintasticMan:
This pull request modifies the values of the discharge curve to slightly more accurately estimate the battery discharge. To calculate these values, I have run different tests on my 3 PineTimes, and then used some code to find the best fit. I can provide the raw data from the watches as well as the code I used on request.

fit

straight

In this last graph, the blue line is the original linear algorithm, the red line is with AlexXZero's values, and the yellow line is with my values.

### DIFF
--- a/src/components/battery/BatteryController.cpp
+++ b/src/components/battery/BatteryController.cpp
@@ -61,14 +61,8 @@ void Battery::SaadcInit() {
 }
 
 void Battery::SaadcEventHandler(nrfx_saadc_evt_t const* p_event) {
-  static const Utility::LinearApproximation<uint16_t, uint8_t, 6> aprox {{{
-    {3500, 0},  // Minimum voltage before shutdown (depends on the battery)
-    {3600, 10}, // Keen point that corresponds to 10%
-    {3700, 25},
-    {3750, 50},
-    {3900, 75},
-    {4180, 100} // Maximum voltage during charging is 4.21V
-  }}};
+  static const Utility::LinearApproximation<uint16_t, uint8_t, 6> approx {
+    {{{3500, 0}, {3616, 3}, {3723, 22}, {3776, 48}, {3979, 79}, {4180, 100}}}};
 
   if (p_event->type == NRFX_SAADC_EVT_DONE) {
 
@@ -83,7 +77,8 @@ void Battery::SaadcEventHandler(nrfx_saadc_evt_t const* p_event) {
 
     uint8_t newPercent = 100;
     if (!isFull) {
-      newPercent = std::min(aprox.GetValue(voltage), isCharging ? uint8_t {99} : uint8_t {100});
+      // max. voltage while charging is higher than when discharging
+      newPercent = std::min(approx.GetValue(voltage), isCharging ? uint8_t {99} : uint8_t {100});
     }
 
     if (isPowerPresent) {


Riksu9000:
The first chart is a bit confusing. Can you create a chart of your test results with voltage on the vertical axis and time on the horizontal? I'm guessing the most significant improvement comes from the program you used to find the best fitting values, what do you think?

FintasticMan:
I agree that the first chart is confusing. The vertical axis is the "desired reported charge", based on the elapsed time since the measurement started. It is the time, normalised between 0 and 100, and then inverted (0 s elapsed should correspond to 100% remaining). I realise now that it makes more sense for it to be on the horizontal axis. Each line represents the theoretical best mapping of voltage to charge, for each test. This way, you can compare if a given approximation fits the lines well.

I can make a chart with voltage plotted against the time, but I'm not sure how useful it would be. This is because I then won't be able to incorporate how my values would change the reported charge. The time elapsed is also not very comparable between runs of the test, because I ran the tests each with different constant loads.

The program I used to calculate the values is quite a simplistic one, it just brute-forces all possible values and chooses the one that fits best. This is very slow (it took almost 6 hours using all 16 threads on my PC) but I can be certain that it definitely doesn't miss anything.

FintasticMan:
I agree that the first chart is confusing. The vertical axis is the "desired reported charge", based on the elapsed time since the measurement started. It is the time, normalised between 0 and 100, and then inverted (0 s elapsed should correspond to 100% remaining). I realise now that it makes more sense for it to be on the horizontal axis. Each line represents the theoretical best mapping of voltage to charge, for each test. This way, you can compare if a given approximation fits the lines well.

I can make a chart with voltage plotted against the time, but I'm not sure how useful it would be. This is because I then won't be able to incorporate how my values would change the reported charge. The time elapsed is also not very comparable between runs of the test, because I ran the tests each with different constant loads.

The program I used to calculate the values is quite a simplistic one, it just brute-forces all possible values and chooses the one that fits best. This is very slow (it took almost 6 hours using all 16 threads on my PC) but I can be certain that it definitely doesn't miss anything.

FintasticMan:
I can't figure out a good graph that shows everything neatly. I have uploaded my results to a GitHub Gist, along with a writeup on how I got the data. It can be found [here](https://gist.github.com/FintasticMan/72d3aa1fc5c7cb9739f45b81197836c7).

FintasticMan:
@Riksu9000 have you had any time to look over my results? I think this is a definite improvement, and would love to get that verified.

Riksu9000:
Everything looks good to me, but I still want to test this on a watch before approving it.

src/components/battery/BatteryController.cpp
```diff
@@ -61,12 +61,12 @@ void Battery::SaadcInit() {
...
-    {3700, 25},
-    {3750, 50},
-    {3900, 75},
+  static const Utility::LinearApproximation<uint16_t, uint8_t, 6> approx {{{
+    {3500, 0}, // Minimum voltage before shutdown (depends on the battery)
+    {3616, 3},
+    {3723, 22},
+    {3776, 48},
+    {3979, 79},
     {4180, 100} // Maximum voltage during charging is 4.21V
```
Riksu9000, line 70: Let's remove the comments, because they're not accurate.

FintasticMan: I agree about the first one, but I think it's useful to have some info about the max charge voltage being different than the voltage at 100% when not charging.

During charging, I saw 4.20V on one PT, 4.21V on another, and then 4.25V on my 3rd.

Riksu9000: You can add a comment outside of the values that says the voltage is higher during charging, although it's normal behaviour for battery charging.

Riksu9000 (approved):

JF002:
@Riksu9000 I presume you had the opportunity to check this new discharge curve on an actual device? Do you confirm @FintasticMan 's observations?

Riksu9000:
Yes, I've been running these changes for a while. The difference in practice isn't enormous, but I feel it's a bit better below 50%, and I trust the provided data, which shows the reported charge being overall closer to the real charge.

JF002 (approved):
Great, thanks! I guess I can trust both of you, then :)

PR #1497, opened 2022-12-28
Improve devcontainer

emosenkis:
- Installs lv_font_conv and cbors
- Installs sudo and adds user to sudo group in case they need to make changes later

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -20,6 +20,8 @@ RUN apt-get update -qq \
       clang-tidy \
       locales \
       libncurses5 \
+      npm \
+      sudo \
 # aarch64 packages
       libffi-dev \
       libssl-dev \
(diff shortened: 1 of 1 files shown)


Riksu9000:
I don't think the currently active maintainers use the devcontainer, so you'll have to be more descriptive about the purpose of these changes, so we can properly review them.

FintasticMan:
From reading the docs, I think it should be possible to adapt the devcontainer to use the main Dockerfile, which will reduce code duplication and maintenance cost. I don't use VS Code myself though, so I don't think I can help too much with that. If someone who is more familiar with devcontainers can shed some light, that would be appreciated.

BloodStainedCrow:
You are totally correct. We currently seem to have two almost identical copies of Dockerfile and build.sh script in /.devcontainer and /docker. The differences seem to be:
- Installing locales and formatting tools like clang-tidy in devcontainer
- installing all required build components in docker (That should happen in .devcontainer anyway, see o.p.)
- adding an additional user in devcontainer for vscode to connect to
- in devcontainer the get scripts from build.sh are explicit instead of calling build.sh

None of these should be a problem to unify these two Dockerfiles. I would propose to add a argument so that the parts only necessary for devcontainer are only installed when needed. 
On the same note I would propose unifying the two build.sh scripts in /.devcontainer and /docker for that someone else will have to have a look at them to determine which is correct since they have weird differences like gcc-version where I do not know which is correct.

BloodStainedCrow:
If any currently active maintainers use docker for compilation I would assume everything used there is more accurate/up to date.

FintasticMan:
The main Docker image is up-to-date, and is used for the GitHub CI.

.devcontainer/Dockerfile
```diff
@@ -58,7 +60,11 @@ RUN rm /tmp/nRF5_SDK_15.3.0_59ac345
 RUN git clone https://github.com/mcu-tools/mcuboot.git
 RUN pip3 install -r ./mcuboot/scripts/requirements.txt
 
+RUN npm install -g lv_font_conv
+RUN pip install cbor  # Upstream mcuboot has moved on to cbor2
```
jmanning2k, line 64: `pip3` for consistency with above.

PR #1498, opened 2022-12-28
Count up, set background red, and ring when timer completes.

### LINKED ISSUE #1374: Have timer go negative after it hits 0:00
### Verification

### Pitch us your idea!

I'd like the timer to go negative so I know how long it has been since it hit 0:00

### Description

It would be nice that after the timer hits 0:00, a separate screen shows up with the timer counting negative to let you know how long it has been since the timer hit 0:00.

emosenkis:
InfiniSim_2022-12-28_222533

Fixes #1374
Fresh implementation of similar features of #557


emosenkis:
Closing in favor of #1499 since the `/` in my branch name broke CI

PR #1501, opened 2022-12-30
Enable unused parameter warning. Fix warnings.

Riksu9000:
Unused parameters make the code confusing and intimidating. They can also cause bugs, like in the case of MotionService. `MotionService::OnNewMotionValues(x, y, z)` is called before updating the values inside MotionController, but these parameters are ignored, instead using the values from `MotionController`, which haven't been updated yet.

DebugPins was removed, since it's not used anywhere and was causing a warning.

Include directories from other projects were marked `SYSTEM`, to suppress their warnings.


src/CMakeLists.txt
```diff
@@ -779,7 +778,7 @@ link_directories(
 
 
 set(COMMON_FLAGS -MP -MD -mthumb -mabi=aapcs -ftree-vrp -ffunction-sections -fdata-sections -fno-strict-aliasing -fno-builtin -fshort-enums -mcpu=cortex-m4 -mfloat-abi=hard -mfpu=fpv4-sp-d16 -fstack-usage -fno-exceptions -fno-non-call-exceptions)
-set(WARNING_FLAGS -Wall -Wextra -Warray-bounds=2 -Wformat=2 -Wformat-overflow=2 -Wformat-truncation=2 -Wformat-nonliteral -Wno-unused-parameter -Wno-missing-field-initializers -Wno-unknown-pragmas -Wno-expansion-to-defined -Wreturn-type -Werror=return-type)
```
Avamander, line 782: Why this?

Riksu9000: What do you mean? Is there a problem with enabling the warning?

Avamander: I think there are places where it generated a lot of noise, e.g. in NimBLE bindings, does it not?

JF002 (approved):
> DebugPins was removed, since it's not used anywhere and was causing a warning.

Right! I used them when I was mostly working on a devboard, but I don't use this class anymore, and I doubt anyone does.

> They can also cause bugs, like in the case of MotionService.

Oof! Good catch! 

I think this is a good idea to explictly ignore parameters like you did in this PR, and the other code cleanings are more than welcome, thanks!

PR #1502, opened 2022-12-31
Revert "Enable CMake unity build"

Riksu9000:
Reverts InfiniTimeOrg/InfiniTime#1395

The CMake unity build can sometimes cause the build to fail, so I don't think this feature is worth maintaining.

https://github.com/InfiniTimeOrg/InfiniTime/actions/runs/3809205628/jobs/6480398882


JF002 (approved):
I confirm that Unity build breaks many builds. I don't think the minor build-time performances improvements are worth the time needed to debug those issues... Let's disable it.

PR #1276, opened 2022-08-11
colored battery icon "juice"

mashuptwice:
Battery status icon "juice" colored depending on the reported battery percentage. 
This is a first attempt to introduce a few colored elements into the UI, in accordance with some of the reactions to #1265 

```
>=60 GREEN
>40 && <60 ORANGE
<=40 RED
```

6pnwi3

PR #1265, opened 2022-08-03
Improve status icon visibility

Riksu9000:
Making the icons white Improves the visibility, but also makes them more subdued instead of attention grabbing. I was planning on making this change regardless of #1263.

Relates to #958. One of the necessary changes for #1263.

InfiniSim_2022-08-03_092844
InfiniSim_2022-08-03_092845
InfiniSim_2022-08-03_092847
InfiniSim_2022-08-03_092851

PR #1503, opened 2022-12-31
Revert low battery alert. Add low battery indicator.

Riksu9000:
This reverts PR https://github.com/InfiniTimeOrg/InfiniTime/pull/1352

There's no need to break the focus of the user to alert them that the battery will be empty in 24h, when they can see that the next time they check the watch. I've made the battery icon turn orange at 15, and red at 5, for a stronger warning to charge the watch.

InfiniSim_2022-12-31_080939

### DIFF
--- a/src/displayapp/screens/WatchFaceAnalog.cpp
+++ b/src/displayapp/screens/WatchFaceAnalog.cpp
@@ -49,6 +49,7 @@ WatchFaceAnalog::WatchFaceAnalog(Controllers::DateTime& dateTimeController,
                                  Controllers::NotificationManager& notificationManager,
                                  Controllers::Settings& settingsController)
   : currentDateTime {{}},
+    batteryIcon(true),
     dateTimeController {dateTimeController},
     batteryController {batteryController},
     bleController {bleController},
--- a/src/components/battery/BatteryController.cpp
+++ b/src/components/battery/BatteryController.cpp
@@ -81,20 +81,10 @@ void Battery::SaadcEventHandler(nrfx_saadc_evt_t const* p_event) {
       newPercent = std::min(approx.GetValue(voltage), isCharging ? uint8_t {99} : uint8_t {100});
     }
 
-    if (isPowerPresent) {
-      batteryLowNotified = false;
-    }
-
     if ((isPowerPresent && newPercent > percentRemaining) || (!isPowerPresent && newPercent < percentRemaining) || firstMeasurement) {
       firstMeasurement = false;
       percentRemaining = newPercent;
       systemTask->PushMessage(System::Messages::BatteryPercentageUpdated);
-
-      // warn about low battery when not charging and below threshold
-      if (BatteryIsLow() && !isPowerPresent && !batteryLowNotified) {
-        systemTask->PushMessage(System::Messages::LowBattery);
-        batteryLowNotified = true;
-      }
     }
 
     nrfx_saadc_uninit();
--- a/src/components/battery/BatteryController.h
+++ b/src/components/battery/BatteryController.h
@@ -18,10 +18,6 @@ namespace Pinetime {
         return percentRemaining;
       }
 
-      bool BatteryIsLow() const {
-        return percentRemaining <= lowBatteryThreshold;
-      }
-
       uint16_t Voltage() const {
         return voltage;
       }
@@ -43,7 +39,6 @@ namespace Pinetime {
       static constexpr nrf_saadc_input_t batteryVoltageAdcInput = NRF_SAADC_INPUT_AIN7;
       uint16_t voltage = 0;
       uint8_t percentRemaining = 0;
-      bool batteryLowNotified = false;
 
       bool isFull = false;
       bool isCharging = false;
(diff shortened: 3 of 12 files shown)


minacode:
Obviously I don't like the reversion. But I like the coloring of the icon.

Low battery is one the few things every device alerts about. Companion apps implemented this feature before I did for the watch, so I'm sure some people wanted it. I do. I want to be actively notified and not having to look often enough by myself.

The loss of focus would be approximately once per week if you don't charge often enough.

Regarding the 24h, we could just change the threshold to a lower value. But people may not carry their charging device for the watch with them and are only able to charge once per day (e.g. during the night). We could also add a second "critical" threshold like you did to warn a second time.

Riksu9000:
It's a bit different when my laptop warns me that there's 20 minutes remaining. In this case this isn't something that needs immediate attention, so there's no reason to grab attention immediately. This color change makes it so you don't have to try and count the number of pixels, but you know exactly when battery is below 15%, just like with the notification. I'm assuming you'll check the watch often enough if you're wearing it.

Also people most likely cannot charge the watch immediately when the notification comes, in which case they must ignore the alert and possibly forget about it. The notification is saved, but there's nothing to remind about that either.

minacode:
What about a notification at 5% or maybe even 1%. I would at least take it as a "be careful now and maybe don't use the flashlight anymore". We can go more for the 20-minutes scenario with your visual clues present.

Also, we should keep the low battery functionality besides the notification, because we could do more things with it like lowering the screen brightness or turning off heart rate measurement for example.

It's a bit sad to work on a PR for a time and have a discussion and then get it completely "reverted" by a maintainer a couple of days after it got merged. I know that I have no rights for my code to be included, but it doesn't feel good. You even changed a clean function call to a random magic number in `BatteryInfo.cpp` line 62.

Riksu9000:
> I would at least take it as a "be careful now and maybe don't use the flashlight anymore"

You would still see the battery icon.

I used tools to revert the commits of the PR. I made no changes myself.

I think the alert was mainly inspired by the poor percentage calculation that was present at the time, which caused the watch to power off unexpectedly. This however shouldn't be an issue anymore with the new approximation.

The watch will keep working even after reaching 0%. What I've had in mind is to enter a very low power mode after reaching 0%.

I don't think all options were thoroughly considered before merging the PR and I'm sorry about how this happened. Let's discuss. @InfiniTimeOrg/core-developers.

minacode:
> What I've had in mind is to enter a very low power mode after reaching 0%.

Maybe this is something we should notify? Especially if the watch turns off features that people rely on (I don't know what you have in mind).

> I don't think all options were thoroughly considered before merging the PR and I'm sorry about how this happened. Let's discuss.

It's in my interest to do what is best for the software, so if your approach is better, we should use it.

Boteium:
Duplicate of https://github.com/InfiniTimeOrg/InfiniTime/pull/1276 ?
Personally, I want both. But I wonder why #1276 isn't merged.
I thought #1352 is a replacement of #1276.

Riksu9000:
It seems to me like #1276 was simply  made in protest of #1265. The thought process and design here is different, so not exactly a duplicate.

minacode:
Ok, thank you. A few days passed and my initial feeling is gone. It seems to make sense to do it like this PR and that is ok.

JF002 (approved):

src/displayapp/screens/WatchFaceAnalog.cpp
```diff
@@ -183,6 +183,15 @@ void WatchFaceAnalog::UpdateClock() {
...
   auto batteryPercent = batteryPercentRemaining.Get();
   batteryIcon.SetBatteryPercentage(batteryPercent);
+  static constexpr int lowBatteryThreshold = 15;
+  static constexpr int criticalBatteryThreshold = 5;
+  if (batteryPercent > lowBatteryThreshold) {
+    batteryIcon.SetColor(LV_COLOR_WHITE);
+  } else if (batteryPercent > criticalBatteryThreshold) {
+    batteryIcon.SetColor(LV_COLOR_ORANGE);
+  } else {
+    batteryIcon.SetColor(LV_COLOR_MAKE(0xff, 0x40, 0));
```
NeroBurner, line 193: multiple definitions of the same hex color. Can we move that to the theme file?

lman0:
some people are not able to see the red (but the orange is)
adding for the red coloring  
and make the battery icon blink
 could help people feel the urgency

Riksu9000:
They can still see the icon. I don't think making it blink is necessary.

Riksu9000:
I tried to move the colorization feature into BatteryIcon so the code wouldn't need to be duplicated, however it ended up using a bit more memory. Let me know if I should move it regardless.

Here's a patch with the change.
[deduplication.patch.txt](https://github.com/InfiniTimeOrg/InfiniTime/files/10412211/deduplication.patch.txt)

JF002:
> I tried to move the colorization feature into BatteryIcon so the code wouldn't need to be duplicated, however it ended up using a bit more memory. Let me know if I should move it regardless.

According to my test, the patch only adds 32 bytes. I do not really understand how removing duplicated code ends up in bigger binary (I guess that's the magic of the optimizer), but since the additional memory usage is very small, I think we could apply this patch as it actually removes code duplication :)

PR #1504, opened 2022-12-31
Update workflow dependencies

Riksu9000:
ubuntu-latest vm has been updated from 20.04 to 22.04. To avoid sudden issues, use 22.04 explicitly.
CMake doesn't need to be updated on 22.04.
actions/checkout@v2 uses deprecated Node.js 12. Update to v3 which uses 16

Fixes warnings in builds. https://github.com/InfiniTimeOrg/InfiniTime/actions/runs/3811961913

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -14,7 +14,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     env:
       USERNAME: infinitime
     steps:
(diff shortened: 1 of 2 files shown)


FintasticMan:
Now that it's using Ubuntu 22.04, we don't need to use `clang-format-12` anymore, because it uses llvm 14 by default now. I suggest switching to either `clang-{format,tidy}` or `clang-{format,tidy}-14`.

Riksu9000:
I'll leave updating clang-format for another PR, as there may be new options we want to enable and it'll require discussion.

JF002 (approved):

PR #1505, opened 2023-01-01
Gray settings list

Riksu9000:
Update settings list style. Keep yellow icons for a bit of spice.

InfiniSim_2023-01-01_082201

InfiniSim_2023-01-01_082045

### DIFF
--- a/src/displayapp/screens/List.cpp
+++ b/src/displayapp/screens/List.cpp
@@ -1,6 +1,7 @@
 #include "displayapp/screens/List.h"
 #include "displayapp/DisplayApp.h"
 #include "displayapp/screens/Symbols.h"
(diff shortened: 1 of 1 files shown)


lman0:
> Update settings list style. Keep yellow icons for a bit of spice.

since the icon are in orange inside the settings panel ( check the exact color inside date , or steps .... setting) 
it would me more consistent to make the yellow icons , orange

NeroBurner (approved):

Riksu9000:
> a suggestion @Riksu9000 : since the icon are in orange inside each settings panels ( check the exact color inside date , steps .... settings) it would be ,for me, consistent to make the yellow icons , orange as well

It's a fair point. I'll think about it, but I think this can be merged and improved in the future if necessary.

PR #1506, opened 2023-01-02
Update StopWatch UI

Riksu9000:
Above is the new look, below is the old. Let me know if you have feedback regarding the new colors.

InfiniSim_2023-01-02_083619

InfiniSim_2023-01-02_083844

### DIFF
--- a/src/displayapp/screens/StopWatch.cpp
+++ b/src/displayapp/screens/StopWatch.cpp
@@ -1,6 +1,4 @@
 #include "displayapp/screens/StopWatch.h"
-
-#include "displayapp/screens/Symbols.h"
 #include "displayapp/InfiniTimeTheme.h"
 
 using namespace Pinetime::Applications::Screens;
@@ -18,12 +16,16 @@ namespace {
 
   void play_pause_event_handler(lv_obj_t* obj, lv_event_t event) {
     auto* stopWatch = static_cast<StopWatch*>(obj->user_data);
-    stopWatch->playPauseBtnEventHandler(event);
+    if (event == LV_EVENT_CLICKED) {
+      stopWatch->playPauseBtnEventHandler();
+    }
   }
 
   void stop_lap_event_handler(lv_obj_t* obj, lv_event_t event) {
     auto* stopWatch = static_cast<StopWatch*>(obj->user_data);
-    stopWatch->stopLapBtnEventHandler(event);
+    if (event == LV_EVENT_CLICKED) {
+      stopWatch->stopLapBtnEventHandler();
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


minacode:
Like: the darker button background and getting rid of the green text.

Neutral: the colored buttons. 

Don't like: why text instead of symbols?

ncartron:
I have a slight preference for text over symbols, as it makes things a bit easier to understand / don't need to figure out what the symbols mean.

Riksu9000:
I also find that symbols look better on square buttons, but text looks better on rectangular buttons.

Elements6007:
I like the changes overall but for some reason the blue looks slightly purple to me(looks purple on device too). I think maybe a less purply blue? or maybe green? 
InfiniSim_2023-01-03_013329  InfiniSim_2023-01-03_015421

Riksu9000:
I've made a slight adjustment to the color to make it a bit more green.

I should mention that I want you to test these changes on your PineTime. This is because different screens can display colors in very different ways.

Please excuse the terrible image quality. From this picture you can somewhat see how my old PineTime has much colder colors than the new one. Comparing to an sRGB monitor, I can say that the old one is too cold and the new one is very close, if not a bit warm or yellow, but both are much less saturated.
smallpic

For example, #0068ff looks even better on my monitor, but doesn't look good on my old PT, so I have to use something a bit different.

minacode:
Do you have a reasoning for which button you colored? Just curious because I don't know much about UI.

Riksu9000:
In this case blue is essentially "proceed", "move forward", "continue", and red is used for the destructive option. The rest are default. I experimented with coloring the lap and pause buttons as well, but I think it looks better when only red and blue are used on this screen.

JF002 (approved):
LGTM! I specifically like the new blinking effect!

Elements6007:
@Riksu9000 I tested it on my PT, i cannot really see a difference. Its probably just me, so its fine.

src/displayapp/screens/StopWatch.cpp
```diff
@@ -119,26 +142,29 @@ void StopWatch::Refresh() {
...
+      } else {
+        lv_obj_set_state(time, LV_STATE_DEFAULT);
+        lv_obj_set_state(msecTime, LV_STATE_DEFAULT);
+      }
+    }
   }
 }
 
-void StopWatch::playPauseBtnEventHandler(lv_event_t event) {
-  if (event != LV_EVENT_CLICKED) {
```
NeroBurner, line 126: why are the events removed? Doesn't this mean a long press event could start the timer as well as a normal click?

FintasticMan: The if-statement has just been moved to the calling function.

NeroBurner: ah thanks!

NeroBurner (approved):

lman0:
the overall ui change is good to the eyes , that 's  the good point!

Personally ,  i think the symbol are better than text 
not because of visualy better 
but because
when internationalization will go inside infinitime 
the translated word might be bigger than the button itself
where as symbol remain the same whatever the langage is.


it will be  sad and/or confusing 
to change in the future the text to symbol when 
no english people will want to change text for their langage

lman0:
french for instance , would not fit  in some case
 (start  mean in french either "commencer" /"demarrer"/"depart"
some would use "depart ", but then resume is only translatable in " reprendre")
i think some language like deutch , maybe be longer
the same for nordic langage
arabic , albei may be able to remain inside , but the button would need to be bigger to be readable

minacode:
German would be ok: Start, Stop, Pause, Weiter, Runde

Less things to translate would still be a good though.

NeroBurner:
I like the color and design changes. But I'm in favor of keeping symbols instead of words. Symbols don't need to be translated

Riksu9000:
While text on wide buttons isn't something written in the ui guidelines, it's something I've had in mind for a while, and I'd rather avoid compromising the design. Therefore we can either make an exception because of designing for a small screen, or try to redesign the screen. I'll be looking into the options.

Riksu9000:
Closing in favor of #1561

PR #1508, opened 2023-01-02
Add SettingScreenBase to create a generic checkbox screen

Riksu9000:
This will reduce duplication greatly once applied to other settings screens as well.

The checkbox object array is stored in the class creating the checkbox screen. This is so there can be any number of checkboxes, without using templates.

Do you have ideas for a better name for this class?

Let me know what you think about this approach.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -412,6 +412,7 @@ list(APPEND SOURCE_FILES
         ## Settings
         displayapp/screens/settings/QuickSettings.cpp
         displayapp/screens/settings/Settings.cpp
+        displayapp/screens/settings/SettingScreenBase.cpp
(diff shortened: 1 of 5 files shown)


minacode:
How does this compare against `displayapp/screens/CheckboxList`?

Riksu9000:
Oops, I forgot about that. I should take a look at that soon.

Riksu9000:
#1510 practically obsoletes this PR, but I'd still like to hear your feedback since I could implement these ideas in CheckboxList. Passing unlimited arrays and an array of objects allows any number of checkboxes to be used, and a function doesn't need to be passed to save the settings.

minacode:
The code looks cleaner to me from a quick look and I cannot think of a settings screen right now where what you showed with the refactored screen would not be enough.

PR #1509, opened 2023-01-03
Update clang-{format,tidy} to 14

FintasticMan:
Now that the GH Actions VMs have updated to Ubuntu 22.04, clang can be updated to 14 over 12. I've looked through all the clang-format options that have been added in 13 and 14, and added the ones I believe fit our coding style. These options are: `EmptyLineAfterAccessModifier: Never`, `IndentAccessModifiers: false`, `PackConstructorInitializers: Default`, `ReferenceAlignment: Pointer` and `SeparateDefinitionBlocks: Always`. All of these except the last one don't actually change the formatting at all, just prevent any incorrect formatting in the future. The last one adds some new lines between definition blocks in some cases.

This change fixes a compiler error in the clang-tidy check, about the `-fstack-usage` flag.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -60,6 +60,7 @@ DeriveLineEnding: false
 DerivePointerAlignment: false
 DisableFormat: false
 EmptyLineAfterAccessModifier: Never
+EmptyLineBeforeAccessModifier: LogicalBlock
 ExperimentalAutoDetectBinPacking: true
 FixNamespaceComments: false
 ForEachMacros:
@@ -113,6 +114,7 @@ SpaceAfterCStyleCast: true
 SpaceAfterLogicalNot: false
 SpaceAfterTemplateKeyword: true
 SpaceBeforeAssignmentOperators: true
+SpaceBeforeCaseColon: false
 SpaceBeforeCpp11BracedList: true
 SpaceBeforeCtorInitializerColon: true
 SpaceBeforeInheritanceColon: true
--- a/.clang-format
+++ b/.clang-format
@@ -59,6 +59,7 @@ Cpp11BracedListStyle: true
 DeriveLineEnding: false
 DerivePointerAlignment: false
 DisableFormat: false
+EmptyLineAfterAccessModifier: Never
 ExperimentalAutoDetectBinPacking: true
 FixNamespaceComments: false
 ForEachMacros:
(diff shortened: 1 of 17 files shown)


.clang-format
```diff
@@ -92,6 +94,7 @@ MacroBlockBegin: ''
 MacroBlockEnd: ''
 MaxEmptyLinesToKeep: 1
 NamespaceIndentation: All
+PackConstructorInitializers: BinPack
```
Riksu9000, line 97: The example on https://clang.llvm.org/docs/ClangFormatStyleOptions.html doesn't make it seem like `BinPack` is what we want, but it doesn't seem to format it like that. `NextLine` also doesn't create new changes, but the example looks like what we have currently, so maybe we should use that instead.

.github/workflows/format.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       - name: Install clang-format
         run: |
           sudo apt-get update
-          sudo apt-get -y install clang-format-12
+          sudo apt-get -y install clang-format
```
Riksu9000, line 26: It might not be a bad idea to keep using a certain version explicitly, to avoid different versions potentially behaving differently, so this would be `clang-format-14` for example.

.clang-format
```diff
@@ -123,7 +128,7 @@ SpacesInCStyleCastParentheses: false
 SpacesInParentheses: false
 SpacesInSquareBrackets: false
 SpaceBeforeSquareBrackets: false
-# Needs new Clang: SpaceAroundPointerQualifiers: After
+SpaceAroundPointerQualifiers: After
```
Riksu9000, line 131: Since this has been disabled, it essentially isn't part of our style, and I'm not sure we should change it. I don't really understand how this option works either.

Riksu9000 (commented):
I really like `SeparateDefinitionBlocks` 😀

.clang-format
FintasticMan: Alright, I'll disable it. It basically adds a space after all qualifiers, for example `const`, disregarding the fact that pointer asterisks are left-aligned.

.clang-format
FintasticMan: Oops, I misread the option. I agree that `NextLine` is what we want.

Riksu9000 (approved):

Avamander (approved):

FintasticMan:
Turns out, not all options that apply to this codebase from clang-format 12 have been added to the configuration, `EmptyLineBeforeAccessModifier: LogicalBlock` and `SpaceBeforeCaseColon: false` were missed.

JF002 (approved):

PR #1511, opened 2023-01-03
Update pre-commit hook

FintasticMan:
This updates the pre-commit hook to use git-clang-format.

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -1,25 +1,8 @@
-#!/bin/bash
-if clang-format --version | grep -q 'version 11\.'; then
-   CLANG_FORMAT_EXECUTABLE="clang-format"
-else
-   CLANG_FORMAT_EXECUTABLE="clang-format-11"
-fi
+#!/bin/sh
 
-if ! command -v $CLANG_FORMAT_EXECUTABLE &> /dev/null
-then
-    echo $CLANG_FORMAT_EXECUTABLE does not exist, make sure to install it
-    exit 1
-fi
+changedFiles="$(git clang-format --extensions cpp,h --style file --staged -q --diffstat -- ':!src/FreeRTOS' ':!src/libs')"
+git clang-format --extensions cpp,h --style file --staged -q -- ':!src/FreeRTOS' ':!src/libs'
 
-for FILE in $(git diff --cached --name-only)
-do
-	if [[ "$FILE" =~ src/[A-Za-z0-9\ \-]+*\.(c|h|cpp|cc)$ ]]; then
-		echo Autoformatting $FILE with $CLANG_FORMAT_EXECUTABLE
-		$CLANG_FORMAT_EXECUTABLE -style=file -i -- $FILE
-		git add -- $FILE
-	elif [[ "$FILE" =~ src/(components|displayapp|drivers|heartratetask|logging|systemtask)/.*\.(c|h|cpp|cc)$ ]]; then
-		echo Autoformatting $FILE with $CLANG_FORMAT_EXECUTABLE
-		$CLANG_FORMAT_EXECUTABLE -style=file -i -- $FILE
-		git add -- $FILE
-	fi
+echo "$changedFiles" | head -n -1 | cut -d ' ' -f 1 | while read -r file; do
+  git add -- "$file"
 done
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -1,23 +1,29 @@
 #!/bin/sh
 
+name="clang-format"
+
+if [ -z "$(command -v "git-$name")" ]; then
+  name="$(basename -a $(find $(echo "$PATH" | tr ':' ' ') -maxdepth 1 -type f -executable -name 'git-clang-format*') | sort | tail -n 1 | sed 's/^git-//')"
+fi
+
 minVersion="14.0.0"
 
-for file in $(find $(echo "$PATH" | tr ':' ' ') -maxdepth 1 -type f -executable -name 'git-clang-format*'); do
-  curName="$(basename "$file" | sed 's/^git-//')"
-  curVersion="$("$curName" --version | cut -d ' ' -f 3)"
+for file in $(find $(echo "$PATH" | tr ':' ' ') -maxdepth 1 -type f -executable -name 'clang-format*'); do
+  curBin="$file"
+  curVersion="$("$curBin" --version | cut -d ' ' -f 3)"
 
   if [ "$(printf '%s\n' "$curVersion" "$version" "$minVersion" | sort -V | tail -n 1)" = "$curVersion" ]; then
-    name="$curName"
+    bin="$curBin"
     version="$curVersion"
   fi
 done
 
-if [ -z "$name" ]; then
+if [ -z "$name" ] || [ -z "$bin" ]; then
   echo "Could not find a suitable clang-format installation. Install clang-format that includes the git-clang-format script, with at least version $minVersion"
   exit 1
 fi
 
-args='-q --extensions cpp,h --style file --staged -- :!src/FreeRTOS :!src/libs'
+args="--binary $bin -q --extensions cpp,h --style file --staged -- :!src/FreeRTOS :!src/libs"
 
 changedFiles="$(git "$name" --diffstat $args)"
 git "$name" $args
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -1,8 +1,27 @@
 #!/bin/sh
 
-changedFiles="$(git clang-format --extensions cpp,h --style file --staged -q --diffstat -- ':!src/FreeRTOS' ':!src/libs')"
-git clang-format --extensions cpp,h --style file --staged -q -- ':!src/FreeRTOS' ':!src/libs'
+minVersion="14.0.0"
 
-echo "$changedFiles" | head -n -1 | cut -d ' ' -f 1 | while read -r file; do
+for file in $(find $(echo "$PATH" | tr ':' ' ') -maxdepth 1 -type f -executable -name 'git-clang-format*'); do
+  curName="$(basename "$file" | sed 's/^git-//')"
+  curVersion="$("$curName" --version | cut -d ' ' -f 3)"
+
+  if [ "$(printf '%s\n' "$curVersion" "$version" "$minVersion" | sort -V | tail -n 1)" = "$curVersion" ]; then
+    name="$curName"
+    version="$curVersion"
+  fi
+done
+
+if [ -z "$name" ]; then
+  echo "Could not find a suitable clang-format installation. Install clang-format that includes the git-clang-format script, with at least version $minVersion"
+  exit 1
+fi
+
+args='-q --extensions cpp,h --style file --staged -- :!src/FreeRTOS :!src/libs'
+
+changedFiles="$(git "$name" --diffstat $args)"
+git "$name" $args
+
+echo "$changedFiles" | head -n -1 | cut -d ' ' -f 2 | while read -r file; do
   git add -- "$file"
 done


Avamander:
Detecting older and incompatible clang-format would be kinda nice.

FintasticMan:
I could add a simple check that makes sure the version is at least 14.0.0. I could also add some slightly more complicated code that goes through all `clang-format*` executables in the PATH, and chooses the one with the highest version, and then makes sure that it's at least 14.0.0. Which of these do you think would be better?

FintasticMan:
I've just implemented the more complicated option. It isn't very elegant sadly, if someone can come up with a better way to do it, I would appreciate it a lot. It should work in basically all situations though (except if the user has a space in their PATH, which they really shouldn't have and will probably break other things as well).

hooks/pre-commit
```diff
@@ -1,25 +1,27 @@
...
-    exit 1
+for file in $(find $(echo "$PATH" | sed 's/:/ /g') -maxdepth 1 -type f -executable -name 'git-clang-format*'); do
+  cur_name="$(basename "$file" | sed 's/^git-//')"
+  cur_version="$("$cur_name" --version | awk '{print $3}')"
+
+  if [ "$(printf '%s\n' "$cur_version" "$version" | sort -V | tail -n 1)" = "$cur_version" ]; then
+    version="$cur_version"
+    name="$cur_name"
+  fi
+done
```
Riksu9000, line 14: Wouldn't just `clang-format` always be the highest available version?

FintasticMan: No, in the case of Ubuntu 20.04, the version of `clang-format` is 10, even though `clang-format-12` is available in the repositories.

hooks/pre-commit
```diff
@@ -1,25 +1,27 @@
...
+
+name=""
+version="14.0.0"
 
-if ! command -v $CLANG_FORMAT_EXECUTABLE &> /dev/null
-then
-    echo $CLANG_FORMAT_EXECUTABLE does not exist, make sure to install it
-    exit 1
+for file in $(find $(echo "$PATH" | sed 's/:/ /g') -maxdepth 1 -type f -executable -name 'git-clang-format*'); do
+  cur_name="$(basename "$file" | sed 's/^git-//')"
```
Riksu9000, line 7: Is there any difference between `git clang-format` and `git-clang-format`? If not, we can remove this step.

hooks/pre-commit
```diff
@@ -1,25 +1,27 @@
-#!/bin/bash
-if clang-format --version | grep -q 'version 11\.'; then
-   CLANG_FORMAT_EXECUTABLE="clang-format"
-else
-   CLANG_FORMAT_EXECUTABLE="clang-format-11"
-fi
+#!/bin/sh
+
+name=""
+version="14.0.0"
```
Riksu9000, line 4: Make this variable contain the minimum accepted version, and don't modify it in the script.

```suggestion
minVersion="14.0.0"
```

hooks/pre-commit
```diff
@@ -1,25 +1,27 @@
...
 
-if ! command -v $CLANG_FORMAT_EXECUTABLE &> /dev/null
-then
-    echo $CLANG_FORMAT_EXECUTABLE does not exist, make sure to install it
-    exit 1
+for file in $(find $(echo "$PATH" | sed 's/:/ /g') -maxdepth 1 -type f -executable -name 'git-clang-format*'); do
```
Riksu9000, line 6: ShellCheck also complains about this line. You should take a look at that.

Riksu9000 (commented):
This should probably follow the same conventions regarding variable naming.

hooks/pre-commit
FintasticMan: This is in there on the off-chance that a clang-format installation doesn't include `git-clang-format`. The reason for removing the `git-` prefix is because `git-clang-format` itself doesn't have a way to print the version, so we have to use `clang-format --version`. If you think it's safe to assume every clang-format installation has an accompanying `git-clang-format`, we can very slightly simplify this code.

hooks/pre-commit
```diff
@@ -1,25 +1,29 @@
...
+#!/bin/sh
+
+minVersion="14.0.0"
+
+name=""
+version=""
+
+for file in $(find $(echo "$PATH" | tr ':' ' ') -maxdepth 1 -type f -executable -name 'git-clang-format*'); do
+  curName="$(basename "$file" | sed 's/^git-//')"
+  curVersion="$("$curName" --version | awk '{print $3}')"
```
Riksu9000, line 10:
```suggestion
  curVersion="$("$curName" --version | cut -d ' ' -f 3)"
```

hooks/pre-commit
```diff
@@ -1,25 +1,29 @@
...
-   CLANG_FORMAT_EXECUTABLE="clang-format"
-else
-   CLANG_FORMAT_EXECUTABLE="clang-format-11"
-fi
+#!/bin/sh
+
+minVersion="14.0.0"
+
+name=""
+version=""
```
Riksu9000, line 6: Variables don't need to be declared in shell.

Riksu9000 (changes requested):

hooks/pre-commit
Riksu9000: > The reason it needs to be a for loop rather than a while read loop or find -execdir is because it needs to modify the name and version variables.

Are you sure that's a problem? Shell is pretty lenient about accessing variables.

hooks/pre-commit
Riksu9000: Well this script depends on `git-clang-format`, but regarding this I think I'm just confusing the different commands.

hooks/pre-commit
```diff
@@ -1,25 +1,29 @@
...
-    exit 1
+  if [ "$(printf '%s\n' "$curVersion" "$version" "$minVersion" | sort -V | tail -n 1)" = "$curVersion" ]; then
+    name="$curName"
+    version="$curVersion"
+  fi
+done
+
+if [ -z "$name" ]; then
+  echo "Could not find a suitable clang-format installation. Install clang-format with at least version $minVersion"
+  exit 1
```
Riksu9000, line 20: You should store the entire found `git-clang-format` script name, and the error message should say it couldn't find a suitable `git-clang-format` script.

Riksu9000 (changes requested):

hooks/pre-commit
FintasticMan: A while read loop could work, but the only way to do it in a POSIX-compliant way is by sending the output from `find` to a temporary file, then reading that file for the loop. This is because if you use the standard pipe from `find` to the while loop, the loop becomes a subshell, and variables inside won't be available outside.

Do you think that that's a good solution?

hooks/pre-commit
FintasticMan: I don't agree that it should store the whole `git-clang-format` script name, because I think it should store the name of the suitable clang-format installation. That name is  used to get the version, and then used to for the `git clang-format` command. I will add a note to the error message that the clang-format installation needs to include the `git-clang-format` script.

hooks/pre-commit
```diff
@@ -1,25 +1,26 @@
...
-		echo Autoformatting $FILE with $CLANG_FORMAT_EXECUTABLE
-		$CLANG_FORMAT_EXECUTABLE -style=file -i -- $FILE
-		git add -- $FILE
-	fi
+changedFiles="$(git "$name" --extensions cpp,h --style file --staged -- ':!src/FreeRTOS' ':!src/libs')"
+
+echo "$changedFiles"
+
+echo "$changedFiles" | tail -n +2 | while read -r file; do
+  git add -- "$file"
```
Riksu9000, line 25: If I change multiple files and leave a formatting error in DisplayApp.cpp, the script will output the following, which is ambiguous and confusing. Do you think it needs to output anything?
```
changed files:
    src/displayapp/DisplayApp.cpp
```

Riksu9000 (changes requested):

hooks/pre-commit
Riksu9000: I guess this sort of applies to the error as well, since it might not be obvious it is coming from the pre-commit hook.

FintasticMan: Because someone has to enable the commit hook manually, I think that the error is fine as it is. What do you think?

The old pre-commit hook printed every file that it formatted quite explicitly. I personally don't think that's necessary, so I will remove that.

FintasticMan:
I've switched to a more robust method of getting the files that have changed. This doesn't rely on using human-readable output that could potentially be changed in an update.

I have a git stash with a bit more logic that more conclusively finds the highest clang-format version, by using `git clang-format`'s `--binary` option. It does make the code less readable, and I think that this is probably alright as it is.

Riksu9000:
I'd be interested to see your new clang-format finding code regardless, if you could post a patch.

FintasticMan:
```diff
diff --git a/hooks/pre-commit b/hooks/pre-commit
index 344464c5..fe167682 100755
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -1,24 +1,30 @@
 #!/bin/sh
 
+name="clang-format"
+
+if [ -z "$(command -v "git-$name")" ]; then
+  name="$(basename -a $(find $(echo "$PATH" | tr ':' ' ') -maxdepth 1 -type f -executable -name 'git-clang-format*') | sort | tail -n 1)"
+fi
+
 minVersion="14.0.0"
 
-for file in $(find $(echo "$PATH" | tr ':' ' ') -maxdepth 1 -type f -executable -name 'git-clang-format*'); do
-  curName="$(basename "$file" | sed 's/^git-//')"
-  curVersion="$("$curName" --version | cut -d ' ' -f 3)"
+for file in $(find $(echo "$PATH" | tr ':' ' ') -maxdepth 1 -type f -executable -name 'clang-format*'); do
+  curBin="$file"
+  curVersion="$("$curBin" --version | cut -d ' ' -f 3)"
 
   if [ "$(printf '%s\n' "$curVersion" "$version" "$minVersion" | sort -V | tail -n 1)" = "$curVersion" ]; then
-    name="$curName"
+    bin="$curBin"
     version="$curVersion"
   fi
 done
 
-if [ -z "$name" ]; then
+if [ -z "$bin" ]; then
   echo "Could not find a suitable clang-format installation. Install clang-format that includes the git-clang-format script, with at least version $minVersion"
   exit 1
 fi
 
-changedFiles="$(git "$name" --extensions cpp,h --style file --staged -q --diffstat -- ':!src/FreeRTOS' ':!src/libs')"
-git "$name" --extensions cpp,h --style file --staged -q -- ':!src/FreeRTOS' ':!src/libs'
+changedFiles="$(git "$name" --binary "$bin" --extensions cpp,h --style file --staged -q --diffstat -- ':!src/FreeRTOS' ':!src/libs')"
+git "$name" --binary "$bin" --extensions cpp,h --style file --staged -q -- ':!src/FreeRTOS' ':!src/libs'
 
 echo "$changedFiles" | head -n -1 | cut -d ' ' -f 1 | while read -r file; do
   git add -- "$file"
```

Here is a patch. The code for finding the clang-format binary is mostly the same, but the script now stores both the git-clang-format script name and the actual clang-format binary.

hooks/pre-commit
```diff
@@ -1,25 +1,25 @@
...
-		echo Autoformatting $FILE with $CLANG_FORMAT_EXECUTABLE
-		$CLANG_FORMAT_EXECUTABLE -style=file -i -- $FILE
-		git add -- $FILE
-	elif [[ "$FILE" =~ src/(components|displayapp|drivers|heartratetask|logging|systemtask)/.*\.(c|h|cpp|cc)$ ]]; then
-		echo Autoformatting $FILE with $CLANG_FORMAT_EXECUTABLE
-		$CLANG_FORMAT_EXECUTABLE -style=file -i -- $FILE
-		git add -- $FILE
-	fi
+changedFiles="$(git "$name" --extensions cpp,h --style file --staged -q --diffstat -- ':!src/FreeRTOS' ':!src/libs')"
+git "$name" --extensions cpp,h --style file --staged -q -- ':!src/FreeRTOS' ':!src/libs'
```
Riksu9000, line 21: Moving the duplicated options to a variable could help avoid future mistakes. If the --diffstat option can be placed at the beginning, you should only need a single variable.

FintasticMan: Good idea. ShellCheck gives an "info" warning about it, but I think it's quite nice.

JF002 (approved):

PR #1513, opened 2023-01-05
Add a simple low pass filter for battery voltage

AlexXZero:
Sometimes I noticed that battery percentage might drops to 5-10%, then stay at the same level during a full day. During a little investigation I found that this behavior is related to noise in voltage measuring data. This PR should improve showing battery percentage by using low pass filter for battery voltage.

There is some data which I measured and used for comparing old and a new versions:
Clear plot:
image
Findings:
image
Source data: [battery.csv](https://github.com/InfiniTimeOrg/InfiniTime/files/10350577/battery_04.01.22.csv)

On the plot you can see blue line is old data and orange line is a new data.

Summary of improvements:
1. Battery discharge curve looks more stable without significant drops
2. The start of the percentage drop is a little slow to make it more stable and match the rest of the discharge time
3. 0% battery is when the watch stops working or about 1-2 hours after that.

### DIFF
--- a/src/components/battery/BatteryController.cpp
+++ b/src/components/battery/BatteryController.cpp
@@ -62,12 +62,12 @@ void Battery::SaadcInit() {
 
 void Battery::SaadcEventHandler(nrfx_saadc_evt_t const* p_event) {
   static const Utility::LinearApproximation<uint16_t, uint8_t, 6> aprox {{{
-    {3500, 0},  // Minimum voltage before shutdown (depends on the battery)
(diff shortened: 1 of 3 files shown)


Riksu9000:
0% is set to 3.5V, because you shouldn't discharge the battery too deep. This filtering causes the percentage to be overreported. SAADC has features to reduce noise. Let's try these before adding software filtering. https://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.nrf52832.ps.v1.1%2Fsaadc.html

PR #1514, opened 2023-01-05
Add maintainer's guide

Riksu9000:
This document clarifies how the maintainers should manage the project.

We should write about using the milestones as well, but I want to discuss that first. I'll open an issue about that. If that doesn't get resolved in time, it can be another PR. EDIT: Here's the issue #1515

I've added new instructions on handling stale PRs. Three months is quite conservative, as it concerns any activity, and can be reduced. Let me know what you think.

### DIFF
--- a/doc/maintainer-guide.md
+++ b/doc/maintainer-guide.md
@@ -1,0 +1,45 @@
+# Reviewing PRs
+
+Approving a PR means that the reviewer has deemed the PR ready to be merged.
+
+There are two steps to reviewing PRs.
+
+- Review the feature:
+
+  - Consider if the feature aligns with the [InfiniTime vision](InfiniTimeVision.md)
+  - Discuss alternative ideas or implementations
+
+- Review the code:
+
+  - Check the quality of the code and make sure it conforms to the [coding conventions](coding-convention.md)
+  - Consider the maintainability of the code
+  - Test the code with at least InfiniSim or a PineTime
+
+# Merging PRs
+
+Two approvals from core developers is usually required to merge a PR.
+Exceptions include urgent fixes
+and small maintenance PRs by core developers,
+that don't affect the apparent behaviour of the watch in any way.
+
+All but the simulator check must be passed before merging a PR.
+
+PRs are either rebase or squash merged,
+depending on whether the commits satisfy the following requirements:
+
+### Commits
+
+- Commits that fix mistakes from previous commits must be squashed before merging a PR.
+- The title and description of the commit must sufficiently explain the changes in the commit.
+
+If these requirements are not met,
+the PR must be squash merged,
+and the merger must write a satisfactory description.
+
+# Stale PRs
+
+Work-in-Progress PRs shall be marked as draft.
+
+Draft PRs with no activity by the author for 3 months may be closed as stale.
+
+PRs with changes requested, but no activity by the author for 3 months may be closed as stale.
--- a/README.md
+++ b/README.md
@@ -31,10 +31,6 @@ Fast open-source firmware for the [PineTime smartwatch](https://www.pine64.org/p
 - [Tips on designing an app UI](doc/ui_guidelines.md)
 - [Bootloader, OTA and DFU](bootloader/README.md)
 - [External resources](doc/ExternalResources.md)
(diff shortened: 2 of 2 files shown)


doc/maintainer-guide.md
```diff
@@ -0,0 +1,30 @@
+# Reviewing PRs
+
+Approving a PR means that the reviewer has deemed the PR ready to be merged.
+
+While reviewing
+
```
JF002, line 6: I would add 2 points here:
 - Questions whether the new functionality is needed in the project and if it follows the vision of the projects ("Do we really want/need that feature?").
 - Questions whether the maintainer will be willing to maintain this feature on the long term ("Do we have the resources to maintain this feature?")

doc/maintainer-guide.md
```diff
@@ -0,0 +1,30 @@
...
+- Consider and discuss alternative ideas or implementations.
+
+# Merging PRs
+
+Two approvals from core developers is usually required to merge a PR.
+Exceptions include urgent fixes
+and when the author is specified as the maintainer of the code in question.
+Then only a single approval is needed.
+
+Commits that fix code formatting from previous commits must be squashed before merging a PR,
```
JF002, line 18: I think we could ask the author of the PR to make sure the history is clean before we merge the PR, indeed!

JF002 (commented):
This is a great initiative!

doc/maintainer-guide.md
Riksu9000: We haven't had any requirements for commits so far, and this is an idea I had to have just slightly cleaner commits. Requirements for commits is probably something worth discussing, as it relates to #1515, so I think I'll open a new issue for that as well.

Riksu9000:
I modified the merging section. I removed the part about the author being specified as the maintainer of the code, as that's not something we do currently, and replaced it with allowing small maintenance PRs by core developers to be merged with a single approval. This should reduce the wait times for maintenance PRs.

JF002:
Looks good to me. That's a good start, and we'll probably refine it later on :)

doc/maintainer-guide.md
Riksu9000: I think rebase merging gets rid of the merge commits anyway, so a linear history is already implicitly enforced by this guide. And we should definitely enforce a linear history. https://docs.github.com/en/github/administering-a-repository/about-protected-branches#require-linear-history

doc/maintainer-guide.md
```diff
@@ -0,0 +1,37 @@
...
+Two approvals from core developers is usually required to merge a PR.
+Exceptions include urgent fixes
+and small maintenance PRs by core developers that don't affect the apparent behaviour of the watch in any way.
+Then only a single approval is needed.
+
+Commits that fix code formatting from previous commits must be squashed before merging a PR,
+or the PR must be squash merged.
+Otherwise the PR shall be squash or rebase merged.
+
+All but the simulator check must be passed before merging a PR.
```
Riksu9000, line 29: @JF002 There are branch protection rules which make it so checks must be passed before merging. If these are enabled for build-firmware, clang-format and clang-tidy, this line can be removed.

JF002: Do you mean this for the **develop** branch? 
image

Enabling as many auto checks as possible is probably a good idea to make our lifes easier, but I also think it's a good idea to mention those checks in the maintainer's guide, as a documentation of the process.

JF002:
Regarding the merging section, and since @Riksu9000 brought to my attention that we can enable protection rules on branches, do we want to add a "linear history" rule to this guide, and enforce it with this check ?

image

doc/maintainer-guide.md
Riksu9000: Okay, I'll keep the note here then. I'll also keep the wording, even though it doesn't exactly match the branch protection rules.

Riksu9000:
I think we should enable the branch protection rule, but I'm not sure we need to explicitly mention it in the guide. Following the instructions in the guide will maintain a linear history, and the branch protection rule makes it so you couldn't break the rule anyway. I'll be updating the merging section soon.

JF002 (approved):

PR #1518, opened 2023-01-07
Add buildsize workflow

Riksu9000:
Add workflow jobs that create and update a comment on PRs with the build size and difference from the base branch (develop).

It will not work on this PR, because develop doesn't contain the getSize.sh script.

Here's an example of the jobs in action https://github.com/Riksu9000/InfiniTime/pull/15

### DIFF
--- a/.github/workflows/getSize.sh
+++ b/.github/workflows/getSize.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+# Requires environment variables from docker/build.sh
+
+set -e
+
+SIZE_BIN="$TOOLS_DIR/$GCC_ARM_PATH/bin/arm-none-eabi-size"
+[ ! -x "$SIZE_BIN" ] && exit 1
+
+[ -z "$1" ] && exit 1
+SIZE_OUTPUT=$($SIZE_BIN "$1" | tail -n1)
+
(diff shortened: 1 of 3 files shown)


JF002 (approved):
This will be very helpful to evaluate the impact on memory usage for future PR, awesome!

FintasticMan:
How will this work if a pull request isn't based off of the latest develop, but an older version? If there have been commits that increase space usage in develop, it might look like a PR adds less than it actually does, or even that it lowers space usage when it actually increases it.

Riksu9000:
Great point. I looked into it and it seems that the checkout action automatically merges the branches when checking out the head branch, so the only difference between base and ref are the commits of the PR.

https://github.com/Riksu9000/InfiniTime/pull/16

develop is at 6439189373f2e70e25a130bccc7d4a6b8a7538d1
test-change is at d7d7856f0dba4dfa7e3e43b544c5391fd74e123e

I tested it by pushing a different commit to develop and test-change branches, then inspecting the build-firmware check log, which reads:

```
HEAD is now at e8cb4b2 Merge d7d7856f0dba4dfa7e3e43b544c5391fd74e123e into 6439189373f2e70e25a130bccc7d4a6b8a7538d1
```

Issue #1522, opened 2023-01-12
Disable notifications pop-ups while using torch app


rambonette:
First of all I want to point out that I use the torch app quite a lot and I find it very useful.
Allow me to take a few moments to explain the situation to you:
Last night I was using the torch app in the dark to find my storage's lock keyhole. All of the sudden a message arrrives on my phone which rightly triggers a notification on the watch. Not only the torch was momentarily deactivated, but once the notification timer expired it returned to the watch face screen. As you can imagine it was quite inconvenient having to re-open the torch while having a bunch of keys in my hand.

Shouldn't we make sure that when the torch is active, notifications are blocked? Or that if notifications come in, when they expire, we revert to the torch app?

Steps to reproduce issue:
- Open torch app and turn on torch
- Trigger a test notification from a companion app (like Gadgetbridge)

I really don't know if this we should label this as bug or enhancement.

I'd tackle this myself, but I'm sadly overworked right now. So I want to thank in advance anyone who wants to address this issue!

FintasticMan:
It would definitely be possible to prevent notifications while the torch is on. The other issue of it not returning to the torch is actually fixed in the latest development branch (PR #1350), so it will be available in the next release.

rambonette:
Sorry

Riksu9000:
Duplicate of #610

PR #1525, opened 2023-01-14
block notifications when flashlight is on - first quick try

rambonette:
**Premise:** _I don't know if this is the right approach to accomplish this._

Quick (possible) solution to this #1522 issue/enhancement. Made a quick before/after video comparison to demonstrate the solution.

Before:

before

After:

after

The basic strategy here is to check if  `settingsController.GetNotificationStatus()` is `On` when Flashlight is toggled, then act accordingly keeping track of the previous notification status with `bool notificationsStatus` to be able to revert the setting once Flashlight is off.

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -459,7 +459,7 @@ void DisplayApp::LoadScreen(Apps app, DisplayApp::FullRefreshDirections directio
                                                             touchPanel);
       break;
     case Apps::FlashLight:
-      currentScreen = std::make_unique<Screens::FlashLight>(this, *systemTask, brightnessController);
+      currentScreen = std::make_unique<Screens::FlashLight>(this, *systemTask, brightnessController, settingsController);
       break;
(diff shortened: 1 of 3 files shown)


FintasticMan:
Looks good! One concern I have is that every change to the settings means a write to the flash storage, and doing that every time the flashlight is turned on or off might not be great for the lifespan of the storage. Maybe it could instead be implemented as an extra check whether the flashlight app is running before sending a notification.

rambonette:
Thanks!

> Looks good! One concern I have is that every change to the settings means a write to the flash storage, and doing that every time the flashlight is turned on or off might not be great for the lifespan of the storage. Maybe it could instead be implemented as an extra check whether the flashlight app is running before sending a notification.

That's actually a very smart idea. I'll definitely try this reverse approach!

Riksu9000:
Considering the before animation is outdated, I think the current behaviour is fine. We shouldn't make unnecessary exceptions for showing notifications, and by having the flashlight turn off, you won't get blinded by the light when you close the notification.

PR #1533, opened 2023-01-16
Improve SetDate and SetTime UI

Elements6007:
Expands 'set' button to the length of the counter widgets and adds the use of the InfiniTimeTheme in `SettingSetDate`
InfiniSim_2023-01-15_041650


kieranc:
Looks good, is there a way to make the selectors the same size between the 2 screens?

Elements6007:
> Looks good, is there a way to make the selectors the same size between the 2 screens?

Well if `jetbrains_mono_42` is changed to `jetbrains_mono_bold_20`, which is the same as in `SettingSetDate` then they will be the same size. Although it definitely looks more uniform, i think the larger counters look better.
InfiniSim_2023-01-16_163511

Riksu9000:
Why not make them different sizes on each screen?

src/displayapp/screens/settings/SettingSetDate.cpp
```diff
@@ -80,9 +81,9 @@ SettingSetDate::SettingSetDate(Pinetime::Applications::DisplayApp* app,
 
   btnSetTime = lv_btn_create(lv_scr_act(), nullptr);
   btnSetTime->user_data = this;
-  lv_obj_set_size(btnSetTime, 120, 48);
+  lv_obj_set_size(btnSetTime, LV_HOR_RES - 35, 48);
```
Riksu9000, line 84: LV_HOR_RES_MAX is more efficient, because it is can be compile time evaluated.

```suggestion
  lv_obj_set_size(btnSetTime, LV_HOR_RES_MAX - 35, 48);
```

Riksu9000 (changes requested):

Elements6007:
> Why not make them different sizes on each screen?

could you elaborate plz?

Riksu9000:
I mean the buttons on the different screens. They don't need to be the exact same width. Being aligned with the visible counters is more important I think.

src/displayapp/screens/settings/SettingSetTime.cpp
```diff
@@ -52,7 +52,7 @@ SettingSetTime::SettingSetTime(Pinetime::Applications::DisplayApp* app,
     hourCounter.EnableTwelveHourMode();
   }
   hourCounter.SetValue(dateTimeController.Hours());
-  lv_obj_align(hourCounter.GetObject(), nullptr, LV_ALIGN_CENTER, -75, POS_Y_TEXT);
+  lv_obj_align(hourCounter.GetObject(), nullptr, LV_ALIGN_CENTER, -72, POS_Y_TEXT);
```
Riksu9000, line 55: I think this was an unintentional change.

Elements6007: Screenshot 2023-02-05 163641 (Custom) (1)
it aligns the counter better with the button (left is new change)

Riksu9000: This is not an improvement. The button width should be adjusted instead.

Untitled

src/displayapp/screens/settings/SettingSetTime.cpp
```diff
@@ -67,7 +67,7 @@ SettingSetTime::SettingSetTime(Pinetime::Applications::DisplayApp* app,
 
   btnSetTime = lv_btn_create(lv_scr_act(), nullptr);
   btnSetTime->user_data = this;
-  lv_obj_set_size(btnSetTime, 120, 50);
+  lv_obj_set_size(btnSetTime, LV_HOR_RES_MAX - 30, 48);
```
Riksu9000, line 70: Please don't add unexplained changes. The button height should be 50, so you may update these accordingly. Please read the Commits section of the maintainer's guide to see the requirements for commits. These will soon be added to the contributing guide #1570. https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/doc/maintainer-guide.md#commits

Riksu9000: It should be 50, because 50 is standard on many screens, and I think it being different here is a mistake. I agree there should be some more distance between the buttons, however it should be addressed some other way, perhaps by adjusting the counter size or position, and ideally in a separate PR.

Elements6007: Ah, Ok. Should I make the button height for setDate 50 also?

Riksu9000: You could. You don't have to.

Riksu9000:
The button size is not correct on the SettingSetDate screen.

Elements6007:
The height? It is 50 because that is what it should be, you said it would be ok to change it. 
> You could. You don't have to.

unless I misunderstood.

Riksu9000:
The width.

image

Elements6007:
Ah ok, I never checked down to that small. It should be fixed now.
infiniui2 (Custom) 
(left side)
infiniui (Custom)
(right)

PR #1547, opened 2023-01-20
Tile/List: Code improvements

Riksu9000:
Auto deduce number of screens in ApplicationList and Settings(list). Replace #define with static constexpr. Reduce magic numbers


JF002 (approved):
I do really like how fluent you are with `constexpr`! Awesome!

src/displayapp/screens/ApplicationList.h
```diff
@@ -33,26 +33,23 @@ namespace Pinetime {
...
-
-        static constexpr std::array<Tile::Applications, appsPerScreen * nScreens> applications {{
+        static constexpr Tile::Applications applications[] {
           {Symbols::stopWatch, Apps::StopWatch},
           {Symbols::clock, Apps::Alarm},
           {Symbols::hourGlass, Apps::Timer},
           {Symbols::shoe, Apps::Steps},
           {Symbols::heartBeat, Apps::HeartRate},
           {Symbols::music, Apps::Music},
-
```
FintasticMan, line 48: I personally think that it's better for readability to leave the empty lines between screens.

Riksu9000: The purpose of this PR is to create separation between the list of applications and the created screens. Otherwise there should just be an array per screen.

Riksu9000:
Not perfectly happy with this code, so I'll close this for now.

PR #1548, opened 2023-01-21
Add Gauge Watch Face

Leon-Plickat:
Basically an analog clock, but with gauges instead of watch-hands.

image

This is implemented using the arc object. Unfortunately, switching to the clock screen from any other screen is tremendously slow when this face is active. I wonder if the arc object is inherently intensive to render? Not quite sure what to do about this. Unfortunately this only happens on the physical device, not the simulator.

* [ ] make sreen switching faster
* [ ] display battery icon when battery gets low
* [ ] display heart rate icon when monitor is active
* [ ] display bluetooth icon when connected


Leon-Plickat:
Seems like the screen switching can be sped up to reasonable times by reducing the amount of pixels drawn by the arc object.

Leon-Plickat:
Turns out the amount of pixels is not really relevant, the issue is the background of the arc. Without the background, switching to the clock behaves as expected.

image

Even a background with line size of 1 caused this.
Not quite sure why that is, I have no experience with embedded graphics, only desktop graphics.

While removing the background doesn't look *quite* as good, it has the added benefit of increasing the contrast of the gauges, probably making them easier to read for someone who has colourblindness or other vision problems.

minacode:
I think it looks very cool without the background! 😎

Leon-Plickat:
Now with switchable themes.

image
image
image
image
image
image

Leon-Plickat:
Considering the requirement too keep it small as well as the size of a theme (`5 * 3 * 8 = 120` bits), I think it makes sense to limit the amount of themes to maybe three or four. I have come to prefer the multi-colour ones over the single-colour ones, so I'll probably go with a few common terminal colour themes.

minacode:
What do you think about using the PineTimeStyle settings that already exist for the theming? Could that be useful?

minacode:
You could make it depend on the time 😁

Leon-Plickat:
I played around with making it depend on time, but wasn't satisfied with the results.

So I am leaving the colours as is right now, meaning that other than finding a better name, I am pretty much done here.

Thinking about "futuristic" for the name or something like that.

Leon-Plickat:
Yeah, I think this is done.

src/displayapp/screens/WatchFaceFuturistic.h
```diff
@@ -0,0 +1,78 @@
...
+        Controllers::Ble& bleController;
+        Controllers::NotificationManager& notificationManager;
+        Controllers::HeartRateController& heartRateController;
+
+        // Based on the tango terminal colour theme.
+        static constexpr int innerTextColor = 0xeeeeec;
+        static constexpr int hoursArcColor = 0xef2929;
+        static constexpr int minutesArcColor = 0x729fcf;
+        static constexpr int secondsArcColor = 0x8ae234;
+        static constexpr int outerTextColor = 0xfce94f;
```
Riksu9000, line 72: Initializing the colors to lv_color_t with LV_COLOR_MAKE is more efficient. Move these to an anonymous namespace in the cpp file.

src/displayapp/screens/WatchFaceFuturistic.h
```diff
@@ -0,0 +1,78 @@
...
+        DirtyValue<bool> heartbeatRunning {};
+        DirtyValue<bool> isCharging {};
+        DirtyValue<uint8_t> batteryPercentRemaining {0};
+
+        lv_style_t arc_fg_style;
+        lv_style_t arc_bg_style;
+
+        lv_obj_t* arc_sec;
+        lv_obj_t* arc_min;
+        lv_obj_t* arc_hour;
```
Riksu9000, line 53: Fix variable naming scheme.
```suggestion
        lv_obj_t* arcSec;
        lv_obj_t* arcMin;
        lv_obj_t* arcHour;
```

src/displayapp/screens/WatchFaceFuturistic.cpp
```diff
@@ -0,0 +1,185 @@
...
+  // low battery alert message
+  batteryAlert = lv_label_create(lv_scr_act(), nullptr);
+  lv_label_set_text(batteryAlert, "low bat");
+  lv_obj_set_style_local_text_color(batteryAlert, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hex(outerTextColor));
+  lv_obj_align(batteryAlert, nullptr, LV_ALIGN_IN_BOTTOM_RIGHT, 0, 0);
+
+  sHour = 99;
+  sMinute = 99;
+  sSecond = 99;
+
```
Riksu9000, line 60: Initialize these in the header.

src/displayapp/screens/WatchFaceFuturistic.cpp
```diff
@@ -0,0 +1,185 @@
...
+  lv_style_init(&arc_fg_style);
+  lv_style_set_line_rounded(&arc_fg_style, LV_STATE_DEFAULT, false);
+  lv_style_set_line_width(&arc_fg_style, LV_STATE_DEFAULT, 8);
+
+  lv_style_init(&arc_bg_style);
+  lv_style_set_line_width(&arc_bg_style, LV_STATE_DEFAULT, 0);
+
```
Riksu9000, line 67: Using styles is usually less efficient than setting local style. Please test the difference in build size.

src/displayapp/screens/WatchFaceFuturistic.h
```diff
@@ -0,0 +1,78 @@
...
+        lv_obj_t* label;
```
Riksu9000, line 54: Give this label a better name.

src/displayapp/screens/WatchFaceFuturistic.cpp
```diff
@@ -0,0 +1,185 @@
...
+  currentDateTime = dateTimeController.CurrentDateTime();
+  if (currentDateTime.IsUpdated()) {
+    uint8_t day = dateTimeController.Day();
+    uint8_t hour = dateTimeController.Hours();
+    uint8_t minute = dateTimeController.Minutes();
+    uint8_t second = dateTimeController.Seconds();
+
+    if (sSecond != second) {
+      lv_arc_set_value(arc_sec, second * (360 / 60));
```
Riksu9000, line 169: You can modify the arc ranges, so you can pass the time without modification.
```suggestion
      lv_arc_set_value(arc_sec, second);
```

Riksu9000 (changes requested):
It's a neat looking watchface, and very cheap in terms of memory use.

Just wondering, why's the time displayed both with the gauges and digitally?

Leon-Plickat:
All done.

> Just wondering, why's the time displayed both with the gauges and digitally?

The digital display is for getting the accurate time, but I personally find that analog watches - and by extension things that emulate them like these gauges - allow for a better feel / intuition for small(-ish) timespans (probably because it maps time to something spatial, which humans are quite good at developing an intuition for).

src/displayapp/screens/WatchFaceFuturistic.cpp
```diff
@@ -0,0 +1,182 @@
...
+    lv_obj_realign(heartbeatValue);
+  }
+
+  batteryPercentRemaining = batteryController.PercentRemaining();
+  isCharging = batteryController.IsCharging();
+  if (isCharging.IsUpdated() || batteryPercentRemaining.IsUpdated()) {
+    if (isCharging.Get()) {
+      lv_obj_set_hidden(plugIcon, false);
+      lv_obj_set_hidden(batteryStatus, false);
+      lv_label_set_text_fmt(batteryStatus, "%03d%%", batteryPercentRemaining.Get());
```
Riksu9000, line 145: I wouldn't force leading zeroes here and probably not in the bpm either. It looks a bit weird to me.

Riksu9000 (commented):
Looks like I'll need to fix the buildsize comparison workflow again. You should look into rebasing this PR and cleaning up the history of commits according our requirements. https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/doc/maintainer-guide.md#commits

PR #1550, opened 2023-01-21
Infineat optimization

Riksu9000:
Infineat was poorly optimized. This PR removes a lot of code and saves a lot of memory.

Fixed 12-hour mode, because the font didn't have the necessary characters.


JF002 (approved):
Great optimizations! And the comment from the bot is really cool! Thanks for this!

PR #1554, opened 2023-01-22
Use checkboxlist in Bluetooth and Chimes settings

Riksu9000:
Only timeout and wakeup screens left, which will require changes in CheckboxList.

### DIFF
--- a/src/displayapp/screens/settings/SettingChimes.cpp
+++ b/src/displayapp/screens/settings/SettingChimes.cpp
@@ -4,70 +4,62 @@
 #include "displayapp/screens/Styles.h"
 #include "displayapp/screens/Screen.h"
 #include "displayapp/screens/Symbols.h"
+#include <array>
 
 using namespace Pinetime::Applications::Screens;
 
 namespace {
-  void event_handler(lv_obj_t* obj, lv_event_t event) {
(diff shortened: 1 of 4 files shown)


src/displayapp/screens/settings/SettingChimes.cpp
```diff
@@ -8,66 +8,57 @@
...
+  struct Option {
+    Pinetime::Controllers::Settings::ChimesOption chimesOption;
+    const char* name;
+  };
 
-  lv_obj_set_style_local_bg_opa(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, LV_OPA_TRANSP);
-  lv_obj_set_style_local_pad_all(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, 10);
-  lv_obj_set_style_local_pad_inner(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, 5);
-  lv_obj_set_style_local_border_width(container1, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, 0);
+  constexpr std::array<Option, 3> options = {{
```
NeroBurner, line 16: missing `<array>` include, the include was removed from the header, but never added to the cpp file

JF002 (approved):

PR #1561, opened 2023-01-23
Update Stopwatch UI *new*

Riksu9000:
Updated design after discussion in #1506. This has the added benefit of making the buttons easier to press.

InfiniSim_2023-01-23_085115


JF002 (approved):

yusufmte:
Awesome!! This may be a separate PR, but while changing the formatting, adding hours to the stopwatch would be awesome. There are a number of uses of stopwatches where that would be helpful, think speedruns, or marathons, or timing a longer commute.

PR #1564, opened 2023-01-24
Comment build size on all PRs

Riksu9000:
This works by uploading the data from the main workflow with low permissions as an artifact, then downloading the data in a workflow with higher permissions to post the comment.

Third party actions are fixed at a commit, in case they get compromised.

Also set the build-firmware VM to ubuntu-22.04, which was missed when updating workflow deps earlier.

This will not run on this PR, because pull_request_target runs in the context of the base branch, but this file doesn't yet exist in develop.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ on:
 
 jobs:
   build-firmware:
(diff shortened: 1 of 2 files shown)


JF002 (approved):
:+1:

PR #1565, opened 2023-01-25
clang-format: Always break template declarations

Riksu9000:
I find this format easier to read, because the definitions are at the expected indentation, making it easier to find what I'm looking for.


JF002 (approved):
I also prefer to add a break after a template declaration!

PR #1566, opened 2023-01-26
infineat: Create colors at compile time.

Riksu9000:
lv_color_hex can't be evaluated at compile time, but LV_COLOR_MAKE can.

I feel like there must be a better approach that I'm missing. Let me know if you have ideas.

### DIFF
--- a/src/displayapp/screens/WatchFaceInfineat.cpp
+++ b/src/displayapp/screens/WatchFaceInfineat.cpp
@@ -17,6 +17,106 @@ namespace {
     auto* screen = static_cast<WatchFaceInfineat*>(obj->user_data);
     screen->UpdateSelected(obj, event);
   }
+
+  enum class colors {
+    orange,
+    blue,
+    green,
+    rainbow,
+    gray,
+    nordBlue,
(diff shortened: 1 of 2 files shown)


NeroBurner (approved):
I like the use of variable names for colors instead of hard coding the color values in the setting code. Letting the compiler do the heavy lifting, and then the PineTime can relax a bit is a very welcome change

I couldn't think of anything better, LGTM

lman0:
@Riksu9000 could you check if your PR , don't impact / increase the ram of infinitime?
Because
I remember in the past that a PR/proposition similar was done but in the end was rejected because live memory in pinetime increased a lot.
(And in contrast the PR was removing firmware memory usage)

JF002:
@lman0 From what I can tell, this PR reduces the flash usage (by removing code) and also RAM usage : it removes the struct WatchFaceInfineat::InfineatColors which was stored in flash and allocated in ram at runtime. It also reduce the stack usage (in ram) by not calling a function.
So I don't think it'll have any negative side effect.

JF002 (approved):

PR #1567, opened 2023-01-27
build-size-workflow: Fix when base gets updated

Riksu9000:
The `github.event.pull_request.base.sha` in the workflow doesn't get updated when there are new commits in the base branch. Instead always checkout the branch to check the sha manually and cache only the build.

The ownership fix seems to still be necessary


JF002 (approved):
What kind of issue is this bug causing?

Riksu9000:
It's comparing to older develop, causing incorrect values.

JF002:
Oh yes, I see... This could be confusing, indeed!

PR #1573, opened 2023-01-31
workflow: Updated workflow comment for all

Riksu9000:
The previous version failed, because it looked for the check from the branch in the fork, but looked for it in the main repo. Now uses the sha to get the commit on which checks were run.

Previous description

This works by uploading the data from the main workflow with low permissions as an artifact, then downloading the data in a workflow with higher permissions to post the comment.

Third party actions are fixed at a commit, in case they get compromised.

Also set the build-firmware VM to ubuntu-22.04, which was missed when updating workflow deps earlier.

### DIFF
--- a/.github/workflows/pr-comment.yml
+++ b/.github/workflows/pr-comment.yml
@@ -1,0 +1,49 @@
+# THIS WORKFLOW HAS WRITE PERMISSIONS TO THE REPO.
+# MAKE SURE IT NEVER RUNS ANY CODE FROM THE FORK
+
+name: PR comment
+
+on:
+  pull_request_target:
+    branches: [ develop ]
+    paths-ignore:
+      - 'doc/**'
+      - '**.md'
+
+jobs:
+  comment:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-comment.yml
```diff
@@ -0,0 +1,49 @@
...
+  comment:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Wait for builds to finish
+      id: wait-for-build
+      uses: fountainhead/action-wait-for-check@297be350cf8393728ea4d4b39435c7d7ae167c93
+      with:
+        checkName: 'Compare build size'
+        token: ${{ secrets.GITHUB_TOKEN }}
+        ref: ${{ github.event.pull_request.head.sha }}
```
Riksu9000, line 23: This is the line that was changed from `${{ github.head_ref }}` and should hopefully fix this issue https://github.com/InfiniTimeOrg/InfiniTime/actions/runs/4051707478/jobs/6970305575

JF002 (approved):

PR #1580, opened 2023-02-10
Move Files needed by factory to a separate document

Riksu9000:
These are instructions for maintainers, and should be in a separate document.

I didn't modify the text, only adjusted the headers.


doc/files-needed-by-factory.md
```diff
@@ -0,0 +1,37 @@
...
+
+```
+/opt/mergehex/mergehex -m ./bootloader.hex ./pinetime-mcuboot-app-image-1.6.0.hex  -o merged-internal.hex
+```
+
+This file must be flashed at offset **0x00** of the internal memory of the NRF52832.
+
+## spinor.bin
+
+This file is the MCUBoot image of the last stable version of the recovery firmware. It must be flashed at offset **0x00** of the external SPINOR flash memory.
```
Riksu9000, line 37: This section could be improved. There's no information on how this file is generated @JF002

JF002 (approved):

PR #1584, opened 2023-02-12
docs: Limit number of jobs used for compilation

FintasticMan:
This is a small change to prevent the use of too many threads locking up the system. This is an issue I've seen in the chat room a few times, especially when people are running in a virtual machine with a small amount of RAM allocated.

All it does is only let `make` use as many jobs as there are threads.

### DIFF
--- a/doc/buildAndProgram.md
+++ b/doc/buildAndProgram.md
@@ -75,7 +75,7 @@ If you just want to build the project and run it on the Pinetime, using *pinetim
 Build:
 
 ```
-make -j pinetime-app
+make -j4 pinetime-app
 ```
 
 List of files generated:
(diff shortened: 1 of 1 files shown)


doc/buildAndProgram.md
```diff
@@ -75,7 +75,7 @@ If you just want to build the project and run it on the Pinetime, using *pinetim
 Build:
 
 ```
-make -j pinetime-app
+make -j $(nproc) pinetime-app
```
Riksu9000, line 78: I'm debating between removing this, keeping this or writing a better explanation. Setting the number of jobs isn't necessary for the build, and this unexplained option could confuse newbies, however it should probably always be set anyway. We could add a comment below about using the `-j` option to speed up the build, but I'm not sure it's our docs job to explain.

FintasticMan: While it doesn't seem like it should be our job to explain what the `-j` option means, I think it would be more confusing to have a build run really slowly, than entering an unknown option. I propose either adding a very small note about that the `-j` option makes it use more threads, just have an unexplained option, or even switch to Ninja, which uses all threads by default.

Riksu9000: I think we should set it to 4 for now. This is better than nothing, less intimidating and easier to type out multiple times.

Riksu9000 (approved):

JF002 (approved):

PR #1587, opened 2023-02-14
Unify docker devcontainer with dockerfile used for CI

BloodStainedCrow:
As discussed in #1497 this removes the need for duplicate Dockerfiles for Docker compilation and .devcontainer development. This will reduce variance in compiler environments and make maintenance easier.


docker/Dockerfile
```diff
@@ -32,6 +32,15 @@ RUN apt-get update -qq \
     && apt-get install -y nodejs \
     && rm -rf /var/cache/apt/* /var/lib/apt/lists/*;
 
+# Add the necessary apt-gets for devcontainer if this is a devcontainer FIXME: This currently does not work
+RUN if [[ "$devcontainer" == "true" ]] ; then \
+    apt-get update -qq \
+        && apt-get install -y \
+            clang-format-12 \
```
FintasticMan, line 39: We've switched to clang-format 14 now, clang-format 12 won't work

.devcontainer/devcontainer.json
```diff
@@ -4,19 +4,23 @@
...
+		"--cap-add=SYS_PTRACE",
+		"--security-opt",
+		"seccomp=unconfined"
+	],
 	// Set *default* container specific settings.json values on container create.
-	"settings": { 
+	"settings": {
 		"terminal.integrated.shell.linux": "/bin/bash",
 		"editor.formatOnSave": true,
 		"clang-format.executable": "clang-format-12"
```
FintasticMan, line 22: Same here

FintasticMan (commented):
I think it would be nice to remove the `.devcontainer` directory, and move the JSON to `.devcontainer.json` in the root. We would need to move the documentation somewhere else though...

BloodStainedCrow:
I would put the .devcontainer/README.md into docs/usingDevcontainers.md, maybe?

FintasticMan:
> I would put the .devcontainer/README.md into docs/usingDevcontainers.md, maybe?

Seems good to me. One more thing, make sure to add a trailing line feed to files, so that the diff doesn't complain.

BloodStainedCrow:
Oh, I thought that would be automatically done via the linter. Will do

BloodStainedCrow:
Nevermind. The linter automatically removes trailing newlines... Is that intended?

FintasticMan:
Which linter are you using? clang-format should add trailing newlines, or at the very least not remove them. We mostly stick to having trailing newlines, and I think that almost all files already have them.

BloodStainedCrow:
Well, I am using the standard linter integrated into the devcontainer haha,
It seems some configuration is busted here....

FintasticMan:
I don't think there should be any difference in build size... The build size it's being compared to is 512ebf07043510930256e3c12498046b510ef8a1, rather than 8e2dcda14fac3e9ebc3b592439ad1f2afbbc7076. For some reason it is comparing against the latest develop rather than the actual base commit.

This doesn't explain the build size difference though, as there are no code changes in develop since the base commit, nor any code changes in this PR. @Riksu9000, have you got any idea what's going on here?

Riksu9000:
The workflow is intended to merge the latest develop and compare to the latest develop. I think the reason for the different commits relates to https://github.com/actions/checkout/issues/1036. It probably picked the commit when commits were pushed, not when running the workflow was approved, but in get-base-ref-size it always picks the latest one. This should be fixed by manually merging with the latest base_ref and a check to make sure the different jobs got the same commit.

As for the build size difference, I can only think of there being some variability between builds. 27c241c7ee51aa3b6e46c6472f7d975d9c702795 and 6dc49e5bdb86ffd5730eb10eb9c477276e79923c are the same size, as is the build from this PR. Only the get-base-ref-size differs.

To avoid straying off topic, let's ignore the build size in this PR.

.vscode/cmake-kits.json
```diff
@@ -0,0 +1,10 @@
+[
+    {
+        "name": "InfiniTime Compiler",
+        "environmentSetupScript": "${workspaceFolder}/docker/build.sh",
+        "compilers": {
+            "CC": "/usr/bin/gcc",
+            "CXX": "/usr/bin/g++"
+        }
+    }
+]
```
Riksu9000, line 10: What's the purpose of this file?

docker/Dockerfile
```diff
@@ -32,6 +33,15 @@ RUN apt-get update -qq \
     && apt-get install -y nodejs \
     && rm -rf /var/cache/apt/* /var/lib/apt/lists/*;
 
+# Add the necessary apt-gets for devcontainer if this is a devcontainer
+RUN if [ -n "$DEVCONTAINER" ] ; then \
+    apt-get update -qq \
+        && apt-get install -y \
+            clang-format-14 \
+            clang-tidy; fi
+
```
Riksu9000, line 42: Why's this needed in the devcontainer, but not in the docker image?

docker/Dockerfile
```diff
@@ -54,5 +64,8 @@ RUN bash -c "source /opt/build.sh; GetNrfSdk;"
 # McuBoot
 RUN bash -c "source /opt/build.sh; GetMcuBoot;"
 
+# If we are a devcontainer add the infinitime user
+RUN if [ -n "$DEVCONTAINER" ] ; then adduser infinitime ; fi
+
```
Riksu9000, line 69: What's this used for in the devcontainer?

.vscode/c_cpp_properties.json
```diff
@@ -8,7 +8,7 @@
                 "${workspaceFolder}/src"
             ],
             "defines": [],
-            "compilerPath": "${env:ARM_NONE_EABI_TOOLCHAIN_PATH}/bin/arm-none-eabi-gcc",
+            "compilerPath": "/usr/bin/gcc",
```
Riksu9000, line 11: Why is this changed?

Riksu9000 (commented):
The currently active maintainers don't use the devcontainer, so you'll need to explain what these changes do.

Please read the commit conventions https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/doc/contribute.md#commit-conventions

docker/Dockerfile
BloodStainedCrow: After the devcontainer is built, VSCode effectively remotes into it. It is convention to not use the root user for this, but instead create another user.

docker/Dockerfile
BloodStainedCrow: The docker image itself is used effectively as a compiler. The container is built, then the compile commands are run and the container is destroyed again. When developing inside a devcontainer the container stays up as the development enviromment.
Since the container doubles as a dev environment we need all tools normally used on the host for "normal" development on the host, like clang-format

.vscode/cmake-kits.json
BloodStainedCrow: This defines the standard compiler used inside the devcontainer. It is a mistake though that the compiler is set to /usr/bin/gcc, that is the standard compiler included in ubuntu.

BloodStainedCrow:
Question: what exactly is the correct compiler we are using? I need to specify the compiler "executable" for Intellisense

Riksu9000:
It's the one specified in the build instructions. https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/doc/buildAndProgram.md

BloodStainedCrow:
This pull request should be ready to merge now

FintasticMan (changes requested):
Sorry it's taken so long to get around to reviewing this. I think that most of it is pretty good. I think it should be possible to get the executable paths etc. from `build.sh` by exporting them to the environment.

.devcontainer.json
```diff
@@ -4,19 +4,24 @@
...
+			"DEVCONTAINER": "true"
+		}
 	},
-	"runArgs": [ "--cap-add=SYS_PTRACE", "--security-opt", "seccomp=unconfined"],
-
+	"runArgs": [
+		"--cap-add=SYS_PTRACE",
+		"--security-opt",
+		"seccomp=unconfined"
+	],
```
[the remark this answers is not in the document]
BloodStainedCrow: You are correct, it seems like these are not necessary

.devcontainer.json
```diff
@@ -25,14 +30,11 @@
...
 		"mjohns.clang-format"
 	],
-
 	// Use 'forwardPorts' to make a list of ports inside the container available locally.
 	// "forwardPorts": [],
-
 	// Use 'postCreateCommand' to run commands after the container is created.
-	// "postCreateCommand": "bash /opt/create_build_openocd.sh",
-
+	"postCreateCommand": "",
```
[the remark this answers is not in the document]
BloodStainedCrow: I thought this was necessary to override the CMD set in the dockerfile. After testing it seems I was wrong

BloodStainedCrow:
Why are the checks failing? The logs imply a check called `Compare build size` is missing? I doubt that is intended behaviour?

BloodStainedCrow:
I have looked into extracting the compilerPath from `build.sh`. I do not see a way to do that currently. It was possible with the cmake extension for vscode since it allows to set a .sh script for setting up env variables. But the Intellisense (C/Cpp) Extension does not seem to support that sadly.
As far as I know this means either there will continue to be a possible desync of the two files or defines like the compiler, clang-format versions or paths need to be centrally defined in a .env file. In either case, that would be out of scope for this PR I think.

If anyone knows if there is way to set VSCode extension settings based on exported variables in a .sh script that would be ideal, I did not find a way.

FintasticMan (approved):
While it's not perfect, mostly due to the copying of variables, this fixes the devcontainer and improves its maintainability.

BloodStainedCrow:
Uh oh, I see the PR now has 176 changed files, did I mess up the merge...?

.vscode/launch.json
```diff
@@ -1,45 +1,33 @@
...
-            "type":"cortex-debug",
-            "cortex-debug.armToolchainPath":"${env:ARM_NONE_EABI_TOOLCHAIN_PATH}/bin",
+            "type": "cortex-debug",
             "cwd": "${workspaceRoot}",
             "executable": "${command:cmake.launchTargetPath}",
             "request": "launch",
             "servertype": "external",
-            // This may need to be arm-none-eabi-gdb depending on your system
-            "gdbPath" : "${env:ARM_NONE_EABI_TOOLCHAIN_PATH}/bin/arm-none-eabi-gdb",
+            // FIXME: This is hardcoded. I have no idea how to use the values set in build.sh here
```
NeroBurner, line 11: Those hard coded paths replace currently working setups that use the environment variables. This is a blocking issue.

Maybe instead of overwriting currently available configs add a new one for the dev container? If this is possible

.vscode/launch.json
```diff
@@ -1,45 +1,33 @@
...
             // Connect to an already running OpenOCD instance
             "gdbTarget": "host.docker.internal:3333",
             "svdFile": "${workspaceRoot}/nrf52.svd",
-            "runToMain": true,
-            // Work around for stopping at main on restart
-            "postRestartCommands": [
-                "break main",
-                "continue"
-            ]
+            "runToEntryPoint": "main"
```
NeroBurner, line 16: Why is this changed?

.vscode/launch.json
```diff
@@ -59,6 +47,5 @@
                 "target/nrf52.cfg"
             ],
         }
-
     ]
-}
+}
```
NeroBurner, line 51: Add newline at end of file again

.vscode/settings.json
```diff
@@ -56,4 +67,4 @@
         "cinttypes": "cpp",
         "typeinfo": "cpp"
     }
-}
+}
```
NeroBurner, line 70: Add newline at end of file again

docker/Dockerfile
```diff
@@ -1,5 +1,6 @@
 FROM ubuntu:22.04
 
+ARG DEVCONTAINER
```
NeroBurner, line 3: What does this do? Is this something specific for the devcontainer to work?

NeroBurner (changes requested):
Thanks for reducing redundancy. But the current version with hard coded paths breaks existing setups. Which isn't desirable

BloodStainedCrow:
Sorry for the long radio silence. Life happened D:
Also I screwed up the merge AGAIN. It shouldn't be so hard, should it...

docker/Dockerfile
BloodStainedCrow: This is an old artifact that is no longer used. I will remove it

.vscode/launch.json
BloodStainedCrow: I will add another launch configuration for launching in a devcontainer. That does not solve the problem of having the path/version be hardcoded but it will not break the existing workflows, which rely on the environment variables.

.vscode/launch.json
BloodStainedCrow: I was under the false illusion it was no longer necessary, I will re-add the "postRestartCommand". runToMain is deprecated and was replaced by runToEntryPoint.

BloodStainedCrow:
If someone is willing to explain the dark art of updating a fork, without having all changes from the source branch showing up in my PR, to an apparent newbie, I would be forever grateful...

mark9064:
The best way is probably to rebase:
- Get InfiniTimeOrg/InfiniTime main up to date (pull that branch)
- Switch to your branch
- `git rebase <name of branch tracking InfiniTimeOrg/InfiniTime main>`
- Resolve conflicts as needed

Just in case things go super wrong, I'd make a copy of the entire repository folder before as a backup

BloodStainedCrow:
Okay, I did as you described. According to git I can now pull 22 commits and push 44 commits onto my unifyDockerDevcontainer branch. Does that sound right? The git log output suggests my changes will be applied after all changes on main, which seems correct.

FintasticMan:
That's right, but you shouldn't pull those 22 commits, you should force push the current state of your local branch to GitHub.

mark9064:
Sounds good to me. The commit history should be something like
Oldest
- Fix code formatting
- Application selection at build time
- Fix link in build with VS
...
- Clean outdated comment
- (branch merge may still be here, doesn't matter if it is or isn't really)
Newest

If so you're good to force push

BloodStainedCrow:
@NeroBurner I have fixed all the issues you pointed out, as far as I know.
There are 3 things though:

1. I noticed, in launch.json is a old launch configuration that no longer works. It has version pinetime-app-1.3.0 hardcoded and has therefore been broken since 1.4. It also reference a bunch of other files that I believe to no longer exist.
2. The docs for using VSCode/Devcontainers are very wrong/outdated. They are without any changes from this PR, and will be even more so afterwards. I plan on updating the docs, question is: Is this part of this PR or should I make another one for it?
3. There a couple of instances where I could not avoid hardcoding variables into launch-configs/settings etc. (It might be possible, but I have not found a way). If the version of clang-format/gdb ever changes, the devcontainer WILL break. Ideally we could have a test that fails if the hardcoded envs I use differ from the ones used in build.sh which I used as the baseline for everything. Then whenever the version is changed there, it must also be changed in the files for the devcontainer.

BloodStainedCrow:
Also, if someone who does NOT use a devcontainer could verify, that I did not break anything with my changes, that would be great.

BloodStainedCrow:
I also still need to test debugging from a devcontainer

BloodStainedCrow:
When debugging I get the error `error while loading shared libraries: libncurses.so.5: cannot open shared object file: No such file or directory` when starting gdb. Will investigate further

BloodStainedCrow:
Okay, debugging seems to work now. I do not seem to get any source mapping, though. Is that normal?

agittins:
It would be great to see this merged. I notice that the issues raised by @NeroBurner's review look to be resolved, and I was able to rebase this PR against main on my system without any issues, and the container builds for me OK.

Are any maintainers able to merge this, or will @BloodStainedCrow need to rebase and push first? It would be great to get the devcontainer setup working again so we can iterate any improvements from there.

.vscode/c_cpp_properties.json
```diff
@@ -10,6 +15,21 @@
...
+        {
+            "name": "nrfCC Devcontainer",
+            "includePath": [
+                "${workspaceFolder}/**",
+                "${workspaceFolder}/src/**",
+                "${workspaceFolder}/src"
+            ],
+            "defines": [],
+            "compilerPath": "${TOOLS_DIR}/${GCC_ARM_PATH}/bin/arm-none-eabi-gcc",
+            "cStandard": "c11",
```
NeroBurner, line 32: Infinitime C standard is 99
https://github.com/InfiniTimeOrg/InfiniTime/blob/9a5f516c427f329bbc161f794bae18908fcd33c5/CMakeLists.txt#L7

.vscode/launch.json
```diff
@@ -51,14 +47,37 @@
...
+        },
+        {
+            "name": "Debug - Openocd Devcontainer",
+            "type": "cortex-debug",
+            "cwd": "${workspaceRoot}",
+            "executable": "${command:cmake.launchTargetPath}",
+            "request": "launch",
+            "servertype": "external",
+            // FIXME: This is hardcoded. I have no idea how to use the values set in build.sh here
+            "gdbPath": "/opt/gcc-arm-none-eabi-10.3-2021.10/bin/arm-none-eabi-gdb",
```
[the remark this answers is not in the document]
BloodStainedCrow: As far as I know (see my previous attempts above) VSCode currently does not support loading variables for launch.json from anything but launch.json itself, which sadly means without some serious hacks there will be this reference here which might become outdated.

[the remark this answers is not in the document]
BloodStainedCrow: Looking at this though, the very least I should do is unify the occurrences of it in a variable in launch.json so only one value can grow stale

BloodStainedCrow:
I tried using the devcontainer again, and found a couple snags:

1. Building the devcontainer failed due to bad CRLF (docker wants LF) line endings in `build.sh`
2. Compiling failed because of bad CRLF (the container/python/`src/displayapp/fonts/generate.py`) wants LF) in `src/displayapp/fonts/jetbrains_mono_bold_20.*`
3. Connecting the OpenOCD debugger failed (silently *rage*) because of a missing dependency: `libncurses5`

1. and 2. were already known by me: I currently attribute them to me using WSL2 on Windows and Docker Desktop which is known to have some bugs leading to weirdness with line endings. If either someone else who knows a little about devcontainers (or me depends on who is faster) can test it on a Linux with docker machine and see if that changes anything.
I presume changing the line endings on these files is a breaking change for some people and therefore not desired?

3. is simply fixed by adding the dependency to the dockerfile.

BloodStainedCrow:
Additionally there does not seem to be any source mapping for debugging, I will have to look into that aswell

BloodStainedCrow:
Okay, I have now found the correct setting to add source/symbol maps to gdb for debugging. As far as I can tell: `pinetime-app-1.14.0.map` is the correct symbol map. But gdb errors out with: `can't read symbols: file format not recognized`.
I was unable to find **any** info on using .map files with gdb so either I do not know what the correct symbol map file is, or I am missing something else. Is `pinetime-app-1.14.0.map` correct?

agittins:
> 4\. I currently attribute them to me using WSL2 on Windows

Indeed. The line endings however is due to git using the local encoding when you check out the repo (ie, in windows, so it gives them CR/LFs) but then when that repo is accessed inside the container all the tools expect unix textfiles.

So the actual repo doesn't contain any CR/LF files, they get converted to that when you check them out under windows.

We hit this when I was first trying to tackle the whole devcontainers thing. I think this comment provided the best general solution:

https://github.com/InfiniTimeOrg/InfiniTime/pull/1401#issuecomment-1296383632
> Even Microsoft itself recommends setting eol=lf when working with devcontainers on Windows.
>
> Another option could be to instruct Windows users to clone with --config core.eol=lf if they want to use devcontainers. > That wouldn't be seamless, but I suppose there really isn't any way to make this completely seamless (thanks Windows).

Ie, if you do the initial checkout in windows with `eol=lf` then it all works, and not doing so doesn't have any effects on those not using devcontainers.

BloodStainedCrow:
Yikes, thanks windows indeed

NeroBurner:
@BloodStainedCrow could you recheck your open points with a `LF` checked out repo on WSL?

BloodStainedCrow:
Both the failed container build and the failed compile are solved with this.

The only remaining problem I encountered is the missing source map/ symbol file for debugging.
Do you know which is the correct symbol file for gdb?

.vscode/launch.json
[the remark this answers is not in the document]
BloodStainedCrow: I was mistaken, it seems is not possible to share variables (outside a .env file, but that would be out of scope for this PR IMO) between mutiple launch.json configurations.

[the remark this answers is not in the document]
BloodStainedCrow: Additionally I currently have no way to test the other configuration on, whether they even still work in their current state, from what I can tell, at least one of them is around 3 years out of date...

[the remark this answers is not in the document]
NeroBurner: Not a different compiler? We just need one (the currently officially supported one)

Or what configurations are you talking about. If they are deprecated the maybe we can remove those?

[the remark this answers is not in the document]
BloodStainedCrow: In `launch.json` there are different launch configurations usable via a button in VSCode. These are what I meant.

While I am open to removing them/replacing them with ones that actually work, I would probably like some more input from people who might use them, so maybe in another PR...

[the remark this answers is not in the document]
NeroBurner: Thanks for the clarification. I agree. Let's merge this!

NeroBurner (approved):
Thanks for keep on going, improving and keeping this PR up to date!

BloodStainedCrow:
No problem, one question before I agree this can be merged: 

> The only remaining problem I encountered is the missing source map/ symbol file for debugging. Do you know which is the correct symbol file for gdb?

NeroBurner:
> No problem, one question before I agree this can be merged:
> 
> > The only remaining problem I encountered is the missing source map/ symbol file for debugging. Do you know which is the correct symbol file for gdb?

I don't sorry. Let's fix that in a separate PR 😁

PR #1591, opened 2023-02-18
lvgl: Use InfiniTime fork of LVGL

Riksu9000:
This is an updated fork of lvgl 7.11 including the patches by joaquim.


JF002:
What is the starting point of this branch? Did you create it from a specific tag? I noticed that the last commit on tag 7.11.0 from LVGL dates from june 2021, while the last commit in the InfiniTimeOrg branch is from 2 weeks ago.

Riksu9000:
LVGL is still maintaining v7 in the release/v7 branch. This branch is based on that.

JF002:
Do you think it's a good idea to use a non-tagged version of the lib?

Riksu9000:
All of the commits since the last tag are fixing issues, so I think using it is fine. I'm not sure they're going to make releases anymore.

JF002:
Ok, great you've already checked everything :+1:

JF002 (approved):

Issue #1099, opened 2022-04-20
No steps are counted


jeena:
### Verification

- [X] I searched for similar bug reports and found none was relevant.

### What happened?

My step counter is always at 0

### What should happen instead?

The step counter should count the steps

### Reproduction steps

I'm not sure how to reproduce it. I can only say what happened. One day I woke up and my PineTime, which I until then had for a couple of weeks and didn't update the software for it, showed me that I had walked 536 bilion steps that morning already.

![steps](https://user-images.githubusercontent.com/151764/164152798-a5042332-df80-4fec-a78c-f1995ea6aa9f.jpg)

I didn't do anything to it and it ran out of battery some day and after that once I charged it and booted it it showed 0 steps, even after walking for a long time with it on my wrist.

![20220420_135247](https://user-images.githubusercontent.com/151764/164153147-26231627-66d8-469b-a933-ce61aabbab29.jpg)

### More details?

I did a factory reset which didn't help. I also Upgraded the firmware to the newest version 1.9.0 and it didn't help either.

There is this gyroscope application on it, and before that happened it showed different colored lines when I was moving the watch. This application does not show anything now anymore either.

https://user-images.githubusercontent.com/151764/164153369-a74d3846-5eec-4d8d-ab70-6a2bdc4819e7.mp4

I'm not sure if this is a hardware error or a software error. It didn't fall down or anything, I just woke up one day with it on my wrist and it was like that.

### Version

1.9.0

### Companion app

Gadgetbridge 0.65.0

alwayslurking:
I saw exactly the same symptoms, and found this discussion on the pine forums:

https://forum.pine64.org/showthread.php?tid=16182

Somewhere there unsealed their watch and found a loose chip. I now wonder if an intermittent accelerometer connection was the cause of other random problems I had with my watch, now all resolved at the expense of nothing that uses the accelerometer working.

jeena:
Very interesting. I have no other issues but the accelerometer, thanks for the link, good to know that I'm not the only one.

JF002:
It looks like InfiniTime cannot initialize the accelerometer or that it's not working at all.
Could you have a look at the 'About' page (last entry in Settings). On the 2nd page, there's a field  'Accel.' What does it show (BMA421, BMA425 or ???).

Anyway, I think this is mostly a hardware issue, and I don't really know if there's anything we can do on the software side to solve that issue :/

sssmack:
??? for me too.
[...]

PR #1603, opened 2023-02-20
motion: Disable Motion app

Riksu9000:
This is a debugging app, not useful for most people. Also remove the app icon.


JF002:
Oh wow! 6KB for this app? What's using all this space? The graph widget from LVGL, maybe?

Riksu9000:
It must be the chart widget. LVGL things always seem like candidates for memory savings to me.

JF002:
Interesting! Most of the time, I do not like removing features from the firmware but... this one makes sense : the memory usage for this test/debug app is not reasonable.

I'm wondering : should we provide another (simpler) test app for the motion sensor?

Riksu9000:
I don't think removing features is a bad idea if it will improve the polish of the firmware, which removing a debug app from release firmware definitely does. I don't think a test app needs to be added, as the detected accelerometer in SystemInfo seems to usually be ??? when the accelerometer is not working #1099.

RageGamerBoi:
Possibly just a live updating readout of the raw accelerometer values in Settings could serve the same purpose.

JF002:
>  I don't think a test app needs to be added, as the detected accelerometer in SystemInfo seems to usually be ??? when the accelerometer is not working.

Fair enough !

> Possibly just a live updating readout of the raw accelerometer X, Y, and Z values in Settings (and Steps too, I guess) could serve the same purpose.

Yes, it could, but it won't bring any other value than checking that the sensors returns some values. As Riksu said, we can check that the motion sensor is detected in SystemInfo so we probably don't need another app for that purpose.

JF002 (approved):

PR #1622, opened 2023-02-21
libs: Disable clang-format and -tidy

Riksu9000:
Disable linters in the libs folder.


NeroBurner (approved):
yes, not our code to fix ;)

PR #1623, opened 2023-02-21
inactivity: Use LVGL inactivity timers

Riksu9000:
Replace custom FreeRTOS inactivity timers with LVGL inactivity timers.

There may be some more optimizations possible, for example the restore brightness message can probably be replaced with `lv_disp_trig_activity(nullptr);`

~~- [ ] Fix an issue where setting the timer to the exact timeout setting causes the watch to go to sleep at the time the timer expires.~~ Looks like this isn't a new issue and shouldn't be blocking.

### DIFF
--- a/src/systemtask/SystemTask.h
+++ b/src/systemtask/SystemTask.h
@@ -84,6 +84,10 @@ namespace Pinetime {
       void OnIdle();
       void OnDim();
 
+      bool IsSleepDisabled() {
+        return doNotGoToSleep;
+      }
+
       Pinetime::Controllers::NimbleController& nimble() {
         return nimbleController;
       };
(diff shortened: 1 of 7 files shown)


src/systemtask/SystemTask.h
```diff
@@ -90,6 +90,10 @@ namespace Pinetime {
       void OnIdle();
       void OnDim();
 
+      bool IsSleepDisabled() {
+        return doNotGoToSleep;
+      }
+
```
Riksu9000, line 96: This should probably be moved to displayapp, but I think this would require a lot more rework, so it should be done another time.

JF002: I have to admit that the way "go to sleep" was implemented is a bit... tricky (multiple messages are exchanged between SystemTask and DisplayApp). This is because we need to ensure that DisplayApp stops sending new frames to the display and go to sleep before SystemTask can put devices to sleep (including the display and  SPI bus).

JF002 (approved):
LGTM, I didn't know LVGL provided these APIs.

FintasticMan (approved):
LGTM!

PR #1629, opened 2023-02-22
Remove DisplayApp parameter from Screen base

Riksu9000:
DisplayApp was passed to all screens due to the Screen class requiring it, but it didn't use it for anything. Here I removed DisplayApp from the base class and removed all unused DisplayApp parameters.


src/displayapp/screens/settings/Settings.h
```diff
@@ -20,6 +20,7 @@ namespace Pinetime {
         bool OnTouchEvent(Pinetime::Applications::TouchEvents event) override;
 
       private:
+        DisplayApp* app;
```
NeroBurner, line 23: I think this reference is unused and can be removed, and later also the passed variable from the constructor can be removed

src/displayapp/screens/Notifications.h
```diff
@@ -69,6 +69,7 @@ namespace Pinetime {
         };
 
       private:
+        DisplayApp* app;
```
NeroBurner, line 72: unused, may be removed from class

NeroBurner (changes requested):
great work slimming down the interfaces! I think some more classes/constructors may be slimmed down (cascading remove of unused `*app` variable)

src/displayapp/screens/Notifications.h
Riksu9000: This is used to call `SetFullRefresh()` function for the dismissal feature.

src/displayapp/screens/settings/Settings.h
Riksu9000: This is passed to the List class, for calling `app->StartApp()`

NeroBurner: whoops, overlooked that

NeroBurner (approved):
seems like I was wrong, great work :)

JF002 (approved):
That's a huge PR! Good work! I guess Screen and other apps needed a ref to DisplayApp before, but previous refactoring have probably already fixed that!

PR #1631, opened 2023-02-23
Touchhandler rework

Riksu9000:
Moved LVGL handling out of TouchHandler, so it just processes Cst816S data.
Removed Cst816S dependency from TouchHandler. Cst816S::TouchInfo is simply passed to the processing function.

Outside the scope of this PR, but since TouchHandler is just processing Cst816S data to an InfiniTime friendly format, this code could in the future be moved inside the driver itself, so different touch drivers would have their own processing functions.

### DIFF
--- a/src/displayapp/LittleVgl.h
+++ b/src/displayapp/LittleVgl.h
@@ -24,7 +24,8 @@ namespace Pinetime {
       void FlushDisplay(const lv_area_t* area, lv_color_t* color_p);
       bool GetTouchPadInfo(lv_indev_data_t* ptr);
       void SetFullRefresh(FullRefreshDirections direction);
-      void SetNewTouchPoint(uint16_t x, uint16_t y, bool contact);
+      void SetNewTouchPoint(int16_t x, int16_t y, bool contact);
+      void CancelTap();
 
       bool GetFullRefresh() {
         bool returnValue = fullRefresh;
@@ -60,9 +61,9 @@ namespace Pinetime {
       uint16_t writeOffset = 0;
       uint16_t scrollOffset = 0;
 
-      uint16_t tap_x = 0;
-      uint16_t tap_y = 0;
+      lv_point_t touchPoint = {0};
       bool tapped = false;
+      bool isCancelled = false;
     };
   }
 }
--- a/src/displayapp/LittleVgl.h
+++ b/src/displayapp/LittleVgl.h
@@ -61,7 +61,7 @@ namespace Pinetime {
       uint16_t writeOffset = 0;
       uint16_t scrollOffset = 0;
 
-      lv_point_t touchPoint = {0};
+      lv_point_t touchPoint = {};
       bool tapped = false;
       bool isCancelled = false;
     };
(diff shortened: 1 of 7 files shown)


src/touchhandler/TouchHandler.h
```diff
@@ -3,42 +3,34 @@
...
 
       Pinetime::Applications::TouchEvents GestureGet();
 
     private:
-      Pinetime::Drivers::Cst816S::TouchInfos info;
-      Pinetime::Drivers::Cst816S& touchPanel;
-      Pinetime::Components::LittleVgl& lvgl;
       Pinetime::Applications::TouchEvents gesture;
-      bool isCancelled = false;
+      TouchPoint currentTouchPoint = {0};
```
NeroBurner, line 33: without a constructor what is the contents of the initial `currentTouchPoint`. It feels for me as it is `x=0` and `y=undefined` and `touching=undefined`

src/systemtask/SystemTask.cpp
```diff
@@ -342,11 +342,10 @@ void SystemTask::Work() {
           // TODO add intent of fs access icon or something
           break;
         case Messages::OnTouchEvent:
-          if (touchHandler.GetNewTouchInfo()) {
-            touchHandler.UpdateLvglTouchPoint();
+          if (touchHandler.ProcessTouchInfo(touchPanel.GetTouchInfo())) {
+            ReloadIdleTimer();
+            displayApp.PushMessage(Pinetime::Applications::Display::Messages::TouchEvent);
```
NeroBurner, line 347: is the change from "sending all touch events to displayapp including the invalid ones" to "send only valid input events to displayapp" intentional? Please explain why this change is needed as I don't know

src/displayapp/LittleVgl.h
```diff
@@ -60,9 +61,9 @@ namespace Pinetime {
       uint16_t writeOffset = 0;
       uint16_t scrollOffset = 0;
 
-      uint16_t tap_x = 0;
-      uint16_t tap_y = 0;
+      lv_point_t touchPoint = {0};
```
NeroBurner, line 64: I think this should be initialized with `{}`, otherwise I think that only the `x` variable is set to `0`, the other variable `y` is undefined. I'm not sure, this should be tested

NeroBurner (changes requested):

src/systemtask/SystemTask.cpp
Riksu9000: Previously `UpdateLvglTouchPoints()` was called here when touchinfo is valid. Now DisplayApp updates these points, so in order to update the points using valid data only, the message is only pushed when the data is valid. It could've always been like this, it doesn't make a big difference.

src/touchhandler/TouchHandler.h
Riksu9000: I think they get implicitly initialized to zero, but explicitly initializing only the first variable is misleading indeed.

NeroBurner (approved):

PR #1633, opened 2023-02-23
TimerController: Use chrono for durations

Riksu9000:


NeroBurner (approved):
Awesome usage of `std::chrono`! Loving the explicitness of this solution

PR #1634, opened 2023-02-23
DateTimeController: Use std::tm for storing date

Riksu9000:


src/components/datetime/DateTimeController.h
```diff
@@ -132,13 +136,7 @@ namespace Pinetime {
...
 
     private:
-      uint16_t year = 0;
-      Months month = Months::Unknown;
-      uint8_t day = 0;
-      Days dayOfWeek = Days::Unknown;
-      uint8_t hour = 0;
-      uint8_t minute = 0;
-      uint8_t second = 0;
+      std::tm localTime;
```
NeroBurner, line 139: missing include for `std::tm`, please add `#include<ctime>` at the top

NeroBurner (changes requested):
small nitpick because of include, but other than that looking good. Thanks for teaching me the existence of `std::tm` from the `C` world. I only knew about the `std::chrono` and `date` library for datetimes

NeroBurner (approved):

JF002:
> Thanks for teaching me the existence of std::tm from the C world.

Same for me :)

JF002 (approved):

PR #1635, opened 2023-02-23
Remove date lib

Riksu9000:
DateTimeController can provide everything we need. Depends on #1634.

### DIFF
--- a/src/displayapp/screens/WatchFaceCasioStyleG7710.cpp
+++ b/src/displayapp/screens/WatchFaceCasioStyleG7710.cpp
@@ -1,6 +1,5 @@
 #include "displayapp/screens/WatchFaceCasioStyleG7710.h"
 
-#include <date/date.h>
 #include <lvgl/lvgl.h>
 #include <cstdio>
 #include "displayapp/screens/BatteryIcon.h"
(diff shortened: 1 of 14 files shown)


src/displayapp/screens/WatchFaceCasioStyleG7710.cpp
```diff
@@ -280,22 +274,19 @@ void WatchFaceCasioStyleG7710::Refresh() {
...
-      uint8_t weekNumber;
-      uint16_t dayOfYearNumber, daysTillEndOfYearNumber;
-
       time_t ttTime =
         std::chrono::system_clock::to_time_t(std::chrono::time_point_cast<std::chrono::system_clock::duration>(currentDateTime.Get()));
       tm* tmTime = std::localtime(&ttTime);
 
-      dayOfYearNumber = tmTime->tm_yday + 1; //  tm_yday  day of year [0,365] => yday+1
-      daysTillEndOfYearNumber = (yearMonthDay.year().is_leap() ? 366 : 365) - dayOfYearNumber;
+      int daysInCurrentYear = (year % 4 == 0 && year % 100 != 0) || year % 400 == 0 ? 366 : 365;
```
NeroBurner, line 281: Are we C++20 yet? Then we could use https://en.cppreference.com/w/cpp/chrono/year/is_leap

if not we could add a TODO to use the `is_leap()` from C++20 standard as soon as we switch to it

Riksu9000: There's another place where this could be used, so I added two comments.

JF002 (approved):

PR #1636, opened 2023-02-23
watchdog: Replace WatchdogView with const Watchdog

Riksu9000:
Works the same way.


NeroBurner (approved):
good use of `const` keyword. Less code duplication. LGTM

JF002 (approved):

JF002:
> good use of `const` keyword. Less code duplication. LGTM

I agree! It's pretty funny to see @Riksu9000 fixing all my design errors! :D

PR #1637, opened 2023-02-23
displayapp: Make class references const

Riksu9000:
This isn't everything, but just those that were easy to change to const in the entire DisplayApp.


src/displayapp/LittleVgl.h
```diff
@@ -4,15 +4,14 @@
...
   namespace Drivers {
-    class Cst816S;
     class St7789;
   }
 
   namespace Components {
     class LittleVgl {
     public:
       enum class FullRefreshDirections { None, Up, Down, Left, Right, LeftAnim, RightAnim };
-      LittleVgl(Pinetime::Drivers::St7789& lcd, Pinetime::Drivers::Cst816S& touchPanel);
```
NeroBurner, line 15: did you get a new warning that the touchPanel variable is unused because you made it `const`?

I think we should remove this unused variable in a separate PR before merging this constness change

NeroBurner (changes requested):
good constness change, I'd just separate out the `touchPanel` cleanup into its on PR

src/displayapp/LittleVgl.h
Riksu9000: I think I just happened to notice. Is it really a problem having it in this PR though? Seems pointless changing it if it's going to get removed anyway.

NeroBurner: I'd merge the split off PR first and then this one. This way we have one PR doing one thing each. Now we've mixed changes. Furthermore the removal has the description of "making more const references". In the end it doesn't matter. It just feels cleaner this way

NeroBurner (approved):
I'm happy now, thanks :+1:

PR #1639, opened 2023-02-24
DateTimeController: Remove SetTime tick parameter

Riksu9000:
The systick doesn't need to be updated when setting time.

Also removed unused nrf_rtc.h includes.

I guess this change will make it so depending on the previous systick value, the time can shift an amount less than one second. Is this a problem?

### DIFF
--- a/src/components/ble/CurrentTimeClient.cpp
+++ b/src/components/ble/CurrentTimeClient.cpp
@@ -1,5 +1,4 @@
 #include "components/ble/CurrentTimeClient.h"
-#include <hal/nrf_rtc.h>
 #include <nrf_log.h>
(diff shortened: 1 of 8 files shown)


JF002:
> I guess this change will make it so depending on the previous systick value, the time can shift an amount less than one second. Is this a problem?

I think you're right : the time that elapsed between the last call to UpdateTime() with a valid systick and the call to SetTime() will be lost. Is this an issue? Probably not, since was are setting a new time anyway.

And yeah.. it does not make sense that the rest of the application needs to have access to the systick when setting  a new time so I think this PR is valid :+1!

JF002 (approved):

PR #1643, opened 2023-02-25
CMakeLists: Remove version from build filenames

Riksu9000:
Naming the files based on version isn't useful, and results in a lot of clutter.

After this change I could remove some unnecessary variables.


Riksu9000:
Maybe we should remove post_build.sh first? I assume that's causing the build check to fail.

JF002:
The version in the filename is still a bit useful to me : it makes storage and distribution of the file easier : no need to wonder what's the version of a specific build. It also make the publishing of the artifacts easier since I don't need to rename them manually before adding them to the release note.

Do you really think we should remove it?

Riksu9000:
I don't think local builds should have the version in the filename, but CI artifacts are already named after the branch, so we could also make release builds include the version. Would this be a better solution?

Riksu9000:
Although they don't necessarily need to be renamed for the release anyway.

NeroBurner:
I'm in favor of keeping version numbers, as they have some information. I'd even put more information into the generated zip files (I do that at work a lot, as it helps to identify laying around zips for my colleagues and me)

Riksu9000:
I'll open an issue about this later, so we can discuss.

PR #1645, opened 2023-02-26
DateTimeController: Fix integer overflow on day of year

Riksu9000:
`tm_yday` is an int and is a number between 0 and 365.


NeroBurner (commented):
i'd keep the unsigned information, or has that drawbacks?

Riksu9000:
`tm_yday` is an int. We shouldn't change the type.

NeroBurner (approved):

PR #1648, opened 2023-02-26
LVGL / FS : Initialize the LVGL FS driver in LittleVgl (instead of FS).

JF002:
Previously, the LVGL driver for the filesystem was initialized in the class FS. However, since 6f942e2, the order of the initializations was incorrect  : the driver was initialized (FS::LVGLFileSystemInit()) before LVGL (LittleVgl.Init()), which means that the driver registration was probably dropped when LVGL was initialized.

The LVGL driver is now initialized in LittleVgl.Init(), which seems to make much more sense, since all LVGL drivers are initialized there. This way, we ensure that the initialization of the drivers is consistent.

### DIFF
--- a/src/displayapp/LittleVgl.cpp
+++ b/src/displayapp/LittleVgl.cpp
@@ -4,6 +4,8 @@
 #include <FreeRTOS.h>
 #include <task.h>
 #include "drivers/St7789.h"
+#include "littlefs/lfs.h"
+#include "components/fs/FS.h"
 
 using namespace Pinetime::Components;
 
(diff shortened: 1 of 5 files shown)


src/displayapp/LittleVgl.cpp
```diff
@@ -12,6 +14,45 @@ namespace {
     lv_theme_t* theme = lv_pinetime_theme_init();
     lv_theme_set_act(theme);
   }
+
+  namespace {
```
Riksu9000, line 18: Anonymous namespace twice is a bit strange.

JF002: Oh yes, right... let me fix this...

JF002:
Sorry for all these force pushes... I'm fixing formatting errors :)

Riksu9000 (approved):

PR #1652, opened 2023-02-27
Counter: cleanup container widget

some00:


Riksu9000:
What's the purpose of this change? When the screen is closed, all objects belonging to that screen are deleted. This increases the build size a bit.

some00:
I want to add a Counter in a Setting screen. Without deleting the container the list is drawn over the counter when I backout. The label is visible between list items.

some00:
Thanks. Yes, I created my own settings screen and missed the destructor.

PR #1654, opened 2023-02-28
Timer Alert

TannerStocks:
This adds alert functionality to the Timer app. The alert can be dismissed or snoozed using the on screen buttons, snoozed with the watch side button, or auto-snoozed after the alert runs for a time.


Riksu9000:
The timer "alerts" the user, so maybe you mean ringing. There are other PRs and an issue about this. We need to come to a solution before we can discuss implementations.
https://github.com/InfiniTimeOrg/InfiniTime/issues/391#issuecomment-1383231455
https://github.com/InfiniTimeOrg/InfiniTime/pull/1499

PR #1659, opened 2023-03-02
motioncontroller: Clean up class and remove unused vars

FintasticMan:
This PR cleans up the MotionController class and some of the code surrounding it. None of the functionality is changed at all.


Riksu9000:
There are quite a few changes here. The ShakeWake changes and RaiseWake variable name change I'm not sure about.

FintasticMan:
It's quite difficult to split these changes into separate commits because of how intertwined they are.

I think that `lastYForRaiseWake` is actually a better name for the variable, so I'll rename it to that. I would have liked to have just used the `lastY` variable, but that's not easy to do without changing any behaviour.

The ShakeWake changes make the code easier to read IMO, the only behaviour change is that the first time the `ShouldShakeWake` function is called after the watch goes to sleep, the time delta is lower than it would have been before this change. This means that it's possible for it to wake up immediately after going to sleep if the user is shaking the watch. Before it would only wake up a hundred milliseconds or so later.

Riksu9000:
I don't see how they're that intertwined. It's difficult to isolate what's been moved where with unrelated changes, while also reasoning about them.

FintasticMan:
@Riksu9000 I've just split my changes into separate commits, so now it should be easier to understand what's going on.

src/components/motion/MotionController.cpp
```diff
@@ -22,11 +22,10 @@ void MotionController::Update(int16_t x, int16_t y, int16_t z, uint32_t nbSteps)
...
 
-  int32_t deltaSteps = nbSteps - this->nbSteps;
-  this->nbSteps = nbSteps;
-  if (deltaSteps > 0) {
-    currentTripSteps += deltaSteps;
+  if (nbSteps > this->nbSteps) {
+    currentTripSteps += nbSteps - this->nbSteps;
   }
+  this->nbSteps = nbSteps;
 }
```
Riksu9000, line 29: The previous code could handle nbSteps overflowing... in case someone walks 2^32 steps in a day... Regardless maybe we don't need to change this.

Riksu9000 (commented):
Thanks. This is much better.

src/components/motion/MotionController.cpp
FintasticMan: I don't see how my code is any worse at handling overflows than the old code. Seeing as nbSteps is stored as a `uint32_t`, overflows just reset it to 0, and so I believe the code does the exact same.

Anyway, I mostly wanted to change this because the conversion from unsigned to signed didn't sit entirely right to me, and this new code is more obvious to me. If the old code makes more sense to you, I'd be happy to revert it though.

Riksu9000: If nbSteps overflows, it's not larger than the previous steps, so the trip steps isn't updated. Using the deltaSteps variable, subtracting 4294967276 from 20 results in 40, so it doesn't miss the trip steps.

FintasticMan: Oh, yeah you're right. Oops. Anyway, seeing as that is an impossibility in reality (unless the accelerometer is returning bogus values, in which case it doesn't matter anyway), I think what matters more is how nicely the code reads. What do you think?

Riksu9000: I think the old code is the most readable after moving this->nbSteps away from the middle.

Riksu9000 (approved):

src/components/motion/MotionController.h
```diff
@@ -1,8 +1,12 @@
 #pragma once
 
 #include <cstdint>
-#include <drivers/Bma421.h>
-#include <components/ble/MotionService.h>
+
+#include <FreeRTOS.h>
+#include <portmacro_cmsis.h>
```
JF002, line 6: Why do we need this include (asking because I don't think we use it anywhere else in the code)? And should'nt we include `portmacro.h` instead?

FintasticMan: That is the file that defines `TickType_t`, and my editor just automatically added an include for it. It seems that `FreeRTOS.h` includes `portable.h` which includes `portmacro.h` which includes `portmacro_cmsis.h`. Because of this, it doesn't need to be explicitly included in our code, and I think it's better to remove it.

PR #1660, opened 2023-03-02
MultiPurposeButton: Add new widget

Riksu9000:
New widget for potentially improving readability on code, where a button is used for many purposes.

Please let me know if this improves the readability of code. Otherwise there's probably no reason to add this. The index could use an enum, but the name of the variable can also be helpful, for example in `btnStopLap` Stop is the first word, and it is index 0.

Use MultiPurposeButton in StopWatch.

### DIFF
--- a/src/displayapp/screens/StopWatch.cpp
+++ b/src/displayapp/screens/StopWatch.cpp
@@ -16,41 +16,54 @@ namespace {
     return TimeSeparated_t {mins, secs, hundredths};
   }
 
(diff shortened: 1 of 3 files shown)


FintasticMan:
I personally don't think this implementation makes the code easier to read, mostly due to the magic numbers used as IDs for the buttons. Perhaps if you make an enum for the buttons, so that it becomes `btnPlayPause.SetActiveButton(PlayButton);`, it might make it a worthwhile change IMO.

PR #1667, opened 2023-03-03
WatchFacePineTimeStyle: Simplify time update check

Riksu9000:
This isn't as thorough as the other PRs, because simply using another time point to check if a minute has passed would've used more memory.

### DIFF
--- a/src/displayapp/screens/WatchFacePineTimeStyle.cpp
+++ b/src/displayapp/screens/WatchFacePineTimeStyle.cpp
@@ -446,16 +446,13 @@ void WatchFacePineTimeStyle::Refresh() {
     AlignIcons();
   }
 
-  currentDateTime = dateTimeController.CurrentDateTime();
+  currentDateTime = std::chrono::time_point_cast<std::chrono::seconds>(dateTimeController.CurrentDateTime());
   if (currentDateTime.IsUpdated()) {
-    auto hour = dateTimeController.Hours();
-    auto minute = dateTimeController.Minutes();
-    auto second = dateTimeController.Seconds();
-    auto year = dateTimeController.Year();
-    auto month = dateTimeController.Month();
-    auto dayOfWeek = dateTimeController.DayOfWeek();
-    auto day = dateTimeController.Day();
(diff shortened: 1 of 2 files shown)


src/displayapp/screens/WatchFacePineTimeStyle.cpp
```diff
@@ -480,21 +477,13 @@ void WatchFacePineTimeStyle::Refresh() {
...
-    if (displayedSecond != second) {
-      displayedSecond = second;
-      lv_label_set_text_fmt(timeDD3, ":%02d", second);
-    }
-
-    if ((year != currentYear) || (month != currentMonth) || (dayOfWeek != currentDayOfWeek) || (day != currentDay)) {
+    currentDate = std::chrono::time_point_cast<days>(currentDateTime.Get());
+    if (currentDate.IsUpdated()) {
       lv_label_set_text_static(dateDayOfWeek, dateTimeController.DayOfWeekShortToString());
+      auto day = dateTimeController.Day();
```
FintasticMan, line 483: Don't use auto for integers.

src/displayapp/screens/WatchFacePineTimeStyle.h
```diff
@@ -42,22 +42,19 @@ namespace Pinetime {
       private:
         uint8_t displayedHour = -1;
         uint8_t displayedMinute = -1;
```
FintasticMan, line 44: I would say it's better to use `UINT8_MAX` here instead of relying on underflow.

FintasticMan (changes requested):
This looks alright, but I think the question here is whether consistency and readability is more important than a slight decrease in memory usage. I haven't tested it myself, but I assume the increase from using another time_point isn't very much, seeing as it decreases memory usage in other watch faces.

src/displayapp/screens/WatchFacePineTimeStyle.h
Riksu9000: There's no guideline about this, so it's more subjective and not really in the scope of this PR.

NeroBurner (approved):
LGTM

PR #1669, opened 2023-03-03
Fix double tap behaviour on PineTimeStyle settings screen

### LINKED ISSUE #1467: Double tapping PineTimeStyle steps style button sends watch to sleep
### Verification

### What happened?

When I tap the buttons of the watch face settings too fast it sends the watch to sleep.

The watch should stay awake and handle two button presses.

### Reproduction steps

1. Set watch face to PineTimeStyle.
2. Long press watch face for settings.
3. Choose lower button (gear icon).
4. See "x" button and "Steps style" button.
5. Double tap "Steps style" button.
6. Watch goes to sleep.

### More details?

This does not happen with the color settings or the settings of the Infineat watch face.

### Version

1.11

### Companion app

_No response_

kieranc:
Fixes #1467 "Double tapping PineTimeStyle steps style button sends watch to sleep"
Double tap is disabled on the color settings screen by checking if the Rnd button is visible, but this didn't work for the options screen as the Rnd button isn't visible. I've changed it to check if the Close button is visible instead, which is used on both settings screens, and resolves the bug. I also changed the button used to disable the long press behaviour which was an as-yet-undiscovered bug which would have allowed the long press action to be used when the options screen was visible.

### DIFF
--- a/src/displayapp/screens/WatchFacePineTimeStyle.cpp
+++ b/src/displayapp/screens/WatchFacePineTimeStyle.cpp
@@ -364,13 +364,13 @@ WatchFacePineTimeStyle::~WatchFacePineTimeStyle() {
 }
 
 bool WatchFacePineTimeStyle::OnTouchEvent(Pinetime::Applications::TouchEvents event) {
-  if ((event == Pinetime::Applications::TouchEvents::LongTap) && lv_obj_get_hidden(btnRandom)) {
+  if ((event == Pinetime::Applications::TouchEvents::LongTap) && lv_obj_get_hidden(btnClose)) {
     lv_obj_set_hidden(btnSetColor, false);
     lv_obj_set_hidden(btnSetOpts, false);
     savedTick = lv_tick_get();
     return true;
   }
-  if ((event == Pinetime::Applications::TouchEvents::DoubleTap) && (lv_obj_get_hidden(btnRandom) == false)) {
+  if ((event == Pinetime::Applications::TouchEvents::DoubleTap) && (lv_obj_get_hidden(btnClose) == false)) {
     return true;
   }
   return false;


src/displayapp/screens/WatchFacePineTimeStyle.cpp
```diff
@@ -364,13 +364,13 @@ WatchFacePineTimeStyle::~WatchFacePineTimeStyle() {
...
 bool WatchFacePineTimeStyle::OnTouchEvent(Pinetime::Applications::TouchEvents event) {
-  if ((event == Pinetime::Applications::TouchEvents::LongTap) && lv_obj_get_hidden(btnRandom)) {
+  if ((event == Pinetime::Applications::TouchEvents::LongTap) && lv_obj_get_hidden(btnClose)) {
     lv_obj_set_hidden(btnSetColor, false);
     lv_obj_set_hidden(btnSetOpts, false);
     savedTick = lv_tick_get();
     return true;
   }
-  if ((event == Pinetime::Applications::TouchEvents::DoubleTap) && (lv_obj_get_hidden(btnRandom) == false)) {
+  if ((event == Pinetime::Applications::TouchEvents::DoubleTap) && (lv_obj_get_hidden(btnClose) == false)) {
```
FintasticMan, line 373: I think it's more clear and concise to use the `!` operator instead of checking equality to false. Those parentheses are also unnecessary and just make the code look more dense IMHO.

```suggestion
  if (event == Pinetime::Applications::TouchEvents::DoubleTap && !lv_obj_get_hidden(btnClose)) {
```

FintasticMan (approved):
Nice to have this fixed. Just a couple of minor style changes I would suggest.

Riksu9000 (approved):
We should find a more reliable approach to disabling touch events. Long tapping while the setting buttons are on the screen, the code gets executed multiple times. This could've been a problem if it created the objects multiple times.

In the mean time, I'll merge this to fix the bug. Thanks.

src/displayapp/screens/WatchFacePineTimeStyle.cpp
Riksu9000: While I agree with the suggestion, and it's also a clang-tidy warning, I don't think it's necessarily in the scope of this PR, so I'll merge without this.

PR #1675, opened 2023-03-07
Activity notifications

FintasticMan:
This adds a feature that sends the user a notification if they haven't walked a certain number of steps within the past hour. The majority of the size increase comes from the settings screen, so if I can make that smaller, that would be good.

I'm not sure if a notification is the best way to handle alerting the user. Suggestions welcome.


src/components/activity/ActivityController.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+ActivityController::ActivityController(Controllers::DateTime& dateTimeController, Controllers::Settings& settingsController)
+  : dateTimeController {dateTimeController}, settingsController {settingsController} {
+}
+
+bool ActivityController::UpdateSteps(uint32_t nbSteps) {
+  int8_t utcMinutes = (dateTimeController.Minutes() + dateTimeController.UtcOffset() * 15) % 60;
+  if (utcMinutes < 0) {
+    utcMinutes += 60;
+  }
```
Riksu9000, line 13: Why does this class need to handle UTC? Shouldn't that happen automatically in DateTimeController?

src/systemtask/SystemTask.cpp
```diff
@@ -200,6 +203,20 @@ void SystemTask::Work() {
...
+      constexpr char message[] = "Low activity\0Fewer steps than threshold in last hour. Try walking around.";
+      constexpr uint8_t messageSize =
+        std::min(sizeof message / sizeof(char), Pinetime::Controllers::NotificationManager::MaximumMessageSize());
+      std::memcpy(notif.message.data(), message, messageSize * sizeof(char));
+      notif.size = messageSize;
+      notif.category = Pinetime::Controllers::NotificationManager::Categories::SimpleAlert;
+      notificationManager.Push(std::move(notif));
+      PushMessage(Messages::OnNewNotification);
+    }
+
```
Riksu9000, line 219: This should be moved to at least one function. Also I'm wondering if this should be part of DisplayApp. What DisplayApp and SystemTask should do is not defined at all, but considering that DisplayAppRecovery does not benefit from this, it probably should not be part of the "system", but rather the "application". I'm working on fixing the ambiguity.

src/components/activity/ActivityController.cpp
```diff
@@ -0,0 +1,34 @@
+#include "components/activity/ActivityController.h"
+
+using namespace Pinetime::Controllers;
+
+ActivityController::ActivityController(Controllers::DateTime& dateTimeController, Controllers::Settings& settingsController)
+  : dateTimeController {dateTimeController}, settingsController {settingsController} {
+}
+
+bool ActivityController::UpdateSteps(uint32_t nbSteps) {
```
Riksu9000, line 9: This function is very obscure. It's called UpdateSteps, yet it returns a value based on whether there's been enough activity recently. You'd have to examine the entire function thoroughly to figure that out.

Riksu9000 (changes requested):
The functionality of ActivityController is hard to understand. Consider giving the variables and functions better names, and extracting functions to give labels to the steps.

src/components/activity/ActivityController.cpp
FintasticMan: DateTimeController's Minutes() function gives it in the local timezone, but for this, we want it in UTC, so that if the user changes to a fractional (i.e. not whole hour difference) timezone, it doesn't mess with the steps buffer.

src/components/activity/ActivityController.cpp
FintasticMan: Fixed. Sorry about that, I just started writing without thinking about the structure first.

src/systemtask/SystemTask.cpp
FintasticMan: I agree that this should be moved to a function, but as you say, I'm not sure whether that function should be in NotificationManager, DisplayApp or even SystemTask. Because of this I won't move this to a function yet.

#1482 makes constructing a notification a lot simpler as well, so when that gets cleaned up and merged it might not even need to be moved to its own function.

src/components/activity/ActivityController.cpp
```diff
@@ -0,0 +1,43 @@
...
+  idx = utcMinutes * stepHistory.size() / 60;
+  if (prevIdx != idx && prevIdx != UINT8_MAX) {
+    oldestSteps = olderSteps;
+    olderSteps = stepHistory[idx];
+    prevIdx = idx;
+    canNotify = true;
+  }
+  stepHistory[idx] = motionController.NbSteps();
+
+  if (stepHistory[idx] < stepHistory[(idx + stepHistory.size() - 1) % stepHistory.size()]) {
```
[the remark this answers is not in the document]
FintasticMan: Good idea. I'll have a look into making that.

src/components/activity/ActivityController.h
```diff
@@ -0,0 +1,34 @@
...
+                         Controllers::Settings& settingsController,
+                         Controllers::MotionController& motionController);
+
+      void UpdateSteps();
+      bool ShouldNotify();
+
+    private:
+      Controllers::DateTime& dateTimeController;
+      Controllers::Settings& settingsController;
+      Controllers::MotionController& motionController;
```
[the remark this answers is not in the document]
FintasticMan: I agree with you for both the Settings controller and the Motion controller, but to me it seems that having the number of minutes since the hour in UTC is a bit of a strange parameter to a function. Do you think that it makes sense to do that? It would also mean moving the local time to UTC calculation to the caller, which seems a bit odd to me.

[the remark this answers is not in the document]
Riksu9000: I don't see how local time to UTC calculation belongs in this class either. If it needs UTC, it should be passed UTC. That's all it needs right? There should just be a function to get UTC time from DateTimeController.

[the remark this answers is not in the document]
Riksu9000: Actually, if you use UTC, depending on the timezone, the notification may come at 0 or 30 minutes in local time, or 15 or 45 in another timezone right? I still don't fully understand the code, but that seems a bit weird. Maybe it should just use local time.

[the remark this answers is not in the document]
FintasticMan: If someone is in a timezone with a 15 or 45 minute offset from UTC, they will receive the notification at 15 and 45 past the hour, yes. I agree that's weird, and I think the issues from using local time are infrequent enough that I'll switch to local time. The issues with local time are that if you switch to a fractional timezone, the index will be offset and so the buffer accesses will be strange. This would also happen if you change the time by a non-hour amount.

Riksu9000:
Actually I was mistaken. This is not quite like a circular buffer. A class that tracks the changes made to a variable, with template parameters specifying the resolution and how much history is stored would be more appropriate for this use case. You would access the value an hour ago through a function that you pass a time point to.

PR #1681, opened 2023-03-10
Remove use of atomic

Riksu9000:
PineTime and InfiniTime are single threaded.

### DIFF
--- a/src/components/ble/HeartRateService.h
+++ b/src/components/ble/HeartRateService.h
@@ -2,7 +2,6 @@
 #define min // workaround: nimble's min/max macros conflict with libstdc++
 #define max
 #include <host/ble_gap.h>
-#include <atomic>
 #undef max
 #undef min
 
@@ -38,7 +37,7 @@ namespace Pinetime {
       struct ble_gatt_svc_def serviceDefinition[2];
 
       uint16_t heartRateMeasurementHandle;
-      std::atomic_bool heartRateMeasurementNotificationEnable {false};
+      bool heartRateMeasurementNotificationEnable {false};
     };
   }
 }
--- a/src/components/ble/MotionService.h
+++ b/src/components/ble/MotionService.h
@@ -33,8 +33,8 @@ namespace Pinetime {
 
       uint16_t stepCountHandle;
       uint16_t motionValuesHandle;
-      bool stepCountNoficationEnabled {false};
-      bool motionValuesNoficationEnabled {false};
+      bool stepCountNotificationEnabled {false};
+      bool motionValuesNotificationEnabled {false};
     };
   }
 }
--- a/src/components/ble/MotionService.h
+++ b/src/components/ble/MotionService.h
@@ -2,7 +2,6 @@
 #define min // workaround: nimble's min/max macros conflict with libstdc++
 #define max
 #include <host/ble_gap.h>
-#include <atomic>
 #undef max
 #undef min
 
@@ -34,8 +33,8 @@ namespace Pinetime {
 
       uint16_t stepCountHandle;
       uint16_t motionValuesHandle;
-      std::atomic_bool stepCountNoficationEnabled {false};
-      std::atomic_bool motionValuesNoficationEnabled {false};
+      bool stepCountNoficationEnabled {false};
+      bool motionValuesNoficationEnabled {false};
     };
   }
 }
--- a/src/components/ble/NotificationManager.cpp
+++ b/src/components/ble/NotificationManager.cpp
@@ -124,8 +124,8 @@ bool NotificationManager::AreNewNotificationsAvailable() const {
   return newNotification;
 }
 
-bool NotificationManager::ClearNewNotificationFlag() {
-  return newNotification.exchange(false);
+void NotificationManager::ClearNewNotificationFlag() {
+  newNotification = false;
 }
 
 size_t NotificationManager::NbNotifications() const {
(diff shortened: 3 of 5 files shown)


minacode:
Do you know how the concurrent tasks from RTOS work? Are they still safe?

Riksu9000:
https://freertos.org/implementation/a00004.html It's not possible to execute code concurrently without multiple cores.

minacode:
That is wrong and your link describes that. A scheduler switches between the tasks, even on one core. It's not happening at the same time, but a switch at an unfortunate time may still cause harm. Therefore my question is if the atomic is needed because there is concurrency going on in InfiniTime.

Riksu9000:
That's contradictory. If the tasks are switched, i.e. "not happening at the same time", they're by definition not concurrent. The processor is not able to execute a read and a write operation concurrently. I think if a task switch occurs while copying a large amount of shared data, the other task can potentially access incomplete data, but that's not what this PR is about.

minacode:
You are confusing parallel with concurrent. [Link](https://stackoverflow.com/questions/1050222/what-is-the-difference-between-concurrency-and-parallelism#1050257)

The atomic bools in question here are used across tasks (looks like that, haven't checked). Using something that synchronizes well in such a place is a good idea and we should maybe even have more of that.

We should find out what exactly atomic does and if it helps preventing an issue in our case. But neither has InfiniTime only one thread nor are we safe of data races because the processor is not capable of parallelism.

Riksu9000:
I wasn't aware of the definitions, but I still can't see the issue. Reading and writing a bool is a single instruction, which is completed before handling the interrupt, so there's no point where a task switch occurring can cause undefined behavior.

minacode:
With atomics you can read *and* write the bool without interrupt for example. The question is if that is something we do. I will take a closer look at the code.

src/components/ble/MotionService.h
```diff
@@ -34,8 +33,8 @@ namespace Pinetime {
 
       uint16_t stepCountHandle;
       uint16_t motionValuesHandle;
-      std::atomic_bool stepCountNoficationEnabled {false};
-      std::atomic_bool motionValuesNoficationEnabled {false};
+      bool stepCountNoficationEnabled {false};
+      bool motionValuesNoficationEnabled {false};
```
minacode, line 37: I agree with those, because there is no point where the atomic is used. But please correct them to "Notification" 😄

src/components/ble/HeartRateService.h
```diff
@@ -38,7 +37,7 @@ namespace Pinetime {
       struct ble_gatt_svc_def serviceDefinition[2];
 
       uint16_t heartRateMeasurementHandle;
-      std::atomic_bool heartRateMeasurementNotificationEnable {false};
+      bool heartRateMeasurementNotificationEnable {false};
```
minacode, line 40: I agree with this.

src/components/ble/NotificationManager.cpp
```diff
@@ -124,8 +124,8 @@ bool NotificationManager::AreNewNotificationsAvailable() const {
   return newNotification;
 }
 
-bool NotificationManager::ClearNewNotificationFlag() {
-  return newNotification.exchange(false);
+void NotificationManager::ClearNewNotificationFlag() {
+  newNotification = false;
```
minacode, line 128: You remove the only use of atomic here. I didn't check for all sites where the function is called, but because it builds, I guess it is ok and the return value was never used.
So I am ok with changing this variable as well.

src/components/ble/NotificationManager.h
```diff
@@ -70,7 +69,7 @@ namespace Pinetime {
       size_t beginIdx = TotalNbNotifications - 1; // index of the newest notification
       size_t size = 0;                            // number of valid notifications in buffer
 
-      std::atomic<bool> newNotification {false};
+      bool newNotification;
```
minacode, line 72: I think it would be better to explicitly initialize this to false as you did with the other variables.

minacode (changes requested):
After looking closer at the code I am ok with the changes, but for the right reasons 😉
Two minor improvements regarding the typos and the one initialization are possible.

Riksu9000:
Getting slightly off topic, but I'm curious, can you give me an example where we should use std::atomic on a type that fits in a register? I believe the behavior is well defined and any issues that can occur from accessing a variable from multiple places aren't issues that are handled by atomic.

minacode:
I am definitely not an expert in this, but from my understanding the power is in having atomic functions that can do multiple operations without interrupt. 
For example `exchange` can read a value and then write it and guarantee you that both or nothing happened.

Riksu9000:
I think I now understand the second issue that atomic can help with, which is what I believe you were concerned with. For example int += 42 in two places can result in only one effective addition, if the value is read but not written before a task switch. In our case I believe we should instead be using FreeRTOS semaphores and critical sections, which will immediately block and switch the task, whereas atomic will create a loop and do nothing until the kernel takes over.

minacode:
Yes, I agree.
Concurrency safety is a topic that might be worth a technical discussion and a larger goal for the future. We can also see that issues occur when multiple parts of the software simultaneously use the same hardware. The motor is the best example.

FintasticMan (approved):
It should be safe to remove the use of atomic in these situations. We should definitely look at places where concurrency is an issue, and agree to a way of solving that.

PR #1684, opened 2023-03-10
Periodic alert

some00:
A new feature for people, who tend to lose track of time. An hourly polite reminder to do something or take a break.


Riksu9000:
How's this different from chimes?

some00:
Hi @Riksu9000! It isn't any different. I prefer to set the exact minute. If I had known chimes before I would not have done this.

Issue #1011, opened 2022-02-26
Short swipes, long swipes, custom swipes?


Riksu9000:
Currently what constitutes a swipe is entirely determined by the touch controller. For a swipe to be registered, there only needs to be a tiny bit of travel. I experimented with making swipes require more distance in https://github.com/InfiniTimeOrg/InfiniTime/issues/710#issuecomment-932120451 and found it was simple to do in software, but had other consequences due to the way LVGL works.

Now I had the idea that we could have separate touch events for long and short swipes, to get the best of both worlds. We could require long swipes to move away from the watchface, which may fix many issues regarding accidental touches being registered. Would this be useful elsewhere?

Furthermore, since this would require a software solution, it could also report the accurate position and length of the swipe. This could be used to require the swipe to close an app to start from the top edge of the screen for example. However, since the screen is very small, this might not be very useful or worth the complication. Let me know if you have ideas about how this could be useful.

PR #1686, opened 2023-03-12
WIP: DisplayApp: Use LVGL gestures

### LINKED ISSUE #1685: Metronome: Dragging the arc will snap to 98 bpm
### Verification

### What happened?

Dragging the arc will snap to 98 bpm

_No response_

### Reproduction steps

This issue happens since this commit https://github.com/InfiniTimeOrg/InfiniTime/commit/5ef3c0d3be58b6741a9e443e6d2151de2b3a59bc.

### More details?

In Metronome, we don't want the app to close while dragging the arc. This was previously accomplished by setting the return touch event to none, and handling the exiting in app, but there's currently no way to disable the default return touch event. Returning true in OnTouchEvent will cancel LVGL tap, which will move the cursor to {-1, -1}, which is at the top left corner of the screen, setting the bpm to 98. Returning false will allow DisplayApp to close the app if it detects an appropriate touch event.

### Version

develop

### Companion app

_No response_

Riksu9000:
I was investigating issue #1685 and found that it can be fixed by using LVGL gestures.

Communicating hardware gestures to LVGL is tricky, because there's no built in support for it. Using LVGL gestures, we only need the touch coordinates from the touch panel. Following up, OnTouchEvent functions can be replaced by LVGL event handlers.

TODO:

- [ ] ~~Fix ScreenList swipe behavior~~ Find a solution to dragging past multiple objects. https://github.com/InfiniTimeOrg/InfiniTime/pull/1686#issuecomment-1465172083
- [ ] Tweak `lv_conf.h`

Relates to #1011
Fixes #1685

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -57,6 +57,66 @@ namespace {
   inline bool in_isr() {
     return (SCB->ICSR & SCB_ICSR_VECTACTIVE_Msk) != 0;
   }
+
+  void ScreenGestureEventHandlerCallback(lv_obj_t* obj, lv_event_t event) {
+    if (event == LV_EVENT_GESTURE) {
+      lv_gesture_dir_t dir = lv_indev_get_gesture_dir(lv_indev_get_act());
+      auto* displayApp = static_cast<DisplayApp*>(obj->user_data);
+      displayApp->HandleScreenGesture(dir);
+    }
+  }
+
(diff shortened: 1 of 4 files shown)


Riksu9000:
A major issue I noticed is that the full gesture must happen on a single object. If I drag a sufficient distance, but the drag passes multiple objects, a gesture event will not happen. This happens even if the objects' gesture parent is set to true, which means the object ignores gestures and the parent (screen) should handle them. If we can't find a solution to this issue, we need to keep another gesture handler besides LVGL.

PR #1689, opened 2023-03-15
Remove easy SystemTask dependencies in BLE services

Riksu9000:


src/components/ble/weather/WeatherService.h
```diff
@@ -36,15 +36,11 @@
...
-  namespace System {
-    class SystemTask;
-  }
-
   namespace Controllers {
 
     class WeatherService {
     public:
-      explicit WeatherService(System::SystemTask& system, DateTime& dateTimeController);
+      explicit WeatherService(DateTime& dateTimeController);
```
NeroBurner, line 43: Unrelated to this PR, but does the weather service really need a non-const dateTimeController? Should the weather service be able to modify the system time?

NeroBurner (approved):
one unrelated note, otherwise looking good

src/components/ble/NavigationService.h
```diff
@@ -27,15 +27,11 @@
...
-  namespace System {
-    class SystemTask;
-  }
-
   namespace Controllers {
 
     class NavigationService {
     public:
-      explicit NavigationService(Pinetime::System::SystemTask& system);
+      explicit NavigationService();
```
FintasticMan, line 34: The `explicit` specifier isn't necessary here anymore. It was necessary because the constructor had 1 argument to prevent it from being used for implicit type conversions.

FintasticMan (approved):
LGTM

PR #1691, opened 2023-03-16
ShakeWake: Fixed instant wake after sleep issue in certain positions

### LINKED ISSUE #927: Hardware button to put watch to sleep causes instant wake under specific circumstances
### Verification

### What happened?

Pressing the hardware button while holding the watch vertically and with shake to wake enabled with a low sensitivity (<50%) causes the watch to immediately wake up.

_No response_

### Reproduction steps

- hold the watch vertically (with the hardware button facing up or down)
- press the hardware button on the watch face as if you were putting it to sleep

### More details?

- setting the sensitivity to over 50% prevents this issue
- lower sensitivities can cause the issue to occur at less extreme angles (closer to being held horizontally)

### Version

v1.8.0

### Companion app

Gadgetbridge 0.64.0

ic-27:
Added lastX var to track the previous x acceleration for correct calculation of speed. 
Closes #927.

### DIFF
--- a/src/components/motion/MotionController.cpp
+++ b/src/components/motion/MotionController.cpp
@@ -54,7 +54,7 @@ bool MotionController::ShouldRaiseWake(bool isSleeping) {
 
 bool MotionController::ShouldShakeWake(uint16_t thresh) {
   /* Currently Polling at 10hz, If this ever goes faster scalar and EMA might need adjusting */
-  int32_t speed = std::abs(z + (y / 2) + (x / 4) - (lastY / 2) - (lastX / 4) - lastZ) / (time - lastTime) * 100;
+  int32_t speed = std::abs(z - lastZ + (y / 2) - (lastY / 2) + (x / 4) - (lastX / 4)) / (time - lastTime) * 100;
   // implemented without floats as .25Alpha
   accumulatedSpeed = (speed / 5) + ((accumulatedSpeed / 5) * 4);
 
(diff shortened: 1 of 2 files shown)


FintasticMan:
This looks good to me, but seeing as @geekbozu wrote the shake wake algorithm, I'd like to hear what they have to say.

src/components/motion/MotionController.cpp
```diff
@@ -53,8 +54,7 @@ bool MotionController::ShouldRaiseWake(bool isSleeping) {
 
 bool MotionController::ShouldShakeWake(uint16_t thresh) {
   /* Currently Polling at 10hz, If this ever goes faster scalar and EMA might need adjusting */
-  int32_t speed = std::abs(z + (y / 2) + (x / 4) - lastY / 2 - lastZ) / (time - lastTime) * 100;
-  //(.2 * speed) + ((1 - .2) * accumulatedSpeed);
```
FintasticMan, line 57: And I don't think you intended to remove this comment.

isaacc27: Good catch, I must have accidentally deleted it when debugging.

FintasticMan (approved):
I think these changes are good. I would suggest using `shakewake` or `ShakeWake` as the module name in the commit, to show that it only affects shake to wake and not raise to wake.

JF002 (approved):

Issue #394, opened 2021-05-23
Display an hours column in stopwatch


Soundtoxin:
Currently if the stopwatch goes past 60 minutes, it will show you're at 70 minutes instead of 1 hour 10 minutes. I think timing things in the hours is fairly common (speedrunning) and it would be good to have this extra column. I believe the Pebble displayed a # of hours.

Avamander:
The Pebbles display years if you leave the stopwatch running that long, heh.

grayrattus:
I'm not sure if adding hours is a good idea. Currently, the vibration is soo light that if you set a stopwatch for like 20 min you can easily forget that you set it and you won't even feel that it was vibrating. Also adding more numbers to the screen means it will be harder to set it up. Maybe it's only with my PineTime but not every touch responds to the time change in the stopwatch app :|

If it has some custom vibrating style for example vibrating JaJa Ding Dong then it would make sense. Or stronger vibration in general. I can't imagine after 1h of training to feel current vibration on the wrist :D

Riksu9000:
Instead of simply shrinking the font to add hours, the time display should adapt to the required number of digits.

JF002:
This was added in #1692.

PR #1692, opened 2023-03-16
stopwatch: Add hours tracking

fossison:
Addresses https://github.com/InfiniTimeOrg/InfiniTime/issues/394

Screenshot of changes

It adds hours without changing any of the existing formatting, but by re-using it as needed.

Once the timer is > 1 hour, the main display becomes Hours:Minutes, and the part beneath becomes Seconds:Milliseconds. Lap timer will display "Hours:Minutes:Seconds.Milliseconds" when a lengthy timer is recorded.

### DIFF
--- a/src/displayapp/screens/StopWatch.cpp
+++ b/src/displayapp/screens/StopWatch.cpp
@@ -110,7 +110,7 @@ void StopWatch::SetInterfaceStopped() {
 
   lv_label_set_text_static(time, "00:00");
   lv_label_set_text_static(msecTime, "00");
-  
+
   if (isHoursLabelUpdated) {
     lv_obj_set_style_local_text_font(time, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, &jetbrains_mono_76);
     lv_obj_realign(time);
--- a/src/displayapp/screens/StopWatch.cpp
+++ b/src/displayapp/screens/StopWatch.cpp
@@ -152,7 +152,7 @@ void StopWatch::Refresh() {
     } else {
       lv_label_set_text_fmt(time, "%02d:%02d:%02d", currentTimeSeparated.hours, currentTimeSeparated.mins, currentTimeSeparated.secs);
       lv_obj_set_style_local_text_font(time, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, &jetbrains_mono_42);
-      lv_obj_realign(time, msecTime, LV_ALIGN_OUT_TOP_MID, 0, 0);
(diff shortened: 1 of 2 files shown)


Riksu9000:
I think I was unclear when writing the comment on the issue. I was thinking that the font would shrink, but not by default. Instead it would shrink once the time to display is large enough. I think with this idea it can be a bit hard to tell what the numbers mean. Also you need to apply a modulo operation to the minutes, otherwise the displayed time will be wrong.

fossison:
Revision

Revised based on feedback above.

src/displayapp/screens/StopWatch.cpp
```diff
@@ -146,7 +147,13 @@ void StopWatch::Refresh() {
     laps[lapsDone] = oldTimeElapsed + xTaskGetTickCount() - startTime;
 
     TimeSeparated_t currentTimeSeparated = convertTicksToTimeSegments(laps[lapsDone]);
-    lv_label_set_text_fmt(time, "%02d:%02d", currentTimeSeparated.mins, currentTimeSeparated.secs);
+    if (currentTimeSeparated.hours == 0) {
+      lv_label_set_text_fmt(time, "%02d:%02d", currentTimeSeparated.mins, currentTimeSeparated.secs);
+    } else {
+      lv_label_set_text_fmt(time, "%02d:%02d:%02d", currentTimeSeparated.hours, currentTimeSeparated.mins, currentTimeSeparated.secs);
+      lv_obj_set_style_local_text_font(time, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, &jetbrains_mono_42);
+      lv_obj_align(time, msecTime, LV_ALIGN_OUT_TOP_MID, 0, 0);
```
Riksu9000, line 155: Use lv_obj_realign instead.

Riksu9000, line 155: Because these are in the Refresh function, they are called constantly. It would be nice to find a solution to this. A bool or the DirtyValue template class will probably do the trick.

fossison: I implemented it with a simple bool. Tried to name it my best :sweat_smile:

fossison:
> I set oldTimeElapsed to 3'680'000 to skip ahead. Don't forget to test your code. The following happens because the label is aligned before setting the text to "00:00" in SetInterfaceStopped().

Thanks for the tip, much easier than what I was doing (or not doing in the case of that last commit lol). All should be fixed now

Riksu9000 (approved):
Thanks for implementing the feature. This can be squash merged.

JF002 (approved):
Thanks @fossison

PR #1695, opened 2023-03-17
Use extra char in notification message array for null terminator

FintasticMan:
The message array in the Notification struct has a size of MaximumMessageSize + 1, for a trailing null terminator if the message is MaximumMessageSize chars. This extra char isn't used for notifications coming from bluetooth, meaning that notifications are truncated 1 character earlier if they are too long. This change is minor and just means that notifications can be 1 character longer.


FintasticMan (commented):
Another solution to this would be to not make the message array 1 char bigger, which would make it more obvious when a size/len does or doesn't include the null terminator.

Riksu9000:
I think the terminator will always cause confusion as long as the truncation needs to happen outside of NotificationManager. I think NotificationManager should accept any std::string and truncate it internally.

Riksu9000:
Would string_view be better?

PR #1696, opened 2023-03-17
circularbuffer: Add circular buffer utility struct

FintasticMan:
Implements a simple circular buffer template. `buf[0]` is the value at the current index. You can use the `++` and `--` operators to move the index. You can initialise the buffer using aggregate initialisation, meaning you can just initialise with a curly-brace list.

### DIFF
--- a/src/utility/CircularBuffer.h
+++ b/src/utility/CircularBuffer.h
@@ -1,0 +1,51 @@
+#pragma once
+
+#include <array>
+#include <cstddef>
+
+namespace Pinetime {
+  namespace Utility {
+    template <class T, size_t S>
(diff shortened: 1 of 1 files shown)


FintasticMan:
I've just pushed a change that switches NotificationManager to use this new CircularBuffer. IMO it makes the code easier to understand from a glance.

Riksu9000:
Maybe a static sized vector class is more along the lines of what we need for NotificationManager, along with a helper function that will erase the oldest entry if pushing a new element wouldn't fit otherwise.

FintasticMan:
Merged with #826.

Issue #1166, opened 2022-06-03
[feature request] add "date format" settings


ITCactus:
### Verification

- [X] I searched for similar feature request and found none was relevant.

### Pitch us your idea!

add  the "date format" settings with "mm/dd" and "dd/mm" options

### Description

i noticed, that on "Digital face" watchface changing the "time format" from 12h to 24h also changes the date format.
it's really neat. 
however, it will be really great, to have the option to choose the "date format" globally, for every watchface and dates.
so, please, add the "date format" settings with e.g. "mm/dd" and "dd/mm" options.

PR #1698, opened 2023-03-18
Add D/M/Y, M/D/Y format support

Brod8362:
This PR adds the baseline support for date formats. As part of it, I merged the time format screen into the Date & Time screen (see below).

This addresses #1166, #1590 

image

This is more of a draft, there will definitely need to be some more changes system wide.

I **did not** implement it across the watchfaces. I'm more than willing to, but I wanted to hold off on that in case it should be a separate PR.

I increased the settings version number (from 4 to 5) as a precaution, since there is now a new setting. I do not know the implications of this, so it should be considered before (if, rather) this gets merged.

I don't feel that it is necessary to add Y/D/M or Y/M/D formats, as not all watchfaces show the year.

**Possible improvements**

- Show a preview of the time/date format in that screen (there's a bit of extra space at the bottom) *OR*
- Increase the size of the buttons to use all of the remaining vertical space

I'm still new to the codebase, so please let me know if I've committed some heinous crime in the commits.

### DIFF
--- a/src/components/settings/Settings.h
+++ b/src/components/settings/Settings.h
@@ -9,6 +9,7 @@ namespace Pinetime {
     class Settings {
     public:
       enum class ClockType : uint8_t { H24, H12 };
+      enum class DateType : uint8_t { DMY, MDY };
       enum class Notification : uint8_t { On, Off, Sleep };
       enum class ChimesOption : uint8_t { None, Hours, HalfHours };
       enum class WakeUpMode : uint8_t {
@@ -172,6 +173,17 @@ namespace Pinetime {
         return settings.clockType;
       };
 
+      void SetDateType(DateType dateType) {
+        if (dateType != settings.dateType) {
+          settingsChanged = true;
+        }
+        settings.dateType = dateType;
+      }
+
+      DateType GetDateType() const {
+        return settings.dateType;
+      }
+
       void SetNotificationStatus(Notification status) {
         if (status != settings.notificationStatus) {
           settingsChanged = true;
@@ -266,7 +278,7 @@ namespace Pinetime {
     private:
       Pinetime::Controllers::FS& fs;
 
-      static constexpr uint32_t settingsVersion = 0x0004;
+      static constexpr uint32_t settingsVersion = 0x0005;
 
       struct SettingsData {
         uint32_t version = settingsVersion;
@@ -274,6 +286,7 @@ namespace Pinetime {
         uint32_t screenTimeOut = 15000;
 
         ClockType clockType = ClockType::H24;
+        DateType dateType = DateType::DMY;
         Notification notificationStatus = Notification::On;
 
         uint8_t clockFace = 0;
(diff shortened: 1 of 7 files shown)


Brod8362:
A couple people mentioned Y/M/D support as an option - this is very doable, I will add it if this PR has some support behind it. At that point, the question of Y/D/M (even if its a silly format) comes up, for the sake of completion.

minacode:
No offense and a little off-topic, but I wonder if we could do this as something like a compile-time setting, because it is a good example of a "choose once" setting. Same would apply for the 12/24-hour format.

src/components/settings/Settings.h
```diff
@@ -9,6 +9,7 @@ namespace Pinetime {
     class Settings {
     public:
       enum class ClockType : uint8_t { H24, H12 };
+      enum class DateType : uint8_t { DMY, MDY };
```
Riksu9000, line 12: It might be interesting to try and combine these, so the first bit would determine the clock type, and the second the date month order. Then settingsVersion wouldn't have to be updated.

Riksu9000 (commented):
Thanks for working on this. I think it is something we need to add to the firmware to support the needs of different people.

You should implement this in at least one watch face, so we can test it.

I would rather use the old TimeFormat screen for now. Combining them is a change independent of this feature. You'd have to remove the checkboxlist and make the code similar to the wakeup setting screen. Checkboxes and radio buttons are preferred over buttons, because they're less ambiguous about the current state.

src/components/settings/Settings.h
Brod8362: I considered this idea. The only reason I didn't do it is because the watchfaces would need to be refactored for that change.

I think it's a great change, but it means we need to make sure to update all the watchfaces in this PR. I'm more than happy to do that, if this is something we wanted to move forward with.

Brod8362:
> You should implement this in at least one watch face, so we can test it.

Sure. May as well implement it in all of them while I'm at it, it's not a ton of work. This also would make moving it to a bitfield more viable later.

> I would rather use the old TimeFormat screen for now.

Sure. I still think it's worth placing the date & time format screens as sub-screens of set date & time screens, what do you think? It feels a bit silly to have one settings for "Set date/time" and another for "Date/Time format"

src/components/settings/Settings.h
Riksu9000: They don't necessarily need to be refactored, if the public enum classes and function behaviors are retained. The fact that they're combined in a single variable doesn't need to be visible outside of the class.

Riksu9000:
I think the time format is not likely to be changed more than once, unlike the current time, so the use cases can be a bit different, but If you think that having separate setting entries for the time format and current time can be considered an issue, you can open an issue about it and we'll continue there.

src/components/settings/Settings.h
Brod8362: Ah, that's true. I'll combine them then.

RageGamerBoi:
> No offense and a little off-topic, but I wonder if we could do this as something like a compile-time setting, because it is a good example of a "choose once" setting. Same would apply for the 12/24-hour format.

Remember, not everyone is going to compile their own build. Most people aren't developers, and/or don't care enough to.

RageGamerBoi:
This is a good idea, not sure about the "highlighted buttons" though. A little unreadable. Radio buttons should work fine.

Y/M/D is now the new international standard btw, as xkcd helpfully points out:

image

PR #1710, opened 2023-03-26
Refactor, document and fix the Watchdog driver

JF002:
Refactor and document the Watchdog driver to make it more readable.

Fix the configuration of the behaviours configuration that was not properly implemented (but it didn't cause any side effect since the correct value was eventually set in NRF_WDT->CONFIG).

Fix the wrong interpretation of the reset reasons caused by implicit conversions of int to bool.

Watchdog is now probably one of the most documented class of the project. What do you think of this format? It's quite helpful in IDEs that support Doxygen like CLion:
image

This PR is an alternative solution to https://github.com/InfiniTimeOrg/InfiniTime/pull/1705.

### DIFF
--- a/src/drivers/Watchdog.cpp
+++ b/src/drivers/Watchdog.cpp
@@ -2,74 +2,148 @@
 #include <mdk/nrf.h>
 using namespace Pinetime::Drivers;
 
-void Watchdog::Setup(uint8_t timeoutSeconds) {
-  NRF_WDT->CONFIG &= ~(WDT_CONFIG_SLEEP_Msk << WDT_CONFIG_SLEEP_Pos);
-  NRF_WDT->CONFIG |= (WDT_CONFIG_HALT_Run << WDT_CONFIG_SLEEP_Pos);
+namespace {
+  /// The watchdog is always driven by a 32768kHz clock
+  constexpr uint32_t ClockFrequency = 32768;
+  /// Write this value in the reload register to reload the watchdog
+  constexpr uint32_t ReloadValue = 0x6E524635UL;
 
-  NRF_WDT->CONFIG &= ~(WDT_CONFIG_HALT_Msk << WDT_CONFIG_HALT_Pos);
-  NRF_WDT->CONFIG |= (WDT_CONFIG_HALT_Pause << WDT_CONFIG_HALT_Pos);
+  /// Configures the behaviours (pause or run) of the watchdog while the CPU is sleeping or halted by the debugger
+  ///
+  /// @param sleepBehaviour Configure the watchdog to either be paused, or kept running, while the CPU is sleeping
+  /// @param haltBehaviour Configure the watchdog to either be paused, or kept running, while the CPU is halted by the debugger
+  void SetBehaviours(Watchdog::SleepBehaviours sleepBehaviour, Watchdog::HaltBehaviours haltBehaviour) {
+    // NRF_WDT->CONFIG : only the 1st and 3rd bits are relevant.
+    // Bit 0 : Behavior when the CPU is sleeping
+    // Bit 1 : Behavior when the CPU is halted by the debugger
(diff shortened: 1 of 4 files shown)


src/drivers/Watchdog.cpp
```diff
@@ -2,74 +2,148 @@
...
+    case Watchdog::ResetReasons::SoftReset:
       return "Soft reset";
-    case ResetReasons::NFC:
+    case Watchdog::ResetReasons::NFC:
       return "NFC";
-    case ResetReasons::HardReset:
+    case Watchdog::ResetReasons::HardReset:
       return "Hard reset";
     default:
       return "Unknown";
```
FintasticMan, line 149: This isn't a very safe way to return a string because the string is stack-allocated. I would suggest an array of strings in the anonymous namespace, and returning the string at the index of the ResetReason.

FintasticMan (commented):
I personally prefer this solution to #1705.

Riksu9000 (commented):
I like this. There are a lot of comments, but in low level code some are necessary.

There are a few potential improvements I can see, which nonetheless shouldn't be blocking. The timeout could be set using std::chrono durations, and the reset reason enum class could contain the values that the reset reason register itself may contain, so it can also be used in GetResetReason() and the value in the register doesn't need to be converted to another format.

src/drivers/Watchdog.cpp
Riksu9000: I'm pretty sure this just returns a pointer to static data.

src/drivers/Watchdog.h
```diff
@@ -3,22 +3,65 @@
...
+      /// Indicates the reasons of a reset of the MCU
       enum class ResetReasons { ResetPin, Watchdog, SoftReset, CpuLockup, SystemOff, LpComp, DebugInterface, NFC, HardReset };
-      void Setup(uint8_t timeoutSeconds);
+
+      /// Behaviours of the watchdog when the CPU is sleeping
+      enum class SleepBehaviours : uint8_t {
+        /// Pause watchdog while the CPU is sleeping
+        Pause = 0,
+        /// Keep the watchdog running while the CPU is sleeping
+        Run = 1
```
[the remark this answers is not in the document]
Riksu9000: I don't think we want this if we're avoiding using the SDK.

src/drivers/Watchdog.cpp
FintasticMan: After looking it up, it does seem that string literals have static duration. I was taught that you shouldn't return a string literal, but I guess that isn't correct.

src/drivers/Watchdog.h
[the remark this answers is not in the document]
JF002: I also wondered if I should use the constants from the MDK or not. 

I wrote the code in such a way that it's a direct translation of the datasheet, which only mentions bits position, and not labels like `WDT_CONFIG_SLEEP_Pos`.

Now, if you think it makes the code more readable, I can use them.

JF002:
> I like this. There are a lot of comments, but in low level code some are necessary.

Thanks! Some comments might be a bit overkill, but they explain how the datasheet was translated into code :)

> There are a few potential improvements I can see, which nonetheless shouldn't be blocking. The timeout could be set using std::chrono durations, 

I also had this idea, but that would make the implementation of `SetTimeout()` much more complex if we want to handle all the possible values.

> and the reset reason enum class could contain the values that the reset reason register itself may contain, so it can also be used in GetResetReason() and the value in the register doesn't need to be converted to another format.

We cannot simply cast the value of the register to the value of the enum because the register might have multiple bits that are set (especially if the register was not cleared between 2 resets). 
We can use a bitfield instead of an enum for ResetReason, but in this case, the application will need to check for all the bits instead of a single value here. That's why I decided to 'convert' the value of the register to a single value - the 1st reset reason specified in the register.

Riksu9000 (approved):

FintasticMan (approved):

PR #1726, opened 2023-04-05
shakewake: Slightly improve accuracy

FintasticMan:
The (accumulated) speed was calculated by dividing first and multiplying/adding after, which results in more rounding errors than if you multiply/add first and then divide. The values aren't big enough to overflow.


JF002:
@FintasticMan What improvement does this PR bring to the user? Or how should we test it?

FintasticMan:
This change doesn't affect the user much, if at all. It's just a small change that makes the maths closer what was intended according to the comment in the code.

FintasticMan:
Merged with #826.

PR #1728, opened 2023-04-05
Watchfaces: motionsensor is ok

Riksu9000:
The application shouldn't be concerned with the sensor being initialized, because the values just wouldn't be updated. DisplayApp also isn't started before the sensor is initialized.

Extracted from #1700

### DIFF
--- a/src/components/motion/MotionController.h
+++ b/src/components/motion/MotionController.h
@@ -50,14 +50,6 @@ namespace Pinetime {
         return accumulatedSpeed;
       }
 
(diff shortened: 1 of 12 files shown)


FintasticMan (approved):
This looks good to me!

JF002 (approved):
I originally added this `isOk()` flag to allow apps to do something appropriate when the sensor would become not ok at runtime. But since we only set this flag at initialization time, this PR makes sense :+1:

Issue #1540, opened 2023-01-16
CMake detecting system toolchain


Riksu9000:
### Verification

- [X] I searched for similar issues and found none was relevant.

### Introduce the issue

When generating the project, CMake claims the compiler in use is one from the system, not the arm crosscompiler. I don't know if this causes side-effects, but it seems incorrect nonetheless.

Output from a build workflow.
```
-- The C compiler identification is GNU 11.3.0
-- The CXX compiler identification is GNU 11.3.0
-- The ASM compiler identification is GNU
-- Found assembler: /usr/bin/cc
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
```

### Preferred solution

Moving the inclusion of the toolchain file earlier seems to fix the issue, but I'm not very familiar with CMake, so I want to give an opportunity to someone who knows better to look at it and find a good solution to this issue. This can be an opportunity to also check the structure of the CMake files, which seems like it might be quite outdated.

PR #1741, opened 2023-04-12
Added CMake changes to allow for project building and debugging on MacOS ARM processor

p-riebs:
I recently bought an ARM-based Mac and wasn't able to build or debug the InfiniTime project.

These changes allow the project to be built and debugged on ARM-based processors. Most of these changes are actually already in the cmake-nRF5x toolchain but aren't set early enough to fix the issues. My guess is most contributors are using Intel-based processors so these issues aren't being caught.

For example, the line:
```
set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
```
is set below 
https://github.com/InfiniTimeOrg/InfiniTime/blob/main/cmake-nRF5x/arm-gcc-toolchain.cmake#L27-L32
but isn't added in the CMakeLists.txt until the end of the build process.
https://github.com/InfiniTimeOrg/InfiniTime/blob/main/CMakeLists.txt#L74

From my research these lines need to be added before the `project()` line:
https://github.com/InfiniTimeOrg/InfiniTime/blob/main/CMakeLists.txt#L5

I'm new to this so if there are any better ways or something I missed please let me know.

Sources:
https://github.com/microsoft/vscode-cmake-tools/issues/662
https://gitlab.kitware.com/cmake/cmake/-/issues/23105
https://stackoverflow.com/a/12794195

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,12 @@ cmake_minimum_required(VERSION 3.10)
 
 set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose Debug or Release")
 
+# Override the default system to cross-compile for ARM processors
+set(CMAKE_SYSTEM_NAME Linux)
+set(CMAKE_SYSTEM_PROCESSOR ARM)
+
+set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
+
 project(pinetime VERSION 1.12.0 LANGUAGES C CXX ASM)
 
 set(CMAKE_C_STANDARD 99)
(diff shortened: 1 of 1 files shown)


Riksu9000:
I'm pretty sure people are able to compile InfiniTime on Pinebooks and raspberries. This might be related to #1540.

JF002:
On my setup (CLion on x86_64 computer), I need to add `
set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)`  at the top of the main CMake file to make CMake happy. I haven't tried to understand why though, since it's seems to be working fine for many other developers and also in the Docker container.

So I'm not sure those changes are specific to MacOS on ARM CPU.

@p-riebs could you test if adding only `
set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)` is working on your setup?

Setting the system name to Linux (`set(CMAKE_SYSTEM_NAME Linux)`) does not feel right since we are not using Linux and build flags are managed in the `src` subfolder (https://github.com/InfiniTimeOrg/InfiniTime/blob/main/src/CMakeLists.txt#L776).

p-riebs:
My knowledge of CMake is very limited here. 

@JF002 I tried to move the `set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)` line to the top without any other changes and that still has the issue of `-arch` being automatically appended
```
arm-none-eabi-gcc: error: unrecognized command-line option '-arch'; did you mean '-march='?
```
Setting the CMAKE_SYSTEM_NAME to `Generic` fixes the issue. So Linux isn't necessary. (I'll update the PR if we find we want to merge this)
```
set(CMAKE_SYSTEM_NAME Generic)
```

@Riksu9000 - Moving the toolchain up in the CMakeList file would only require moving the line `add_subdirectory(src)` further up right? I tried that but wasn't successful.

PR #1742, opened 2023-04-15
Fix potential buffer overflows when calling sprintf

szsam:


minacode:
Welcome to InfiniTime! 😊 I think it would be helpful if you write a little description about the motivations of the changes you made.

Avamander:
These aren't overflows, truncations at best and modulo is relatively quite expensive. 

The other parts, would be very nice if you would explain them.

src/components/datetime/DateTimeController.cpp
szsam, line 1: First, please note that the `2` in `"%02i"` specifies the *minimum* (not maximum) field width. Thus, the result of `"%02i"` can be as lengthy as `"-2147483648"`. Hence, there is indeed a potential risk of buffer overflow.

To prevent the potential buffer overflow, I use modulo here to ensure that the integers (hour, min, etc) passed to `sprintf` contain no more than 2 digits. However, I agree that modulo is relatively expensive on embedded MCUs. I think a better approach to prevent buffer overflow here is to use `snprintf` instead of `sprintf`. What do you think?

szsam (commented):
I have left some pull request comments on files changed. PTAL.

szsam:
A recent commit introduces a *real* one-byte buffer overflow.
https://github.com/InfiniTimeOrg/InfiniTime/blob/40f7e1c7be6882e01058b5ccf64d5005c6105346/src/displayapp/screens/StopWatch.cpp#L201-L206
The format str `"#%2d %2d:%02d:%02d.%02d\n"` requires 17 bytes, but the buffer length is only 16. This is fixed in this PR.

FintasticMan (changes requested):
I don't think that replacing `sprintf` with `snprintf` is necessarily a good idea. I believe that because it adds an overhead and that, with the current configuration, I don't think it does anything if it detects an overflow, we shouldn't use it.

szsam:
> I don't think that replacing `sprintf` with `snprintf` is necessarily a good idea. I believe that because it adds an overhead and that, with the current configuration, I don't think it does anything if it detects an overflow, we shouldn't use it.

I agree that `snprintf` adds overhead. However, when `snprintf(buffer, bufsz, fmt, ...)` detects an overflow, it will write at most `bufsz-1` characters, effectively preventing buffer overflow.

That said, if you insist on using `sprintf`, I will change the code and use it.

src/displayapp/screens/settings/SettingDisplay.cpp
```diff
@@ -43,10 +43,10 @@ SettingDisplay::SettingDisplay(Pinetime::Applications::DisplayApp* app, Pinetime
   lv_label_set_align(icon, LV_LABEL_ALIGN_CENTER);
   lv_obj_align(icon, title, LV_ALIGN_OUT_LEFT_MID, -10, 0);
 
-  char buffer[12];
+  char buffer[4];
   for (unsigned int i = 0; i < options.size(); i++) {
     cbOption[i] = lv_checkbox_create(container1, nullptr);
-    sprintf(buffer, "%2ds", options[i] / 1000);
+    snprintf(buffer, sizeof(buffer), "%2" PRIu8 "s", options[i] / 1000);
```
FintasticMan, line 49: I've just tried running these changes on my watch (without using `snprintf`), and these changes break the settings menu. The checkboxes are stacked with only 1 column, rather than 2, and the text on all of them is `hus`. I don't currently have too much time to look in to what's happening here.

FintasticMan (changes requested):
After looking into the overhead that snprintf adds, I've come to the conclusion that it is fine to use it over sprintf as an added layer of safety. We should of course still make sure that our values can never be higher than what we expect and that our buffers are big enough.

For some reason changing the format specifier macro to PRIu16 fixes the issue I mentioned before. Once that is fixed, I'm willing to get this merged.

src/displayapp/screens/settings/SettingDisplay.cpp
FintasticMan: To fix this, the format specifier macro used here needs to be changed from `PRIu8` to `PRIu16`.

FintasticMan: Yes, indeed. I think the reason that it was made to be u8 was because any u16 / 1000 can be fit into a u8.

NeroBurner: This is done to keep the character buffer small, I think?

Then if we want the `PRIu8` the result of `u16/1000` could be wrapped by a `static_cast<uint8_t>()`

szsam: I changed it to `PRIu16`. `char buffer[4]` saves buffer space and `snprintf` ensures security. Unnecessary to use `PRIu8`.

FintasticMan (approved):

src/displayapp/screens/Twos.cpp
```diff
@@ -241,8 +241,8 @@ void Twos::updateGridDisplay() {
     const unsigned int row = i / nCols;
     const unsigned int col = i % nCols;
     if (grid[row][col].value > 0) {
-      char buffer[7];
-      sprintf(buffer, "%d", grid[row][col].value);
+      char buffer[11];
```
NeroBurner, line 244: nitpick: could we leave this buffer to be of size 7. I don't want to increase the buffer size for the game as the buffers are copied into the `lv_table`.

Keeping the  size of 7 results in 6 usable bytes. Then the biggest possible value in the game would be `2**19` -> `524288`.

I doubt that anyone would reach this insane number while playing the game

Even buffer size of 6 (with 5 usable bytes) would be enough to hold `2**16 = 65536`. But I think keeping the size that works is less intrusive than saving a few bytes

mark9064: Because there are 16 squares and the largest number you can get is 4, it is not possible to have a tile with a value above 2^17 (131072). So 7 bytes is guaranteed to be sound

NeroBurner (approved):
LGTM

Machiry:
Hello @FintasticMan, @Avamander,  @NeroBurner ,

My name is Aravind Machiry, Assistant Professor at Purdue's ECE Department.

Thank you for considering this pull request. This pull request was the result of our on-going research work (along with @szsam) to improve the security and quality of open-source embedded projects.

In addition to scanning codebases with CodeQL, we are also doing a short (~4 minutes) survey to understand the use of static analysis tools like `gcc -Wall` and CodeQL in embedded software projects.

It would greatly benefit our research if you could fill this anonymous survey: https://purdue.ca1.qualtrics.com/jfe/form/SV_0OnXfr5plPe1QCa

Thank you,
Aravind

PR #1752, opened 2023-04-27
Update Apps.md

Alex-EEE:
Example code will not compile.  Screen no longer takes an arg


doc/code/Apps.md
```diff
@@ -74,7 +74,7 @@ MyApp.cpp:
 
 using namespace Pinetime::Applications::Screens;
 
-MyApp::MyApp(DisplayApp* app) : Screen(app) {
+MyApp::MyApp() : Screen() {
```
FintasticMan, line 77: The constructor doesn't need to init a Screen instance.

```suggestion
MyApp::MyApp() {
```

FintasticMan (changes requested):
We should keep the docs updated.

FintasticMan (approved):

kieranc (approved):
Looks good :+1:

PR #1755, opened 2023-05-07
Power optimization - enable the DC/DC regulator.

JF002:
The DC/DC regulator reduce the power consumption of the MCU compared to the default LDO regulator. The DC/DC regulator needs additional HW that is mounted on the PineTime.

This change reduces the power usage by 380µA during fast advertising, by 200µA during slow advertising and by 186µA when BLE is disabled.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -293,7 +293,12 @@ void calibrate_lf_clock_rc(nrf_drv_clock_evt_type_t /*event*/) {
   nrf_drv_clock_calibration_start(16, calibrate_lf_clock_rc);
 }
 
+void enable_dcdc_regulator() {
+  NRF_POWER->DCDCEN = 1;
+}
+
 int main() {
+  enable_dcdc_regulator();
   logger.Init();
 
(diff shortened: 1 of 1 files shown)


aykevl:
Apparently this is supported in Mynewt - but not in FreeRTOS I guess:

https://github.com/apache/mynewt-core/blob/24d9c3b3c1699fba734a2dc3e37935ff27ee04e3/hw/bsp/pinetime/syscfg.yml#L54

```yaml
    # Enable DC/DC regulator
    MCU_DCDC_ENABLED: 1
```

> [...] and by 186µA when BLE is disabled.

If DCDCEN has any effect, it means that something is still draining power: DCDCEN does not (significantly) reduce current consumption when the nrf52832 is fully asleep. (But I guess the other PRs should help with that).

JF002:
According to my measurement, the watch uses ~240µA more during fast BLE advertising, and ~15µA during slow advertising. However, the measurements are the same when BLE is disabled, exactly as you said!

PR #1756, opened 2023-05-07
Power optimization - Enable HRS3300 (heart rate sensor) sleep mode.

JF002:
Put the HRS3300 to sleep mode when InfiniTime is going to sleep. This change reduces the power consumption by 130µA when the heart rate sensor is disabled.


JF002:
> I'm not able to verify the changes, but I don't see any issues.

It works fine on my side, power usage is reduced and the HR sensor still works as expected :)

PR #1757, opened 2023-05-07
Power optimization - Increase SystemTaskPeriod.

JF002:
Increase the timeout on the message queue in SystemTask. This reduces the power consumption by 60-70µs in sleep mode.

### DIFF
--- a/src/systemtask/SystemTask.cpp
+++ b/src/systemtask/SystemTask.cpp
@@ -181,7 +181,7 @@ void SystemTask::Work() {
     UpdateMotion();
 
     Messages msg;
-    if (xQueueReceive(systemTasksMsgQueue, &msg, 100) == pdTRUE) {
+    if (xQueueReceive(systemTasksMsgQueue, &msg, 4000) == pdTRUE) {
       switch (msg) {
         case Messages::EnableSleeping:
           // Make sure that exiting an app doesn't enable sleeping,


mark9064:
Super cool to see all the power optimisations inbound!
Unfortunately this PR means that the time is only updated every 4s when the watch is on, which is a problem for watch faces with seconds on them. It also means chimes can arrive 4s late, though I suppose this is less of a problem (it's still nice to have them arrive right on the hour though)
Maybe it could be changed to update more often when the display is on?

src/systemtask/SystemTask.cpp
```diff
@@ -181,7 +181,7 @@ void SystemTask::Work() {
     UpdateMotion();
 
     Messages msg;
-    if (xQueueReceive(systemTasksMsgQueue, &msg, 100) == pdTRUE) {
```
FintasticMan, line 184: How come there was no need to convert the 100ms to ticks before, but now we do?

mark9064: `configTICK_RATE_HZ` is set to 1024, so 1 millisecond is 1024/1000 ticks=1.024 ticks
So previously it was set to 97.66ms rather than 100ms.
i.e the existing ratio is close enough to 1:1 that it wasn't noticeable, but it should be using the conversion for correctness

FintasticMan: Ah I see, thanks for the explanation.

src/systemtask/SystemTask.cpp
```diff
@@ -504,3 +504,14 @@ void SystemTask::PushMessage(System::Messages msg) {
     xQueueSend(systemTasksMsgQueue, &msg, portMAX_DELAY);
   }
 }
+
+TickType_t SystemTask::GetQueueTimeout() const {
+  // By default, the timeout on the queue is 100ms.
+  // It's extended to 4s in sleep mode, when no motion based wake up option is enabled.
+  TickType_t timeout = pdMS_TO_TICKS(100);
+  if (state == SystemTaskState::Sleeping && !settingsController.isWakeUpModeOn(Pinetime::Controllers::Settings::WakeUpMode::RaiseWrist) &&
+      !settingsController.isWakeUpModeOn(Pinetime::Controllers::Settings::WakeUpMode::Shake)) {
```
FintasticMan, line 513: `settingsController.GetNotificationStatus() == Controllers::Settings::Notification::Sleep` could be added as an additional clause to make the timeout longer when sleep mode is enabled (which disables the motion wake algorithms).

```suggestion
  if (state == SystemTaskState::Sleeping && ((!settingsController.isWakeUpModeOn(Pinetime::Controllers::Settings::WakeUpMode::RaiseWrist) &&
      !settingsController.isWakeUpModeOn(Pinetime::Controllers::Settings::WakeUpMode::Shake)) ||
      settingsController.GetNotificationStatus() == Controllers::Settings::Notification::Sleep)) {
```

FintasticMan (changes requested):

Riksu9000:
There are many things that rely on the SystemTask loop period, not just motion. Have you considered all of them so they don't cause issues? Ideally all periodic tasks should be moved away from the message handling loop, so the period can be set to infinite.

In your measurement, did you have motion wake up methods enabled, i.e. did UpdateMotion() return early? If not, I'd like to see the power consumption measured again with two different periods, without the automatic period switching, but motion disabled in both tests by disabling motion wake up methods.

JF002:
> There are many things that rely on the SystemTask loop period, not just motion. Have you considered all of them so they don't cause issues? Ideally all periodic tasks should be moved away from the message handling loop, so the period can be set to infinite.

The period cannot be set to Infinite, it must actually be shorter than the period of the watchdog. That's exactly why I chose to run all background tasks in SystemTask : if SystemTask does not run correctly, we cannot guarantee the good state of the system, so we let the watchdog reset the board.

> In your measurement, did you have motion wake up methods enabled, i.e. did UpdateMotion() return early? If not, I'd like to see the power consumption measured again with two different periods, without the automatic period switching, but motion disabled in both tests by disabling motion wake up methods.

I've just done the measurements :  
 - Wake on wrist / shake wake enabled : period set to 100ms : 890µA.
 - Wake on wrist / shake wake disabled : period set to 4s : 780µA

Note that motion algorithms and low level driver could probably be improved by using the hardware FIFO and the "motion/no-motion" interrupt pin to reduce the power usage.

JF002:
> I would like to see the power consumption difference between the different periods set statically, with motion disabled in both tests. 

 - Period set to 100ms : 850µA
 - Period set to 4000ms : 770µA

> I went through the tasks that are run in the loop and the noteworthy observations [...]

This are valid concerns. I'll check them and see if I can solve those issue in this PR or if they'll need more work.

PR #1758, opened 2023-05-07
Power optimization - Improve SPI sleep mode

JF002:
Ensure that all pins are set to their default configuration during sleep mode. Disable the workaround for FTPAN58 (SPI freezes when transfering a single byte) at the end of the transfer. This disables the resources needed for the workaround. Those changes reduce the power usage by 430-490µA.

### DIFF
--- a/src/drivers/SpiNorFlash.cpp
+++ b/src/drivers/SpiNorFlash.cpp
@@ -10,7 +10,6 @@ SpiNorFlash::SpiNorFlash(Spi& spi) : spi {spi} {
 }
 
 void SpiNorFlash::Init() {
-  spi.Init();
   device_id = ReadIdentificaion();
   NRF_LOG_INFO("[SpiNorFlash] Manufacturer : %d, Memory type : %d, memory density : %d",
                device_id.manufacturer,
@@ -24,12 +23,10 @@ void SpiNorFlash::Uninit() {
 void SpiNorFlash::Sleep() {
   auto cmd = static_cast<uint8_t>(Commands::DeepPowerDown);
   spi.Write(&cmd, sizeof(uint8_t));
-  spi.Sleep();
   NRF_LOG_INFO("[SpiNorFlash] Sleep")
 }
 
 void SpiNorFlash::Wakeup() {
-  spi.Wakeup();
   // send Commands::ReleaseFromDeepPowerDown then 3 dummy bytes before reading Device ID
   static constexpr uint8_t cmdSize = 4;
   uint8_t cmd[cmdSize] = {static_cast<uint8_t>(Commands::ReleaseFromDeepPowerDown), 0x01, 0x02, 0x03};
--- a/src/drivers/SpiNorFlash.cpp
+++ b/src/drivers/SpiNorFlash.cpp
@@ -10,6 +10,7 @@ SpiNorFlash::SpiNorFlash(Spi& spi) : spi {spi} {
 }
 
 void SpiNorFlash::Init() {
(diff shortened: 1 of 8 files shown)


src/drivers/St7789.cpp
```diff
@@ -187,10 +187,13 @@ void St7789::HardwareReset() {
 void St7789::Sleep() {
   SleepIn();
   nrf_gpio_cfg_default(pinDataCommand);
+  nrf_gpio_cfg_default(26);
```
FintasticMan, line 190: It might be good to define a constant, rather than having a magic number here.

src/drivers/SpiNorFlash.cpp
```diff
@@ -23,10 +24,12 @@ void SpiNorFlash::Uninit() {
 void SpiNorFlash::Sleep() {
   auto cmd = static_cast<uint8_t>(Commands::DeepPowerDown);
   spi.Write(&cmd, sizeof(uint8_t));
+  spi.Sleep();
```
Riksu9000, line 27: Since "sleep" in this instance only means changing the pin configuration, can this be done automatically inside the read and write functions?

Riksu9000 (commented):
How much do each of the changes contribute to reducing the power consumption separately?

src/drivers/SpiNorFlash.cpp
JF002: See my other comment, calling Sleep here might not be a good idea. I'll do the change :)

JF002:
> How much do each of the changes contribute to reducing the power consumption separately?

This PR basically contains 2 changes : reset the CS pin configuration during sleep (and restore it at wakeup) and disable the "workaround for FTPAN58" (disable the PPI channel). 

As I said in responses to your reviews, resetting the CS pin config is probably wrong as CS pins must be kept high whenever we don't want devices to listen to the bus, so I'll remove those change.
Most of the power usage reduction comes from disabling the PPI channel : from 900µA to 410µA in sleep mode with BLE disabled.

src/drivers/St7789.cpp
```diff
@@ -179,14 +178,15 @@ void St7789::DrawBuffer(uint16_t x, uint16_t y, uint16_t width, uint16_t height,
 }
 
 void St7789::HardwareReset() {
-  nrf_gpio_pin_clear(26);
+  nrf_gpio_pin_clear(pinReset);
   nrf_delay_ms(10);
-  nrf_gpio_pin_set(26);
+  nrf_gpio_pin_set(pinReset);
 }
```
Riksu9000, line 184: Can we configure the pin only at the beginning of this function, and reset back to default before returning?

JF002: The pins needs to be kept at a high level to ensure the display controller won't reset unexpectedly, so I guess we can't configure the pins to default (input, no pull up/down) here.

Riksu9000: If that's true, wouldn't that mean we shouldn't default configure the pin in the sleep function either or anywhere else?

JF002: Yes, you are right, I fixed this in 3c668ca01bb95e02dadb7a9bb1e7f43ddf9071e5.

Riksu9000: No, there's still nrf_gpio_cfg_default in the St7789::Sleep() function.

JF002: Yes, right, I guess we shouldn't default the reset pin!

PR #1764, opened 2023-05-19
node: Add lv_font_conv and lv_img_conv to node package

FintasticMan:
This means that these can be installed by simply running `npm i`.


package-lock.json
```diff
@@ -0,0 +1,2525 @@
+{
+  "name": "InfiniTime",
```
NeroBurner, line 2: SO MANY LINES! I have no idea how node works, but why do we need this huge file? Shouldn't it be enough to list the two needed packages?

NeroBurner:
If we just need to do `npm install`, then we could simplify the workflow a bit. Could you update those files in `.github` as well?

package-lock.json
FintasticMan: This file is the package lock, that specifies the versions of all the packages that are needed (which is more than just the 2 direct dependencies), so that builds are reproducible. It prevents npm pulling newer versions of packages, which might have changed things we (or other packages) rely on. The `package.json` specifies our direct dependencies.

NeroBurner: Reproducibility is good! Please document how this file is created, or is it always created by `npm install`?

FintasticMan: It's created and updated by any npm command that affects the `node_modules` directory or the `package.json`, so it is just created by running npm install. See https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json.

FintasticMan:
> If we just need to do `npm install`, then we could simplify the workflow a bit. Could you update those files in `.github` as well?

I've just done this, by installing the packages in the InfiniTime directory, and telling InfiniSim to also search there for the binaries.

I've also updated the docker image to install the npm packages in the InfiniTime directory.

docker/build.sh
```diff
@@ -33,6 +33,8 @@ main() {
   [ ! -d "$TOOLS_DIR/$NRF_SDK_VER" ] && GetNrfSdk
   [ ! -d "$TOOLS_DIR/mcuboot" ] && GetMcuBoot
 
+  npm i
```
NeroBurner, line 36: why move the install from the docker file where it ran once to the build script where it has to run each build.

please try to move it back to the dockerfile

FintasticMan: To do that, it would need to clone the InfiniTime repo during the Docker image build. This doesn't really make sense, and I think that it doesn't make sense to need to rebuild the Docker image just to use updated npm packages. It also shouldn't affect the build time much, because if the packages are up to date it doesn't do anything.

NeroBurner:
But it would require me to have a working internet connection for each build. Previously that was a requirement for just the docker image creation

I think you could use a `COPY packages.json` command and delete it after the npm command again
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

docker/build.sh
FintasticMan: > But it would require me to have a working internet connection for each build. Previously that was a requirement for just the docker image creation
>
> I think you could use a `COPY packages.json` command and delete it after the npm command again

Good point. In this case, npm actually runs even if there's no internet connection as long as the packages are up-to-date with the `package-lock.json`.

You're right, I could just download the `package.json` and `package-lock.json` from the repo and use those in the Docker image itself. I personally think that installing the modules in the repo and running `npm i` for every build is a better solution.

mark9064:
Might want to remove image conv from the title of the PR to clarify that it's no longer used, I was a bit confused at first

FintasticMan:
Oh yeah, forgot to do that.

docker/build.sh
JF002: I would also prefer to have all dependencies installed during the build of the Docker image rather than during the build of the project. This ensures that once the Docker image is generated and tested, it will be able to build the project no matter what.

FintasticMan: After looking into it, npm doesn't support installing packages globally from a package.json. If we do want the Docker container to contain the npm packages, we'll have to install them like we do now, without a package.json.

PR #1768, opened 2023-06-05
Remove the vibration while charging

### LINKED ISSUE #724: Constant vibration when charging
### 

### What happened?

While charging my PineTime, the watch sonstantly vibrates

Should not vibrate

### Reproduction steps

Just charge the watch

### More details?

_No response_

### Version

InfiniTime 1.6.0

### Companion app

GadgetBridge 0.60.0

minacode:
This fixes #724 where the vibration interupts the physical connection with the charger and therefore triggers a new charging event and vibration, ending in a cycle of vibrations while charging.


Itai-Nelken:
I like this, I have experienced this problem many times. But I would still like some way to be sure the watch is charging other than the tiny icon that is hard to see. Maybe the battery settings page opens when charging?

minacode:
> I like this, I have experienced this problem many times. 

Nice :) It seems to be a common issue.

> But I would still like some way to be sure the watch is charging other than the tiny icon that is hard to see. Maybe the battery settings page opens when charging?

I think that would belong to another PR. This one is only about removing the bug.

everypizza1:
That's not a problem I have had on the sealed watch, but the vibration is very loud.

mark9064:
Just FYI there is also #1876

minacode:
Which I like even more! 😄 I hope your gets merged. We can close this one if it did.

RageGamerBoi:
I do like the tactility though, what about only doing the vibration if there hasn’t been a charging event in the last ~1 minute?

minacode:
I will keep this PR as simple as possible and just remove the one line.

mark9064 (commented):
I think it also makes sense to remove the charging event from displayapp as apps pull charging state directly from the batterycontroller; after this change the message would be unused.

minacode:
Totally. I removed it.

mark9064 (approved):

JF002 (approved):

Issue #1473, opened 2022-12-05
backport lvgl fallback font to enable support for more language (via external flash font loading)


Boteium:
### Verification

- [X] I searched for similar feature request and found none was relevant.

### Pitch us your idea!

Let's backport font fallback to enable more languages and emoji

### Description

Due to the small size of os storage, we can only fit a very small number of glyphs in the default font.

To support other non-English language, numerous forks have been created to support different language by adding glyphs to jetbrains_mono_bold_20 such as <a href="https://github.com/yehoshuapw/InfiniTime">this one</a>

Loading custom font from external storage is another idea that have been proposed several times. (e.g. https://github.com/InfiniTimeOrg/InfiniTime/issues/212 ) There are also watchfaces in current release that already use external flash to store fonts.

However, if a LV_LABEL is set to use a certain font. (as stated <a href="https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/doc/ExternalResources.md">here</a>)
`lv_obj_set_style_local_text_font(label, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, font);`
It can only use the _font_ specified here and cannot fallback to system font (e.g. jetbrains_mono_bold_20) if a glyph is not in this font. This means the custom font will have to duplicate everything in jetbrains_mono_bold_20 including ascii alnum, which is not very efficient.

Starting from LVGL 8.1, <a href="https://docs.lvgl.io/8.1/overview/font.html#use-font-fallback">a fallback font can be selected.</a> This way, only the extended glyphs (non-english alphabet, emoji ...) needs to be put in the custom font. Unfortunately, InfiniTime is currently using LVGL 7.

I think backporting this function should be considered since previous <a href="https://github.com/InfiniTimeOrg/InfiniTime/issues/616">attempt</a> to upgrade to LVGL 8 is stalled.

If I didn't miss anything, the following patch should be enough.
```diff
diff --git a/src/lv_draw/lv_draw_label.c b/src/lv_draw/lv_draw_label.c
index 970791ff..7cbf099b 100644
--- a/src/lv_draw/lv_draw_label.c
+++ b/src/lv_draw/lv_draw_label.c
@@ -433,6 +433,10 @@ LV_ATTRIBUTE_FAST_MEM static void lv_draw_letter(const lv_point_t * pos_p, const
         return;
     }
 
+    if (g.resolved_font) {
+        font_p = g.resolved_font;
+    }
+
[...]

PR #1773, opened 2023-06-18
initial i18n support for notification

Boteium:
I decide to write a PoC for notification i18n support https://github.com/InfiniTimeOrg/InfiniTime/issues/1473
This requires https://github.com/InfiniTimeOrg/lvgl/pull/3

This will enable people to start working on i18n support for their own languages.
The design uses external flash so that people can flash the language pack of their choice.
To reduce memory footprint, only the very basic glyph should be added, which means the example provided here (latin.json) should further be split to Spanish, French, Sweden ... etc

We should also limit the file size of this font to prevent people from building a huge font that will crash and brick their watches. So, It is not possible to add support for Chinese or Japanese using this method.

InfiniSim_2023-06-18_180701

.


everypizza1:
Can we also have this on the music screen? A lot of my playlists are Japanese music.

Boteium:
> Can we also have this on the music screen? A lot of my playlists are Japanese music.

Well, It depends,

Yes, we can make this method system wide.
But no, Japanese contains too much glyphs, which won't fit in the memory.

However, limited support is possible. (e.g. only includes hiragana and katakana)

Boteium:
> Can we also have this on the music screen? A lot of my playlists are Japanese music.

Well, It depends,

Yes, we can make this method system wide.
But no, Japanese contains too many glyphs, which won't fit in the memory.

However, limited support is possible. (e.g. only includes hiragana and katakana)

everypizza1:
Can we at least romanize it?

FintasticMan:
We can't romanize it on-device, that would require a lot of logic we don't want to implement, but the companion app would be able to.

ThatOneCalculator:
I think hiragana and katakana support would be better than nothing

minimaxwell:
I've been testing this for a while, this looks like a good step towards proper i18n support. I've been using the latin pack, without it, notifications in french are barely readable. Thanks for that PR, hopefully this can be merge at some point :)

OddKMS:
I support the splitting up the glyphs by language, it's a good way of getting the ball rolling on accent support for notifications.

While it would be nice to support "all" languages down the line (and I have no doubt someone will find a cleve way of achieving that), this would solve one of the main pain points for me: Messed-up text and message notifications in my own language.

Universal i18n: Nice to have.
Modular language pack so I can easily read incoming messages from my friends without transliteration: A pretty damn good step forward.

Issue #1571, opened 2023-01-29
Hardcoded apps


Riksu9000:
### Verification

- [X] I searched for similar issues and found none was relevant.

### Introduce the issue

There's an entry for every single screen in DisplayApp. To add an app, the app must be added in the app enum and the switch statement in LoadScreen must always be updated. This is too rigid, if our goal is to support third party apps, be it selectable at compile time, or dynamically.

### Preferred solution

Removing the Apps enum.

RageGamerBoi:
"At compile time" is something I hear a lot when referring to 3rd party apps, and I think we need to remember that most people don't want to recompile the OS every time an app is installed...
Dynamically is probably best, but I'm not 100% sure exactly what that would entail.

Avamander:
@RageGamerBoi 
> I think we need to remember that most people don't want to recompile the OS every time an app is installed...

I'm sure people are well-aware... It's just an inevitability at this point in time.

minacode:
Is this issue targeting some kind of AppController class? I guess for compile-time we could just import everything once in there and make everything else that handles apps use that controller. For dynamically loaded apps we would then reimplement the class to search the file system and load the apps from there.

brainrom:
My idea:
1. Split `screens` to different apps with different CMakeLists 
2. Write code generator, which "register" all apps to a map (name and pointer to construction method)
3. Store apps names as list on SPI flash and dynamically generate menu. This also allows to hide or rearrange apps without recompile.

According to recompilation for new apps installation, this isn't so bad. I have some developments, which allows to compile InfiniTime on Android smartphone with fancy UI. With ccache, it can be fast.

minacode:
Splitting in apps would be nice. 

Regarding the rest, do you have a good understanding how this this would work with FreeRTOS on the watch? There is some discussion about position independent code. Do you mean that?

brainrom:
Position independent code allows to install binary apps inside already built firmware. This is the best solution, but at this point, this is too complicated.
In my proposed solution, InfiniTime stays single binary, just compiled differently and allows slight modification of main menu without recompilation. Combining with "compilation on smartphone", it's almost enough.

minacode:
Ok, I get it now. So your code generator would more or less generate something like an AppController. But then you add the extra capability of rearranging the menu via a config file. 

I would say we do the following to work in little steps:
1. Implement the Controller with the app code base as is and verify that it works.
2. Maybe add the config file, if necessary.
[...]

PR #1775, opened 2023-06-20
Draft [Help wanted]: Unified app interface

minacode:
This is an experiment based on #1571. It is about reducing all the different places in the code where apps are declared.

I am not experienced with this, so any input would be really helpful. Maybe we can figure something out together. Take this PR as a basis for discussion. I would like to learn from this. And hopefully we can find a good solution on how to handle apps :relieved:  

The idea is to implement an `App` class that gets inherited by all `Screen`s we consider to be apps. This means all the screens that could one day be loaded dynamically (watch faces maybe their own thing...).
For example `QuickSettings` is not an `App`, but `Twos` is.

There are some issues to be resolved with the current state:
1. The apps are explicitly instanciated in the `DisplayApp.cpp` and with custom arguments. This PR tries to implement an `AppInterface` that provides the same way of creation for all apps. Right now, this is done via a `Get` method that is known to be present with all apps. We can further on reduce the capabilities that we give each app, but I would like to focus on getting this to run the easy way first.
2. The `ApplicationList` explicitly lists all the apps and their symbols. This PR implements a `GetSymbol` method for the `App` class. This moves the declaration of the symbol inside the app.
3. All apps are handled individually. This is not fixed right now, but I would like to have an `AppController` in the later progress that can handle the `App` objects. The interface of the `App` class must be enough to do all the things the controller has to do.

I would really like your ideas and feedback! :)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -595,6 +595,7 @@ set(INCLUDE_FILES
         displayapp/Messages.h
         displayapp/TouchEvents.h
         displayapp/screens/Screen.h
+        displayapp/screens/App.h
         displayapp/screens/Clock.h
         displayapp/screens/Tile.h
         displayapp/screens/InfiniPaint.h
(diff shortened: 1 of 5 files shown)


everypizza1:
Did you mean to close this?

minacode:
Yes, I am working on a newer version which isn't public yet.

Issue #1783, opened 2023-06-26
Weather: temp won't display in Fahrenheit


5iver:
### Verification

- [X] I searched for similar bug reports and found none was relevant.

### What happened?

Weather only reports in Celsius

### What should happen instead?

When specified, weather should report in Fahrenheit

### Reproduction steps

I am using Gadgetbridge (0.74.0) with the QuickWeather (2.3.2) app on Android. GadgetBridge is configured for Imperial and QuickWeather is configured for Fahrenheit. My family all have PineTimes running InfiniTime 1.13.0 and they are displaying the temperature in Celsius.

### More details?

I do not see any relevant issues reported for either of their apps. A quick look through their repositories did not lead me to believe the issue was elsewhere, especially because PineTime recently implemented weather functionality. Though, I could absolutely be incorrect and apologize in advance if that is the case.

Has anyone seen Fahrenheit temperatures using the new weather feature?

### Version

1.13.0

### Companion app

Gadgetbridge 0.74.0

kieranc:
This is correct, there's no imperial support yet, I believe it should be fairly easy to implement so I'll leave some notes here in case anyone gets to it before me.

[This](https://codeberg.org/Freeyourgadget/Gadgetbridge/src/commit/cc9c8d4207f746d5df79f62ca7b90f2d624a30fe/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/pebble/PebbleProtocol.java#L1151) is how it's implemented for the Pebble. It checks if GB is set to use imperial units and if so, recalculates the temperature sent to the watch. The watch deliberately doesn't include units after the degrees symbol so it should display fahrenheit just fine, at least below 99 degrees, after that there probably isn't space to show 3 digits. So don't go anywhere too hot.

[This](https://codeberg.org/Freeyourgadget/Gadgetbridge/src/commit/cc9c8d4207f746d5df79f62ca7b90f2d624a30fe/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/pinetime/PineTimeJFSupport.java#L753) is the relevant section for Pinetime.

Brod8362:
@kieranc and I talked a bit on discord and I want to summarize our conversation here:

- It seems like it could be possible to approach this from either the companion app side, or performing the conversion on the watch
- By performing it on the watch, any companion app that publishes weather data should be supported
- It should be easy for the user to switch their preferred unit (e.g no recompiling)
[...]

PR #1785, opened 2023-06-28
weather: Add functions for temperature in Fahrenheit

FintasticMan:
This adds some functions to the `WeatherService` class for getting the temperature in Fahrenheit. This could theoretically overflow, but only if the temperature is 164.27 °C or higher (or -164.28 °C or lower), which I believe is unlikely enough that we don't need to worry about it. Please tell me if you think that is an issue.

This doesn't add a setting to display the temperature in Fahrenheit in PTS.

Related to #1783.

### DIFF
--- a/src/components/ble/weather/WeatherService.cpp
+++ b/src/components/ble/weather/WeatherService.cpp
@@ -606,6 +606,13 @@ namespace Pinetime {
       return result;
     }
 
+    int16_t WeatherService::GetTempFahrenheit(int16_t temp) const {
+      if (temp == -32768) {
+        return -32768;
+      }
+      return temp * 9 / 5 + 3200;
+    }
+
     void WeatherService::CleanUpQcbor(QCBORDecodeContext* decodeContext) {
       QCBORDecode_ExitMap(decodeContext);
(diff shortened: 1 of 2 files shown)


Avamander:
I really wouldn't add the helper to the data struct definitions, that file is ideally just about the data.

There are other helper functions in the weather controller, that seems like a better place for it?

FintasticMan:
Ah OK, I was just looking at what you said [here](https://github.com/InfiniTimeOrg/InfiniTime/pull/353#issuecomment-993349135), but I've just pushed a change that adds the functions to the WeatherService instead.

Avamander (approved):
Looks good in this form. I'd appreciate if for example @kieranc could test it out with PTS, if he has the time.

matoro:
Hi, wondering if there's anything missing to allow this and the PTS change to be merged?  Was very surprised to discover weather service does not support Fahrenheit.

FintasticMan:
I don't think there's anything blocking this in particular. There are a couple of issues to be worked out with the weather implementation, which take priority over problems like Fahrenheit not being supported.

FintasticMan:
This is superseded by #1805.

PR #1791, opened 2023-07-03
new game: bird and cactus

Boteium:
I made a flappy bird clone.
It runs surprisingly smooth on the watch. Much smoother than InfiniSim.

https://github.com/InfiniTimeOrg/InfiniTime/assets/8587031/8094f736-3761-4da0-8385-98f609678f17


everypizza1:
Better name, please.

ProgramminCat:
cactusy bird??

minacode:
The only name in the code is "Bird", right?

Boteium:
Flappy Bird is a good name. But that's already taken.
flappy-something-else is also very likely to be in the Google search result.
Maybe we should avoid the word "flappy" altogether.

a13xie:
I might have an objection towards the scoring. Comparing the number of passed obstacles with other players might make more sense than number of seconds in-game.

Although in the end this likely doesn't matter all that much. Could also improve performance, but that's just my guess.

ProgramminCat:
InfiniBird??

JF002:
@Boteium Great job! This looks awesome, and yeah, the display is very smooth on the PineTime !

https://github.com/InfiniTimeOrg/InfiniTime/assets/2261652/6e885b29-bcf1-458b-9b66-483366399ea6

Not sure when we'll be able to review this PR, but the result is really good!

JustScott:
Updated to work on v1.14, available on my branch: https://github.com/JustScott/InfiniTime/tree/bird_app

This game is challenging!

src/CMakeLists.txt
```diff
@@ -383,6 +384,7 @@ list(APPEND SOURCE_FILES
         displayapp/screens/ApplicationList.cpp
         displayapp/screens/Notifications.cpp
         displayapp/screens/Twos.cpp
+        displayapp/screens/Bird.cpp
```
minacode, line 387: This was added two times.

src/displayapp/DisplayApp.cpp
```diff
@@ -14,7 +14,9 @@
 #include "displayapp/screens/FirmwareUpdate.h"
 #include "displayapp/screens/FirmwareValidation.h"
 #include "displayapp/screens/InfiniPaint.h"
+#include "displayapp/screens/Bird.h"
 #include "displayapp/screens/Paddle.h"
+#include "displayapp/screens/Bird.h"
```
minacode, line 19: Duplicate, see above.

src/displayapp/screens/Bird.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+  pos += accel / 6;
+
+  if (is_ascending) {
+    if (accel >= -18) {
+      accel -= 3;
+    }
+  } else {
+    if (accel <= 42) {
+      accel++;
```
minacode, line 71: Same as your other PR. I think adding variable names to the magic numbers would be a good idea.

src/displayapp/screens/Bird.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+  // checks if it has hit the floor or ceiling
+  if (pos <= 1 || pos >= LV_VER_RES - BIRD_SIZE) {
+    GameOver();
+    return;
+  }
+
+  // checks if it has rammed into cacti
+  // BIRD_X-CACTUS_WIDTH to BIRD_X+BIRD_SIZE
+  if (90 < cactus_x && cactus_x < 130) {
```
minacode, line 83: Make this more explicit with `constexpr`.

src/displayapp/screens/Bird.h
```diff
@@ -0,0 +1,63 @@
...
+#include <cstdint>
+#include "displayapp/screens/Screen.h"
+#include "Symbols.h"
+#include "systemtask/SystemTask.h"
+
+#define BIRD_X        110
+#define BIRD_SIZE     20
+#define CACTUS_HEIGHT 160
+#define CACTUS_WIDTH  30
+#define CACTUS_GAP    80
```
minacode, line 13: Use `constexpr` instead of the preprocessor.

PR #1795, opened 2023-07-07
new game: cat vs dog

Boteium:
Classic 2000s games.

https://github.com/InfiniTimeOrg/InfiniTime/assets/8587031/5506ab12-712b-463c-88a8-172ac0c60409


src/CMakeLists.txt
```diff
@@ -383,6 +384,7 @@ list(APPEND SOURCE_FILES
         displayapp/screens/ApplicationList.cpp
         displayapp/screens/Notifications.cpp
         displayapp/screens/Twos.cpp
+        displayapp/screens/CatDog.cpp
```
minacode, line 387: You are adding this file twice (see line 371)

src/displayapp/fonts/fonts.json
```diff
@@ -7,7 +7,7 @@
          },
          {
             "file": "FontAwesome5-Solid+Brands+Regular.woff",
-            "range": "0xf294, 0xf242, 0xf54b, 0xf21e, 0xf1e6, 0xf017, 0xf129, 0xf03a, 0xf185, 0xf560, 0xf001, 0xf3fd, 0xf1fc, 0xf45d, 0xf59f, 0xf5a0, 0xf027, 0xf028, 0xf6a9, 0xf04b, 0xf04c, 0xf048, 0xf051, 0xf095, 0xf3dd, 0xf04d, 0xf2f2, 0xf024, 0xf252, 0xf569, 0xf06e, 0xf015, 0xf00c, 0xf743"
```
minacode, line 10: Which icon (0xf743) gets removed here and why?

src/displayapp/screens/CatDog.cpp
```diff
@@ -0,0 +1,283 @@
...
+      // add a delay to reject previous press event
+      if (animationTime > 0) {
+        animationTime--;
+      } else {
+        if (isGameStopped) {
+          state = CatDog::State::Stopped;
+        } else {
+          state = CatDog::State::Prompt;
+        }
+      }
```
minacode, line 84:
```suggestion
      } else if (isGameStopped) {
        state = CatDog::State::Stopped;
      } else {
        state = CatDog::State::Prompt;
      }
```

src/displayapp/screens/CatDog.cpp
```diff
@@ -0,0 +1,283 @@
...
+      boneVX += wind; // acceleration caused by wind
+      boneX += boneVX;
+      boneY += boneVY;
+      lv_obj_set_pos(bone, boneX / 100, boneY / 100);
+      if (boneY >= HORIZON || boneX > 24000 || boneX < -2400) {
+        if (-2400 < boneX && boneX < 24000) {
+          if (isDogsTurn) {
+            handleDamage(abs(boneX - 21500) / 100);
+          } else {
+            handleDamage(abs(boneX - 500) / 100);
```
minacode, line 149: Where are all those magic numbers coming from? Can we replace them by a variable with a good name and `constexpr`?

src/displayapp/screens/CatDog.cpp
```diff
@@ -0,0 +1,283 @@
...
+void CatDog::Throw(int16_t speed) {
+  lv_obj_set_pos(powerBar, 240, 240); // hide it
+  boneY = HORIZON;
+  boneVX = speed;
+  boneVY = 0;
+  if (speed > 0) {
+    boneX = 500;
+    boneVY -= 25 * speed / 10;
+  } else {
+    // 100*(240-BONE_SIZE-5)
```
minacode, line 218: I think it is better to turn this into a `constexpr` instead.

src/displayapp/screens/CatDog.h
```diff
@@ -0,0 +1,86 @@
...
+
+#include <lvgl/lvgl.h>
+#include <cstdint>
+#include "displayapp/screens/Screen.h"
+#include "displayapp/apps/Apps.h"
+#include <displayapp/Controllers.h>
+#include "Symbols.h"
+#define HORIZON   18000
+#define MIN_POWER 200
+#define MAX_POWER 500
```
minacode, line 11: We try to avoid the preprocessor when possible. This can be a `constexpr` as well.

minacode (commented):
Found some small things. There are a lot of magic numbers that could benefit a lot from being either documented better or (even better) calculated so we know where they are coming from and they must not be hand-tuned again in case of later code changes.

Issue #1803, opened 2023-07-14
CasioStyleG7710's notification icon is only useful if notifications are muted


borkymcgee:
### Verification

- [X] I searched for similar issues and found none was relevant.

### Introduce the issue

CasioStyleG7710's notification icon appears to only show the user if they have new notifications since they opened the 'notifications' app.

Since the app opens automatically whenever a new notification comes in, unless the user has their notifications muted, the icon will never show up.

Additionally, if the user gets several notifications while notifications are muted, and then opens the notifications app, the notification icon will be dismissed even if they only see one out of those several messages.

### Preferred solution

I suggest that rather than show the user if they have any 'unread' notifications, the icon should show whether they have any 'undismissed' notifications, and potentially even show the number of undismissed notifications that there are.

This way if someone receives a notification and is unable to check their wrist at the moment, there is a reminder on the watchface.

### Version

v1.13.0

FintasticMan:
Duplicate of #1616 and related to #1234. The conclusion we've come to is to not change the behaviour until notifications are synced with the phone (i.e. notifications get dismissed on the phone when you dismiss them on the watch and vice versa).

PR #1804, opened 2023-07-14
Changed CasioStyleG7710 notification icon to show number of undismissed notifications

borkymcgee:
Previously, CasioStyleG7710's notification icon only showed an 'i' if there were any unread notifications (i.e. notifications that the watch recieved while notifications were muted)

This change implements the fix i proposed in #1803, displaying the number of undismissed notifications present instead


FintasticMan:
As mentioned in #1616, we will implement this in all watch faces once notifications are synced with the connected device.

borkymcgee:
Apologies, I forgot to show closed issues while searching for duplicates *facepalm*

FintasticMan:
No worries!

PR #1805, opened 2023-07-15
PTS: Add setting to display temperature in Fahrenheit

### LINKED ISSUE #1783: Weather: temp won't display in Fahrenheit

FintasticMan:
This adds a "weather format" setting, which can be either Metric or Imperial.

Fixes #1783.

@kieranc does this look good to you?


JF002:
@FintasticMan Thanks for this PR! I usually refrain from adding new settings but I think we cannot avoid this one : the world use 2 units to measure the temperature, we need to adapt :-)

Would you like to port those changes to the new Simple Weather Service implementation (https://github.com/InfiniTimeOrg/InfiniTime/pull/1924) ? We'll probably merge it soon.

FintasticMan:
Yes, I'll rebase this onto #1924 👍

src/components/ble/SimpleWeatherService.h
```diff
@@ -0,0 +1,143 @@
...
+          Icons iconId;
+        };
+
+        std::array<Day, MaxNbForecastDays> days;
+      };
+
+      std::optional<CurrentWeather> Current() const;
+      std::optional<Forecast> GetForecast() const;
+
+      static constexpr int16_t CelsiusToFahrenheit(int16_t celsius);
```
JF002, line 107: I don't think we need `constexpr` here, the function will most probably be evaluated at runtime anyway.

FintasticMan: We don't need it, you're right, but there's no harm in adding it.

JF002: In my opinion, `constexpr` adds some semantic to the code, it tells the reader that this method returns a constant value that might be computed at compile time. But this is not the case for this method : it can only be evaluated at runtime, and the `const` semantic does not apply here either.

I was a bit confused when I read it, so I think it would be better to remove it, if that's ok for you.

JF002 (commented):
Thanks, we'll merge this one after https://github.com/InfiniTimeOrg/InfiniTime/pull/1924 :)

Issue #1806, opened 2023-07-15
Metronome bpb dropdown cannot scroll up


borkymcgee:
### Verification

- [X] I searched for similar bug reports and found none was relevant.

### What happened?

The 'bpb' dropdown in the metronome app will scroll down, but not up

### What should happen instead?

The dropdown should be able to scroll up as well as down

### Reproduction steps

1. Open Metronome app
2. Tap bpb button
3. Scroll to the bottom of the list
4. Attempt to scroll back up to the top of the list

### More details?

Bug exists on both InfiniSim and on actual hardware

### Version

v1.13.0

PR #1808, opened 2023-07-15
Swiping up on metronome switches arc input to counter for higher precision

### LINKED ISSUE #1191: Metronome: add up/down buttons to increase/decrease BPM by 1 unit
### Verification

### Pitch us your idea!

Add buttons to UI to allow the user to increase or decrease BPM by 1 unit

### Description

While I appreciate the speed at which tempo can be adjusted in the 1.9.0 version of the metronome app, this comes at the cost of precision. For example, I find it difficult to precisely set a target BPM to a specific tempo. If I am aiming for 121 bpm for example, I can waste many attempts by swiping, landing on nearby tempos, from 119 through 127 bpm. It strikes me that by reducing the size of the radial tempo controller and shifting it to the left of the UI, two more buttons could be added to the interface- an up arrow and a down arrow. These buttons would either increase or decrease the BPM by one unit.

borkymcgee:
Swipe up to switch the arc to a counter to make precise BPMs easier to input
Swipe down to switch back to the arc

Implements fix proposed in https://github.com/InfiniTimeOrg/InfiniTime/issues/1191#issuecomment-1408550849
Resolves #1191

image image

### DIFF
--- a/src/displayapp/screens/Metronome.cpp
+++ b/src/displayapp/screens/Metronome.cpp
@@ -47,8 +47,8 @@ Metronome::Metronome(Controllers::MotorController& motorController, System::Syst
   bpmTap->user_data = this;
   lv_obj_set_event_cb(bpmTap, eventHandler);
   lv_obj_set_style_local_bg_opa(bpmTap, LV_BTN_PART_MAIN, LV_STATE_DEFAULT, LV_OPA_TRANSP);
-  lv_obj_set_height(bpmTap, 80);
-  lv_obj_align(bpmTap, bpmCounter.GetObject(), LV_ALIGN_CENTER, 0, 0);
+  lv_obj_set_height(bpmTap, 130);
+  lv_obj_align(bpmTap, bpmCounter.GetObject(), LV_ALIGN_CENTER, 0, 30);
 
   bpbDropdown = lv_dropdown_create(lv_scr_act(), nullptr);
   bpbDropdown->user_data = this;
--- a/src/displayapp/screens/Metronome.cpp
+++ b/src/displayapp/screens/Metronome.cpp
@@ -39,15 +39,14 @@ Metronome::Metronome(Controllers::MotorController& motorController, System::Syst
   createLabel("bpm", bpmValue, LV_ALIGN_OUT_BOTTOM_MID, &jetbrains_mono_bold_20, 0, 0);
 
   bpmCounter.Create();
-  bpmCounter.SetValue(120);
+  bpmCounter.SetValue(bpm);
   lv_obj_align(bpmCounter.GetObject(), lv_scr_act(), LV_ALIGN_IN_TOP_MID, 0, 0);
   lv_obj_set_hidden(bpmCounter.GetObject(), true);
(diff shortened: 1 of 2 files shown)


minacode:
Two UIs seem like one too much for such a restricted device. I agree with the solution, though. Do you think the bpb setting could be made into a slimmer counter widget right of the tempo counter as well? With a long play/pause button underneath?

borkymcgee:
@minacode I could be wrong but i think the counter widget

borkymcgee:
Oops! Misclicked *blush* sorry!

borkymcgee:
@minacode I think you can have arbitrarily sized counter elements, so you COULD probably do that.

I'm not sure it's a good idea though:

-I think it'd look garish
-That's a lot of little touch buttons for one small screen
-I'm not sure there are any problems with the dropdown as-is, it seems a good UI fit for the application

minacode:
Ok, fair points. I almost never use the metronome, so maybe I should not argue too much about its usability.

1. This is the only drop-down in the whole code base, right? Could we make the UI more consistent and remove some lvgl code if we replace it? The diff of the resulting image would be interesting.
2. The argument regarding too much stuff is not that valid, because the time settings or the alarm have the same UI with two counters.

borkymcgee:
I don't use the metronome function either, so it's hard to know what usability is like

I was imagining the bpb counter being smaller than the bpm one, since it's less important. I like the way the current design puts the bpm front and center and big, with bpb and start/stop smaller. It gives the app a simple, clean feel.

If we can figure out a good alternative, removing the dropdown might be a good idea, especially since it's a bit buggy (#1806).

Baggypants:
I want to use the metronome but it's current input is utterly hopeless, so I don't. Tapping the requested bpm in binary coded decimal using morse code would be better ux than what you have currently.

Having an 'up' only button that increments every value between 0 and 600 and loops round to zero would be more accurate, and faster.

Baggypants:
What about using the input method for the stop watch, move to colon to be between the 3rd and 4th element so the first three would set hundreds, tens and units and the 4th one would be 0-9 for beat. Sure anyone who has a piece in 11/4 or 17/4 might complain but I'd suggest they already have bigger problems.

mark9064:
The one thing I like about the arc is that keying in the BPM approximately is pretty fast - ideally you want to be able to quickly select any frequency (roughly) in the range and then refine it.

I'm not totally sure what'd work better though. Does tap tempo still work with this branch? I think tapping to set it approximately and then refining with the spinner would be OK

NeroBurner:
just got bit by the not precise enough input to the metronome. I don't know if this PR is the best way to solve the issue, but it is much better than just the arc.

Adding it to Milestone 1.16.0 to get some eyes on the PR

@borkymcgee would you be interested to rebase the PR on the current main branch?

minacode:
To give this some direction: if you will build the app only with the second screen with the +-number input with tapping on the number as alternative input and the bpm as a simple button that cycles through numbers (2-8?; like the "once"/"mon-fri"/"daily" button for the alarm), I will approve it.

Issue #1788, opened 2023-06-30
Out of memory using weather with Gadgetbridge, causing font corruption


Hegz:
### Verification

- [X] I searched for similar bug reports and found none was relevant.

### What happened?

Clock text becomes corrupt

### What should happen instead?

I was expecting to see readable fonts

### Reproduction steps

I've updated to 1.13, and enabled weather with gadget bridge and quickweather.  After about 9 hours being on on my wrist, I glanced at my watch and the clock font's were corrupt.  This has been happening for the past several days after updating to 1.13 and enabling weather support.  I don't know the time periods for the other times.

The following are corrupted:
- Pinetimestyle -- Clock text unreadable.  Date, weather, steps are OK.
- Infineat face -- Clock text unreadable.  Day of week, date, and steps are OK. 
- Casio G7710 -- Clock text unreadable.  Date/week/steps battery text are OK.

Unaffected faces
- Digital
- Analog
- Terminal

### More details?

I can switch to an unaffected face, and the watch will continue to work as normal.  Attempting to flip back to Pinetimestyle causes the screen to lock before displaying the Pinetimestyle face, and then it reboots.

The previous reboot was at about 6:00am, which was a spontaneous reboot.  I didn't see if there was corruption prior to the reboot.  The corruption kicked in at around 3:00pm the same day.

I've attempted to reinstall the assets pack, as I thought it may have changed since I first installed it.  This didn't change the behaviour.

### Version

1.13

### Companion app

Gadget Bridge

Hegz:
The text corruption appears the same as shown on #1376

FintasticMan:
When this happens, can you check the values shown under "Memory heap" on the 3rd page of the About page (you can get there quickly by holding the button for 2-3 seconds)? I have a feeling that that is caused by running out of memory.

kieranc:
> The text corruption appears the same as shown on #1376

I disagree, this is a new and different corruption, sadly. It's now more transparent, with vertical banding. The extra strange bit is that PTS fonts are internal, previously only fonts loaded from ext flash were affected but now it seems to be affecting internal fonts as well. I saw this behaviour a couple of days ago but it only affected the temperature label on PTS, while I was provoking it, so I thought I was to blame... but apparently not.

JF002:
Yep, I agree with @kieranc, I don't think this issue has the same cause than #1376 since PTS does not load anything from the external storage.

[...]

PR #1822, opened 2023-08-07
Improve weather service

faxe1008:
I wanted to investigate this issue:
[https://github.com/InfiniTimeOrg/InfiniTime/issues/1788](https://github.com/InfiniTimeOrg/InfiniTime/issues/1788)

I found some things that could be refactored, and also some issues with the current implementation:

- Already expired events are added to the timeline and deleted right afterwards. This causes unnecessary alloc/free churn and I suspect more heap fragmentation. 
- The tidying of the timeline happens after the addition of the pending event. This means that if the vector is at its capacity limit it will need to reallocate even if there are expired events that could be cleaned beforehand. 

Further investigation needs to be done here:
- Is the CBOR code (especially the one using UsefullBufC) leaking memory? (Couldnt figure out that quickly what the library expects in terms of cleanup)
- Maybe using a fixed size array and no heap allocation? This will increase the flash size, but there is less potential for runtime alloc errors?

### DIFF
--- a/src/components/ble/weather/WeatherService.cpp
+++ b/src/components/ble/weather/WeatherService.cpp
@@ -66,6 +66,14 @@ namespace Pinetime {
           CleanUpQcbor(&decodeContext);
           return BLE_ATT_ERR_INVALID_ATTR_VALUE_LEN;
         }
+
+        // Do not add already expired events
+        uint64_t currentTimestamp = GetCurrentUnixTimestamp();
+        if (static_cast<uint64_t>(tmpTimestamp + tmpExpires) < currentTimestamp) {
+          CleanUpQcbor(&decodeContext);
+          return BLE_ATT_ERR_INVALID_ATTR_VALUE_LEN;
+        }
+
         int64_t tmpEventType = 0;
         QCBORDecode_GetInt64InMapSZ(&decodeContext, "EventType", &tmpEventType);
         if (QCBORDecode_GetError(&decodeContext) != QCBOR_SUCCESS || tmpEventType < 0 ||
@@ -73,6 +81,8 @@ namespace Pinetime {
           CleanUpQcbor(&decodeContext);
           return BLE_ATT_ERR_INVALID_ATTR_VALUE_LEN;
         }
+
+        TidyTimeline();
 
         switch (static_cast<WeatherData::eventtype>(tmpEventType)) {
           case WeatherData::eventtype::AirQuality: {
@@ -363,9 +373,9 @@ namespace Pinetime {
           }
         }
 
+        std::sort(std::begin(timeline), std::end(timeline), CompareTimelineEvents);
+
         QCBORDecode_ExitMap(&decodeContext);
-        GetTimelineLength();
-        TidyTimeline();
 
         if (QCBORDecode_Finish(&decodeContext) != QCBOR_SUCCESS) {
           return BLE_ATT_ERR_INSUFFICIENT_RES;
--- a/src/components/ble/weather/WeatherService.cpp
+++ b/src/components/ble/weather/WeatherService.cpp
@@ -401,103 +401,52 @@ namespace Pinetime {
       return 0;
     }
 
-    std::unique_ptr<WeatherData::Clouds>& WeatherService::GetCurrentClouds() {
+    template <class T>
+    std::unique_ptr<T>& WeatherService::GetEventOfType(WeatherData::eventtype eventType) {
       uint64_t currentTimestamp = GetCurrentUnixTimestamp();
       for (auto&& header : this->timeline) {
-        if (header->eventType == WeatherData::eventtype::Clouds && IsEventStillValid(header, currentTimestamp)) {
-          return reinterpret_cast<std::unique_ptr<WeatherData::Clouds>&>(header);
+        if (header->eventType == eventType && IsEventStillValid(header, currentTimestamp)) {
+          return reinterpret_cast<std::unique_ptr<T>&>(header);
         }
       }
 
-      return reinterpret_cast<std::unique_ptr<WeatherData::Clouds>&>(*this->nullHeader);
+      return reinterpret_cast<std::unique_ptr<T>&>(*this->nullHeader);
+    }
+
+    std::unique_ptr<WeatherData::Clouds>& WeatherService::GetCurrentClouds() {
+      return GetEventOfType<WeatherData::Clouds>(WeatherData::eventtype::Clouds);
     }
 
     std::unique_ptr<WeatherData::Obscuration>& WeatherService::GetCurrentObscuration() {
-      uint64_t currentTimestamp = GetCurrentUnixTimestamp();
-      for (auto&& header : this->timeline) {
-        if (header->eventType == WeatherData::eventtype::Obscuration && IsEventStillValid(header, currentTimestamp)) {
-          return reinterpret_cast<std::unique_ptr<WeatherData::Obscuration>&>(header);
-        }
-      }
-
-      return reinterpret_cast<std::unique_ptr<WeatherData::Obscuration>&>(*this->nullHeader);
+      return GetEventOfType<WeatherData::Obscuration>(WeatherData::eventtype::Obscuration);
     }
 
     std::unique_ptr<WeatherData::Precipitation>& WeatherService::GetCurrentPrecipitation() {
-      uint64_t currentTimestamp = GetCurrentUnixTimestamp();
-      for (auto&& header : this->timeline) {
-        if (header->eventType == WeatherData::eventtype::Precipitation && IsEventStillValid(header, currentTimestamp)) {
-          return reinterpret_cast<std::unique_ptr<WeatherData::Precipitation>&>(header);
-        }
-      }
-
-      return reinterpret_cast<std::unique_ptr<WeatherData::Precipitation>&>(*this->nullHeader);
+      return GetEventOfType<WeatherData::Precipitation>(WeatherData::eventtype::Precipitation);
     }
 
     std::unique_ptr<WeatherData::Wind>& WeatherService::GetCurrentWind() {
-      uint64_t currentTimestamp = GetCurrentUnixTimestamp();
-      for (auto&& header : this->timeline) {
-        if (header->eventType == WeatherData::eventtype::Wind && IsEventStillValid(header, currentTimestamp)) {
-          return reinterpret_cast<std::unique_ptr<WeatherData::Wind>&>(header);
-        }
-      }
-
-      return reinterpret_cast<std::unique_ptr<WeatherData::Wind>&>(*this->nullHeader);
+      return GetEventOfType<WeatherData::Wind>(WeatherData::eventtype::Wind);
     }
 
     std::unique_ptr<WeatherData::Temperature>& WeatherService::GetCurrentTemperature() {
-      uint64_t currentTimestamp = GetCurrentUnixTimestamp();
-      for (auto&& header : this->timeline) {
-        if (header->eventType == WeatherData::eventtype::Temperature && IsEventStillValid(header, currentTimestamp)) {
-          return reinterpret_cast<std::unique_ptr<WeatherData::Temperature>&>(header);
-        }
-      }
-
-      return reinterpret_cast<std::unique_ptr<WeatherData::Temperature>&>(*this->nullHeader);
+      return GetEventOfType<WeatherData::Temperature>(WeatherData::eventtype::Temperature);
     }
 
     std::unique_ptr<WeatherData::Humidity>& WeatherService::GetCurrentHumidity() {
-      uint64_t currentTimestamp = GetCurrentUnixTimestamp();
-      for (auto&& header : this->timeline) {
-        if (header->eventType == WeatherData::eventtype::Humidity && IsEventStillValid(header, currentTimestamp)) {
-          return reinterpret_cast<std::unique_ptr<WeatherData::Humidity>&>(header);
-        }
-      }
-
-      return reinterpret_cast<std::unique_ptr<WeatherData::Humidity>&>(*this->nullHeader);
+      return GetEventOfType<WeatherData::Humidity>(WeatherData::eventtype::Humidity);
     }
 
     std::unique_ptr<WeatherData::Pressure>& WeatherService::GetCurrentPressure() {
-      uint64_t currentTimestamp = GetCurrentUnixTimestamp();
-      for (auto&& header : this->timeline) {
-        if (header->eventType == WeatherData::eventtype::Pressure && IsEventStillValid(header, currentTimestamp)) {
-          return reinterpret_cast<std::unique_ptr<WeatherData::Pressure>&>(header);
-        }
-      }
-
-      return reinterpret_cast<std::unique_ptr<WeatherData::Pressure>&>(*this->nullHeader);
+      return GetEventOfType<WeatherData::Pressure>(WeatherData::eventtype::Pressure);
     }
 
     std::unique_ptr<WeatherData::Location>& WeatherService::GetCurrentLocation() {
-      uint64_t currentTimestamp = GetCurrentUnixTimestamp();
-      for (auto&& header : this->timeline) {
-        if (header->eventType == WeatherData::eventtype::Location && IsEventStillValid(header, currentTimestamp)) {
-          return reinterpret_cast<std::unique_ptr<WeatherData::Location>&>(header);
-        }
-      }
-
-      return reinterpret_cast<std::unique_ptr<WeatherData::Location>&>(*this->nullHeader);
+      return GetEventOfType<WeatherData::Location>(WeatherData::eventtype::Location);
     }
 
     std::unique_ptr<WeatherData::AirQuality>& WeatherService::GetCurrentQuality() {
-      uint64_t currentTimestamp = GetCurrentUnixTimestamp();
-      for (auto&& header : this->timeline) {
-        if (header->eventType == WeatherData::eventtype::AirQuality && IsEventStillValid(header, currentTimestamp)) {
-          return reinterpret_cast<std::unique_ptr<WeatherData::AirQuality>&>(header);
-        }
-      }
-
-      return reinterpret_cast<std::unique_ptr<WeatherData::AirQuality>&>(*this->nullHeader);
+      return GetEventOfType<WeatherData::AirQuality>(WeatherData::eventtype::AirQuality);
     }
 
     size_t WeatherService::GetTimelineLength() const {
--- a/src/components/ble/weather/WeatherService.h
+++ b/src/components/ble/weather/WeatherService.h
@@ -146,6 +146,9 @@ namespace Pinetime {
       static bool CompareTimelineEvents(const std::unique_ptr<WeatherData::TimelineHeader>& first,
                                         const std::unique_ptr<WeatherData::TimelineHeader>& second);
 
+      template <class T>
+      std::unique_ptr<T>& GetEventOfType(WeatherData::eventtype eventType);
+
       /**
        * Returns current UNIX timestamp
        */
--- a/src/components/ble/weather/WeatherService.h
+++ b/src/components/ble/weather/WeatherService.h
@@ -87,6 +87,10 @@ namespace Pinetime {
        */
       bool HasTimelineEventOfType(WeatherData::eventtype type) const;
 
+      const std::vector<std::unique_ptr<WeatherData::TimelineHeader>>& GetEventTimeline() const {
+        return timeline;
+      }
+
     private:
       // 00040000-78fc-48fe-8e23-433b3a1942d0
       static constexpr ble_uuid128_t BaseUuid() {
--- a/src/displayapp/Apps.h
+++ b/src/displayapp/Apps.h
@@ -37,6 +37,7 @@ namespace Pinetime {
       SettingChimes,
       SettingShakeThreshold,
       SettingBluetooth,
+      Weather,
       Error
     };
   }
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -540,11 +540,10 @@ void DisplayApp::LoadScreen(Apps app, DisplayApp::FullRefreshDirections directio
     case Apps::Metronome:
       currentScreen = std::make_unique<Screens::Metronome>(motorController, *systemTask);
       break;
-    /* Weather debug app
+    /* Weather debug app*/
     case Apps::Weather:
       currentScreen = std::make_unique<Screens::Weather>(this, systemTask->nimble().weather());
       break;
-    */
     case Apps::Steps:
       currentScreen = std::make_unique<Screens::Steps>(motionController, settingsController);
       break;
--- a/src/displayapp/screens/ApplicationList.h
+++ b/src/displayapp/screens/ApplicationList.h
@@ -52,7 +52,7 @@ namespace Pinetime {
           {"2", Apps::Twos},
           {Symbols::drum, Apps::Metronome},
           {Symbols::map, Apps::Navigation},
-          {Symbols::none, Apps::None},
+          {"W", Apps::Weather},
 
           // {"M", Apps::Motion},
         }};
--- a/src/displayapp/screens/Weather.cpp
+++ b/src/displayapp/screens/Weather.cpp
@@ -31,6 +31,9 @@ Weather::Weather(Pinetime::Applications::DisplayApp* app, Pinetime::Controllers:
     screens {app,
              0,
              {[this]() -> std::unique_ptr<Screen> {
+                return CreateTimelineStatsPage();
+              },
+              [this]() -> std::unique_ptr<Screen> {
                 return CreateScreenTemperature();
               },
               [this]() -> std::unique_ptr<Screen> {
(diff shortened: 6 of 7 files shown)


Avamander:
> Already expired events are added to the timeline and deleted right afterwards. This causes unnecessary alloc/free churn and I suspect more heap fragmentation.

Great point.

> Is the CBOR code (especially the one using UsefullBufC) leaking memory? (Couldnt figure out that quickly what the library expects in terms of cleanup)

It really shouldn't, I did try and make sure everything is checked, closed and cleaned up (according to the documentation). Adding timeline support to InfiniSim using the same library might shed some light though.

> Maybe using a fixed size array and no heap allocation? This will increase the flash size, but there is less potential for runtime alloc errors?

I did initially consider an array per type of event (because different types of events have different-sized data structures), but that's IMO way more wasteful.

src/components/ble/weather/WeatherService.cpp
```diff
@@ -401,103 +409,52 @@ namespace Pinetime {
...
     }
 
-    std::unique_ptr<WeatherData::Clouds>& WeatherService::GetCurrentClouds() {
+    template <class T>
+    std::unique_ptr<T>& WeatherService::GetEventOfType(WeatherData::eventtype eventType) {
       uint64_t currentTimestamp = GetCurrentUnixTimestamp();
       for (auto&& header : this->timeline) {
-        if (header->eventType == WeatherData::eventtype::Clouds && IsEventStillValid(header, currentTimestamp)) {
-          return reinterpret_cast<std::unique_ptr<WeatherData::Clouds>&>(header);
+        if (header->eventType == eventType && IsEventStillValid(header, currentTimestamp)) {
```
Avamander, line 416: Oops. :sweat:

src/components/ble/weather/WeatherService.h
```diff
@@ -146,6 +146,9 @@ namespace Pinetime {
       static bool CompareTimelineEvents(const std::unique_ptr<WeatherData::TimelineHeader>& first,
                                         const std::unique_ptr<WeatherData::TimelineHeader>& second);
 
+      template <class T>
+      std::unique_ptr<T>& GetEventOfType(WeatherData::eventtype eventType);
```
Avamander, line 150: Templates... this probably adds a version of this function of each event of type `eventtype`. Which is why I kinda just added the few that might be immediately needed.

It might be better to deprecate the named functions and force everything to use this one, to avoid at least one level of duplication.

src/components/ble/weather/WeatherService.cpp
```diff
@@ -401,103 +409,52 @@ namespace Pinetime {
...
     std::unique_ptr<WeatherData::Temperature>& WeatherService::GetCurrentTemperature() {
-      uint64_t currentTimestamp = GetCurrentUnixTimestamp();
-      for (auto&& header : this->timeline) {
-        if (header->eventType == WeatherData::eventtype::Temperature && IsEventStillValid(header, currentTimestamp)) {
-          return reinterpret_cast<std::unique_ptr<WeatherData::Temperature>&>(header);
-        }
-      }
-
-      return reinterpret_cast<std::unique_ptr<WeatherData::Temperature>&>(*this->nullHeader);
+      return GetEventOfType<WeatherData::Temperature>(WeatherData::eventtype::Temperature);
```
Avamander, line 441: This might actually be slightly more involved than picking the first valid event.

Ideally we'd pick the most recently (ideally added, but that might be a waste) valid event, with the shortest expiricy but not the one soonest to expire. Goal being to pick the freshest forecast for the near immediate future rather than a daily forecast, should there be one.

Keeping the current behaviour with this PR would be fine though.

src/components/ble/weather/WeatherService.cpp
```diff
@@ -66,6 +66,14 @@ namespace Pinetime {
           CleanUpQcbor(&decodeContext);
           return BLE_ATT_ERR_INVALID_ATTR_VALUE_LEN;
         }
+
+        // Do not add already expired events
```
Avamander, line 70: Even though it would be a valid approach to be more defensive here, I would really really rather have the companions not to write :hankey: to the timeline at all.

src/components/ble/weather/WeatherService.h
faxe1008: Nope, the template is only depedant on the `class T` type, and it will only create a version for each one of those. I purposely did not make the event type a template parameter. I checked the final binary. The compiler is even smart enough to inline the call to this function for each instance.

src/components/ble/weather/WeatherService.cpp
faxe1008: Hmm, you mean the previous formatting?

src/components/ble/weather/WeatherService.cpp
faxe1008: Yeah, I did not want to change any behaviour, just doing some clean up.
Previously the invariant was that after insertion the events are tidied (removed expired and sorted) therefore the first pick was actually the nearest not yet expired one :^). My change broke this a wee bit, we still need to sort after insertion.

src/components/ble/weather/WeatherService.cpp
faxe1008: Its kinda cheap flashwise and I figured, why not early abort here :^)

src/components/ble/weather/WeatherService.h
Avamander: > and it will only create a version for each one of those

Yep.

> The compiler is even smart enough to inline the call to this function for each instance.

That's nice. Though without LTO the unused ones will remain right now, right?

Avamander:
> Sure this will increase flash size, but when the alternative is to have alloc fails when loading font or something similar

Not just flash though, it will consume that amount of RAM (permanently). Which is already spread really quite thin.

> I'd rather not have the most up to date weather :^).

The intent is to extend the timeline to contain events of other types as well. Such as calendar events or reminders, but we'll have to carefully consider if we can afford that.

faxe1008:
> The intent is to extend the timeline to contain events of other types as well. Such as calendar events or reminders, but we'll have to carefully consider if we can afford that.

Would love that one. However I think it might be smart to have a fixed max size for the vector, thats tried and tested to leave some breathing room for other allocations :^). Also maybe there is a way to tune the `std::vector` behaviour in regards to load and growth factor to be more conservative somehow? My knowledge of the STL is a bit rusty.

faxe1008:
@Avamander 
So what I found is that at most the TinyWeatherForeCast app pushed ~98 events to the timeline, increasing the vector capacity to 128. Also what I saw is that if you hit the refresh button in the app the events are pushed to the watch again even if the values did not change. 

I think to validate strategies the best course of action would be to gather "real life" data at what point in time the app pushes which data and do some simulations on the vector behaviour. I checked the app and it offers protocolling functionality but it does not log the expiration and the fields "verbatim" as the bluetooth communication uses it. There is some work to be done to change the logging a tad bit.

Things that I think of adding:
- Hard limit the vector capacity at `128`. 256 is still possible but this only leaves a minimum free of `328 bytes` for me which is not ideal.  
- Deduplicate the events, meaning that the same event can not be added twice
- Maybe add a conditional `shrink_to_fit` when tidying the vector if we cleaned elements and the load factor dropped below say `70%`?

What do you think?

minacode:
I have a (semi-informed) question: 
Could you put all the stuff from the timeline into files, store only the most necessary metadata, eg. timestamp and file id and then read everything else dynamically from external memory when needed?

faxe1008:
@minacode any input is welcome :^). I think this is maybe possible, to store the data  on the external flash. My thoughts on this:

- Querying and housekeeping gets more complicated
- Possible battery life will be affected
- There will be more wear on the external flash this way
- Even if we move to external flash we still need a limiting mechanism of some kind

Eventually if we ever retrieve calender events storing the to external flash is promising so they would persist across reboots. For the weather I am unsure if that makes sense.

I think that for the weather data we need some clear target to go for. Should we only care about the most current not expired event?

JF002:
Thanks everyone for this analysis so far! I'm now joining the party!

I've just had a look at `WeatherService` and `WeatherData` while keeping in mind the memory issues that were reported regarding the weather feature. I also found a few points in the implementation that could explain those issues:

 - most of  `WeatherData` types are structs where the compiler seems to add some padding bytes.
 - Some of those types uses std::string, which pre-allocates a few bytes for the string and dynamically allocates a bigger buffer if needed
 - Here are the sizes returned by `sizeof()` on the data types (not on initialized object with actual strings). As you can see those data types need between 16 and 56B of RAM storage.

```
<info> app: Size TimelineHeader, 16
<info> app: Size Location, 56
<info> app: Size Clouds, 24
<info> app: Size Obscuration, 24
<info> app: Size Precipitation, 24
<info> app: Size Wind, 24
<info> app: Size Temperature, 24
<info> app: Size Humidity, 24
<info> app: Size Pressure, 24
<info> app: Size AirQuality, 48
```

 - Let's say we receive 98 events from the weather apps, this means that we'll need between 1568B and 5488B to store those events (without accounting for the actual strings). With a vector of 128 elements, the total data size would be, at worse, 7168B, which is more than 10% of the total RAM memory available!
 - Those objects are dynamically allocated/de-allocated. Since they are quite small, this probably increases memory fragmentation. If the memory is too fragmented, bigger memory allocations (fonts, for example) might fail.
 - As @faxe1008 pointed out, `WeatherService` uses an std::vector<> of those data types. This vector will automatically grow when needed, and potentially use twice the amount of RAM during the growing step.

All in all, we can see that `WeatherService` is quite hungry in RAM space, and that it'll use as much memory as needed, potentially at the expense of other functionalities (like fonts and watchfaces).

So, how can we fix/improve this?

I know that this timeline implementation was designed to be as generic as possible. We are currently using it to store weather data, but the bigger plan was to use it for other functionalities like navigation directions, agenda events,... which is a good idea!

However, now that we face those memory issues, the current implementation does not seem optimal due to its memory usage and the limited amount of RAM memory available on the PineTime.

I guess we can explore a few options 

 - Find other ways to store the data that would be more memory efficient
 - Avoid padding using `#pragma pack`, replace `std::string` with `const char*`
 - Avoid using dynamic allocation to avoid memory fragmentation and (unlimited) growing memory usage
 - Put some limits on the number of events we want to handle
 - ...

As an example, I designed the notification functionality with strong limitations : max 5 notifications of max 100 characters. This implementation is not the most effective one since we cannot store 10 notifications of 15 characters even if there's theoretically enough room for them, but it's at least simple and deterministic : it uses 500B of RAM memory. And since this memory is statically allocated, it'll never interfere with another feature.

I'm not saying we should implement the weather feature this way, but I think that compromises and limitations are probably unavoidable here.

Let's have a look at the current usage of the weather data in InfiniTime : we only want to display the current temperature and precipitation. The timeline allows the companion app to send multiple events (of the same time) with different expiry time, but we currently use only the 1st valid event available in the timeline.

I know that we *could* use more functionalities and data in the future, but maybe we should think about which data we'll realistically use. Do we really think we'll display all these data? Or, put another way, maybe we should decide which data we really want to handle, and remove the other ones.

I'm also wondering *why* would the weather companion apps send 98 events for a single update? Do all these events contains actual data? Even if I think InfiniTime should be able to "defend" itself against a companion app sending too much data, there are maybe a few improvements to be done on the companion app side?

faxe1008:
> I'm also wondering _why_ would the weather companion apps send 98 events for a single update?

Sorry if I did not phrase it correctly, but what I meant is that at one point in time there were 98 elements in the vector. A single update from the TWFG app yields ~6-9 events.


> I know that we _could_ use more functionalities and data in the future, but maybe we should think about which data we'll realistically use. Do we really think we'll display all these data? Or, put another way, maybe we should decide which data we really want to handle, and remove the other ones.

+1 for me on this one. For now I think keeping the value closest to the current time would be sufficient, but I do not know if we reject the others if the apps do push previous values again?

Avamander:
> So what I found is that at most the TinyWeatherForeCast app pushed ~98 events to the timeline, increasing the vector capacity to 128. Also what I saw is that if you hit the refresh button in the app the events are pushed to the watch again even if the values did not change.

That seems to be indeed quite excessive and ideally we'd handle this misbehaviour.

> Deduplicate the events, meaning that the same event can not be added twice

That would be a nice addition.

> Eventually if we ever retrieve calender events storing the to external flash is promising so they would persist across reboots. For the weather I am unsure if that makes sense.

Especially with longer string content (descriptions, full titles) this seems like a logical approach.

Though it wouldn't hurt too much to periodically flush the entire timeline to storage, deallocate the entire list and then reallocate. This could reduce fragmentation, in theory.

> Avoid padding using #pragma pack

This seems like a fairly simple and safe change, though it might need some testing.

>  it uses 500B of RAM memory. And since this memory is statically allocated, it'll never interfere with another feature.

That's a large allocation for a feature that might not be used. It would also mean a custom allocator (considering the different-sized elements), which is a significant amount of complexity.

> there are maybe a few improvements to be done on the companion app side?

Certainly. Though it seems quite weird there are that many events at the same time. Has anyone gotten a dump of all the events?

> I know that we could use more functionalities and data in the future, but maybe we should think about which data we'll realistically use. Do we really think we'll display all these data? Or, put another way, maybe we should decide which data we really want to handle, and remove the other ones.

We could possibly discard the ones we currently certainly don't use, but the compatibility exists for the explicit reason to make it possible to write things that do use this data. Plus this wide compatibility has now revealed certain assumptions, such as well-behaving companions, to be false. I wouldn't make large alternations to our current approach, rather fortify it against abuse (upper sanity limits), improve storage (deduplication, packing) and data usage (which events do we use, clearer instructions about which data should be pushed and how).

faxe1008:
Since discussing abstract ideas is quite difficult I opted to go for a more methodical approach. I want to actually see what is going on in regards to memory consumption etc. without having to build the watch firmware, reflash it and wait. 

I modified the Gadgetbridge app to log the data send to the watch as a json. In parallel I started work on a benchmarking simulator for the behaviour where I can plot different implementations against each other:
[https://github.com/faxe1008/InifiniteWeather](https://github.com/faxe1008/InifiniteWeather)

The idea is to take the heap allocator from infinitime and make it available as a C++ allocator to be used with STL-containers and smart-pointers. Overriding the malloc/free symbols globally  I avoided on purpose since this might capture unwanted test infrastructure noise. Anyways the plan is to feed the captured data into the simulators which will yield effectively this:

```
Strategy:             SimpleVectorTimeline
Allocations:        1
Deallocations:   0
OOM:                  0
Lowest Ever:      40920
Fragmentation: 0
Leaking:              false
```

This will give an actual point of comparison for different implementations, which should be close to (if not identical) to the watches behaviour.

faxe1008:
Finished implementation of the simulator and captured 3 days of weather log data, that I will use to benchmark

JF002:
Thanks for continuing the analysis! Your new testing method will probably provide valuable data!

I do plan on having a closer look on this issue soon so we can try to figure out the best solution for this issue!

JF002:
I added a few debug logs to see what exactly happens when weather data is received by [WeatherService](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/src/components/ble/weather/WeatherService.cpp#L43). For each event, I log the following info:

 - Timestamp, expiration and type of the command
 - Heap memory monitoring (same info than the one displayed in the settings page)
 - The current size and max size of the timeline queue.
 - malloc/realloc/free from the [heap manager](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/src/FreeRTOS/heap_4_infinitime.c#L118) (I added a debug log that is enabled only during the call to [WeatherService::OnCommand()](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/src/components/ble/weather/WeatherService.cpp#L43).

```
<info> app: [Weather Service] Received weather command : 
        Timestamp : 1694357103
        Expiration(h) : 6
        Type : Location
<info> app: [Weather Service] Memory : free : 13376 - min free : 13080, malloc errors : 0 - stack overflow errors : 0
<info> app: [FreeRTOS HEAP] ALLOC 64B -> 536896376
<info> app: [FreeRTOS HEAP] ALLOC 16B -> 536896440
<info> app: [Weather Service] AddEventToTimeline - Current size: 1 - Max size: 536870911
```

Here is the complete patch with those logs : [Debug_weather_service.patch.txt](https://github.com/InfiniTimeOrg/InfiniTime/files/12568486/Debug_weather_service.patch.txt)

I ran the following tests using Gadgetbridge 0.75.1 and Tiny Weather Forecast Germany 0.61.0.

Shortly after connecting the watch, 7 events are received and 296B were allocated : 

```
<info> app: -------------------------------
<info> app: [Weather Service] Received weather command : 
        Timestamp : 1694357103
        Expiration(h) : 6
        Type : Location
<info> app: [Weather Service] Memory : free : 13376 - min free : 13080, malloc errors : 0 - stack overflow errors : 0
<info> app: [FreeRTOS HEAP] ALLOC 64B -> 536896376
<info> app: [FreeRTOS HEAP] ALLOC 16B -> 536896440
<info> app: [Weather Service] AddEventToTimeline - Current size: 1 - Max size: 536870911
<info> app: -------------------------------
<info> app: [Weather Service] Received weather command : 
        Timestamp : 1694357103
        Expiration(h) : 6
        Type : Humidity
<info> app: [Weather Service] Memory : free : 13296 - min free : 13080, malloc errors : 0 - stack overflow errors : 0
<info> app: [FreeRTOS HEAP] ALLOC 32B -> 536896456
<info> app: [FreeRTOS HEAP] ALLOC 16B -> 536896488
<info> app: [FreeRTOS HEAP] Free 536896440
<info> app: -------------------------------
<info> app: [Weather Service] Received weather command : 
        Timestamp : 1694357103
        Expiration(h) : 6
        Type : Temperature
<info> app: [Weather Service] Memory : free : 13264 - min free : 13080, malloc errors : 0 - stack overflow errors : 0
<info> app: [FreeRTOS HEAP] ALLOC 32B -> 536896504
<info> app: [FreeRTOS HEAP] ALLOC 24B -> 536896536
<info> app: [FreeRTOS HEAP] Free 536896488
<info> app: -------------------------------
<info> app: [Weather Service] Received weather command : 
        Timestamp : 1694357103
        Expiration(h) : 24
        Type : Temperature
<info> app: [Weather Service] Memory : free : 13224 - min free : 13080, malloc errors : 0 - stack overflow errors : 0
<info> app: [FreeRTOS HEAP] ALLOC 32B -> 536896560
<info> app: [Weather Service] AddEventToTimeline - Current size: 4 - Max size: 536870911
<info> app: -------------------------------
<info> app: [Weather Service] Received weather command : 
        Timestamp : 1694357103
        Expiration(h) : 6
        Type : Wind
<info> app: [Weather Service] Memory : free : 13192 - min free : 13080, malloc errors : 0 - stack overflow errors : 0
<info> app: [FreeRTOS HEAP] ALLOC 32B -> 536896592
<info> app: [FreeRTOS HEAP] ALLOC 40B -> 536896664
<info> app: [FreeRTOS HEAP] Free 536896536
<info> app: -------------------------------
<info> app: [Weather Service] Received weather command : 
        Timestamp : 1694357103
        Expiration(h) : 6
        Type : Precipitation
<info> app: [Weather Service] Memory : free : 13136 - min free : 13080, malloc errors : 0 - stack overflow errors : 0
<info> app: [FreeRTOS HEAP] ALLOC 32B -> 536896704
<info> app: [Weather Service] AddEventToTimeline - Current size: 6 - Max size: 536870911
<info> app: -------------------------------
<info> app: [Weather Service] Received weather command : 
        Timestamp : 1694357103
        Expiration(h) : 6
        Type : Clouds
<info> app: [Weather Service] Memory : free : 13104 - min free : 13080, malloc errors : 0 - stack overflow errors : 0
<info> app: [FreeRTOS HEAP] ALLOC 32B -> 536896736
<info> app: [Weather Service] AddEventToTimeline - Current size: 7 - Max size: 536870911
```

Then I hit the refresh button in TWFG: 13 events & 552B

```
<info> app: -------------------------------
<info> app: [Weather Service] Received weather command : 
        Timestamp : 1694357117
        Expiration(h) : 6
        Type : Location
<info> app: [Weather Service] Memory : free : 13072 - min free : 13072, malloc errors : 0 - stack overflow errors : 0
<info> app: [FreeRTOS HEAP] ALLOC 64B -> 536896768
<info> app: [Weather Service] AddEventToTimeline - Current size: 8 - Max size: 536870911
<info> app: -------------------------------
<info> app: [Weather Service] Received weather command : 
        Timestamp : 1694357117
        Expiration(h) : 6
        Type : Humidity
<info> app: [Weather Service] Memory : free : 13008 - min free : 13008, malloc errors : 0 - stack overflow errors : 0
<info> app: [FreeRTOS HEAP] ALLOC 32B -> 536896832
<info> app: [FreeRTOS HEAP] ALLOC 72B -> 536896864
<info> app: [FreeRTOS HEAP] Free 536896664
<info> app: -------------------------------
<info> app: [Weather Service] Received weather command : 
        Timestamp : 1694357117
        Expiration(h) : 6
        Type : Temperature
<info> app: [Weather Service] Memory : free : 12944 - min free : 12904, malloc errors : 0 - stack overflow errors : 0
<info> app: [FreeRTOS HEAP] ALLOC 32B -> 536896664
<info> app: [Weather Service] AddEventToTimeline - Current size: 10 - Max size: 536870911
<info> app: -------------------------------
<info> app: [Weather Service] Received weather command : 
        Timestamp : 1694357117
        Expiration(h) : 24
        Type : Temperature
<info> app: [Weather Service] Memory : free : 12904 - min free : 12904, malloc errors : 0 - stack overflow errors : 0
<info> app: [FreeRTOS HEAP] ALLOC 32B -> 536896936
<info> app: [Weather Service] AddEventToTimeline - Current size: 11 - Max size: 536870911
<info> app: -------------------------------
<info> app: [Weather Service] Received weather command : 
        Timestamp : 1694357117
        Expiration(h) : 6
        Type : Wind
<info> app: [Weather Service] Memory : free : 12872 - min free : 12872, malloc errors : 0 - stack overflow errors : 0
<info> app: [FreeRTOS HEAP] ALLOC 32B -> 536896968
<info> app: [Weather Service] AddEventToTimeline - Current size: 12 - Max size: 536870911
<info> app: -------------------------------
<info> app: [Weather Service] Received weather command : 
        Timestamp : 1694357117
        Expiration(h) : 6
        Type : Precipitation
<info> app: [Weather Service] Memory : free : 12840 - min free : 12840, malloc errors : 0 - stack overflow errors : 0
<info> app: [FreeRTOS HEAP] ALLOC 32B -> 536897000
<info> app: [Weather Service] AddEventToTimeline - Current size: 13 - Max size: 536870911
<info> app: -------------------------------
<info> app: [Weather Service] Received weather command : 
        Timestamp : 1694357117
        Expiration(h) : 6
        Type : Clouds
<info> app: [Weather Service] Memory : free : 12808 - min free : 12808, malloc errors : 0 - stack overflow errors : 0
<info> app: [FreeRTOS HEAP] ALLOC 32B -> 536897032
<info> app: [Weather Service] AddEventToTimeline - Current size: 14 - Max size: 536870911
<info> app: -------------------------------
<info> app: [Weather Service] Received weather command : 
        Timestamp : 1694357118
        Expiration(h) : 6
        Type : Location
<info> app: [Weather Service] Memory : free : 12776 - min free : 12776, malloc errors : 0 - stack overflow errors : 0
<info> app: [FreeRTOS HEAP] ALLOC 64B -> 536897064
<info> app: [Weather Service] AddEventToTimeline - Current size: 15 - Max size: 536870911
<info> app: -------------------------------
<info> app: [Weather Service] Received weather command : 
        Timestamp : 1694357118
        Expiration(h) : 6
        Type : Humidity
<info> app: [Weather Service] Memory : free : 12712 - min free : 12712, malloc errors : 0 - stack overflow errors : 0
<info> app: [FreeRTOS HEAP] ALLOC 32B -> 536897128
<info> app: [Weather Service] AddEventToTimeline - Current size: 16 - Max size: 536870911
<info> app: -------------------------------
<info> app: [Weather Service] Received weather command : 
        Timestamp : 1694357118
        Expiration(h) : 6
        Type : Temperature
<info> app: [Weather Service] Memory : free : 12680 - min free : 12680, malloc errors : 0 - stack overflow errors : 0
<info> app: [FreeRTOS HEAP] ALLOC 32B -> 536897160
<info> app: [FreeRTOS HEAP] ALLOC 136B -> 536897192
<info> app: [FreeRTOS HEAP] Free 536896864
<info> app: -------------------------------
<info> app: [Weather Service] Received weather command : 
        Timestamp : 1694357118
        Expiration(h) : 24
        Type : Temperature
<info> app: [Weather Service] Memory : free : 12584 - min free : 12512, malloc errors : 0 - stack overflow errors : 0
<info> app: [FreeRTOS HEAP] ALLOC 32B -> 536896864
<info> app: [Weather Service] AddEventToTimeline - Current size: 18 - Max size: 536870911
<info> app: -------------------------------
<info> app: [Weather Service] Received weather command : 
        Timestamp : 1694357118
        Expiration(h) : 6
        Type : Wind
<info> app: [Weather Service] Memory : free : 12552 - min free : 12512, malloc errors : 0 - stack overflow errors : 0
<info> app: [FreeRTOS HEAP] ALLOC 32B -> 536896896
<info> app: [Weather Service] AddEventToTimeline - Current size: 19 - Max size: 536870911
<info> app: -------------------------------
<info> app: [Weather Service] Received weather command : 
        Timestamp : 1694357118
        Expiration(h) : 6
        Type : Precipitation
<info> app: [Weather Service] Memory : free : 12512 - min free : 12512, malloc errors : 0 - stack overflow errors : 0
<info> app: [FreeRTOS HEAP] ALLOC 32B -> 536897328
<info> app: [Weather Service] AddEventToTimeline - Current size: 20 - Max size: 536870911
<info> app: -------------------------------
<info> app: [Weather Service] Received weather command : 
        Timestamp : 1694357118
        Expiration(h) : 6
        Type : Clouds
<info> app: [Weather Service] Memory : free : 12480 - min free : 12480, malloc errors : 0 - stack overflow errors : 0
<info> app: [FreeRTOS HEAP] ALLOC 32B -> 536897360
<info> app: [Weather Service] AddEventToTimeline - Current size: 21 - Max size: 536870911
```

With those test, I can confirm that
 - TWFG & Gadgetbridge send a lot of events 
 - The amount of memory allocated to store those event is probably not the most efficient (but it provides more flexibility so... it's just a comprise).

But the following points struck me first:
 - The timeline is virtually infinite : it currently accepts up to 536870911 (`std::vector::max_size()`) elements. There's no way the PineTime can store that much events
 - The weather service stores and keeps in memory multiple "versions" of the each type of events
 - There are barely no memory that is freed : events are removed from the timeline only when they are expired.

With those observations in mind, I think we do a few easy things to reduce and keep under control the memory usage of the weather service:
 - Add to the timeline only the events that will actually be used by the current version of InfiniTime, and ignore everything else. If I'm not wrong, we currently process only the temperature, cloud and precipitations.
 - Remove older version of an event when a new event of the same type/expiration is received. We probably don't need to keep more than 1 or 2 versions of the events (short term, long term). 
 - Improve `WeatherService::GetCurrentTemperature()`, `WeatherService::GetCurrentCloud()` so they select *the best* version of the element to return (based on the timestamp and expiry).
 - Set a saner maximum size of the timeline. Since we need only 3 different event types and assuming we keep 2 versions of them, the maximum size could be set to 6.

This should fix the issue on the short term : the memory usage will be reduced by a lot and should not grow uncontrollably. And this will give us more time to improve the current implementation to reduce the number of allocations, the storage overhead, and ensure that the memory usage stays under control when we add new event types.

What do you think of those suggestions?

JF002:
Thanks for your work, @faxe1008. However, I decided to rework the weather service in #1924 so we can fix those issues more easily. So I'm closing this one in favor of #1924.

Issue #1246, opened 2022-07-25
Heart Rate UI Update Concept


RageGamerBoi:
### Verification

- [X] I searched for similar feature request and found none was relevant.

### Pitch us your idea!

Replace the heart rate start button with the new timer UI button.

### Description

Update the heart rate app from this:

![image](https://user-images.githubusercontent.com/105567407/180809065-a8b798e5-39e7-46e8-b7fb-61f9d2a84d6c.png)

To something like this:

![image](https://user-images.githubusercontent.com/105567407/180808888-9e6a69e4-4a39-458e-89e2-befcc48a4001.png)

mashuptwice:
It would definitely be more consistent with the timers button design, but not with the stepcounter. 

I'm not sure if a bigger button makes sense, because it would be easier to accidentally disable the HRM when keeping the app open.

Apart from that IMHO it would look nicer.

ITCactus:
there are ["UI design guidelines"](https://github.com/InfiniTimeOrg/InfiniTime/blob/develop/doc/ui_guidelines.md),
and according to that, the "start" button should be full width indeed

RageGamerBoi:
> * Make interactable objects big

and 

> * Align objects all the way to the edge or corner

Should cover the wide buttons.

As for the other question, the Start/Stop button already enables/disables the HRM through the entire OS, none of the functionality needs to be different here.

RageGamerBoi:
As for the stepcounter concern;

> It would definitely be more consistent with the timers button design, but not with the stepcounter.

Check issue #1254 that I just made :)

Idcrafter:
also would a moving graph make it look nicer while also serving as visual feedback for it being active

PR #1828, opened 2023-08-14
Heart Rate: Updated UI

judgie97:
Updated start button in Heart Rate app to match timer app. Fixes Issue #1246


minacode:
Since multiple buttons should follow this design, is it possible to define the style globally and just import and apply it here? Maybe this saves some bytes.

judgie97:
Sounds good. Still trying to work my way through the code to see how everything works but I will have a look.

PR #1838, opened 2023-08-20
Move Navigation font to external memory

JF002:
The TTF font used by the navigation app is ~20KB and is stored in internal flash memory. To free this space, the TTF font is now converted in 2 "atlas pictures" (pictures that contain multiple concatenated images) stored in the external flash memory. The navigation app now accesses one of those 2 files and apply an offset to display the desired picture.

The corresponding documentation has also been updated.

The final binary file is now 19380B lighter. 

Here is a quick video I made while testing the implementation : 

https://github.com/InfiniTimeOrg/InfiniTime/assets/2261652/da878cb6-520a-4083-bf45-169f7015a17f


src/displayapp/screens/Navigation.cpp
```diff
@@ -23,105 +23,123 @@
...
+  };
+
+  constexpr uint16_t iconHeight = -80;
+  constexpr uint8_t flagIndex = 18;
+  constexpr uint8_t maxIconsPerFile = 25;
+  const char* iconsFile0 = "F:/images/navigation0.bin";
+  const char* iconsFile1 = "F:/images/navigation1.bin";
+
+  constexpr std::array<std::pair<const char*, uint8_t>, 86> iconMap = {{
+    {"arrive-left", 1},
```
FintasticMan, line 39: It would be good if we could get rid of these magic numbers. Either by sorting by the index, or using maths to calculate the index (not sure if that's possible).

FintasticMan (changes requested):
I like this! It doesn't look to me like you're checking that the image files exist in the filesystem before using them. That would be good to add so that if people haven't updated their resources yet, it doesn't cause issues. Other than that, this code looks good to me.

doc/buildAndProgram.md
```diff
@@ -48,7 +48,7 @@ CMake configures the project according to variables you specify the command line
 #### (\*) Note about **CMAKE_BUILD_TYPE**
 By default, this variable is set to *Release*. It compiles the code with size and speed optimizations. We use this value for all the binaries we publish when we [release](https://github.com/InfiniTimeOrg/InfiniTime/releases) new versions of InfiniTime.
 
-The *Debug* mode disables all optimizations, which makes the code easier to debug. However, the binary size will likely be too big to fit in the internal flash memory. If you want to build and debug a *Debug* binary, you'll need to disable some parts of the code. For example, the icons for the **Navigation** app use a lot of memory space. You can comment the content of `m_iconMap` in the [Navigation](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/src/displayapp/screens/Navigation.h#L148) application to free some memory.
```
mark9064, line 51: Shouldn't the example of the navigation app be removed or changed? Since icons no longer reside in the main binary, removing them won't free up binary space?

If this example is being kept, the link URL needs to be updated

mark9064 (commented):
Builds with -DBUILD_RESOURCES=1 fail for me with this PR. Could totally be a problem with my dependency setup if it works for you.

```
/home/laptop/Local/pinetime/InfiniTime/node_modules/lv_img_conv/lib/image-q/utils/point.ts:85
    this.r = point.r;
                   ^
TypeError: Cannot read properties of undefined (reading 'r')
    at Point.from (/home/laptop/Local/pinetime/InfiniTime/node_modules/lv_img_conv/lib/image-q/utils/point.ts:85:20)
    at ErrorDiffusionArray.<anonymous> (/home/laptop/Local/pinetime/InfiniTime/node_modules/lv_img_conv/lib/image-q/image/array.ts:128:15)
    at step (/home/laptop/Local/pinetime/InfiniTime/node_modules/lv_img_conv/lib/image-q/image/array.ts:205:23)
    at Object.next (/home/laptop/Local/pinetime/InfiniTime/node_modules/lv_img_conv/lib/image-q/image/array.ts:146:20)
    at Immediate.next (/home/laptop/Local/pinetime/InfiniTime/node_modules/lv_img_conv/lib/image-q/basicAPI.ts:163:33)
    at processImmediate (node:internal/timers:476:21)
Traceback (most recent call last):
  File "/home/laptop/Local/pinetime/InfiniTime/src/resources/generate-img.py", line 56, in <module>
    main()
  File "/home/laptop/Local/pinetime/InfiniTime/src/resources/generate-img.py", line 51, in main
    subprocess.check_call(line)
  File "/usr/lib/python3.11/subprocess.py", line 413, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['/home/laptop/Local/pinetime/InfiniTime/node_modules/.bin/lv_img_conv', '/home/laptop/Local/pinetime/InfiniTime/src/resources/images/navigation0.png', '--force', '--output-file', 'navigation0.bin', '--color-format', 'CF_INDEXED_1_BIT', '--output-format', 'bin', '--binary-format', 'ARGB8565_RBSWAP']' returned non-zero exit status 1.
```
Using `"lv_font_conv": "^1.5.2","lv_img_conv": "^0.3.0"` (npm chose the versions for me)

src/displayapp/screens/Navigation.cpp
JF002: >  It doesn't look to me like you're checking that the image files exist in the filesystem before using them

Good point, I definitely need to add this!

> It would be good if we could get rid of these magic numbers

Those indexes heavily depends on how the picture was generated (the index correspond to the position of the icon in the whole picture, knowing that each icon has a height of 80px). I don't have any ideas to improve this, except maybe add a comment to explain this.

doc/buildAndProgram.md
JF002: Yes, you are right, removing `iconMap` won't free much memory now. I'll remove the example.

src/displayapp/screens/Navigation.cpp
FintasticMan: Yeah, it doesn't seem like much can be done about it other than adding a comment.

JF002: I've just added a comment about the layout of the pictures and the indexing of the icons.

About checking for the file availability : LVGL will gracefully handle the case where the file is not available (no error, no crash, simply displays nothing). So the application will work but won't display the icon.
I'll however work on something similar than what we did for the watchfaces so that user won't be able to run the app if the files are not available. Since this change will impact other parts of the firmware, I'll do this in another PR in the next few days :)

doc/buildAndProgram.md
JF002: I've just edited that part of the doc and provided an alternative way to build (parts of) the firmware in debug mode.

FintasticMan (approved):
LGTM now! Shall we merge the PR that prevents starting the navigation app if the files aren't present after this one?

doc/buildAndProgram.md
mark9064: Looks great :+1:

JF002:
@FintasticMan 
I've just pushed #1847 which disables the Navigation app if the resource files are not available. We'll be able to use this implementation for other applications when needed as well.

If you're OK with this, we can merge #1847 after we merge this one.

JF002:
> Builds with -DBUILD_RESOURCES=1 fail for me with this PR. Could totally be a problem with my dependency setup if it works for you.
> [...]
> Using `"lv_font_conv": "^1.5.2","lv_img_conv": "^0.3.0"` (npm chose the versions for me)

This is probably not related to this PR since it doesn't to any change on how the images and fonts are converted. Does it work fine on the `main` branch ?

These are the versions I'm using on my dev setup:

```
$ lv_font_conv -v
1.5.2
$lv_img_conv --version
0.3.0
```

mark9064:
> > Builds with -DBUILD_RESOURCES=1 fail for me with this PR. Could totally be a problem with my dependency setup if it works for you.
> > [...]
> > Using `"lv_font_conv": "^1.5.2","lv_img_conv": "^0.3.0"` (npm chose the versions for me)
> 
> This is probably not related to this PR since it doesn't to any change on how the images and fonts are converted. Does it work fine on the `main` branch ?
> 
> These are the versions I'm using on my dev setup:
> 
> ```
> $ lv_font_conv -v
> 1.5.2
> $lv_img_conv --version
> 0.3.0
> ```
- Built fine before this commit
- I've tried nodejs 16,18,20 with the dependency versions you use, no dice
- Creates the pine_logo image without problem, only breaks on the navigation atlases
- Broken on both my laptop and desktop (though they run pretty much the same software)

But it works on your machine and CI, so I have no clue. I've basically never used node before so I guess I will have to find out how to debug it... strange behaviour all round

FintasticMan:
Maybe try using the node lockfile from #1764.

mark9064:
Good idea, I gave this a go. After installing from the package(-lock), the build failed with error message asking for swc/core, which I then installed. Unfortunately the original error then returned :(

I've been digging through lv_img_conv and it looks like it's trying to create a palette with no colours in it. Still working on figuring out the root problem

mark9064:
> I experienced this issue in #1819, try pinning `@swc/core` to 1.2.160 (see my changes on that PR). It probably needs adding to the Dockerfile (and anywhere else we install `@swc/core`).

Spot on, this resolved the issue. Cheers!

JF002:
> Spot on, this resolved the issue. Cheers!

Great! I honestly don't know why it doesn't work with more recent version of swc (in fact, I don't know what it is... I don't know Node either, but that's what the LVGL tools are based on). 
The only think I can think of is that it's the 1st time we use `CF_INDEXED_1_BIT` as output format. Maybe there was a breaking change in lv_img_conv in recent versions?

JF002:
> I experienced this issue in #1819, try pinning `@swc/core` to 1.2.160 (see my changes on that PR). It probably needs adding to the Dockerfile (and anywhere else we install `@swc/core`).

If I understand correctly, this change is needed to ensure that the correct version will be installed? If so, I guess we would appreciate a PR for this :)

Issue #1262, opened 2022-08-02
Dynamic apps and custom watchfaces


JF002:
### Verification

- [X] I searched for similar feature request and found none was relevant.

### Pitch us your idea!

I want to install new watchfaces and apps without flashing the whole firmware

### Description

InfiniTime is currently a monolithic firmware : everything is built into a single binary file : OS, drivers, applications & watchfaces, fonts, pictures,...

This feature request is about adding the possibility to create and install new watchfaces and app without having to rebuild and flash the whole firmware.

InfiniTime is not quite ready yet to support these features, but we are slowly but surely moving forward to that: 
 - The PineTime is equipped with a huge 4MB SPI flash memory
 - We've already implemented a [filesystem](https://github.com/InfiniTimeOrg/InfiniTime/pull/438) (based on [LittleFS](https://github.com/littlefs-project/littlefs) on that SPI flash memory 
 - We've also implemented a [BLE FS API](https://github.com/InfiniTimeOrg/InfiniTime/pull/756) to allow companion apps to access to this FS via BLE
 - We are working on [moving assets [fonts and picture](https://github.com/InfiniTimeOrg/InfiniTime/issues/321) to that external memory

There's of course still a lot of researches, experiments and work to do before we can enable dynamic app and custom watchfaces.

Over the time, a few feature requests on this topic were opened, and I create this one to group them all. 
Those existing feature requests are : 
 - https://github.com/InfiniTimeOrg/InfiniTime/issues/113
 - https://github.com/InfiniTimeOrg/InfiniTime/issues/421
 - https://github.com/InfiniTimeOrg/InfiniTime/issues/755
 - https://github.com/InfiniTimeOrg/InfiniTime/issues/1055

Arsen6331:
I had an idea for this a while ago. It may be possible to use WebAssembly for this. Searching online, I see that there are a few projects that have successfully run WASM on microcontrollers. It will be slower than native, but if it can be done, should be an improvement over something like MicroPython. My concern would be memory usage, but I don't know just how much memory it uses.

There seems to be an example for ESP-IDF here, which uses FreeRTOS: https://github.com/bytecodealliance/wasm-micro-runtime/tree/main/product-mini/platforms/esp-idf. I'm not sure how applicable it would be to InfiniTime, but it might be useful.

Obviously, native code would be preferable, but I'm just not sure how to accomplish that without having to recompile every time you add something.

Tiggilyboo:
[...]

PR #1840, opened 2023-08-22
App controller

minacode:
This PR stems from #1571 and is related to #1262.
It is a draft, because there is work left to do (see the open problems below).

This PR implements an `AppController` class as a single place in the code where all non-core apps are registered.
Short-term this allows us to implement optional apps (see #1408).
Long-term this allows us to implement dynamically loadable apps.

Currently the app `Twos` is handled by the `AppController` as an example, mostly because it has no dependencies and is therefore simple to handle.

The implementation uses the `Apps` enum (see also #1760), but removes all entries of apps that will be handled by the `AppController`. Instead, the entry `Dynamic` is added to the enum.
The ids of dynamic apps are then calculated by `Dynamic + <offset>` where `<offset>` is the id of an app inside the `AppController` (e.g. an array index). This leads to a lot of `static_cast` everywhere.

All places, where the current apps where hardcoded before have to use the AppController now.

Apps that can be used with the `AppController` must implement a `Get` method and its symbol like this:
```cpp
...
public:
  static std::unique_ptr<Twos> Get() {
    return std::make_unique<Twos>();
  }

  static constexpr const char* Symbol {"2"};
```
The `AppController` saves those contructors internally and can reference them via the `Dynamic + <offset>` ids described above.

There are also already fragments of an `AppInterface` in the code that every `Get` method must receive. The `AppInterface` bundles all the controllers that apps can require.

The current version runs with InfiniSim if you use [my branch](https://github.com/minacode/InfiniSim/tree/app-controller-3).

Feedback and ideas would be very apprechiated :blush:

###### Open Problems:
- [ ] The `AppController` should be a component and live in the respective directory.
- [ ] How are we handling the core apps? Are we registering them in the `AppController` as well or are we hardcoding them (for now)?
- [ ] Implement the `AppInterface`.
- [ ] Move all/more apps to the AppController.
- [ ] Currently, there are a lot of `std::vector` in the code. Is it better to swap them for `std::array`? This is a tradeoff between dynamically allocating a collection when we know how many apps we have vs. hardcoding an array with reasonable bounds. Using only vectors seems to overflow the stack.


JF002:
@minacode Thanks for your work on this topic! Since you created this PR, we merged #1894 which, I think, does mostly the same functionality at build time. Is this OK for you if we close this PR in favor of #1894 ?

minacode:
Yes, absolutely :)

PR #1841, opened 2023-08-25
Music: Replace placeholder text

Commenter25:
When starting the Music app, you can see a flash of the placeholder developer text that appears.

[A video showing entering and exiting the music app, and the flash of placeholder text when entering](https://github.com/InfiniTimeOrg/InfiniTime/assets/35016761/a3ccc4bc-8164-47f2-b5bf-10dbd0e8d07e)

In some cases, the placeholder can actually stay. It appears to be when a companion app is connected but no music data is available.

This simply changes the placeholder text to the waiting text, to prevent that extra flash changing from the placeholder to the waiting state.

In the case where the placeholder stays, I also found that it will change to `00:00/00:00`, rather than `--:--/--:--`. This is a similar unnecessary flash of text, so when the total length is zero, it will show dashes instead.

### DIFF
--- a/src/displayapp/screens/Music.cpp
+++ b/src/displayapp/screens/Music.cpp
@@ -117,7 +117,7 @@ Music::Music(Pinetime::Controllers::MusicService& music) : musicService(music) {
   lv_obj_align(txtArtist, nullptr, LV_ALIGN_IN_LEFT_MID, 12, MIDDLE_OFFSET + 1 * FONT_HEIGHT);
   lv_label_set_align(txtArtist, LV_ALIGN_IN_LEFT_MID);
   lv_obj_set_width(txtArtist, LV_HOR_RES - 12);
-  lv_label_set_text_static(txtArtist, "Artist Name");
+  lv_label_set_text_static(txtArtist, "Waiting for");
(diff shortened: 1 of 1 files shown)


FintasticMan:
It would be good to get rid of the placeholder text, but the way this is done means that the `"Waiting for"` and `"track information.."` strings are duplicated. I'd rather avoid that.

Commenter25:
Very understandable. I've moved the repeated strings into variables. The text was removed from `MusicService`. In my opinion, it makes more sense for the generalized service to simply return nothing when there is nothing, and have the UI handle that. Currently it appears to be only used for the music app anyway, but any new implementations might benefit, like a watch face that shows the current song.

Instead, if the total duration is zero, it will draw the waiting text. This way, it can't show a partial placeholder; and when a companion app sends empty data, the waiting text will reappear. Previously, it was possible for it to say "Waiting for [track name]", and blank data would show nothing until the app restarts.

I must admit, my C++ knowledge is limited (non-existent), but I've tried my hand anyway. Hopefully not too poorly, it at least works in my testing, but if I did something wrong lemme know.

PR #1847, opened 2023-09-02
Add is available to navigation app

JF002:
Add `IsAvailable()`  to Navigation App

Navigation app now (since #1838) needs 2 images to be loaded from the resources on the external filesystem. This PR adds an `enabled` field to the Applications struct. This field is true for all applications expect for Navigation which calls `Navigation::IsAvailable()`. This methods returns true if the 2 files are available in the resources. 

The application list disables the application (draws it in grey, disables the touch callback) if the enable flag is not set.

This implementation is similar than the one used by the Watchface Setting app.

Here is the result from the devkit and InfiniSim : 

InfiniSim_2023-09-02_161026
image


FintasticMan (approved):
Looks good!

PR #1850, opened 2023-09-04
dev: Adding shell.nix for development environments

jzbor:
This adds a shell.nix file to easily create a somewhat reproducible development environment with nix. It does require Nix to be installed, but may be a "lighter" alternative to Docker. Accompanying documentation is also provided.


blm768:
The errors seem to be non-deterministic. I'll try this without the package cache.

blm768:
This is downright delightful, at least for us Nix users! I can think of a couple of minor potential downsides, but I'm not sure they're huge ones:

- This is another (small) bit of code to maintain.
- The compiler version may be subtly different from the one used in CI. I'd say the risk of breakage is probably relatively low, though. (?)

On the upside, Nix evaluations can be quite cacheable (especially using flakes); we could build on this to cache dependency installations between CI runs.

FintasticMan:
I wouldn't worry much about the compiler version; InfiniTime works with most versions. What I do worry about is the nRF SDK version. I can't see that this specifies a version to use, and InfiniTime definitely requires the specific version we mention in the docs. Is there a way to specify that?

jzbor:
Yes that should be possible. What Version does InfiniTime require?

FintasticMan:
InfiniTime requires 15.3.0.

jzbor:
I have now pinned the nrf-sdk version to 15.3.0.

It should also be mentioned that it is possible to pin nixpkgs altogether, making the environment completely reproducible. However this requires the use of [Flakes](https://serokell.io/blog/practical-nix-flakes), which are not officially stable yet (although widely used). It also comes with the maintenance burden of keeping the Flake up-to-date, so I think a simple `shell.nix` is still the best fit here.

jzbor:
Are there any updates? Is there still an interest in merging this?

blm768:
I'm certainly interested in using this as a NixOS user. With the errors I'm getting from the Docker container, this is currently my only way to build the project on NixOS.

icewind1991:
I've created a flake for building infinitime to make it easy to create patched builds on nixos: https://github.com/icewind1991/infinitime-builder

PR #1855, opened 2023-09-07
fonts: add missing cyrillic chars from ukrainian alphabet

gagara:
Add missing chars from Ukrainian alphabet.
In particular: 
```
Іі Єє Її Ґґ
```
Tested on PineTime.


FintasticMan:
I'm afraid that we aren't currently interested in adding characters other than alphanumerics and symbols to the system fonts. This is because fonts take up a lot of flash and memory, and we don't want to pick and choose which characters to allow and which to not.

FintasticMan:
Ah, I was unaware that there were any Cyrillic characters... I wonder why those are there. It might indeed be good to get rid of them.

PR #1856, opened 2023-09-07
WatchFaceDigital: Add weather display

vkareh:
If weather is available, display the cloud icon and temperature.

Screenshot at 2023-09-07 16-37-56

### DIFF
--- a/src/displayapp/screens/WatchFaceDigital.cpp
+++ b/src/displayapp/screens/WatchFaceDigital.cpp
@@ -4,11 +4,13 @@
 #include <cstdio>
 #include "displayapp/screens/NotificationIcon.h"
 #include "displayapp/screens/Symbols.h"
+#include "displayapp/screens/WeatherSymbols.h"
 #include "components/battery/BatteryController.h"
 #include "components/ble/BleController.h"
 #include "components/ble/NotificationManager.h"
 #include "components/heartrate/HeartRateController.h"
 #include "components/motion/MotionController.h"
+#include "components/ble/SimpleWeatherService.h"
 #include "components/settings/Settings.h"
 
 using namespace Pinetime::Applications::Screens;
@@ -19,13 +21,15 @@ WatchFaceDigital::WatchFaceDigital(Controllers::DateTime& dateTimeController,
                                    Controllers::NotificationManager& notificationManager,
                                    Controllers::Settings& settingsController,
                                    Controllers::HeartRateController& heartRateController,
-                                   Controllers::MotionController& motionController)
+                                   Controllers::MotionController& motionController,
+                                   Controllers::SimpleWeatherService& weatherService)
   : currentDateTime {{}},
     dateTimeController {dateTimeController},
     notificationManager {notificationManager},
     settingsController {settingsController},
     heartRateController {heartRateController},
     motionController {motionController},
+    weatherService {weatherService},
     statusIcons(batteryController, bleController) {
 
   statusIcons.Create();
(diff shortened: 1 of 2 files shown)


vkareh:
Rebased against `main` and adapted to work with the SimpleWeatherService.

src/displayapp/screens/WatchFaceDigital.cpp
```diff
@@ -153,4 +168,26 @@ void WatchFaceDigital::Refresh() {
...
+
+  if (currentWeather.IsUpdated()) {
+    auto optCurrentWeather = currentWeather.Get();
+    if (optCurrentWeather) {
+      int16_t temp = optCurrentWeather->temperature;
+      if (settingsController.GetWeatherFormat() == Controllers::Settings::WeatherFormat::Imperial) {
+        temp = Controllers::SimpleWeatherService::CelsiusToFahrenheit(temp);
+      }
+      temp = temp / 100 + (temp % 100 >= 50 ? 1 : 0);
+      lv_label_set_text_fmt(temperature, "%d°", temp);
```
FintasticMan, line 182: Given that there is enough space on the screen on the digital watch face, maybe we want to also display the unit? I.e, 17°C rather than just 17°. Just a suggestion, not sure if it's actually better.

FintasticMan (approved):
I like it!

src/displayapp/screens/WatchFaceDigital.cpp
vkareh: I like the idea! Done.

src/displayapp/screens/WatchFaceDigital.cpp
```diff
@@ -153,4 +169,28 @@ void WatchFaceDigital::Refresh() {
...
+        temp = Controllers::SimpleWeatherService::CelsiusToFahrenheit(temp);
+        tempUnit = 'F';
+      }
+      temp = temp / 100 + (temp % 100 >= 50 ? 1 : 0);
+      lv_label_set_text_fmt(temperature, "%d°%c", temp, tempUnit);
+      lv_label_set_text(weatherIcon, Symbols::GetSymbol(optCurrentWeather->iconId));
+      lv_obj_realign(temperature);
+      lv_obj_realign(weatherIcon);
+    }
+  } else {
```
NeroBurner, line 190: Is the else branch one too late. For me it reads: if the current weather isn't updated then set it to empty string. I think this would only be needed of the optional has no value

src/displayapp/screens/WatchFaceDigital.h
```diff
@@ -85,7 +92,8 @@ namespace Pinetime {
                                              controllers.notificationManager,
                                              controllers.settingsController,
                                              controllers.heartRateController,
-                                             controllers.motionController);
+                                             controllers.motionController,
+                                             *controllers.weatherController);
```
NeroBurner, line 96: Why is the weathercontroller the only one that gets dereferenced? Just curious as I don't know the weather code quite well

src/displayapp/screens/WeatherSymbols.h
```diff
@@ -6,7 +6,7 @@ namespace Pinetime {
   namespace Applications {
     namespace Screens {
       namespace Symbols {
-        const char* GetSymbol(const Pinetime::Controllers::SimpleWeatherService::Icons icon) {
+        static const char* GetSymbol(const Pinetime::Controllers::SimpleWeatherService::Icons icon) {
```
NeroBurner, line 9: Curious: what's the reasoning behind using a static variable in a namespace? Wouldn't this just mean that it is instanced into every compilation unit?

vkareh: Right, without it you end up re-defining it and the linker throws a `multiple definition` error.

src/displayapp/screens/WatchFaceDigital.cpp
vkareh: `weatherService.Current()` returns a reference to the current weather instance only if the weather is "recent enough" (quick look seems to show 24 hours). Otherwise it returns an empty value.

In either case it considers it "updated", so I think you're right. The only case when the `else` will get called is if weather data has _never_ been sent since the init of the service.

Therefore it makes sense to move the `else` clause inside the `IsUpdated` conditional check and clear the values when `currentWeather` is empty. I'll test this out and push the change.

src/displayapp/screens/WeatherSymbols.h
FintasticMan: Yeah, this should be marked as static. It means that the function can be called without being tied to an object. This is different from declaring static linkage, which this function won't have.

FintasticMan: The best way to solve the issue of multiple definitions would probably be to declare the function in the .h file and create a .cpp file that implements the function.

src/displayapp/screens/WatchFaceDigital.cpp
vkareh: It seems to work, so I've pushed the change. I also added a new commit to fix the PineTimeStyle face (which has the same issue).

src/displayapp/screens/WeatherSymbols.h
vkareh: This is solved by #1969.

src/displayapp/screens/WatchFaceDigital.cpp
FintasticMan: Fixed by #1965.

NeroBurner (approved):
LGTM!

src/displayapp/screens/WatchFaceDigital.h
FintasticMan: The reason this is like this is because the BLE components are initialised later by the NimbleController, so they can't be stored as references in the AppControllers struct created by DisplayApp. It would probably be good to refactor this, as it might clean up some things, but that's definitely outside the scope of this PR.

PR #1863, opened 2023-09-21
lv_img_conv_py: minimal python port of node module

NeroBurner:
Create a minimal python port of the node.js module `lv_img_conv`. Only
the currently in use color formats `CF_INDEXED_1_BIT` and
`CF_TRUE_COLOR_ALPHA` are implemented.

Output only as binary with format `ARGB8565_RBSWAP` (and `ARGB8888` for debugging).

This is enough to create the `resources-1.13.0.zip`.

In the simulator the pinecone in Casio is visible, and the navigation symbol is visible as well in the simulator when the resource zip is loaded

The new `lv_img_conv.py` script needs `python3-pil` python module as runtime dependency.
The runtime dependency is added to the docker container image.

InfiniSim_2023-09-21_230253_casio
InfiniSim_2023-09-21_230304_nav

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,6 +31,10 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
+      - name: Install resource build dependencies
+        run:  |
+          apt-get update
+          apt-get -y install --no-install-recommends python3-pil
       - name: Build
         shell: bash
         run: /opt/build.sh all
--- a/doc/buildAndProgram.md
+++ b/doc/buildAndProgram.md
@@ -42,7 +42,7 @@ CMake configures the project according to variables you specify the command line
 **NRF5_SDK_PATH**|path to the NRF52 SDK|`-DNRF5_SDK_PATH=/home/jf/nrf52/Pinetime/sdk`|
 **CMAKE_BUILD_TYPE (\*)**| Build type (Release or Debug). Release is applied by default if this variable is not specified.|`-DCMAKE_BUILD_TYPE=Debug`
 **BUILD_DFU (\*\*)**|Build DFU files while building (needs [adafruit-nrfutil](https://github.com/adafruit/Adafruit_nRF52_nrfutil)).|`-DBUILD_DFU=1`
-**BUILD_RESOURCES (\*\*)**| Generate external resource while building (needs [lv_font_conv](https://github.com/lvgl/lv_font_conv) and [lv_img_conv](https://github.com/lvgl/lv_img_conv). |`-DBUILD_RESOURCES=1`
+**BUILD_RESOURCES (\*\*)**| Generate external resource while building (needs [lv_font_conv](https://github.com/lvgl/lv_font_conv) and [python3-pil/pillow](https://pillow.readthedocs.io) module). |`-DBUILD_RESOURCES=1`
 **TARGET_DEVICE**|Target device, used for hardware configuration. Allowed: `PINETIME, MOY-TFK5, MOY-TIN5, MOY-TON5, MOY-UNK`|`-DTARGET_DEVICE=PINETIME` (Default)
 
 #### (\*) Note about **CMAKE_BUILD_TYPE**
--- a/src/resources/generate-img.py
+++ b/src/resources/generate-img.py
@@ -11,6 +11,9 @@ import subprocess
 
 def gen_lvconv_line(lv_img_conv: str, dest: str, color_format: str, output_format: str, binary_format: str, sources: str):
     args = [lv_img_conv, sources, '--force', '--output-file', dest, '--color-format', color_format, '--output-format', output_format, '--binary-format', binary_format]
+    if lv_img_conv.endswith(".py"):
+        # lv_img_conv is a python script, call with current python executable
+        args = [sys.executable] + args
 
     return args
 
--- a/src/resources/lv_img_conv.py
+++ b/src/resources/lv_img_conv.py
@@ -1,0 +1,193 @@
+#!/usr/bin/env python3
+import argparse
+import pathlib
+import sys
+import decimal
+from PIL import Image
+
+
+def classify_pixel(value, bits):
+    def round_half_up(v):
+        """python3 implements "propper" "banker's rounding" by rounding to the nearest
+        even number. Javascript rounds to the nearest integer.
+        To have the same output as the original JavaScript implementation add a custom
+        rounding function, which does "school" rounding (to the nearest integer).
+
+        see: https://stackoverflow.com/questions/43851273/how-to-round-float-0-5-up-to-1-0-while-still-rounding-0-45-to-0-0-as-the-usual
+        """
+        return int(decimal.Decimal(v).quantize(decimal.Decimal('1'), rounding=decimal.ROUND_HALF_UP))
+    tmp = 1 << (8 - bits)
+    val = round_half_up(value / tmp) * tmp
+    if val < 0:
+        val = 0
+    return val
+
+
+def test_classify_pixel():
+    # test difference between round() and round_half_up()
+    assert classify_pixel(18, 5) == 16
+    # school rounding 4.5 to 5, but banker's rounding 4.5 to 4
+    assert classify_pixel(18, 6) == 20
+
+
+def main():
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument("img",
+        help="Path to image to convert to C header file")
+    parser.add_argument("-o", "--output-file",
+        help="output file path (for single-image conversion)",
+        required=True)
+    parser.add_argument("-f", "--force",
+        help="allow overwriting the output file",
+        action="store_true")
+    parser.add_argument("-i", "--image-name",
+        help="name of image structure (not implemented)")
+    parser.add_argument("-c", "--color-format",
+        help="color format of image",
+        default="CF_TRUE_COLOR_ALPHA",
+        choices=[
+            "CF_ALPHA_1_BIT", "CF_ALPHA_2_BIT", "CF_ALPHA_4_BIT",
+            "CF_ALPHA_8_BIT", "CF_INDEXED_1_BIT", "CF_INDEXED_2_BIT", "CF_INDEXED_4_BIT",
+            "CF_INDEXED_8_BIT", "CF_RAW", "CF_RAW_CHROMA", "CF_RAW_ALPHA",
+            "CF_TRUE_COLOR", "CF_TRUE_COLOR_ALPHA", "CF_TRUE_COLOR_CHROMA", "CF_RGB565A8",
+        ],
+        required=True)
+    parser.add_argument("-t", "--output-format",
+        help="output format of image",
+        default="bin", # default in original is 'c'
+        choices=["c", "bin"])
+    parser.add_argument("--binary-format",
+        help="binary color format (needed if output-format is binary)",
+        default="ARGB8565_RBSWAP",
+        choices=["ARGB8332", "ARGB8565", "ARGB8565_RBSWAP", "ARGB8888"])
+    parser.add_argument("-s", "--swap-endian",
+        help="swap endian of image (not implemented)",
+        action="store_true")
+    parser.add_argument("-d", "--dither",
+        help="enable dither (not implemented)",
+        action="store_true")
+    args = parser.parse_args()
+
+    img_path = pathlib.Path(args.img)
+    out = pathlib.Path(args.output_file)
+    if not img_path.is_file():
+        print(f"Input file is missing: '{args.img}'")
+        return 1
+    print(f"Beginning conversion of {args.img}")
+    if out.exists():
+        if args.force:
+            print(f"overwriting {args.output_file}")
+        else:
+            pritn(f"Error: refusing to overwrite {args.output_file} without -f specified.")
+            return 1
+    out.touch()
+
+    # only implemented the bare minimum, everything else is not implemented
+    if args.color_format not in ["CF_INDEXED_1_BIT", "CF_TRUE_COLOR_ALPHA"]:
+        raise NotImplementedError(f"argument --color-format '{args.color_format}' not implemented")
+    if args.output_format != "bin":
+        raise NotImplementedError(f"argument --output-format '{args.output_format}' not implemented")
+    if args.binary_format not in ["ARGB8565_RBSWAP", "ARGB8888"]:
+        raise NotImplementedError(f"argument --binary-format '{args.binary_format}' not implemented")
+    if args.image_name:
+        raise NotImplementedError(f"argument --image-name not implemented")
+    if args.swap_endian:
+        raise NotImplementedError(f"argument --swap-endian not implemented")
+    if args.dither:
+        raise NotImplementedError(f"argument --dither not implemented")
+
+    # open image using Pillow
+    img = Image.open(img_path)
+    img_height = img.height
+    img_width = img.width
+    if args.color_format == "CF_TRUE_COLOR_ALPHA" and args.binary_format == "ARGB8888":
+        buf = bytearray(img_height*img_width*4) # 4 bytes (32 bit) per pixel
+        for y in range(img_height):
+            for x in range(img_width):
+                i = (y*img_width + x)*4 # buffer-index
+                pixel = img.getpixel((x,y))
+                r, g, b, a = pixel
+                buf[i + 0] = r
+                buf[i + 1] = g
+                buf[i + 2] = b
+                buf[i + 3] = a
+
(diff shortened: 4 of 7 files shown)


NeroBurner:
I've fixed part of the differences between the original nodejs script and the python based port. It was the usage of `round()` in the helper function `classify_pixel()`.
Python3 implements "propper" "banker's rounding" by rounding to the nearest
even number. Javascript rounds to the nearest integer as we've learned in school.
To have the same output as the original JavaScript implementation add a custom
rounding function, which does "school" rounding (to the nearest integer).

---

After that there is still a difference between the binaries. Therefore I implemented the `ARGB8888` binary-output-format to compare the rgba values read by the scripts from the png.

```sh
./lv_img_conv.js pine_logo.png --force --output-file pine_small_js_8888.bin --color-format CF_TRUE_COLOR_ALPHA --output-format bin --binary-format ARGB8888
xxd pine_small_js_8888.bin > pine_small_js_8888.bin.hex
```

Same with the current python port

```sh
python3 lv_img_conv.py  --force --output-file pine_small_js_8888.bin --color-format CF_TRUE_COLOR_ALPHA --output-format bin --binary-format ARGB8888
xxd pine_small_py_8888.bin > pine_small_py_8888.bin.hex
```

Uploaded the hex files for comparison:
- [pine_small_js_8888.bin.hex](https://github.com/InfiniTimeOrg/InfiniTime/files/12753707/pine_small_js_8888.bin.hex.txt)
- [pine_small_py_8888.bin.hex](https://github.com/InfiniTimeOrg/InfiniTime/files/12753711/pine_small_py_8888.bin.hex.txt)

Comparing the two hex files shows there are differences in how the PNG file is read in (because ARGB8888 just outputs the read rgba values)
Screenshot_20230928_222816

The first different pixel is at x==10, y==0. Looking at that pixel nodejs returns `rgba=[19, 19, 19, 26]` compared to python `rgba=[20, 20, 20, 26]`

Opened the `pine-logo.png` in Gimp and it shows the same values as the python script

gimp rgba color values

This off-by-one input differences results in some differences in the normal `ARGB8565_RBSWAP` output binary format as well.
As the difference is small and I believe the value from Gimp I'd keep the python script as it is (as I don't know how to "fix" it to be like the node input)

---

The only difference for the `navigation0.png` and `navigation1.png` files is the hard coded color palette. The remaining bits are the same for nodejs and python implementation

src/resources/lv_img_conv.py
```diff
@@ -0,0 +1,214 @@
...
+    img_height = img.height
+    img_width = img.width
+    print(f"loaded image with width x heigth: {img_width} x {img_height}")
+    if args.color_format == "CF_TRUE_COLOR_ALPHA" and args.binary_format == "ARGB8888":
+        buf = bytearray(img_height*img_width*4) # 4 bytes (32 bit) per pixel
+        for y in range(img_height):
+            for x in range(img_width):
+                i = (y*img_width + x)*4 # buffer-index
+                pixel = img.getpixel((x,y))
+                r = pixel[0]
```
mark9064, line 100: You can reduce this to `buf[i:i+4] = pixel` (can also then inline pixel)
Fair enough if you'd rather be more explicit about it though

src/resources/lv_img_conv.py
```diff
@@ -0,0 +1,214 @@
...
+def classify_pixel(value, bits):
+    def round_half_up(v):
+        """python3 implements "propper" "banker's rounding" by rounding to the nearest
+        even number. Javascript rounds to the nearest integer.
+        To have the same output as the original JavaScript implementation add a custom
+        rounding function, which does "school" rounding (to the nearest integer).
+
+        see: https://stackoverflow.com/questions/43851273/how-to-round-float-0-5-up-to-1-0-while-still-rounding-0-45-to-0-0-as-the-usual
+        """
+        return int(decimal.Decimal(v).quantize(decimal.Decimal('1'), rounding=decimal.ROUND_HALF_UP))
```
mark9064, line 18: Is there any advantage to using the JS rounding method rather than round-to-even? If it doesn't matter I think we may as well avoid having the extra code
It looks like GIMP agrees with round-to-even as well (if I'm reading your previous PR comments right), and I trust GIMP above lv_image_conv on correctness

src/resources/lv_img_conv.py
```diff
@@ -0,0 +1,214 @@
...
+        if args.force:
+            print(f"overwriting {args.output_file}")
+        else:
+            print(f"output-file exists, set --force to allow overwriting of file")
+            return 0
+    out.touch()
+
+    # only implemented the bare minimum, everything else is not implemented
+    if args.color_format not in ["CF_INDEXED_1_BIT", "CF_TRUE_COLOR_ALPHA"]:
+        raise NotImplementedError(f"args.color_format '{args.color_format}' not implemented")
```
mark9064, line 83:
```suggestion
        raise NotImplementedError(f"Color format '{args.color_format}' not implemented")
```
and same for others

src/resources/lv_img_conv.py
```diff
@@ -0,0 +1,214 @@
...
+                buf[p] |= (c & 0x1) << (7 - (x & 0x7))
+        # write palette information
+        palette_size = 2
+        bits_per_value = 1
+        # write 8 palette bytes
+        buf[0] = 0
+        buf[1] = 0
+        buf[2] = 0
+        buf[3] = 0
+        # Normally there is much math behind this, but for the current use case this is close enough
```
mark9064, line 174: Could this describe under what scenario this will become a problem (if any)

src/resources/lv_img_conv.py
```diff
@@ -0,0 +1,214 @@
...
+    parser.add_argument("-t", "--output-format",
+        help="output format of image",
+        default="bin", # default in original is 'c'
+        choices=["c", "bin"])
+    parser.add_argument("--binary-format",
+        help="binary color format (needed if output-format is binary)",
+        default="ARGB8565_RBSWAP",
+        choices=["ARGB8332", "ARGB8565", "ARGB8565_RBSWAP", "ARGB8888"])
+    #  --swap-endian, -s    swap endian of image                            [boolean]
+    #  --dither, -d         enable dither                                   [boolean]
```
mark9064, line 64: Should these go? (same for image-name)

src/resources/lv_img_conv.py
```diff
@@ -0,0 +1,214 @@
...
+    out = pathlib.Path(args.output_file)
+    if not img_path.is_file():
+        print(f"Input file is missing: '{args.img}'")
+        return 1
+    print(f"Beginning conversion of {args.img}")
+    if out.exists():
+        if args.force:
+            print(f"overwriting {args.output_file}")
+        else:
+            print(f"output-file exists, set --force to allow overwriting of file")
```
mark9064, line 77: f-string with no variables

src/resources/lv_img_conv.py
```diff
@@ -0,0 +1,214 @@
...
+
+        for y in range(img_height):
+            for x in range(img_width):
+                c, a = img.getpixel((x,y))
+                p = w * y + (x >> 3) + 8  # +8 for the palette
+                #if(!isset(this.d_out[p])) this.d_out[p] = 0  # Clear the bits first
+                buf[p] |= (c & 0x1) << (7 - (x & 0x7))
+        # write palette information
+        palette_size = 2
+        bits_per_value = 1
```
mark9064, line 168: Unused variables

mark9064 (commented):
Looks great overall, performance is a lot better as well. Haven't tested on hardware but according to your earlier tests I trust the output
Mostly code cleanliness suggestions/doc suggestions, only architectural one is whether you want to import the file directly or run it as a subprocess

.github/workflows/main.yml
```diff
@@ -31,6 +31,10 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
+      - name: Install python3-pil
```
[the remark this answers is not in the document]
NeroBurner: once the docker file is updated this install should be removed as it is already done in the used docker image :thinking:

doc/buildAndProgram.md
```diff
@@ -42,7 +42,7 @@ CMake configures the project according to variables you specify the command line
 **NRF5_SDK_PATH**|path to the NRF52 SDK|`-DNRF5_SDK_PATH=/home/jf/nrf52/Pinetime/sdk`|
 **CMAKE_BUILD_TYPE (\*)**| Build type (Release or Debug). Release is applied by default if this variable is not specified.|`-DCMAKE_BUILD_TYPE=Debug`
 **BUILD_DFU (\*\*)**|Build DFU files while building (needs [adafruit-nrfutil](https://github.com/adafruit/Adafruit_nRF52_nrfutil)).|`-DBUILD_DFU=1`
-**BUILD_RESOURCES (\*\*)**| Generate external resource while building (needs [lv_font_conv](https://github.com/lvgl/lv_font_conv) and [lv_img_conv](https://github.com/lvgl/lv_img_conv). |`-DBUILD_RESOURCES=1`
+**BUILD_RESOURCES (\*\*)**| Generate external resource while building (needs [lv_font_conv](https://github.com/lvgl/lv_font_conv) and `python3-pil` module). |`-DBUILD_RESOURCES=1`
```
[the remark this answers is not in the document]
NeroBurner: good catch, gonna mention the Ubuntu package name, but link to the pillow documentation

src/resources/lv_img_conv.py
NeroBurner: not implemented in the python script, but options available in the original node version. I'm undecided whether to keep them or not. Tried to be as close as possible to the original script.

Maybe add the arguments and throwing unimplemented errors instead?

src/resources/lv_img_conv.py
NeroBurner: good catch. using original error message now

src/resources/lv_img_conv.py
```diff
@@ -0,0 +1,214 @@
...
+    if not img_path.is_file():
+        print(f"Input file is missing: '{args.img}'")
+        return 1
+    print(f"Beginning conversion of {args.img}")
+    if out.exists():
+        if args.force:
+            print(f"overwriting {args.output_file}")
+        else:
+            print(f"output-file exists, set --force to allow overwriting of file")
+            return 0
```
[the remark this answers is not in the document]
NeroBurner: yes, good catch. Original message and return value

```js
console.error("Error: refusing to overwrite " + outputPath + " without -f specified.");
process.exit(1);
```

src/resources/lv_img_conv.py
NeroBurner: I'm gonna use the argument directly instead, in this case `--color-format`

src/resources/lv_img_conv.py
```diff
@@ -0,0 +1,214 @@
...
+                #    val_g = round(g / tmp_6) * tmp_6
+                #    val_b = round(b / tmp_5) * tmp_5
+                #    print("rgba:", r, g, b, a)
+                #    print("rgba_act:", r_act, g_act, b_act, a)
+                #    print("classify_pixel(r, 5):", "tmp:", tmp_5, "r/tmp:", r/tmp_5, "round(r/tmp):", round(r/tmp_5), "val:", val_r)
+                #    print("classify_pixel(g, 6):", "tmp:", tmp_6, "g/tmp:", g/tmp_6, "round(g/tmp):", round(g/tmp_6), "val:", val_g)
+                #    print("classify_pixel(b, 5):", "tmp:", tmp_5, "b/tmp:", b/tmp_5, "round(b/tmp):", round(b/tmp_5), "val:", val_b)
+                #    print("c16:", c16)
+                #    print(f"{x},{y}: {buf_i0:02x}, {buf_i1:02x}, {buf_i2:02x}")
+                #    print("found it!")
```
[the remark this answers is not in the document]
NeroBurner: yeah, debug code for finding and debugging the differences

src/resources/lv_img_conv.py
```diff
@@ -0,0 +1,214 @@
...
+            print(f"done line {y}")
```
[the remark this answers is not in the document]
NeroBurner: indeed! :sweat_smile:

NeroBurner (commented):
partial fixes, gonna do the rest at a later time

src/resources/generate-img.py
```diff
@@ -11,6 +11,9 @@
 
 def gen_lvconv_line(lv_img_conv: str, dest: str, color_format: str, output_format: str, binary_format: str, sources: str):
     args = [lv_img_conv, sources, '--force', '--output-file', dest, '--color-format', color_format, '--output-format', output_format, '--binary-format', binary_format]
+    if lv_img_conv.endswith(".py"):
```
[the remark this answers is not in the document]
NeroBurner: my reasoning is that if we need some other functionality from the JS version it is easy to switch between JS and python script

JF002 (approved):
I'm far from a Python expert, so I trust you on that regard :+1: 

This scripts makes the setup of the build environment much easier and it works fine on my setup and in the build Docker container so... LGTM :-)

src/resources/generate-img.py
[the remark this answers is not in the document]
mark9064: Sure, sounds fine

src/resources/lv_img_conv.py
mark9064: Yeah I think explicitly flagging that they're not implemented would be good if we're trying to expose the same command line interface

src/resources/lv_img_conv.py
```diff
@@ -0,0 +1,214 @@
...
+        buf = bytearray(img_height*img_width*3) # 3 bytes (24 bit) per pixel
+        for y in range(img_height):
+            for x in range(img_width):
+                i = (y*img_width + x)*3 # buffer-index
+                pixel = img.getpixel((x,y))
+                r_act = classify_pixel(pixel[0], 5)
+                g_act = classify_pixel(pixel[1], 6)
+                b_act = classify_pixel(pixel[2], 5)
+                a = pixel[3]
+                if r_act > 0xF8:
```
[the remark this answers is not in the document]
NeroBurner: will do, thanks

src/resources/lv_img_conv.py
```diff
@@ -0,0 +1,214 @@
...
+        if img_width & 0x07:
+            w+=1
+        max_p = w * (img_height-1) + ((img_width-1) >> 3) + 8  # +8 for the palette
+        buf = bytearray(max_p+1)
+
+        for y in range(img_height):
+            for x in range(img_width):
+                c, a = img.getpixel((x,y))
+                p = w * y + (x >> 3) + 8  # +8 for the palette
+                #if(!isset(this.d_out[p])) this.d_out[p] = 0  # Clear the bits first
```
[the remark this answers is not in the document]
NeroBurner: yes, the original code uses some datastructure, that extends the memory range if needed. We create the memory range before the image conversion, so this comment (or what it describes) isn't needed for us

src/resources/lv_img_conv.py
NeroBurner: thanks for the hint, but for this one i'd like to be explicit about the r,g,b,a parts

src/resources/lv_img_conv.py
NeroBurner: the GIMP case is for how the PNG is read into memory, the rounding is for how the pixels are rounded before written out. I'd like to keep to be as close as possible to the original here (so I'd like to keep the rounding)

src/resources/lv_img_conv.py
NeroBurner: We currently have only one bit per color, and the input images are either transparent, or black. If we had input pictures with different color this information would be lost. I have no idea if this information would be important for a 1-bit-per-pixel image (and I wasn't able to figure out how to do a correct color palette calculation)

src/resources/lv_img_conv.py
NeroBurner: doing that now :+1:

src/resources/lv_img_conv.py
NeroBurner: removing, adding information into comment above

.github/workflows/main.yml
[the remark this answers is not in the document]
NeroBurner: renamed as recommended, probably should remove in future :thinking: 
but keeping it for now to allow a smooth CI until docker image is updated

NeroBurner:
code review incorporated, squashed and pushed, gonna merge as soon as possible

NeroBurner:
User MorsMortium informed me on Matrix, that there is a full port available at https://github.com/W-Mai/lvgl_image_converter

Still using this minimum port, and if we need more we could either switch to the full port or use the full port as inspiration to implement the newly needed parts

mark9064 (approved):
Looks great :+1:

PR #1869, opened 2023-09-30
Always on display

KaffeinatedKat:
I have added an option in the "Display timeout" for an always on display. Not sure how useful this is, or how good the battery life is (will be testing this and will post a comment with how long the battery lasted).

Any suggestions on how I could possibly increase the battery with the display always on would be great.

image

### DIFF
--- a/src/components/brightness/BrightnessController.h
+++ b/src/components/brightness/BrightnessController.h
@@ -2,11 +2,14 @@
 
 #include <cstdint>
 
+#include "nrf_ppi.h"
+#include "nrfx_gpiote.h"
+
 namespace Pinetime {
   namespace Controllers {
     class BrightnessController {
     public:
-      enum class Levels { Off, Low, Medium, High };
+      enum class Levels { Off, AlwaysOn, Low, Medium, High };
       void Init();
 
       void Set(Levels level);
@@ -20,6 +23,25 @@ namespace Pinetime {
 
     private:
       Levels level = Levels::High;
+      static constexpr uint8_t UNSET = UINT8_MAX;
+      uint8_t lastPin = UNSET;
+      // Maximum time (μs) it takes for the RTC to fully stop
+      static constexpr uint8_t rtcStopTime = 46;
+      // Frequency of timer used for PWM (Hz)
+      static constexpr uint16_t timerFrequency = 32768;
+      // Backlight PWM frequency (Hz)
+      static constexpr uint16_t pwmFreq = 1000;
+      // Wraparound point in timer ticks
+      // Defines the number of brightness levels between each pin
+      static constexpr uint16_t timerPeriod = timerFrequency / pwmFreq;
+      // Warning: nimble reserves some PPIs
+      // https://github.com/InfiniTimeOrg/InfiniTime/blob/034d83fe6baf1ab3875a34f8cee387e24410a824/src/libs/mynewt-nimble/nimble/drivers/nrf52/src/ble_phy.c#L53
+      // SpiMaster uses PPI 0 for an erratum workaround
+      // Channel 1, 2 should be free to use
+      static constexpr nrf_ppi_channel_t ppiBacklightOn = NRF_PPI_CHANNEL1;
+      static constexpr nrf_ppi_channel_t ppiBacklightOff = NRF_PPI_CHANNEL2;
+
+      void ApplyBrightness(uint16_t val);
     };
   }
 }
--- a/src/components/settings/Settings.h
+++ b/src/components/settings/Settings.h
@@ -214,6 +214,17 @@ namespace Pinetime {
         return settings.screenTimeOut;
       };
 
+      void SetAlwaysOnDisplay(bool state) {
+        if (state != settings.alwaysOnDisplay) {
+          settingsChanged = true;
+        }
+        settings.alwaysOnDisplay = state;
+      };
+
+      bool GetAlwaysOnDisplay() const {
+        return settings.alwaysOnDisplay;
+      };
+
       void SetShakeThreshold(uint16_t thresh) {
         if (settings.shakeWakeThreshold != thresh) {
           settings.shakeWakeThreshold = thresh;
@@ -286,12 +297,14 @@ namespace Pinetime {
     private:
       Pinetime::Controllers::FS& fs;
 
-      static constexpr uint32_t settingsVersion = 0x0007;
+      static constexpr uint32_t settingsVersion = 0x0008;
 
       struct SettingsData {
         uint32_t version = settingsVersion;
         uint32_t stepsGoal = 10000;
         uint32_t screenTimeOut = 15000;
+
+        bool alwaysOnDisplay = false;
 
         ClockType clockType = ClockType::H24;
         WeatherFormat weatherFormat = WeatherFormat::Metric;
--- a/src/components/settings/Settings.h
+++ b/src/components/settings/Settings.h
@@ -196,6 +196,14 @@ namespace Pinetime {
         if (status != settings.notificationStatus) {
           settingsChanged = true;
         }
+        // Disable always on screen while sleep mode is enabled
+        if (settings.alwaysOnDisplay.enabled) {
+          if (status == Notification::Sleep) {
+            settings.alwaysOnDisplay.state = false;
+          } else {
+            settings.alwaysOnDisplay.state = true;
+          }
+        }
         settings.notificationStatus = status;
       };
 
@@ -215,15 +223,31 @@ namespace Pinetime {
       };
 
       void SetAlwaysOnDisplay(bool state) {
-        if (state != settings.alwaysOnDisplay) {
+        if (state != settings.alwaysOnDisplay.state) {
           settingsChanged = true;
         }
-        settings.alwaysOnDisplay = state;
+        settings.alwaysOnDisplay.state = state;
       };
 
       bool GetAlwaysOnDisplay() const {
-        return settings.alwaysOnDisplay;
+        return settings.alwaysOnDisplay.state;
       };
+
+      void SetAlwaysOnDisplaySetting(bool state) {
+        if (state != settings.alwaysOnDisplay.enabled) {
+          settingsChanged = true;
+        }
+        settings.alwaysOnDisplay.enabled = state;
+
+        // Don't enable always on if we are currently in notification sleep
+        if (GetNotificationStatus() != Notification::Sleep) {
+          SetAlwaysOnDisplay(state);
+        }
+      }
+
+      bool GetAlwaysOnDisplaySetting() const {
+        return settings.alwaysOnDisplay.enabled;
+      }
 
       void SetShakeThreshold(uint16_t thresh) {
         if (settings.shakeWakeThreshold != thresh) {
@@ -299,12 +323,19 @@ namespace Pinetime {
 
       static constexpr uint32_t settingsVersion = 0x0008;
 
+      // To enable disabling it during notification sleep, differentiate between
+      // the setting being on, and the setting being set by the user
+      struct alwaysOnDisplayData {
+        bool enabled = false;
+        bool state = false;
+      };
+
       struct SettingsData {
         uint32_t version = settingsVersion;
         uint32_t stepsGoal = 10000;
         uint32_t screenTimeOut = 15000;
 
-        bool alwaysOnDisplay = false;
+        alwaysOnDisplayData alwaysOnDisplay;
 
         ClockType clockType = ClockType::H24;
         WeatherFormat weatherFormat = WeatherFormat::Metric;
--- a/src/components/settings/Settings.h
+++ b/src/components/settings/Settings.h
@@ -196,14 +196,6 @@ namespace Pinetime {
         if (status != settings.notificationStatus) {
           settingsChanged = true;
         }
-        // Disable always on screen while sleep mode is enabled
-        if (settings.alwaysOnDisplay.enabled) {
-          if (status == Notification::Sleep) {
-            settings.alwaysOnDisplay.state = false;
-          } else {
-            settings.alwaysOnDisplay.state = true;
-          }
-        }
         settings.notificationStatus = status;
       };
 
@@ -222,31 +214,19 @@ namespace Pinetime {
         return settings.screenTimeOut;
       };
 
-      void SetAlwaysOnDisplay(bool state) {
-        if (state != settings.alwaysOnDisplay.state) {
-          settingsChanged = true;
-        }
-        settings.alwaysOnDisplay.state = state;
-      };
-
       bool GetAlwaysOnDisplay() const {
-        return settings.alwaysOnDisplay.state;
+        return settings.alwaysOnDisplay && GetNotificationStatus() != Notification::Sleep;
       };
 
       void SetAlwaysOnDisplaySetting(bool state) {
-        if (state != settings.alwaysOnDisplay.enabled) {
+        if (state != settings.alwaysOnDisplay) {
           settingsChanged = true;
         }
-        settings.alwaysOnDisplay.enabled = state;
-
-        // Don't enable always on if we are currently in notification sleep
-        if (GetNotificationStatus() != Notification::Sleep) {
-          SetAlwaysOnDisplay(state);
-        }
+        settings.alwaysOnDisplay = state;
       }
 
       bool GetAlwaysOnDisplaySetting() const {
-        return settings.alwaysOnDisplay.enabled;
+        return settings.alwaysOnDisplay;
       }
 
       void SetShakeThreshold(uint16_t thresh) {
@@ -323,19 +303,12 @@ namespace Pinetime {
 
       static constexpr uint32_t settingsVersion = 0x0008;
 
-      // To enable disabling it during notification sleep, differentiate between
-      // the setting being on, and the setting being set by the user
-      struct alwaysOnDisplayData {
-        bool enabled = false;
-        bool state = false;
-      };
-
       struct SettingsData {
         uint32_t version = settingsVersion;
         uint32_t stepsGoal = 10000;
         uint32_t screenTimeOut = 15000;
 
-        alwaysOnDisplayData alwaysOnDisplay;
+        bool alwaysOnDisplay = false;
 
         ClockType clockType = ClockType::H24;
         WeatherFormat weatherFormat = WeatherFormat::Metric;
--- a/src/drivers/St7789.cpp
+++ b/src/drivers/St7789.cpp
@@ -25,6 +25,9 @@ void St7789::Init() {
 #ifndef DRIVER_DISPLAY_MIRROR
   DisplayInversionOn();
 #endif
+  PorchSet();
+  FrameRateNormalSet();
+  IdleFrameRateOff();
   NormalModeOn();
   SetVdv();
   PowerControl();
@@ -149,27 +152,44 @@ void St7789::IdleModeOff() {
   WriteCommand(static_cast<uint8_t>(Commands::IdleModeOff));
 }
 
-void St7789::FrameRateLow() {
-  WriteCommand(static_cast<uint8_t>(Commands::FrameRate));
-  // Enable frame rate control for partial/idle mode, 8x frame divider
+void St7789::PorchSet() {
+  WriteCommand(static_cast<uint8_t>(Commands::Porch));
+  constexpr uint8_t args[] = {
+    0x02, // Normal mode front porch
+    0x03, // Normal mode back porch
+    0x01, // Porch control enable
+    0xed, // Idle mode front:back porch
+    0xed, // Partial mode front:back porch (partial mode unused but set anyway)
+  };
+  WriteData(args, sizeof(args));
+}
+
+void St7789::FrameRateNormalSet() {
+  WriteCommand(static_cast<uint8_t>(Commands::FrameRateNormal));
+  // Note that the datasheet table is imprecise - see formula below table
+  WriteData(0x0a);
+}
+
+void St7789::IdleFrameRateOn() {
+  WriteCommand(static_cast<uint8_t>(Commands::FrameRateIdle));
   // According to the datasheet, these controls should apply only to partial/idle mode
   // However they appear to apply to normal mode, so we have to enable/disable
   // every time we enter/exit always on
   // In testing this divider appears to actually be 16x?
   constexpr uint8_t args[] = {
     0x13, // Enable frame rate control for partial/idle mode, 8x frame divider
-    0x1f, // Idle mode frame rate (lowest possible)
-    0x1f, // Partial mode frame rate (lowest possible, unused)
+    0x1e, // Idle mode frame rate
+    0x1e, // Partial mode frame rate (unused)
   };
   WriteData(args, sizeof(args));
 }
 
-void St7789::FrameRateNormal() {
-  WriteCommand(static_cast<uint8_t>(Commands::FrameRate));
+void St7789::IdleFrameRateOff() {
+  WriteCommand(static_cast<uint8_t>(Commands::FrameRateIdle));
   constexpr uint8_t args[] = {
     0x00, // Disable frame rate control and divider
-    0x0f, // Idle mode frame rate (normal)
-    0x0f, // Partial mode frame rate (normal, unused)
+    0x0a, // Idle mode frame rate (normal)
+    0x0a, // Partial mode frame rate (normal, unused)
   };
   WriteData(args, sizeof(args));
 }
@@ -266,13 +286,13 @@ void St7789::HardwareReset() {
 
 void St7789::LowPowerOn() {
   IdleModeOn();
-  FrameRateLow();
+  IdleFrameRateOn();
   NRF_LOG_INFO("[LCD] Low power mode");
 }
 
 void St7789::LowPowerOff() {
   IdleModeOff();
-  FrameRateNormal();
+  IdleFrameRateOff();
   NRF_LOG_INFO("[LCD] Normal power mode");
 }
 
--- a/src/drivers/St7789.cpp
+++ b/src/drivers/St7789.cpp
@@ -16,6 +16,7 @@ void St7789::Init() {
   nrf_gpio_pin_set(pinReset);
   HardwareReset();
   SoftwareReset();
+  Command2Enable();
   SleepOut();
   PixelFormat();
   MemoryDataAccessControl();
@@ -61,6 +62,17 @@ void St7789::SoftwareReset() {
   sleepIn = true;
   lastSleepExit = xTaskGetTickCount();
   vTaskDelay(pdMS_TO_TICKS(125));
+}
+
+void St7789::Command2Enable() {
+  WriteCommand(static_cast<uint8_t>(Commands::Command2Enable));
+  constexpr uint8_t args[] = {
+    0x5a, // Constant
+    0x69, // Constant
+    0x02, // Constant
+    0x01, // Enable
+  };
+  WriteData(args, sizeof(args));
 }
 
 void St7789::SleepOut() {
@@ -133,6 +145,31 @@ void St7789::IdleModeOn() {
 
 void St7789::IdleModeOff() {
   WriteCommand(static_cast<uint8_t>(Commands::IdleModeOff));
+}
+
+void St7789::FrameRateLow() {
+  WriteCommand(static_cast<uint8_t>(Commands::FrameRate));
+  // Enable frame rate control for partial/idle mode, 8x frame divider
+  // According to the datasheet, these controls should apply only to partial/idle mode
+  // However they appear to apply to normal mode, so we have to enable/disable
+  // every time we enter/exit always on
+  // In testing this divider appears to actually be 16x?
+  constexpr uint8_t args[] = {
+    0x13, // Enable frame rate control for partial/idle mode, 8x frame divider
+    0x1f, // Idle mode frame rate (lowest possible)
+    0x1f, // Partial mode frame rate (lowest possible, unused)
+  };
+  WriteData(args, sizeof(args));
+}
+
+void St7789::FrameRateNormal() {
+  WriteCommand(static_cast<uint8_t>(Commands::FrameRate));
+  constexpr uint8_t args[] = {
+    0x00, // Disable frame rate control and divider
+    0x0f, // Idle mode frame rate (normal)
+    0x0f, // Partial mode frame rate (normal, unused)
+  };
+  WriteData(args, sizeof(args));
 }
 
 void St7789::DisplayOn() {
@@ -208,11 +245,13 @@ void St7789::HardwareReset() {
 
 void St7789::LowPowerOn() {
   IdleModeOn();
+  FrameRateLow();
   NRF_LOG_INFO("[LCD] Low power mode");
 }
 
 void St7789::LowPowerOff() {
   IdleModeOff();
+  FrameRateNormal();
   NRF_LOG_INFO("[LCD] Normal power mode");
 }
 
--- a/src/drivers/St7789.cpp
+++ b/src/drivers/St7789.cpp
@@ -27,6 +27,7 @@ void St7789::Init() {
 #endif
   NormalModeOn();
   SetVdv();
+  PowerControl();
   DisplayOn();
 }
 
@@ -174,6 +175,19 @@ void St7789::FrameRateNormal() {
 
 void St7789::DisplayOn() {
   WriteCommand(static_cast<uint8_t>(Commands::DisplayOn));
+}
+
+void St7789::PowerControl() {
+  WriteCommand(static_cast<uint8_t>(Commands::PowerControl1));
+  constexpr uint8_t args[] = {
+    0xa4, // Constant
+    0x00, // Lowest possible voltages
+  };
+  WriteData(args, sizeof(args));
+
+  WriteCommand(static_cast<uint8_t>(Commands::PowerControl2));
+  // Lowest possible boost circuit clocks
+  WriteData(0xb3);
 }
 
 void St7789::SetAddrWindow(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1) {
--- a/src/drivers/St7789.cpp
+++ b/src/drivers/St7789.cpp
@@ -127,6 +127,14 @@ void St7789::NormalModeOn() {
   WriteCommand(static_cast<uint8_t>(Commands::NormalModeOn));
 }
 
+void St7789::IdleModeOn() {
+  WriteCommand(static_cast<uint8_t>(Commands::IdleModeOn));
+}
+
+void St7789::IdleModeOff() {
+  WriteCommand(static_cast<uint8_t>(Commands::IdleModeOff));
+}
+
 void St7789::DisplayOn() {
   WriteCommand(static_cast<uint8_t>(Commands::DisplayOn));
 }
@@ -196,6 +204,16 @@ void St7789::HardwareReset() {
   sleepIn = true;
   lastSleepExit = xTaskGetTickCount();
   vTaskDelay(pdMS_TO_TICKS(125));
+}
+
+void St7789::LowPowerOn() {
+  IdleModeOn();
+  NRF_LOG_INFO("[LCD] Low power mode");
+}
+
+void St7789::LowPowerOff() {
+  IdleModeOff();
+  NRF_LOG_INFO("[LCD] Normal power mode");
 }
 
 void St7789::Sleep() {
--- a/src/drivers/St7789.cpp
+++ b/src/drivers/St7789.cpp
@@ -28,6 +28,7 @@ void St7789::Init() {
   NormalModeOn();
   SetVdv();
   PowerControl();
+  GateControl();
   DisplayOn();
 }
 
@@ -188,6 +189,12 @@ void St7789::PowerControl() {
   WriteCommand(static_cast<uint8_t>(Commands::PowerControl2));
   // Lowest possible boost circuit clocks
   WriteData(0xb3);
+}
+
+void St7789::GateControl() {
+  WriteCommand(static_cast<uint8_t>(Commands::GateControl));
+  // Lowest possible VGL/VGH
+  WriteData(0x00);
 }
 
 void St7789::SetAddrWindow(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1) {
--- a/src/systemtask/SystemTask.cpp
+++ b/src/systemtask/SystemTask.cpp
@@ -102,7 +102,9 @@ void SystemTask::Work() {
   watchdog.Setup(7, Drivers::Watchdog::SleepBehaviour::Run, Drivers::Watchdog::HaltBehaviour::Pause);
   watchdog.Start();
   NRF_LOG_INFO("Last reset reason : %s", Pinetime::Drivers::ResetReasonToString(watchdog.GetResetReason()));
-  APP_GPIOTE_INIT(2);
+  if (!nrfx_gpiote_is_init()) {
+    nrfx_gpiote_init();
+  }
 
   spi.Init();
   spiNorFlash.Init();
--- a/src/systemtask/SystemTask.cpp
+++ b/src/systemtask/SystemTask.cpp
@@ -194,7 +194,7 @@ void SystemTask::Work() {
           if (!bleController.IsFirmwareUpdating()) {
             doNotGoToSleep = false;
           }
-          displayApp.PushMessage(Pinetime::Applications::Display::Messages::RestoreBrightness);
+          displayApp.PushMessage(Pinetime::Applications::Display::Messages::NotifyDeviceActivity);
           break;
         case Messages::DisableSleeping:
           doNotGoToSleep = true;
@@ -245,7 +245,7 @@ void SystemTask::Work() {
           heartRateApp.PushMessage(Pinetime::Applications::HeartRateTask::Messages::GoToSleep);
           break;
         case Messages::OnNewTime:
-          displayApp.PushMessage(Pinetime::Applications::Display::Messages::RestoreBrightness);
+          displayApp.PushMessage(Pinetime::Applications::Display::Messages::NotifyDeviceActivity);
           displayApp.PushMessage(Pinetime::Applications::Display::Messages::UpdateDateTime);
           if (alarmController.State() == Controllers::AlarmController::AlarmState::Set) {
             alarmController.ScheduleAlarm();
@@ -256,7 +256,7 @@ void SystemTask::Work() {
             if (state == SystemTaskState::Sleeping) {
               GoToRunning();
             } else {
-              displayApp.PushMessage(Pinetime::Applications::Display::Messages::RestoreBrightness);
+              displayApp.PushMessage(Pinetime::Applications::Display::Messages::NotifyDeviceActivity);
             }
             displayApp.PushMessage(Pinetime::Applications::Display::Messages::NewNotification);
           }
@@ -268,7 +268,7 @@ void SystemTask::Work() {
           displayApp.PushMessage(Pinetime::Applications::Display::Messages::AlarmTriggered);
           break;
         case Messages::BleConnected:
-          displayApp.PushMessage(Pinetime::Applications::Display::Messages::RestoreBrightness);
+          displayApp.PushMessage(Pinetime::Applications::Display::Messages::NotifyDeviceActivity);
           isBleDiscoveryTimerRunning = true;
           bleDiscoveryTimer = 5;
           break;
@@ -466,7 +466,7 @@ void SystemTask::HandleButtonAction(Controllers::ButtonActions action) {
     return;
   }
 
-  displayApp.PushMessage(Pinetime::Applications::Display::Messages::RestoreBrightness);
+  displayApp.PushMessage(Pinetime::Applications::Display::Messages::NotifyDeviceActivity);
 
   using Actions = Controllers::ButtonActions;
 
--- a/src/systemtask/SystemTask.cpp
+++ b/src/systemtask/SystemTask.cpp
@@ -236,7 +236,7 @@ void SystemTask::Work() {
           break;
         }
         case Messages::GoToSleep:
-          if (doNotGoToSleep or settingsController.GetAlwaysOnDisplay()) {
+          if (doNotGoToSleep) {
             break;
           }
           state = SystemTaskState::GoingToSleep; // Already set in PushMessage()
@@ -512,7 +512,7 @@ void SystemTask::OnTouchEvent() {
 }
 
 void SystemTask::PushMessage(System::Messages msg) {
-  if (msg == Messages::GoToSleep && !doNotGoToSleep && !settingsController.GetAlwaysOnDisplay()) {
+  if (msg == Messages::GoToSleep && !doNotGoToSleep) {
     state = SystemTaskState::GoingToSleep;
   }
 
--- a/src/systemtask/SystemTask.cpp
+++ b/src/systemtask/SystemTask.cpp
@@ -198,7 +198,10 @@ void SystemTask::Work() {
           doNotGoToSleep = true;
           break;
         case Messages::GoToRunning:
-          spi.Wakeup();
+          // SPI doesn't go to sleep for always on mode
+          if (!settingsController.GetAlwaysOnDisplay()) {
+            spi.Wakeup();
+          }
 
           // Double Tap needs the touch screen to be in normal mode
           if (!settingsController.isWakeUpModeOn(Pinetime::Controllers::Settings::WakeUpMode::DoubleTap)) {
@@ -231,7 +234,7 @@ void SystemTask::Work() {
           break;
         }
         case Messages::GoToSleep:
-          if (doNotGoToSleep) {
+          if (doNotGoToSleep or settingsController.GetAlwaysOnDisplay()) {
             break;
           }
           state = SystemTaskState::GoingToSleep; // Already set in PushMessage()
@@ -323,7 +326,11 @@ void SystemTask::Work() {
             // if it's in sleep mode. Avoid bricked device by disabling sleep mode on these versions.
             spiNorFlash.Sleep();
           }
-          spi.Sleep();
+
+          // Must keep SPI awake when still updating the display for always on
+          if (!settingsController.GetAlwaysOnDisplay()) {
+            spi.Sleep();
+          }
 
           // Double Tap needs the touch screen to be in normal mode
           if (!settingsController.isWakeUpModeOn(Pinetime::Controllers::Settings::WakeUpMode::DoubleTap)) {
@@ -503,7 +510,7 @@ void SystemTask::OnTouchEvent() {
 }
 
 void SystemTask::PushMessage(System::Messages msg) {
-  if (msg == Messages::GoToSleep && !doNotGoToSleep) {
+  if (msg == Messages::GoToSleep && !doNotGoToSleep && !settingsController.GetAlwaysOnDisplay()) {
     state = SystemTaskState::GoingToSleep;
   }
 
--- a/src/components/brightness/BrightnessController.cpp
+++ b/src/components/brightness/BrightnessController.cpp
@@ -2,13 +2,118 @@
 #include <hal/nrf_gpio.h>
 #include "displayapp/screens/Symbols.h"
 #include "drivers/PinMap.h"
+#include <libraries/delay/nrf_delay.h>
 using namespace Pinetime::Controllers;
+
+namespace {
+  // reinterpret_cast is not constexpr so this is the best we can do
+  static NRF_RTC_Type* const RTC = reinterpret_cast<NRF_RTC_Type*>(NRF_RTC2_BASE);
+}
 
 void BrightnessController::Init() {
   nrf_gpio_cfg_output(PinMap::LcdBacklightLow);
   nrf_gpio_cfg_output(PinMap::LcdBacklightMedium);
   nrf_gpio_cfg_output(PinMap::LcdBacklightHigh);
+
+  nrf_gpio_pin_clear(PinMap::LcdBacklightLow);
+  nrf_gpio_pin_clear(PinMap::LcdBacklightMedium);
+  nrf_gpio_pin_clear(PinMap::LcdBacklightHigh);
+
+  static_assert(timerFrequency == 32768, "Change the prescaler below");
+  RTC->PRESCALER = 0;
+  // CC1 switches the backlight on (pin transitions from high to low) and resets the counter to 0
+  RTC->CC[1] = timerPeriod;
+  // Enable compare events for CC0,CC1
+  RTC->EVTEN = 0b0000'0000'0000'0011'0000'0000'0000'0000;
+  // Disable all interrupts
+  RTC->INTENCLR = 0b0000'0000'0000'1111'0000'0000'0000'0011;
   Set(level);
+}
+
+void BrightnessController::ApplyBrightness(uint16_t rawBrightness) {
+  // The classic off, low, medium, high brightnesses are at {0, timerPeriod, timerPeriod*2, timerPeriod*3}
+  // These brightness levels do not use PWM: they only set/clear the corresponding pins
+  // Any brightness level between the above levels is achieved with efficient RTC based PWM on the next pin up
+  // E.g 2.5*timerPeriod corresponds to medium brightness with 50% PWM on the high pin
+  // Note: Raw brightness does not necessarily correspond to a linear perceived brightness
+
+  uint8_t pin;
+  if (rawBrightness > 2 * timerPeriod) {
+    rawBrightness -= 2 * timerPeriod;
+    pin = PinMap::LcdBacklightHigh;
+  } else if (rawBrightness > timerPeriod) {
+    rawBrightness -= timerPeriod;
+    pin = PinMap::LcdBacklightMedium;
+  } else {
+    pin = PinMap::LcdBacklightLow;
+  }
+  if (rawBrightness == timerPeriod || rawBrightness == 0) {
+    if (lastPin != UNSET) {
+      RTC->TASKS_STOP = 1;
+      nrf_delay_us(rtcStopTime);
+      nrf_ppi_channel_disable(ppiBacklightOff);
+      nrf_ppi_channel_disable(ppiBacklightOn);
+      nrfx_gpiote_out_uninit(lastPin);
+      nrf_gpio_cfg_output(lastPin);
+    }
+    lastPin = UNSET;
+    if (rawBrightness == 0) {
+      nrf_gpio_pin_set(pin);
+    } else {
+      nrf_gpio_pin_clear(pin);
+    }
+  } else {
+    // If the pin on which we are doing PWM is changing
+    // Disable old PWM channel (if exists) and set up new one
+    if (lastPin != pin) {
+      if (lastPin != UNSET) {
+        RTC->TASKS_STOP = 1;
+        nrf_delay_us(rtcStopTime);
+        nrf_ppi_channel_disable(ppiBacklightOff);
+        nrf_ppi_channel_disable(ppiBacklightOn);
+        nrfx_gpiote_out_uninit(lastPin);
+        nrf_gpio_cfg_output(lastPin);
+      }
+      nrfx_gpiote_out_config_t gpioteCfg = {.action = NRF_GPIOTE_POLARITY_TOGGLE,
+                                            .init_state = NRF_GPIOTE_INITIAL_VALUE_LOW,
+                                            .task_pin = true};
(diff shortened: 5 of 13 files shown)


JF002:
This will probably reduce the battery life to just a few hours. The LCD uses between 10 and 30mA when it's ON, which is quite high compared to the current power usage of InfiniTime (less than 1mA).

I can't think of any way to reduce power usage of the display when it's on... Lowering the brightness will effectively reduce the power usage, but probably not enough to maintain a reasonable battery life. This is caused by the LCD technology of the display.

KaffeinatedKat:
is there anyway to reduce the brightness lower than the lowest setting right now? Currently on the lowest brightness setting it's gone ~15 hours and it's at ~35%, which was better than I was expecting

JF002:
You could drive the brightness using PWM. There might already be a PR for that.

mark9064:
> You could drive the brightness using PWM. There might already be a PR for that.

Indeed there is, see #575

KaffeinatedKat:
I've changed it so that the "always on" setting prevents the display from turning off, but the watch still goes into the sleep mode, and just sets it to a low brightness with pwm. Now the display will light up when any of the wakeup actions happen, then go dim again after the display timeout. 

My only issue is the watch is in sleep mode and the watchface does not update while sleeping. Because the display is still on the watchface is just frozen till woken up. 
I've been poking around the code and I cannot find where I can make the display update while sleeping

A video of the watchface not updating while asleep:
https://drive.google.com/file/d/1pNt-cbVS1s_8tmAK6B8ZmDR_WU164OI8/view

mark9064:
A common pattern you'll spot all across InfiniTime is using queue read timeout to implement period tasks. Many tasks have an event queue that they read from forever, and when fetching an event from a queue you can specify a timeout, which is how long to wait if nothing arrives. After this timeout, the queue fetch function returns that no items were fetched from the queue as nothing was available in the time specified. So if you have a loop which does:
1. get event with timeout 1 minute
1. process event if one exists
1. execute x task

then x task runs at least every minute.

This is exactly how it's implemented for the display, with x task in this case being refreshing display contents. So change
```c++
    case States::Idle:
        queueTimeout = portMAX_DELAY;
      break;
```
to
```c++
    case States::Idle:
      if (settingsController.GetAlwaysOnDisplay()) {
        queueTimeout = lv_task_handler();
      } else {
        queueTimeout = portMAX_DELAY;
      }
      break;
```
 and you're good :)

mark9064:
Seems to work well with the above change. I haven't been running it long enough to test battery yet though. One suggestion would be to disable always on if InfiniTime is set to sleep mode

KaffeinatedKat:
> Seems to work well with the above change. I haven't been running it long enough to test battery yet though. One suggestion would be to disable always on if InfiniTime is set to sleep mode

This is something I plan to implement, this feature would pair really nicely with #1461

KaffeinatedKat:
> A common pattern you'll spot all across InfiniTime is using queue read timeout to implement periodic tasks. Many tasks have an event queue that they read from forever, and when fetching an event from a queue you can specify a timeout, which is how long to wait if nothing arrives. After this timeout, the queue fetch function returns that no items were fetched from the queue as nothing was available in the time specified. So if you have a loop which does:
> 
>     1. get event with timeout 1 minute
> 
>     2. process event if one exists
> 
>     3. execute x task
> 
> 
> then x task runs at least every minute.
> 
> This is exactly how it's implemented for the display, with x task in this case being refreshing display contents. So [change](https://github.com/InfiniTimeOrg/InfiniTime/blob/3a7dfdba8e5da4202155fef53673240fa0f22f5a/src/displayapp/DisplayApp.cpp#L187)
> 
> ```c++
>     case States::Idle:
>       queueTimeout = portMAX_DELAY;
>       break;
> ```
> 
> to
> 
> ```c++
>     case States::Idle:
>       if (settingsController.GetAlwaysOnDisplay()) {
>         queueTimeout = lv_task_handler(); // returns time until LVGL tasks need running
>       } else {
>         queueTimeout = portMAX_DELAY;
>       }
>       break;
> ```
> 
> and you're good :)

I have applied this, and it doesn't work. Screen still doesn't update while sleeping and when I wake it up the colors invert, which is extremely weird. Any idea why?

KaffeinatedKat:
> I have applied this, and it doesn't work. Screen still doesn't update while sleeping and when I wake it up the colors invert, which is extremely weird. Any idea why?

I have solved this issue, this change worked fine in the simulator but did not work on the watch. Figured out that by not putting the SPI to sleep, these changes work as intended

mark9064:
Nice job, I'm running a patchset that disables SPI sleep and I suspected that might be difference - unfortunately I didn't have time to test and report back so sorry about that

One change I'd suggest: running the LVGL tasks full speed (ie max display refresh) is pretty power hungry, and I've found throttling it to 250ms saves power while keeping watchfaces with seconds up to date (as the display effectively refreshes at 4Hz)
I also thought for correctness it might be good to pull up the screen running check. It also might be good not to reinitialise the SPI inside GoToRunning if it wasn't actually suspended 

So I've been running this - feel free to integrate changes if you fancy
```c++
      if (settingsController.GetAlwaysOnDisplay()) {
        if (!currentScreen->IsRunning()) {
          LoadPreviousScreen();
        }
        int lvglWaitTime = lv_task_handler();
        // while in always on mode, throttle LVGL events to 4Hz
        queueTimeout = std::max(lvglWaitTime, 250);
      } else {
        queueTimeout = portMAX_DELAY;
      }
      break;
```

I've also noticed an issue where sometimes the brightness is on Low rather than Lowest while the watch is asleep - I haven't had time to debug that either but see if you spot it. For reference I have my normal brightness set to Low

src/components/settings/Settings.h
```diff
@@ -184,6 +184,14 @@ namespace Pinetime {
...
           settingsChanged = true;
         }
+	// Disable always on screen while sleep mode is enabled
+	if (settings.alwaysOnDisplay.enabled) {
+	  if (status == Notification::Sleep) {
+	    settings.alwaysOnDisplay.state = false;
+	  } else {
+            settings.alwaysOnDisplay.state = true;
+	  }
+	}
```
KaffeinatedKat, line 194: i'm unsure if this code belongs here, or if we should check the notification status before sleeping instead

escoand:
Wouldn't be an update every minute enough for an always on display? It should increase the battery runtime further.

mark9064:
> Wouldn't be an update every minute enough for an always on display? It should increase the battery runtime further.
> 
> Off course with hiding the seconds hand...

Yeah it's definitely something worth thinking about. On one hand it saves power, but on the other, it allows incorrect information to be onscreen for more than a minute like seconds on watchfaces with them, seconds on the timer app, bluetooth status, music status, steps, heart rate (if continuous/background measuring implemented) etc
Whether these are important enough for the power saving, I'm not sure. Maybe it's worth testing to see how big the power difference is?

KaffeinatedKat:
> > Wouldn't be an update every minute enough for an always on display? It should increase the battery runtime further.
> > Off course with hiding the seconds hand...
> 
> Yeah it's definitely something worth thinking about. On one hand it saves power, but on the other, it allows incorrect information to be onscreen for more than a minute like seconds on watchfaces with them, seconds on the timer app, bluetooth status, music status, steps, heart rate (if continuous/background measuring implemented) etc Whether these are important enough for the power saving, I'm not sure. Maybe it's worth testing to see how big the power difference is?

This might be worth it, we can set the update timer to 4Hz when apps like the timer and music are running, but set it back when not running these apps; perhaps an enum with apps that need a higher refresh timer. We can also disable the seconds on every watchface that has them when always on mode. We could also add a setting to disable this if someone wants seconds on. I think with the display the pinetime has, any power savings we can get is worth it

hexisXz:
> This will probably reduce the battery life to just a few hours. The LCD uses between 10 and 30mA when it's ON, which is quite high compared to the current power usage of InfiniTime (less than 1mA).
> 
> I can't think of any way to reduce power usage of the display when it's on... Lowering the brightness will effectively reduce the power usage, but probably not enough to maintain a reasonable battery life. This is caused by the LCD technology of the display. 

I honestly think this would be pretty useful. You could just have a low quality version of the digital clock display on the screen with like medium brightness and just have everything else suspended in the background.

KaffeinatedKat:
This pr, for the most part, is complete. Only remaining issue is sometimes the display does not go into the lowest setting when going to sleep, and stays at the brightness setting. I have not been able to figure this out because it happens so infrequently.

@JF002, is there any way this feature would be considered for merging, or is the battery life reduction while enabled to high? Maybe there could be some kind of popup warning when enabled, to notify the user of the significant battery drain while enabled. 

It lasts about a full day when enabled, and most smartwatches only last that long anyways. I feel like the people who genuinely find this useful would be willing to charge it everyday

mark9064:
The issues with the screen getting stuck in Low are due to the restorebrightness messages coming from the system task.

I think this PR (with the above issue fixed) is usable. But we should also test what the majority of the power consumption is always on is (I suspect it's the LCD) and if we can reduce it further (the LCD offers a power mode between sleep (display blank) and fully on with reduced colours etc.. Also might be possible to redrive the LCD at a lower refresh rate when in always on, but the datasheet for it on the Pine wiki is an absolute monster and I haven't had the time to fully explore it. As an aside it looks like the solution to the display invert issue might be in there too

KaffeinatedKat:
I took a peek at the datasheet, and implementing switching to reduced colors while always on. The datasheet says this uses less power, but by how much I am unsure. I don't have a devkit, and can't measure power consumption. 

This is also essentially free power reduction. Turns out, InfiniTime doesn't utility many colors at all, and while in reduced colors, you can hardly tell. The reduced brightness also helps hide it. The one app where you can really tell is 2048, and I doubt anybody will be letting it fall asleep on 2048 anyways.

I'm gunna keep poking around the datasheet to see if I can figure out the refresh rate thing, if that is possible it would probably help the power consumption by a sizeable margin

mark9064:
Oh yeah also forgot to mention, the settings version number should be bumped as a new member has been added

mark9064:
https://github.com/InfiniTimeOrg/InfiniTime/pull/1869#issuecomment-1762923811

The patch in question: https://github.com/mark9064/InfiniTime/commit/d43759fcc817be8b1daaadc7bd02f5353be06e47

Locally working well - to reproduce the issue on the original try disconnecting and reconnecting BT (the display should get stuck on after). With the patch it should stay in always on as expected

KaffeinatedKat:
Applied your patch, and it has fixed the issue on my device as well. Bumped the settings version, while I was at it

Also figured out how to modify the refresh rate, it now goes to ~4.8hz while always on, which is the lowest setting available. Hopefully this helps increase the battery life

mark9064:
+1 on the reduced colours thought

Can confirm refresh rate changes have applied properly, well done on figuring that out! I think it should actually be possible to inform LVGL of the exact refresh rate too and that way we shouldn't have to do the timer throttling hack? I'm not totally sure on this, I'd need to dig through the LVGL internals but it might be worth considering

mark9064:
I had a quick look: it seems that `LV_DISP_DEF_REFR_PERIOD` is used as a define to set the refresh period on all the LVGL tasks to 20ms. So I think what we'd have to do is go through all of these tasks (there are many inside InfiniTime and a couple in LVGL itself), and then use `lv_task_set_period` to set them to a longer execution period. Quite a bit of work, but looks possible

KaffeinatedKat:
>Perhaps it would be better to properly understand exactly what lv_task_handler does and see if we should implement our own task handler that runs tasks with efficiency in mind rather than latency (ie batching task execution). What are your thoughts?
I think that this is a good idea, would this new task handler replace lv_task_handler all together, or only be used for the always on display?

mark9064:
Probably just for always on as I suppose we still want responsive inputs etc while on. But I guess it depends on how it's implemented? It might even not be needed at all - I'm not sure how the LVGL scheduler works

A quick list of things I've noticed so far:

- Screen tearing is quite visible. Try with the analogue face, notice how the second hand gets chopped in half occasionally. Also, the lcd pulses with the refresh rate (pretty tricky to spot but try looking at the top right). Maybe the LCD is being driven too slowly? I'm honestly no expert on what refresh rates the panel supports. I think I might try keeping 60Hz but using /8 divider on that and seeing if it does better - will report back. There also might be some VSYNC stuff we can do with the panel to fix this
- Frame pacing is noticeably poor (visible with seconds) but honestly I think that's fine as a sacrifice if it gets battery life back
- There are still some wake/sleep state bugs I need to investigate (display gets stuck on in some scenarios)
- A background transition to the Sleep mode (eg with the auto sleep PR) causes everything to break

Battery life is looking promising though! Not sure if it's the reduced refresh or reduced colours but I'm seeing an improvement for sure

KaffeinatedKat:
>Screen tearing is quite visible

I've noticed the screen tearing, it's really bad if you run it at that refresh rate while awake. I don't personally run a watchface with seconds and so I haven't noticed it much, but it's for sure worth investigating for those who do. I'll do some tinkering with the refresh rate and see about VSYNC

>A background transition to the Sleep mode (eg with the auto sleep PR) causes everything to break

I have encountered this, stopping the spi from sleeping altogether fixes the crashing, and then I just call "GoToRunning" to fix the screen state being wrong afterwards (see [e6d60f1](https://github.com/KaffeinatedKat/InfiniTime/commit/e6d60f1f3e59b1f21c2b32f09fe72c6721d0b8a9)). We should find a way to fix this bug without the autosleep pr however, i'll keep you updated on that

>There are still some wake/sleep state bugs I need to investigate (display gets stuck on in some scenarios)

I have not encountered any of these, if you could provide some scenarios that cause this, I can help debug it

mark9064:
Coming in on 48h runtime now with 20% still in the tank, very impressive! I'd say it's a lot more usable with those changes, awesome job

Another thought I had for power usage is driving the LCD at a slightly lower voltage. From what I can see the LCD driver allows choosing what drive voltage is used for the LCD, so maybe a slightly lower voltage could save some power? It also might cause the LCD to not turn pixels on/off properly so not sure if this is a viable route (and without a power profiler I'm not even sure if the panel draws much power or if it's mostly the driver chip). Also there appears to be about 10 driver voltages :grimacing: so yeah sounds fun 

One other thing: some of the comments from the original PWM PR should probably be addressed. I think the points about not running the PWM engine when the screen is off or at one of the fixed brightness modes has merit, with PWM only being used between brightness settings. I think it also makes sense to PWM just one pin at a time, though I can see the formulas for calculating brightness being annoying as the brightness increase per duty increase is different for each pin
Ideally I think this should take the form of a unified brightness interface where it takes some target brightness value between 0-10000 (or whatever) and the brightness code calculates what pin configuration is needed along with whether PWM is required? I could have a crack at implementing this, what do you reckon?

KaffeinatedKat:
I can go ahead and mess with the voltage stuff, still messing around with the refresh stuff to reduce screen tearing, might as well mess with the voltage while i'm at it. Really glad to see how much potential there is for greater battery life with the display being on all the time, considering it's just a regular LCD screen

About the PWM stuff, I honestly have no idea how it works, but if you wanna take a crack at implementing the unified brightness interface, I would happy merge that into the branch. Sounds like it would be a really nice addition to improving the PWM implementation.

KaffeinatedKat:
I have changed all the voltage settings to their lowest and have been running it for a while and everything seems fine, the display still updates normally and I can't notice any pixels failing to turn on/off. Best case scenario, we save a bit of power, worst case, it does nothing. Really wish I had a devkit and power analysis tools to see if this makes any difference

src/drivers/St7789.cpp
```diff
@@ -119,21 +129,30 @@ void St7789::FrameRateLow() {
...
 }
 
 void St7789::DisplayOn() {
   WriteCommand(static_cast<uint8_t>(Commands::DisplayOn));
 }
 
+void St7789::PowerControl() {
+  WriteCommand(static_cast<uint8_t>(Commands::PowerControl1));
+  WriteData(0xa4);
+  WriteData(0x30);
```
mark9064, line 150: Shouldn't this be 00?
30 is 0011 0000
So AVDD is 0 = 6.4V
AVCL is 3 = -5.0V (0 would give -4.4V)
VDS is 0 = 2.19V

KaffeinatedKat: I would assume that -5.0v would use less power than -4.4v, correct me if i'm wrong

mark9064: Usually a potential difference of 0 volts means no current, so the closer to zero the lower the current should be AFAICT

KaffeinatedKat:
>Frame pacing is noticeably poor

This has really been bothering me. The delay that the DisplayApp runs at is not consistent, and increasing the delay it runs at creates really inconsistent draw times. The longer you make the delay, the worse it gets.

Normally, the DisplayApp updates so often that this is a non-issue for watchfaces with seconds. It grabs the current time so fast and redraws the screen every time, so it's always up to date. When you increase this delay, it only grabs the current time every so often, and because the timing is inconsistent, this leads to the seconds skipping, or moving between seconds quickly every so often.

I don't know why this is, and i'm wondering if there is some kind of workaround to make the DisplayApp update at a very consistent rate

mark9064:
I've poked around too, and in fact I'm pretty sure it's due to the desynchronisation between displayapp and the screen refresh
Adaptive sync on desktop PCs works by refreshing at full scanout speed and then waiting for the next refresh cycle. At maximum refresh rate, the wait is zero, and it becomes longer as the input refresh rate decreases below the monitor refresh rate
I have discovered that the pinetime LCD controller does not work like this. The scanout slows as the refresh rate drops. I find the scanout to be quite visible in the 2s app when in always on mode, look for a sweeping bar going top to bottom. If you can't spot it try a darker room

I think this can be resolved by synchronising the LVGL refresh rate with the display refresh. Ideally, we'd using VSYNC. I think that may be impossible as the pin for VSYNC is not connected as far as I know, and also it disables features like hardware scrolling. Instead the displayApp loop must be changed to be a time-locked loop that runs lv_task_handler exactly at at the display refresh interval. Time locked loops are kinda tricky (I've implemented one for the heart rate sensor as part of some work I've done) as you have to deal with things like system tick count wraparound, being CPU starved, and avoiding overflow and imprecision when calculating the required sleep time. I can definitely take a look at implementing this.

This still isn't a clean win as there will be a tear line at the same place every refresh. But the pacing should be better?

mark9064:
Oh yeah one other issue I've noticed, after leaving the flashlight app brightness is not restored.
Quick todo
- Flashlight brightness bug
- Refactor PWM code (me)
- Refactor display refresh timing (me)
- Test if each change we've made actually affects power draw (someone with a power profiler)
- Write relevant patch for always on in InfiniSim
- Review the many other driver voltages
  - Gate control (B7) - controls voltage sent to LCD panel
  - VCOMS/VDV/VRH - I think these are gamma reference voltages? In this case we wouldn't want to touch them

KaffeinatedKat:
The patch for InfiniSim is really easy, it won't compile because it implements it's own BrightnessController which doesn't have the `Lowest` setting in the enum. I'll go ahead and open a PR for that. I'll allso check out the other voltage stuff

Another issue i've found is sometimes the LCD turns off after waking up, which stems from the refresh rate changing. I am pretty sure this is because the delay after setting it is not quite right. I'm currently testing out different delay times to get rid of this issue.

I can also go ahead and check out the flashlight bug you mentioned

src/components/brightness/BrightnessController.h
```diff
@@ -20,6 +20,9 @@ namespace Pinetime {
 
     private:
       Levels level = Levels::High;
+      uint16_t pwmSequence[1] = {10000};
```
NeroBurner, line 23: couldn't that just be a single value instead of an c-array?

KaffeinatedKat:
>Another issue i've found is sometimes the LCD turns off after waking up

After a day of testing, i've figured out when this issue occurs. If you are running the analog watch face with the always on display, sometimes when you wake up/put to sleep the watch, the screen blanks out till you reboot. This only happens on the analog watch face. Every other watch face works, all the apps that allow sleeping also work. As for why this issue occurs, I have no idea.

src/components/brightness/BrightnessController.h
mark9064: probably :sweat_smile: this was pulled from the previous PWM backlight PR and I'm not sure anyone here has figured out exactly how it's designed yet. Is on my todo list though

mark9064:
Huh, interesting. I'll see if I can figure out why

KaffeinatedKat:
>Looks like the backlight is coming on fine, but nothing from the LCD

Yea, calling SoftwareReset doesn't bring back the LCD after this happens either, which is quite strange. I'm going to attempt to figure out what is happening if you wanna focus on the other stuff instead, you have a bit on your plate already

mark9064:
No exciting progress yet but
- Time locked (constant frequency) loop is implemented for display refresh
  - With no improvement to pacing
  - Figuring out why currently
- I've verified the LVGL scheduler is a batch scheduler and therefore there are not any changes for us to make

mark9064:
> If I remove the second hand on the analog watch face, the LCD going blank stops happening, ran it for a day and I never encountered it

Super weird, maybe there is some implementation issue with the analogue face? Other watchfaces seem to work fine. Could also be that we are violating some assumptions somewhere in the watchface code or LVGL. I doubt it's the display as it gets treated the same no matter the watchface - unless it's something to do with the render time of the watchface

mark9064:
Implemented gate control, you should just be able to hit merge on https://github.com/KaffeinatedKat/InfiniTime/pull/1

KaffeinatedKat:
>unless it's something to do with the render time of the watchface

This is my best guess. It calculates Sin and Cos (these are notoriously slow), and also it draws the hands with lines defined by 2 points. Depending on what kind of line drawing algorithm LVGL uses, this could also be quite slow. It does these every second. It could be that because of the time it takes to draw, it sends the image to the display at an inopportune time while the refresh rate is changing causing the display to freak out (overflow? race condition?). If this was the case, it would explain the inconsistency, and also why removing the calculation and drawing of the second hand prevents the issue

mark9064:
So I took a look at the fix for flashlight brightness, and I thought it could be simplified by making it flashlights job to restore the brightness. I think this works better as forcing the brightness to be set on every button press is restrictive to functionality and creates weirdness for when the flashlight app is left without the button being pressed (eg by notification).

Does the PR workflow work for you btw, it's definitely more convenient for me than copy pasting patches and keeps the blame accurate so I can be called out when everything catches fire https://github.com/KaffeinatedKat/InfiniTime/pull/2

KaffeinatedKat:
>and I thought it could be simplified by making it flashlights job to restore the brightness

This is much more elegant, also fixes an issue where if you wake up the watch and put it to sleep right away with the button it would go to sleep then set the brightness to normal. 

>Does the PR workflow work for you btw

Yes, much better than before. Much easier to test and integrate the changes on my end

KaffeinatedKat:
>It calculates Sin and Cos (these are notoriously slow)

It does not appear to be a rendering time issue, added a simple lookup table for Sin and Cos for the second hand, which eliminates the slowness of these functions, and the screen still blanks out randomly. Gunna have to take a look at how this is rendered. Perhaps it's trying to render something out of bounds, or something of the likes

FintasticMan:
Yeah, sin and cos are already implemented as lookup tables in LVGL.

mark9064:
RE: https://github.com/InfiniTimeOrg/InfiniTime/pull/1869#issuecomment-1776021839
The constant frequency frames work exactly as expected, I've verified this by keeping 60Hz while in always on mode as a test. I can't seem to find anything that displays frames stably, at least with the *8 divider. I guess I'll try some other dividers.

As a sidenote, the frame rate control command (B3) doesn't seem to work how the datasheet says. It says if FRES is clear the idle frame rate is defined by the normal mode frame rate, which in testing isn't true. It also says that the frame rate applies only to idle mode, which also isn't true. So I honestly have no clue what B3 does compared to C6? Did you figure anything out here

mark9064:
I've configured it to refresh at exactly 32Hz/4Hz(div 8) (using the formulas), but instead it's refreshing at 2Hz. No clue why, I'm certain I've got my calculations using the refresh rate formula correct. Anyway it works great and battery life seems better, frame pacing is solid. Just need to figure out what's actually happening refresh wise now

KaffeinatedKat:
Great to hear that, can't wait to merge those changes, good work so far. Interested to see if that happened to fix the analog watch face issue, as that could potentially have something to do with the inconsistent pacing

mark9064:
Still working on this (been busy with work)
Got a bit sidetracked with some changes to time handling, there's currently quite a bit of jitter which causes the 2Hz display to occasionally miss seconds. I want to take the time to refactor things properly, so I'm probably going to have to create a couple other PRs which this one will depend on (as I don't want to include too many unrelated changes). This branch can pull them early though, so it doesn't block dev here

Re the disappearing display issue on wake: I tried navigating to the quick settings menu and switching to sleep mode while the display was blanked (as sleep mode disables AOD). Took a few tries to hit it as I was blind, but switching off and back on again (just with power button) after that brought the display to life (though with inverted colours). Curious to hear if this hack works for you

I've also tightened down all of the AOD display command timings with no ill effects, so the watch is much more responsive now when it comes to leaving/entering AOD. Will try get patches up soon but my branch is a bit of a mess lol

KaffeinatedKat:
> Took a few tries to hit it as I was blind, but switching off and back on again (just with power button) after that brought the display to life (though with inverted colours). Curious to hear if this hack works for you

This does bring back the display for me, colors do not always invert though

KaffeinatedKat:
I merged the new 1.14 changes, and I was hoping that the new changes to how the analog face is drawn would fix the display blank, but unfortunately it did not. Still stuck with no second hand with the aod. @mark9064 Looking forward to seeing your recent changes on your branch, i'd love to help clean it up if you wanted to open a pull request, got a lot of time on my hands

mark9064:
I've been thinking a lot about our usage of PWM, which adds a non-negotiable 150uA consumption. We should be able to replace this with a timer and PPI to the GPIOTE which essentially emulates PWM for us without having the power hungry PWM clock. This should get us down to 5uA. We can also stop the timer when modulation isn't needed, but as a more general approach this should allow efficient backlight modulation

Gonna have a shot at implementing this soon. I'll try get the 2Hz backlight changes up

mark9064:
I've got this working. Been working on this every day, absolute nightmare as the previous SPI erratum 58 workaround was silently overwriting the events - but I've fixed that now so we're good to go. Haven't had a chance to evaluate power usage yet. The refactor only uses TIMER/PPI when needed and only ever controls one pin, so hopefully current consumption should be reduced. It also only re-initialises the channels when switching output pin so calling setBrightness() every frame to create a smooth brightness fade should be performant.

Will push it when I have a chance tomorrow. After that we have just the analogue issue to resolve (confirmed not to be fixed by these changes) and this branch should be good to go after a rebase (to get the commits into a sensible order!)

KaffeinatedKat:
That's great to hear, can't wait to see these changes with a hopefully even better battery life.

Some info on the analog display blank:
I've been messing with the analog face and i've found that it's not the second hand that causes the issue.

If I remove all the if statements from `WatchFaceAnalog::UpdateClock()` so the hands are redrawn every time it's called, then remove the part that does the second hand, the issues still occurs. 

Currently looking for a way to disable the watchface from redrawing the hands while changing sleep states, i'll let you know if I get anywhere

KaffeinatedKat:
i've added you as a collaborator to the repo, i'm not super well versed in git and i'm not sure how to accomplish this

mark9064:
Oh thanks! That'll work. I'll get it rebased, but I should explain what I actually mean first. So when you merge main into this branch, you fundamentally take all the commits on main and append them to this branch. But when you rebase, you take all the commits on the branch, and then append them to a fresh copy the new up to date main branch. So this way you don't have any merge commits in the history, you have only the new commits you added.

To achieve this you run `git rebase <branchname>`. So for main that's `git rebase main` (make sure to update your local main branch first!). But now you have a problem: the branch on the remote and the local branch you've just created from the rebase now have different versions of history. On one of them, the commits you wrote were added at some point in the past, and on the rebased one they were added right on the tip of the main branch. Since you want to keep the history where the commits were added to the fresh main, you have to force push and overwrite what's on the remote. This can cause problems for other contributors as they won't be able to `git pull` due to the diverging histories on the branch (they will have to reset their branches, and copy back over changes they've made which is super painful).

So common practice is only to rebase a branch if you're the only one working on it, because it causes huge problems for others with these diverging histories. The best way to avoid all these problems though? Just don't add new changes from main into the branch being worked on - only merge things in if needed (if you want to try the PR branch as if it were on the latest main, create a local copy of the PR branch and rebase that local copy onto main). This keeps the history clean, and when the PR is merged into the main branch Git automatically figures out that the new commits need to go on the end of main.

--

Now I've written all that, I've realised that rebasing it will break your stuff so I'll just work around it locally, it's not too big of a deal actually. Hey maybe you got something out of this essay at least!

KaffeinatedKat:
thanks for the explanation, I assumed keeping it up to date with main was important but when explained like that it makes sense why you would wait to merge with the main branch assuming there is no major changes to adjust for

mark9064:
PR is up! https://github.com/KaffeinatedKat/InfiniTime/pull/3

src/components/brightness/BrightnessController.h
```diff
@@ -20,6 +24,17 @@ namespace Pinetime {
...
     private:
       Levels level = Levels::High;
+      static constexpr uint8_t UNSET = UINT8_MAX;
+      uint8_t lastPin = UNSET;
+      nrfx_timer_t timer;
+      // warning: nimble reserves some ppis
+      // https://github.com/InfiniTimeOrg/InfiniTime/blob/034d83fe6baf1ab3875a34f8cee387e24410a824/src/libs/mynewt-nimble/nimble/drivers/nrf52/src/ble_phy.c#L53
+      // spimaster uses ppi 0 for an erratum workaround
+      // channel 1, 2 should be free to use
+      nrf_ppi_channel_t ppi_backlight_on = NRF_PPI_CHANNEL1;
```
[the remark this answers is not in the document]
minacode: Let's try from my and the [internets](https://en.cppreference.com/w/cpp/language/constexpr) understanding:

`static` means that this will not be bound to a specific instance of the class. Sounds good for a general pin config.

`constexpr` let the compiler evaluate this at compile-time, so there is no runtime overhead in the resulting image. In classes, this also implies `const`, meaning that it cannot be changed later at runtime. And in `static` declarations it even implies `inline`, so the value will be copied everywhere at compile time instead of loading it from the variable. 

`static constexpr` is fine here, if this value will never change and you want to use it everywhere with as less runtime overhead as possible.

I don't know exactly what the code does, but that seems to be the case here.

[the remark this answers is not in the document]
mark9064: Thanks for the explainer, looks like static constexpr is appropriate here then :+1:

In this case the ppi_backlight_on/off are constants designating the channel to be used for the backlight on/off triggers

mark9064:
Fixed the above two issues, minor changes so hopefully you don't mind me pushing directly (I'll always PR for actual work so you can review)

KaffeinatedKat:
The analog display blank issue is mysterious, and I honestly think the best solution here is to just keep the second hand disabled while the aod is on (as it is now). This issue only happens when waking up. 

I've tried adding in delays, stopping the watchface from updating while waking up, making sure the hands aren't being drawn outside of the screen, removing the call to wakup to spiNorFlash (i'm just throwing shit at the wall and seeing if anything sticks at this point). 

Unless anyone has any ideas on why this issue is happening, removing the second hand appears to be the best solution

mark9064:
There's certainly some deeper display issue here. At the moment I'm reading through all the SPI code to try and get an idea of the exact issue. I suspect it's behind the display invert problem, and also behind the bug where the button text on the timer has a ~10% chance of becoming a bit corrupted when the timer wakes the watch from AOD (set the timer to a few seconds and use lower to sleep to get fast repeats)

I can trigger display inversion without even needing AOD on the analogue face (just using sleep notifications mode to disable it and spamming the button). I've only just discovered this today - I need to test if this happens on the current main

mark9064:
Normal analogue face bug not reproducible on main so for sure there are some SPI race conditions or some overtightened timings - not sure which. Continuing to investigate

KaffeinatedKat:
I think that https://github.com/InfiniTimeOrg/InfiniTime/issues/1984#issuecomment-1919625745 this gives us a little bit more insight on what is happening here. The analog issue only happens when turning the screen on, and this issue happens when turning the screen on. Both of these actions draw to the screen. I think the issue might lie with drawing the screen and also modifying the state of the screen at the same time, causing some kind of race condition and bits are written to the wrong place, so the display freaks out and either turns off or inverts the colors

This whole SPI thing flies right over my head, but i'm going to attempt to learn how it all works so we can hopefully figure out this nasty bug a bit faster. Let me know if you have any more updates on it

mark9064:
Yep, I've arrived at the same conclusion!

My current theory is that the display switches from panel data write mode to command mode too early and interprets the image data as commands somehow. Not sure how close I am though. I think it corroborates with the analogue face issues as well, as the analogue face creates the largest screen paints regularly (repainting the second hand requires changing a good portion of the screen)

FintasticMan:
Thank you people for your hard work in implementing this very useful feature and squashing bugs! I'm sorry I've not been very involved with the PR. When this is ready for review, could you organise the commits into standalone changes, as right now it's quite a lot to try to review in one go 😅. I'm looking forward to using the always on display!

mark9064:
Yeah no worries, we are still trying to get the last display bugs ironed out then after that it can be rebased into something more sensible. Right now it's definitely a bit chaotic

mark9064:
Oh, another issue I just remembered. If always on is changed during sleep notification mode it will enter aod as opposed to turning the display off as usual

KaffeinatedKat:
the always on setting is now correctly displayed as enabled while in notification sleep, and changing it while in notification sleep works correctly

mark9064:
I *think* I might've fixed the display issue. Fix is currently very hacky and needs a refactor, also not 100% sure I've got everything sorted here either (perhaps I just have good/bad luck with triggering the issue right now). Will be a while until I can work on it again as I'm still catching up with things after FOSDEM and I'm also away this weekend too. But hopefully soon...

mark9064:
> the always on setting is now correctly displayed as enabled while in notification sleep, and changing it while in notification sleep works correctly

I might've messed up a local merge but this isn't working quite right for me. If I switch on always on display while in sleep mode, always on will activate in sleep until I toggle notifications to normal and back to sleep again

KaffeinatedKat:
Best way to check for sure is to just run the aod with the analog face for a day or 2. If the issue is still there it'll eventually trigger

- John Crawford

-------- Original Message --------

mark9064:
Pretty confident the fix is good now. Should also clear up the random display inverts people get. Going to PR this in hopefully tomorrow after some code cleanup

KaffeinatedKat:
That's awesome to hear, can't wait to have that fixed. What was the cause of the issue? Very interested to know

- John Crawford

-------- Original Message --------

mark9064:
The problem was that the LCD can be switched to command mode (from data) while the previous SPI transaction was still ongoing, so the pixel data got interpreted as commands. I didn't even know the pin to switch between command mode and data mode existed before investigating this!

KaffeinatedKat:
> > the always on setting is now correctly displayed as enabled while in notification sleep, and changing it while in notification sleep works correctly
> 
> I might've messed up a local merge but this isn't working quite right for me. If I switch on always on display while in sleep mode, always on will activate in sleep until I toggle notifications to normal and back to sleep again

I have actually fixed this now. The previous fix changed the UI toggle to be displayed properly, and disabling aod while in notification sleep. You can now enable the aod setting while in notification sleep and the display will turn off as expected till you leave notification sleep

mark9064:
OK, with https://github.com/KaffeinatedKat/InfiniTime/pull/4 this PR is ready functionally
It won't build right now because it needs the removal of GFX which just got merged into main

Hopefully https://github.com/KaffeinatedKat/InfiniTime/pull/4 looks OK. If so, I'll start rebasing this branch onto main so it can be reviewed properly

mark9064:
Quick list of other issues this happens to resolve
- Display inversion on some watchfaces (SPI race condition fixed)
- Flashlight brightness inconsistency on flashlight exit
- Current time caching (current time is now always fetched fresh from RTC)
  - Paves way to reduce system task frequency when not polling accel/other hw. Bit of a niche gain though
- Off by one error in timekeeping when the RTC overflows
- Workaround for errata 58 which overwrote PPI/GPIOTE channels implicitly
- Spinning the CPU when waiting for the display to process commands
- Waiting unnecessarily long for the display to execute commands
- Weirdness of the RestoreBrightness message (the brightness restoring part has been removed and now this functions only as a message to notify of device activity, renamed as so)

KaffeinatedKat:
awesome stuff man, good job figuring that out and fixing it. PR looks good to me, you can go ahead and rebase everything so this can be reviewed and hopefully merged

SteveAmor:
@mark9064, would it be possible to break out the code for the SPI fix into a separate PR, therefore creating a small, focused pull requests that fulfills a single purpose? This will help with tracking.

KaffeinatedKat:
> @mark9064, would it be possible to break out the code for the SPI fix into a separate PR, therefore creating a small, focused pull requests that fulfills a single purpose? This will help with tracking.

This is probably a good idea, these SPI problems are a problem outside of this PR, this just exposed the issue enough that it needed to be resolved

SteveAmor:
> > @mark9064, would it be possible to break out the code for the SPI fix into a separate PR, therefore creating a small, focused pull requests that fulfills a single purpose? This will help with tracking.
> 
> This is probably a good idea, these SPI problems are a problem outside of this PR, this just exposed the issue enough that it needed to be resolved

Absolutely fantastic work here 😀

mark9064:
Rebasing is in progress, on to main now and down to 32 commits. When it's in shape I'll repush it, and then I think we can discuss what changes we might want to break out

JF002:
I haven't had the opportunity to have a closer look at this PR, but I'm definitely interested in all the work that have been done here!

Splitting the PR into multiple PRs will probably help with the reviews since we'll be able to focus on a single topic with less code involved. Do you think it would be possible to isolate a few of the points mentionned in https://github.com/InfiniTimeOrg/InfiniTime/pull/1869#issuecomment-1937937984 into smaller PRs? 

If this is too much work, maybe wait for me to have a closer look at this PR so we can find the best way to continue the review. I would not want you to do the extra work for nothing. I can't promise anything (I'm still recovering from a flu, it's taking more time than expected...), but I'll try to provide more feedback soon.

Another question : do you have any idea of the battery life with the always-on mode enabled?

mark9064:
Yeah, don't worry about reviewing this yet. I don't have a whole lot of free time right now so it'll take me a few days to reorder/combined/split things. I'll drop another message here when that's done.

After that we can discuss if we want to split parts out into other PRs (I think we will want to split a few)

Battery life is a bit more than 2 days with the watch in sleep mode for ~7h/day

mark9064:
Suggested PRs
- Continuous time updates (https://github.com/InfiniTimeOrg/InfiniTime/pull/1869/commits/7e178ccaa70c2e8b9b22c115527cfecb588b4430)
- Erratum 58 improvements (https://github.com/InfiniTimeOrg/InfiniTime/pull/1869/commits/0792c3dd8fc831ec190762e1c7a21368da9c9021)
- LCD delay improvements (https://github.com/InfiniTimeOrg/InfiniTime/pull/1869/commits/2438927b97d4b584f99d2090888ba382cc85d89a, https://github.com/InfiniTimeOrg/InfiniTime/pull/1869/commits/fb8eb479b29811c007f46cbad5e91a1962825cbc)
- SPI transaction hooks (https://github.com/InfiniTimeOrg/InfiniTime/pull/1869/commits/e5968882a717ad829546a21ef7dbb41cec053431, https://github.com/InfiniTimeOrg/InfiniTime/pull/1869/commits/c1b6ae1f0af93245c804e8fbcf35cc984d9b289b, https://github.com/InfiniTimeOrg/InfiniTime/pull/1869/commits/77327b4da89a18ab4d772d51575b2064e624decf)
- This PR (rest of the commits)

The commits above are all in order with respect to this PR

Rebase patch notes:
- Onto latest main
- All fixup commits etc squashed into originals
- Lowest brightness level renamed to AlwaysOn
- Time updates now protected with mutex for correctness
- Commit naming consistent
- Code comments reworked
- Branch compiles at every commit (tested) and all commits are code style compliant (to the best of my knowledge, but not double checked)

src/systemtask/SystemTask.cpp
```diff
@@ -323,7 +328,11 @@ void SystemTask::Work() {
             // if it's in sleep mode. Avoid bricked device by disabling sleep mode on these versions.
             spiNorFlash.Sleep();
```
mark9064, line 329: Sleeping the flash here should *technically* be fine as once a screen has initialised it doesn't tend to load any resources after (so the flash isn't needed). But the 12μA to keep it in standby is probably worth it, the LCD uses 100x this current

JF002:
I've just setup my devkit with my NRF PPKII (power measurement tool).

image

The PPK is connected between the watch and the battery, so it measures the whole power usage of the watch.

To get a baseline, I first measured the average current of the watch running InfiniTime 1.14

 - Max brightness : 23mA
 - Medium brightness : 13.5mA
 - Low brightness : 8mA
 - IDLE (BLE enabled, slow advertising (sleeping for more than 30s)) : 420µA

Now, using this branch without any change (on 7e178ccaa70c2e8b9b22c115527cfecb588b4430):

 - Max brightness  : 26mA
 - Medium brightness : 13mA
 - Low brightness : 7.75
 - Always on mode : 3.5mA

Here's the capture from the always mode :
image

More zoomed in:
image

The pattern repeats every 500ms : 130ms @ 1.92mA and then 370ms @ 4.02mA

Let me know what other measurements would be useful :)

KaffeinatedKat:
Is there any noticible difference between running the display at 8 colors vs full color while in AlwaysOn?

- John Crawford

-------- Original Message --------

JF002:
@KaffeinatedKat 

> Is there any noticible difference between running the display at 8 colors vs full color while in AlwaysOn?

How can I check this?

mark9064:
I've got a set of test cases I can hopefully send to you in an hour or so

mark9064:
[aod_testing.md](https://github.com/InfiniTimeOrg/InfiniTime/files/14396862/aod_testing.md)
Sorry for the delay. All the branch names refer to the ones on my repository, not this branch here. I spotted a rebase mistake meaning that the previous testing you did might not quite be accurate

https://github.com/mark9064/InfiniTime/branches

JF002:
Here are my measurements! For each branch/ref, I 
 - Rebuilt the whole application (remove the previous build folder and cmake/build again)
 - Flashed the firmware using my debugger
 - Power OFF the debugger
 - Power cycle the watch (to ensure that the debug circuitry is OFF)
 - Disable BLE
 - Measure ON/OFF/AOD power usage (averaged on 10s)

|Test case | Ref | Result |
|--------------|-------|----------|
| InfiniTime 1.14 | ecf2f56 | ON : 7.48mA - OFF :  397µA |
| Test case 1 | main: 6ab512a | 10.57mA |
| Test case 1 | aod: f84b619 | 10.58mA |
| Test case 2 | aod tip | ON : 7.42mA - AOD : 3.45mA - OFF : 398µA
| Test case 3 | aod tip | Digital : 3.45mA - Analog : 3.84mA - PTS : 3.45mA - Terminal : 3.49mA - Infineat : 3.45mA - G7110 : 3.45mA |
| Test case 4 | 4028304 | ON : 7.43mA - OFF : 398µA |
| Test case 4 | ac19f99 | ON : 7.00mA - OFF : 398µA |
| Test case 5 | 0dd662a | ON : 7.51mA - AOD : 3.89mA |
| Test case 5 | 81d25d2 | ON : 7.49mA - AOD : 3.5mA |
| Test case 6 | | ON : 7.14mA - AOD : 7.11mA |
| Test case 7 | | ON : 7.30mA - AOD : 6.27mA |
| Test case 8 | fd92ff3 | Digital ON : 7.38mA - Digital OFF : 3.64mA - Analog ON : 7.80mA - Analog OFF : 4.05mA | 

Test case 6 and 7 show strange behaviors (and I double checked them) : AOD power usage is **higher** than when the display is simply ON.

Let me know if you want me to do more measurements :)

----

On another note, I took the time to have a look at the changes from this branch. I think it would really help my review (and the reviews from other Core developers and contributors) if some of those changes were applied in dedicated PR, especially those that touch the low level drivers (SPI, LCD) and that are not directly linked to this PR (DateTime component, Weather component).

I know that splitting into multiple PR will require a bit of work, but this will allow the reviews to focus on smaller and simpler topics.

Thanks!

mark9064:
Thank you so much for all of the measurements! The data looks great

Case 6 particularly is very odd (7 is probably strange for the same reason). I'll need to design some follow up tests. Not sure when I'll be able to get some test cases going but hopefully by the weekend.

The other measurements are very interesting though
- The changes before AOD have no effect on power (great!)
- The overall impact of all of the AOD changes has pretty much no impact on On power consumption (and none to Off)
- AOD draws 3.45mA when screen content is static, and up to about 3.85mA when consistently redrawing (e.g analogue face draws second hand)
- 75Hz display costs about 400uA. Personally I think that's worth it particularly for the smoother scrolling. 75Hz gives a 3:2 refresh:draw ration (as LVGL draws at ~50Hz) which is definitely paced better than the current situation. I was shooting for 100Hz but the LCD couldn't take it
- The LCD voltage tweaks save about 400uA for both On and AOD
- LVGL throttling saves about 400uA for AOD

Overall it looks like the biggest power savings are related to refresh rate. But to say much more the backlight by itself definitely needs to be tested. Having full colours while in always on seems plausible though power wise provided we can manage the refresh another way

RE splitting into multiple PRs: we already discussed this in the pinetime-dev channel but for anyone else following along here I put forward some ideas in this comment https://github.com/InfiniTimeOrg/InfiniTime/pull/1869#issuecomment-1949583369

JF002:
Here are the result for test case #9 (sorry again for the delay!)

|Test case | Ref | Result |
|--------------|-------|----------|
| 9 | 7ad0bab | ON : 7.48mA - OFF (AOD) :  403µA |
| 9 | 0e265dc1 | ON : 7.47mA - OFF (AOD) :  970µA |
| 9 | 61405ec | ON : 7.37mA - OFF (AOD) :  1.69mA |
| 9 | 80812f8 | ON : 6.71mA - OFF (AOD) :  2.25mA |

Feel free to split this PR into smaller ones (like you mentioned in https://github.com/InfiniTimeOrg/InfiniTime/pull/1869#issuecomment-1949583369) so that we can we review each parts of the code impacted by this PR :)

mark9064:
Thanks for the extra test! The last one looks a bit strange, there are no changes to ON behaviour so the 6.71 is unexpected - not sure what happened there? If that's reproducible, it would be a very interesting result

Aside from that, the numbers are very useful:
- Low brightness uses 1.85mA (as written on the wiki, so good to confirm that!)
- PPI backlight at 10% duty and 50% uses 370µA and 1.29mA respectively. Giving efficiencies of 50% and 72% respectively

It looks like the PPI backlight has more overhead than expected :(

I will think about what tests could be useful here. I suspect efficiency at lower duties is terrible, and 99% duty is probably pretty good. If 50% efficiency is the best we can get it's not the end of the world (doing better than PWM I think, would need to check the datasheet though), but it would be nice to do better. According to the datasheet overhead should be low (I expect 80+% to be possible)

I'll try get the PRs up soon, life is so busy right now!

mark9064:
I've started splitting the PRs now, only had time to do one today but I'll get around to the others when I'm free

mark9064:
OK, the four sub-PRs are now split out. It looks like I need to make PRs for InfiniSim for 3 of them as well, not sure how long this will take as I've never worked on it before. But the PRs themselves here are all ready for review :)

mark9064:
InfiniSim PRs now done, so should be good on that front

mark9064:
I've removed the commits that have been split with the goal of making review simpler. This PR now soft-depends on #2041 #2043 #2042 #2046 (they are required for AOD to work but it compiles without them)

FintasticMan:
The pre-commit hook doesn't affect commits during a rebase (unless you edit them).

mark9064:
Ah makes sense, that's probably it

mark9064:
Just locked down the last annoying issue with frame pacing. Apparently the LVGL task handler can return before it's finished all its work (but it at least tells you that this has happened!)

With that this PR has all outstanding (feature-wise) issues resolved :)

I'm still mulling over how to optimise the backlight power further, it's quite difficult to know how the hardware is going to behave power-wise without a profiler myself (documentation isn't great so I'm making 'educated' guesses here). Same goes for optimising LCD power.

I don't think I've got any more changes to make to AOD now, so this is definitely ready for review (as well as the PRs it depends on). This PR could be split further if review is difficult, but I won't spam the repo with any more AOD PRs for now :sweat_smile:

PS: I also ported #1870 locally and it's looking great!

JF002:
Thanks for your extensive work on this feature and on both InfiniTime and InfiniSim @mark9064 !
I'll definitively have a look at those PR ASAP!

mark9064:
Changelog:
- RTC based backlight rather than using the TIMER peripheral. Means that the high frequency clock doesn't need to be kept running the whole time. Should save some power
- Fixed displayapp spinning infinitely when LVGL is very busy (causes system lockups)

Notes:
- Once #2059 is reviewed/merged I'll update this PR based on that
- This PR can be tested on hardware now since #2042 and #2046 are merged

If someone could do a power comparison between https://github.com/InfiniTimeOrg/InfiniTime/pull/1869/commits/487089a340a246f772c5021e9a415b019521f090 and https://github.com/InfiniTimeOrg/InfiniTime/pull/1869/commits/33c254fafea57ac7c725643e105a0423f7cb9e9a to verify that the RTC backlight is more efficient that'd be super :)

Saarsk:
Nice improvement, this. :) Small note, however.

It might be out of scope for this PR, but I find the spacing to be insufficient for the timeout setting items. It's hard to hit the right one, resulting in me tapping several times until I actually hit the setting I want. :)

There is screen area to space it out a little in both directions, though. It would help a lot. ^^,

mark9064:
Yeah this is out of scope for this PR - please do open an issue though so the design of this page can be discussed

JF002:
Hello @mark9064 ! I've just merged #2059 ;)

I also measured the power usage on 48a61fb and d9eb1a4:

48a61fb 
 - LCD ON (low brightness) : 7.53mA
 - Sleep (display OFF) : 420µA
 - Always on mode : 3.50mA

d9eb1a4
 - LCD ON (low brightness) : 7.53mA
 - Sleep (display OFF) : 420µA
 - Always on mode : 3.20mA

d9eb1a4 reduces the "always on" power usage by 300µA, which is not insignificant :+1


Could you remind me if there are any other PR that are needed before we can finish review and merging this one?

mark9064:
I've just rebased to migrate this PR to batched display arguments as well, and to accept the RTC based implementation.

Thanks for the power testing again, you've been brilliant running so many tests over the course of this PR!
300µA takes AOD backlight PWM from 50% to 100% efficiency :)
I think that's as good we'll get get

All required PRs here are merged

mark9064:
Small rebase mistake (used old version of a comment)

There are no more technical enhancements I can think of - if people want to try this PR out and feedback now is the time :)

The commits should tell a reasonably coherent story of development. We might want to consider breaking out the brightness message refactor (fix brightness getting stuck high / flashlight brightness restore) but I think it's a simple enough change?

Either way, everything here is ready for review

KaffeinatedKat:
awesome to see that this pr is finally ready for review

thanks so much to @mark9064, without your help with all the lower level issues this would not have been possible

JF002:
Thank you very much for you work @mark9064 !

I flashed this branch on my PineTime so I can test it, and I'm pleasantly surprised by the result : the time is well readable in most situation : indoor and even outside if you live in Belgium right now (it's been raining for numerous weeks...).

The battery lasted for 2 days on a single charge (BLE enabled, wake up on double tap on the display), From June 19th to June 22nd : 

Screenshot_20240623_104019_Gadgetbridge
Screenshot_20240623_104031_Gadgetbridge

While it's expected that the always-on display would reduce battery life, 2 full days is a reasonable trade-off

I'll do my best to review this PR ASAP !

mark9064:
And in real use the screen is off ~8h per night, so 3 days is possible (total runtime 48h as tested above - 16h on per day in real usage - 48/16 = 3)

One thing that's come to mind - should the brightness PWM bits belong in a driver rather than the controller?

src/components/brightness/BrightnessController.h
```diff
@@ -20,6 +23,25 @@ namespace Pinetime {
 
     private:
       Levels level = Levels::High;
+      static constexpr uint8_t UNSET = UINT8_MAX;
```
JF002, line 26: Since all those constants are implementation details (nothing except the code in BrightnessController.cpp use them) , they can be defined in `BrightnessController.cpp`. The includes for nrf_ppi.h and nrfx_gpiote.h will also move to the cpp file. 

This won't change anything except that it cleans the header (interface) of the class.

JF002 (changes requested):
Thanks you for this very high quality work (in this PR and all previous PR that were needed for the always on mode)!

I have the feeling that all the low-level code added in BrightnessController should fit better in a driver class rather than in a Controller class (drivers encapsulates all the low-level interactions with the hardware, controllers add higher-level - applicative logic to those driver).
But since there's no brightness driver in InfiniTime right now, I guess we can do this in a future PR :)

I added a few non-critical comments. Would you like to have a look at them? The most important one consists in clarifying the always-on setting (state, enabled).

src/components/settings/Settings.h
```diff
@@ -286,13 +321,22 @@ namespace Pinetime {
     private:
       Pinetime::Controllers::FS& fs;
 
-      static constexpr uint32_t settingsVersion = 0x0007;
+      static constexpr uint32_t settingsVersion = 0x0008;
+
+      // To enable disabling it during notification sleep, differentiate between
```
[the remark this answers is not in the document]
KaffeinatedKat: this is true. `state` is weather or not the setting is enabled, and `enabled` is weather or not it's currently active. This is so that when the watch goes into notification sleep `enabled` is off so the screen turns off, but `state` is still set to true so that the screen stays on again after exiting the notification sleep.

 Any idea for either a better implementation or a better naming scheme for the variables are welcome

src/components/brightness/BrightnessController.h
mark9064: I agree that UNSET and its definition can definitely move in, so can the the gpiote header.

Do you essentially mean that all private constants should move into the .cpp file (as they are private so by definition they can only be used by the corresponding class).
If so, this doesn't follow how a lot of InfiniTime is currently written.

I view constants declared in the private section of the header as constants that one may want to review or change in the future. IMO things like the PWM frequency or brightness steps available are things that should be explicitly configurable (while the exact value of UNSET does not need to be, for example).

Where do you draw the line between a private constant in the header and a constant defined in the implementation?

src/components/settings/Settings.h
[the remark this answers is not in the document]
JF002: Yeah, I think this is just the naming that is a bit confusing. Maybe we can try to find better names or at least clarify their meaning in comments?

src/components/brightness/BrightnessController.h
JF002: > If so, this doesn't follow how a lot of InfiniTime is currently written.

Yes, you're right! It's just that I learn everyday, and I noticed that some project put their private (aka implementation detail) into their .cpp file so that they are not visible at all to the 'users' of the class. This allows to keep the header of the class cleaner (you only expose what is interesting to the users of the class) and it also remove the coupling between classes by hiding dependence (like this include to the NRF SDK in this case). 

This might not be a good idea in all cases and for all variables though... it's just a feeling :)

Also, hiding the dependency to the NRF SDK might help when we'll eventually support multiple platform, but this is not the case right now.

mark9064: OK, so my take is that tunables for a class e.g BLE address for a characteristic, screen width / height, filesystem size should live in the header, as changing these should not require changing the implementation (.cpp). These are all examples of current private fields. Would these in your opinion better belong in the implementation? I'm trying to feel out your exact views here so I know what to do in the future :)

Since they are private, users of the class are unbothered anyway (no access), but being able to glance at the filesystem size for example may well be useful to a programmer using the class.

Regarding coupling, I've got a WIP/RFC resource map prototype
```c++
#pragma once
#include <cstdint>
#include "nrf_ppi.h"

namespace Pinetime {
  namespace ResourceMap {
    // PPI
    // 20 configurable channels, 5 used
    // Channels 3,6-19 are unused
    // Channels 20 through 31 (inclusive) have fixed function (mostly BLE radio related)
    // Refer to https://infocenter.nordicsemi.com/topic/com.nordic.infocenter.nrf52832.ps.v1.1/ppi.html

    static constexpr nrf_ppi_channel_t SpiErratumWorkaround = NRF_PPI_CHANNEL0;
    static constexpr nrf_ppi_channel_t BacklightOn = NRF_PPI_CHANNEL1;
    static constexpr nrf_ppi_channel_t BacklightOff = NRF_PPI_CHANNEL2;

    // FOR INFORMATION ONLY: defined in ble_phy.c
    static constexpr nrf_ppi_channel_t NimbleResponseTimerDisable = NRF_PPI_CHANNEL4;
    static constexpr nrf_ppi_channel_t NimbleResponseTimerWait = NRF_PPI_CHANNEL5;
    // FOR INFORMATION ONLY

    // RTC
    // 3 RTCs, all used

    // FOR INFORMATION ONLY: hardcoded across nimble (ble_phy.c, hal_timer.c)
    static constexpr uint32_t NimbleRTC = NRF_RTC0_BASE;
    // FOR INFORMATION ONLY

    // FOR INFORMATION ONLY: defined in portmacro_cmsis.h
    static constexpr uint32_t FreeRTOSRTC = NRF_RTC1_BASE;
    // FOR INFORMATION ONLY

    static constexpr uint32_t BacklightRTC = NRF_RTC2_BASE;

    // TIMER
    // 5 timers, 2 used
    // Timers 1,3,4 are unused

    // FOR INFORMATION ONLY: hardcoded across nimble (ble_phy.c)
    static constexpr uint32_t NimbleTimer = NRF_TIMER0_BASE;
    // FOR INFORMATION ONLY

    // Timer 2 CC0 is used to store the firmware version
    // Other than that the timer is unused
    static constexpr uint32_t FirmwareVersionStorageTimer = NRF_TIMER2_BASE;
  }
}
```
Which would clean that up. I think a central definition of peripherals used does make sense. Not sure if this form is the ideal one though.

src/components/settings/Settings.h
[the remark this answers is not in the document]
mark9064: Resolved by simplifying the whole setup. There's no need to store the actual state of AOD in the settings file

src/components/settings/Settings.h
```diff
@@ -286,13 +321,22 @@ namespace Pinetime {
...
+      // the setting being on, and the setting being set by the user
+      struct alwaysOnDisplayData {
```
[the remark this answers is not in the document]
mark9064: It's a boolean again so name is now correct :P

mark9064:
Cleaned up code style in a couple other places

src/components/brightness/BrightnessController.h
JF002: > Would these in your opinion better belong in the implementation? I'm trying to feel out your exact views here so I know what to do in the future :)

I'm currently working (on my day job) on big projects with a lot of internal and external dependencies. In this context, I noticed that it's far easier to integrate a class/library that exposes the least amount of types and other dependencies. So yeah, I think that everything that's not needed by the code that uses the class should be hidden in the .cpp file. 
You're right that this could serve as documentation for the developer though so... I guess the idea is not mature enough for me to write an actual coding guideline.

Your resource map looks similar to the current `PinMap` (https://github.com/InfiniTimeOrg/InfiniTime/blob/main/src/drivers/PinMap.h) we have in InfiniTime : I'm not a huge fan on this, since this effectively creates global variables, but I guess this is a effective way to document the resource usage so... let's use it :-)

Anyway, this discussion it probably out of the scope of this PR, let's move the variables to .cpp if we ever feel the need in the future :)

src/systemtask/SystemTask.cpp
JF002: Yep, we could power the spi flash only when we need it, but yeah, power usage of the chip is very low compares to the display or the MCU so I haven't done that yet :-)

JF002:
Thank you so much to everyone who contributed to this PR, and especially to @mark9064 who did a wonderful job and who was very patient during this very long review process (sorry about that!)  :1st_place_medal:

mark9064:
A massive thanks to you too for taking the time to test configurations and review/discuss even the most minor details (including for all the supporting PRs!). Also cheers to @KaffeinatedKat for graciously trusting me with contributing directly to this feature :)

PR #1870, opened 2023-10-03
Smooth brightness transitions with pwm backlight

KaffeinatedKat:
Using the pwm backlight changes from #575, I have changed BrightnessController::Set to smoothly transition from each brightness instead of setting it instantly with the GPIO pins. This small change makes the watch feel much nicer and more polished

Currently, changing the brightness via the setting also does the transition, which I personally think looks fine, but implementing a boolean argument for BrightnessController::Set for the transition to disable this might be a good idea. Would love to hear some thoughts on this

### DIFF
--- a/src/components/brightness/BrightnessController.cpp
+++ b/src/components/brightness/BrightnessController.cpp
@@ -2,39 +2,75 @@
 #include <hal/nrf_gpio.h>
 #include "displayapp/screens/Symbols.h"
 #include "drivers/PinMap.h"
+
+#include "nrf_pwm.h"
+
 using namespace Pinetime::Controllers;
 
 void BrightnessController::Init() {
   nrf_gpio_cfg_output(PinMap::LcdBacklightLow);
   nrf_gpio_cfg_output(PinMap::LcdBacklightMedium);
   nrf_gpio_cfg_output(PinMap::LcdBacklightHigh);
-  Set(level);
-}
 
-void BrightnessController::Set(BrightnessController::Levels level) {
+  nrf_gpio_pin_clear(PinMap::LcdBacklightLow);
+  nrf_gpio_pin_clear(PinMap::LcdBacklightMedium);
+  nrf_gpio_pin_clear(PinMap::LcdBacklightHigh);
+
+  static nrf_pwm_sequence_t seq;
+
+  seq.values.p_common = pwmSequence;
+  seq.length          = 1;
+  seq.repeats         = 0;
+  seq.end_delay       = 0;
+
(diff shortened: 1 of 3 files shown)


Saarsk:
> This small change makes the watch feel much nicer and more polished
I agree. :) It's one of the first things I noticed with this watch. I just assumed it was a limitation of the LCD panel and there was nothing that could be done to fix it.

Note that I'm imagining the transition from the video linked to in this comment (haven't tried this PR (I don't know how, to be honest)): https://github.com/InfiniTimeOrg/InfiniTime/pull/575#issuecomment-898465400


> Currently, changing the brightness via the setting also does the transition, which I personally think looks fine, **but implementing a boolean argument for BrightnessController::Set for the transition to disable this might be a good idea.** Would love to hear some thoughts on this
Do you mind elaborating on this? Does this mean that you allow for the brightness transition code to be executed without affecting the set brightness levels? And that what I'm seeing on my watch today when adjusting brightness setting would be smoother with this PR?

---

Also, it would be nice if the brightness would gradually dim after being awakened. Something like this:
_Backlight on (3 sec) --> slow transition to next lower brightness level --> Another slow transition which continues until display turns off (5 sec)_

Whereas the 2nd step (transition to next lower brightness) could extend depending on display timeout user setting. For very short timeout settings like 5s, the transition could start e.g. from second 4 and span over 3 seconds?

Personally I set a 15s timeout setting just because the other ones tend to catch me off guard and I have to re-wake it, which I find annoying. So most of the time the display is being lit way too long, just to solve the annoyance of it going dark before I'm able to react. If I'm reading, I can still read if it transitions into the next lower brightness level and when it starts to fade to black slowly I have the time to keep it wakened if I need to (sometimes I multitask and look at my phone or press a button or whatever and the display just goes dark without any time to react).

If I'm just checking the time or reading a notification, chances are I'll put my wrist down while the LCD is still being backlit at the user set brightness. So a lot of the time it's being brightly lit for nothing and consuming precious energy.

Currently, it wakes to the user set brightness and stays like that for as long as you set the backlight timeout. After which dims briefly and then quickly goes dark before you have a chance to react.

Gradual dimming would not only extend battery life but it would also act as an indicator that the watch isn't picking up "wake intents" and the third phase of slow fadeout will indicate clearly that it's about to turn off (at which point you could choose to extend the backlight active time, if so desired).

I'm positive this would make users set a lower timeout setting, which would have a positive impact on battery time (which is a win-win situation for everyone).

tituscmd:
If you have the time, could you rebase this to 1.15? :)

mark9064:
Since I reworked the backlight in the AOD PR it can't be easily rebased. But I have already implemented smooth transitions locally, I just need to upstream it. There are a few last things I want to iron out before it's ready

tituscmd:
That's awesome! Make sure to keep me/us in the loop!

PR #1872, opened 2023-10-04
Shows selected Watch face page first

SteveAmor:
When you enter the watch face menu, it will take you to the second page if your current watch face is the one that is active.  I thought it was a bit odd that the first page is shown with no radio buttons selected and you scroll down the the second page.


SteveAmor:
Before
InfiniSim_2023-10-04_202745
After
InfiniSim_2023-10-04_203105

src/displayapp/screens/settings/SettingWatchFace.cpp
```diff
@@ -26,7 +26,10 @@ SettingWatchFace::SettingWatchFace(Pinetime::Applications::DisplayApp* app,
   : app {app},
     settingsController {settingsController},
     filesystem {filesystem},
-    screens {app, 0, CreateScreenList(), Screens::ScreenListModes::UpDown} {
+    screens {app,
+             static_cast<uint8_t>(static_cast<uint8_t>(settingsController.GetWatchFace()) / 4),
```
mark9064, line 61: I think it would make more sense to cast the constant rather than the result

SteveAmor: After a lot of testing, casting the constant results in narrowing compiler errors. The above is the only casting that the compiler would accept (not even a warning thrown).

mark9064: You're totally right. All divisions get promoted to at least `int` size apparently. So I think we can leave this as is. Sorry!

PR #1874, opened 2023-10-04
Removes redundant words to save space

SteveAmor:
The title of the page is Watch face so I felt the word face added to the watch face name is a little redundant and will save some space if it is removed.


mark9064:
Once conflicts resolved, approved by me :)

I'll put on the 1.17 milestone so it doesn't block release (which should hopefully soonish), but it can definitely be merged before if ready IMO

mark9064 (approved):

PR #1876, opened 2023-10-04
Open battery app when charging

KaffeinatedKat:
This change makes the battery info app open upon putting the watch on the charger, if the current app is the clock (as to not close any app that might be running). If the watch is taken off the charger while the battery app is open, it returns to the clock. 

This pr also removes the vibration while charging if the battery app is opened/closed, which should help prevent #724.

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -362,7 +362,16 @@ void DisplayApp::Refresh() {
         break;
       case Messages::OnChargingEvent:
         RestoreBrightness();
-        motorController.RunForDuration(15);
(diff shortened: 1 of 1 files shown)


RageGamerBoi:
This might be better as kind of popup, like a notification with that 5 second timeout.

vkareh (approved):
I like the simplicity of this code. Doesn't add much overhead and works well (have been using it for a few months with no issues) while improving user expectation upon plugging/unplugging.

vkareh:
I've included this patchset as part of a larger configuration that allows auto-opening apps based on certain events: https://github.com/InfiniTimeOrg/InfiniTime/pull/2004/commits

This makes this feature optional/configurable.

Issue #1878, opened 2023-10-06
RFC: Timeline event packing


Avamander:
Problem:
It's somewhat difficult to handle long-term and short-term events overlapping in the timeline in user-facing situations.

Proposal:
Some of this confusion can be removed by implementing predictable conflict resolution. 

We can adjust the `timestamp` value of the longer event to start at the end of shorter/previous element's `timestamp + expiration` and shorten expiration of the longer event by the amount shortened.

This packing shouldn't be done for equally-long events, neither should a shorter event be shortened.

Before:
```
|     |     |     |     |     |     |     |     |     |
|────────────────────────────────────────────────────────────►
  ┌───────────┐
  │Temperature│
  └───────────┘
     ┌────────────────────────────────────────────────┐
     │ Temperature (long-term)                        │
     └────────────────────────────────────────────────┘
```
After:
```
|     |     |     |     |     |     |     |     |     |
|────────────────────────────────────────────────────────────►
  ┌───────────┐┌──────────────────────────────────────┐
  │Temperature││ Temperature (long-term)              │
  └───────────┘└──────────────────────────────────────┘
```

Are there any potential use-cases where this might not work out as intended?

minacode:
> It's somewhat difficult to handle long-term and short-term events overlapping in the timeline in user-facing situations.

Can you explain more, what exactly the problem is?

JF002:
Thanks for your work @Avamander. However, I decided to rework the weather service to fix the issues more easily, so I'll close this one in favor of #1924.

PR #1882, opened 2023-10-07
weather: Add some heuristics to GetCurrent*

FintasticMan:
This means that there is a higher likelihood that the returned event is the most up-to-date current event.
It chooses the event with the lowest expiry, and between events with the same expiry, it chooses the one with the latest timestamp.

Alternative to #1878.

I'm not the best when it comes to working with references to unique_ptrs etc, so if there's a better way to keep track of the best option so far, I'd like to hear it.


FintasticMan:
Superseded by #1890.

PR #1884, opened 2023-10-08
Show alarm controller state in status icons

jmlich:
I would like to know if the alarm is active on Watch Face Digital screen. 

I made an prototype with InfiniSim and it looks that it works as expected:

screenshot-alarm-indicator InfiniSim_2024-12-22_141530

However, I have only sealed version of PineTime and this is my first attempt to make something for embedded device. I will be happy if someone can check if this actually works on real device. I am looking forward for your feedback.

### DIFF
--- a/src/displayapp/widgets/StatusIcons.cpp
+++ b/src/displayapp/widgets/StatusIcons.cpp
@@ -1,10 +1,13 @@
 #include "displayapp/widgets/StatusIcons.h"
 #include "displayapp/screens/Symbols.h"
+#include "components/alarm/AlarmController.h"
 
 using namespace Pinetime::Applications::Widgets;
 
-StatusIcons::StatusIcons(const Controllers::Battery& batteryController, const Controllers::Ble& bleController)
-  : batteryIcon(true), batteryController {batteryController}, bleController {bleController} {
+StatusIcons::StatusIcons(const Controllers::Battery& batteryController,
+                         const Controllers::Ble& bleController,
+                         const Controllers::AlarmController& alarmController)
+  : batteryIcon(true), batteryController {batteryController}, bleController {bleController}, alarmController {alarmController} {
 }
 
 void StatusIcons::Create() {
@@ -19,6 +22,9 @@ void StatusIcons::Create() {
 
   batteryPlug = lv_label_create(container, nullptr);
   lv_label_set_text_static(batteryPlug, Screens::Symbols::plug);
+
+  alarmIcon = lv_label_create(container, nullptr);
+  lv_label_set_text_static(alarmIcon, Screens::Symbols::bell);
 
   batteryIcon.Create(container);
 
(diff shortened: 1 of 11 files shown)


src/displayapp/widgets/StatusIcons.cpp
```diff
@@ -37,6 +43,8 @@ void StatusIcons::Update() {
     batteryIcon.SetBatteryPercentage(batteryPercent);
   }
 
+  lv_obj_set_hidden(alarmIcon, (alarmController.State() == Pinetime::Controllers::AlarmController::AlarmState::Not_Set));
```
minacode, line 46: I would put the equality in its own line and give it a meaningful name.

minacode (commented):
I like this feature! :)

You should be very safe to test this on a sealed device (speaking from experience).
If something goes wrong, which shouldn't happen, you can always:
* reboot to get back to your last firmware (don't verify your debug flash of course)
* hold the button pressed until the color changed twice during boot to get to the fallback firmware and update your watch from there.

src/displayapp/widgets/StatusIcons.cpp
jmlich: Is it better now?

minacode: Yes :)

jmlich:
I have also tested the feature on hardware.

RageGamerBoi:
I like this, but I think the icon has a little too much filled space and the simple circle isn’t as defined as a more standard alarm icon:
image
Of course these might not work with a resolution that small, but I’m sure something similar could…

jmlich:
I wasn't trying to replace clock icon. I was trying to provide an information about alarm status with existing icon.

I agree that rework of icons and design overall may improve look and feel, but this wasn't my intention in this pull request.

RageGamerBoi:
Ah, I forgot it was already the icon for the alarm app.
Sorry about that.

vkareh:
This should be rebased against `main` and updated to reflect the new icon from https://github.com/InfiniTimeOrg/InfiniTime/pull/1996

rnwgnr:
@jmlich I'd like to give this change a try, would you mind rebasing this changeset against the current state of `main`?

jmlich:
Rebased to recent main

jmlich:
Rebased to recent main.

src/displayapp/widgets/StatusIcons.cpp
```diff
@@ -37,6 +43,8 @@ void StatusIcons::Update() {
     batteryIcon.SetBatteryPercentage(batteryPercent);
   }
 
+  lv_obj_set_hidden(alarmIcon, !alarmController.IsEnabled());
```
mark9064, line 46: Not an LVGL expert, but I'm pretty sure this should use DirtyValue like the other icons as LVGL doesn't track what's on the screen

mark9064 (commented):
Looking great!

src/displayapp/widgets/StatusIcons.cpp
jmlich: I see, I haven't much experience with LVGL either. I hope it is better now.

mark9064 (approved):
Code looks great! I still need to test on hardware, other than that all good to go from me :)

vkareh (approved):
LGTM
I've been using this on hardware in my daily driver branch for many months

JF002 (approved):
LGTM, thanks!

PR #1887, opened 2023-10-11
Heartrate app displays --- instead of 000

### LINKED ISSUE #1881: Heartrate Value jumps to 000
### Verification

### What happened?

The heartrate monitor will sporadically jump to 000, and will sometimes stay at that value. for a long time 

_No response_

### Reproduction steps

Wearing the watch, I select the heart rate app and leave it running.

### More details?

_No response_

### Version

v1.13.0

### Companion app

Not currently paired with phone.

KaffeinatedKat:
Fixes #1881, when the heartrate data is not clear, or there is no measurement, display --- instead of 000

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@ set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose Debug or Release")
 project(pinetime VERSION 1.13.0 LANGUAGES C CXX ASM)
 
 set(CMAKE_C_STANDARD 99)
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 20)
 
 # set(CMAKE_GENERATOR "Unix Makefiles")
 set(CMAKE_C_EXTENSIONS OFF)
--- a/doc/code/Apps.md
+++ b/doc/code/Apps.md
@@ -9,49 +9,95 @@ This page will teach you:
 
 The user interface of InfiniTime is made up of **screens**.
 Screens that are opened from the app launcher are considered **apps**.
-Every app in InfiniTime is it's own class.
+Every app in InfiniTime is its own class.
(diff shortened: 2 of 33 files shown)


FintasticMan (approved):
I think this makes it more clear to the user what the situation is.

src/displayapp/screens/HeartRate.cpp
```diff
@@ -82,10 +82,14 @@ void HeartRate::Refresh() {
     case Controllers::HeartRateController::States::NoTouch:
     case Controllers::HeartRateController::States::NotEnoughData:
       // case Controllers::HeartRateController::States::Stopped:
-      lv_label_set_text_static(label_hr, "000");
+      lv_label_set_text_static(label_hr, "---");
       break;
     default:
-      lv_label_set_text_fmt(label_hr, "%03d", heartRateController.HeartRate());
+      if (heartRateController.HeartRate() == 0) {
+        lv_label_set_text_static(label_hr, "---");
```
NeroBurner, line 89: is this last if-else needed? doesn't the state go back to `NotEnoughData`? Or does it go back down to 0 heartrate? I'm curious

FintasticMan: It does look like it goes to `NotEnoughData`, at least, if I let it get a value then remove it from my skin, it displays the not enough data text.

FintasticMan: I'm quite sure it's not needed.

KaffeinatedKat: it is in place so that when you press the stop button is displays "---" instead of "000", as this does not fall into the `NotEnoughData`

FintasticMan: Ah, thanks for the clarification!

NeroBurner (approved):

Issue #1814, opened 2023-07-21
No pairing is needed to re-flash watch (Security issue!)


annoyinganime:
### Verification

- [X] I searched for similar issues and found none was relevant.

### Introduce the issue

Yesterday, when I was flashing my watch using WatchMate I found something that looks like a severe security issue. 
I disabled my phone's bluetooth (I read that when BLE connection is active no other device can connect to that BLE device), opened WatchMate on my laptop, selected there my pinetime watch, then selected "flashing", chose new firmware and began the process. And while the progress bar slowly filled, it dawned on me: in no point in time I EVER touch my watch. I just basically connected, and started flashing.
So looks like if you keep watch disconnected with bluetooth still enabled anyone in range can connect and flash watch with potentially malicious firmware.

### Preferred solution

What I suggest is any pairing process on watch side, even pressing simple "YES" on new connection would be good enough

### Version

v1.13.0

everypizza1:
Sometimes I will shut down my phone overnight and then reconnect to gadgetbridge in the morning. It's much easier than pressing yes on the watch.

annoyinganime:
You didn't understand me, I want it to be done during first-time pairing, and not at every reconnect. I shit down my phone every night, and that's exactly an issue I'm describing: when your phone is disabled, anyone can connect to your watch without your consent, and upload malicious firmware to your watch

minacode:
When pairing was introduced not all companions implemented it, so there had to be a phase where you could (but don't have to) pair your watch. This is where we currently still are. If I remember correctly, you get an encrypted connection while being paired.

Which companions support pairing right now and which don't? I think at least with iOS there is no pairing available.

annoyinganime:
@minacode I am not talking about established connection, and pairing for encryption, and maybe I'm just bad at explaining, what I want to be implemented is something like:
At first connection to new device, whether the connection is encrypted or not, infinitime asks user if new connection should be trusted, and if user presses "yes' - at adds MAC address to "known devices" list. So if there is a new connection from previously unknown device - user has to explicitly allow it.

minacode:
Ah, you mean #1264?

Avamander:
In order for there to be a security issue there has to be either a risk with an impact that can manifest, or it has to bypass some existing guarantee. I don't see either of those being true.

Asking for consent before an update would be a reasonable feature request. 

[...]

PR #1891, opened 2023-10-20
Setting to disable DFU and FS access

DavisNT:
This PR adds setting to disable firmware update and Bluetooth file transfer.
This is 2. part for https://github.com/InfiniTimeOrg/InfiniTime/issues/1814#issuecomment-1763740314


DavisNT:
Both issues discovered by the workflows should now be fixed.

 - I have force-pushed fix for the formatting error in `SettingOTA.cpp`.
 - For InfiniSim I have created https://github.com/InfiniTimeOrg/InfiniSim/pull/125 Building this PR with InfiniSim PR should fix the InfiniSim build issue.

DavisNT:
The build-simulator step is expected to fail, because it needs https://github.com/InfiniTimeOrg/InfiniSim/pull/125

DavisNT:
Made two improvements:

- Added branch prediction hints, to minimize performance impact during real data transfer
- Return `BLE_ATT_ERR_INSUFFICIENT_RES` when access is not allowed

Could it be possible to review this PR?

DavisNT:
Rebased as well.

DavisNT:
Is this PR (and the other PR described in https://github.com/InfiniTimeOrg/InfiniTime/issues/1814#issuecomment-1763740314 ) welcome?
If I will rebase and adjust this PR to work with the latest changes, will it be merged?
@JF002, @FintasticMan, @Avamander Any feedback from project maintainers would be appreciated.

DavisNT:
If there are questions about b2e2690adfbc13cbf4e0778a428baa6b11b35fba please see branch https://github.com/DavisNT/InfiniTime/tree/otaset-oom - this branch has the memory saving as the last commit (trying to build [otaset-oom](https://github.com/DavisNT/InfiniTime/tree/otaset-oom) without last commit should cause the build to fail with `/opt/gcc-arm-none-eabi-10.3-2021.10/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/bin/ld: region RAM overflowed with stack`).

mark9064 (commented):
Generally looking good to me! Works well on hardware
Few suggestions, feel free to challenge any of them

src/components/ble/DfuService.cpp
```diff
@@ -78,6 +80,18 @@ void DfuService::Init() {
...
 int DfuService::OnServiceData(uint16_t connectionHandle, uint16_t attributeHandle, ble_gatt_access_ctxt* context) {
+#ifndef PINETIME_IS_RECOVERY
+  if (__builtin_expect(systemTask.GetSettings().GetDfuAndFsMode() == Pinetime::Controllers::Settings::DfuAndFsMode::Disabled, 0)) {
+    Pinetime::Controllers::NotificationManager::Notification notif;
+    memcpy(notif.message.data(), denyAlert, denyAlertLength);
+    notif.size = denyAlertLength;
+    notif.category = Pinetime::Controllers::NotificationManager::Categories::SimpleAlert;
+    systemTask.GetNotificationManager().Push(std::move(notif));
+    systemTask.PushMessage(Pinetime::System::Messages::OnNewNotification);
+    return BLE_ATT_ERR_INSUFFICIENT_RES;
```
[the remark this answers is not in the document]
DavisNT: Thanks! I somehow thought that I have already used this error code (but for some off reason I had confused and used another one). 
Now it should be fixed.

src/components/ble/DfuService.cpp
```diff
@@ -78,6 +80,18 @@ void DfuService::Init() {
 }
 
 int DfuService::OnServiceData(uint16_t connectionHandle, uint16_t attributeHandle, ble_gatt_access_ctxt* context) {
+#ifndef PINETIME_IS_RECOVERY
+  if (__builtin_expect(systemTask.GetSettings().GetDfuAndFsMode() == Pinetime::Controllers::Settings::DfuAndFsMode::Disabled, 0)) {
```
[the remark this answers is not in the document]
DavisNT: Idea was to have minimal performance impact during actual file transfer and firmware update (by instructing compiler that branching should be optimized for the scenario when the condition is false).
However I tested the performance and in both cases it is around 6.65 KB/s from around 1 m distance.
`__builtin_expect` removed.

src/components/ble/DfuService.h
```diff
@@ -83,6 +85,9 @@ namespace Pinetime {
       DfuImage dfuImage;
       NotificationManager notificationManager;
 
+      static constexpr const char* denyAlert = "InfiniTime\0Firmware update attempted, but disabled.";
+      static constexpr const uint8_t denyAlertLength = 52;
```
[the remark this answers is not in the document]
DavisNT: Done. And added a comment to prevent this from being accidentally broken in future.

src/components/ble/DfuService.h
```diff
@@ -83,6 +85,9 @@ namespace Pinetime {
       DfuImage dfuImage;
       NotificationManager notificationManager;
 
+      static constexpr const char* denyAlert = "InfiniTime\0Firmware update attempted, but disabled.";
```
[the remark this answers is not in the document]
DavisNT: Done with shorter text. I think it is still as understandable.

src/components/ble/FSService.h
```diff
@@ -26,6 +28,10 @@ namespace Pinetime {
     private:
       Pinetime::System::SystemTask& systemTask;
       Pinetime::Controllers::FS& fs;
+
+      static constexpr const char* denyAlert = "InfiniTime\0File access attempted, but disabled.";
+      static constexpr const uint8_t denyAlertLength = 48;
```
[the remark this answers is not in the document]
DavisNT: Done.

src/displayapp/screens/settings/Settings.h
```diff
@@ -45,6 +45,7 @@ namespace Pinetime {
           {Symbols::clock, "Chimes", Apps::SettingChimes},
           {Symbols::tachometer, "Shake Calib.", Apps::SettingShakeThreshold},
           {Symbols::check, "Firmware", Apps::FirmwareValidation},
+          {Symbols::shieldAlt, "Over-the-air", Apps::SettingOTA},
```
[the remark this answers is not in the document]
DavisNT: Sure! Done.

src/components/ble/DfuService.cpp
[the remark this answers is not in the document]
mark9064: I totally agree with you technically (makes sense to optimise the enabled case), the only reason I pointed this out is that the number of packets per second is never high enough for it to make a big difference (it is easy to forget how fast the CPU is - I'm pretty sure it'd be bottlenecked by the max bandwidth of BLE at about 700kbps) and therefore IMO the "less code wins" rule takes precedence

src/components/settings/Settings.h
```diff
@@ -36,6 +36,7 @@ namespace Pinetime {
       };
       enum class PTSGaugeStyle : uint8_t { Full, Half, Numeric };
       enum class PTSWeather : uint8_t { On, Off };
+      enum class DfuAndFsMode : uint8_t { Disabled = 0, Enabled = 1, EnabledTillReboot = 2 };
```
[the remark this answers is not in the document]
DavisNT: I always like to explicitly add enum values (just in case somebody will insert something in the middle of enum and enum is stored in non-volatile memory), but as this is not done in 7 of 8 other enums in this file, I have removed the values.

src/components/ble/FSService.cpp
```diff
@@ -49,6 +51,18 @@ void FSService::Init() {
...
 int FSService::OnFSServiceRequested(uint16_t connectionHandle, uint16_t attributeHandle, ble_gatt_access_ctxt* context) {
+#ifndef PINETIME_IS_RECOVERY
+  if (__builtin_expect(systemTask.GetSettings().GetDfuAndFsMode() == Pinetime::Controllers::Settings::DfuAndFsMode::Disabled, 0)) {
+    Pinetime::Controllers::NotificationManager::Notification notif;
+    memcpy(notif.message.data(), denyAlert, denyAlertLength);
+    notif.size = denyAlertLength;
+    notif.category = Pinetime::Controllers::NotificationManager::Categories::SimpleAlert;
+    systemTask.GetNotificationManager().Push(std::move(notif));
+    systemTask.PushMessage(Pinetime::System::Messages::OnNewNotification);
+    return BLE_ATT_ERR_INSUFFICIENT_RES;
```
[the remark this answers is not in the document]
DavisNT: Done.

mark9064:
Regarding https://github.com/InfiniTimeOrg/InfiniTime/pull/1891/commits/b2e2690adfbc13cbf4e0778a428baa6b11b35fba
So one option is to decrease the heap size a notch so more RAM is free for static alloc. This is going to happen soon anyway as there are very few bytes free
I'm not 100% sure why the original approach was taken, really @JF002 would have to chime in on this. I suspect that it's to ensure that classes only have access to certain controllers etc? The memory savings look good though

mark9064 (approved):
LGTM code wise :)

I haven't had the time to properly consider the alternative ideas put forward in #1814 yet so that might end up being a better user interface, but in terms of implementation this looks solid (aside from discussion about the way of accessing controllers)

src/components/settings/Settings.h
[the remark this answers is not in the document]
mark9064: I agree on this one. I guess we rely on the fact that the entire settings file gets wiped if the version changes, so if new options are added it doesn't matter?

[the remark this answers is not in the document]
DavisNT: Yes - increasing `settingsVersion` should wipe any incorrect enum values stored in settings.

src/components/ble/DfuService.cpp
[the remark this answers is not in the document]
DavisNT: Could very well be that CPU is never the bottleneck in this case (I haven't thought much about this).
Maybe the flash memory speed is the bottleneck (as the average firmware transfer speed is around 50-55 kbps - at least for me).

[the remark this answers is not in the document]
mark9064: The small packet size used for DFU (20 bytes) is probably the biggest factor. Due to the somewhat fragile way DFU is currently implemented, any size that evenly divides 200 bytes is actually acceptable. It would be interesting to see what throughput with 200 byte packets would be, but unfortunately due to a bug with the driver for my laptop's BT chipset I can't test it currently! I need to debug the driver at some point but my kernel skills aren't too sharp...

The flash memory is really fast (>100mbit/s when used properly), it's bottlenecked by the SPI bus for sure (8mbit/s)

DavisNT:
@JF002, @NeroBurner, @FintasticMan, @Avamander Could it be possible to review this PR?
CC: @mark9064 

P.S. If I need to rebase this, please let me know!

src/components/settings/Settings.cpp
```diff
@@ -34,6 +34,9 @@ void Settings::LoadSettingsFromFile() {
   if (bufferSettings.version == settingsVersion) {
     settings = bufferSettings;
   }
+  if (settings.dfuAndFsMode == DfuAndFsMode::EnabledTillReboot) {
```
mark9064, line 37: Although it doesn't make too much of a difference, maybe apply this to `bufferSettings` inside the if clause. That way it's clear that work is being done only when settings are successfully loaded, and it means that settings only gets updated once

src/components/ble/ImmediateAlertService.cpp
```diff
@@ -63,7 +63,8 @@ int ImmediateAlertService::OnAlertLevelChanged(uint16_t attributeHandle, ble_gat
       auto* alertString = ToString(alertLevel);
 
       NotificationManager::Notification notif;
-      std::memcpy(notif.message.data(), alertString, strlen(alertString));
+      std::memcpy(notif.message.data(), alertString, strlen(alertString) + 1);
```
mark9064, line 66: Is this change related to this patchset? Or is it an unrelated bug fix

mark9064 (commented):
Spotted a couple more things on re-review

I think the only questionable thing at the moment is https://github.com/InfiniTimeOrg/InfiniTime/pull/1891/commits/b2e2690adfbc13cbf4e0778a428baa6b11b35fba and whether this architecture is a good idea

Other than those bits this is ready to go IMO

src/components/ble/ImmediateAlertService.cpp
DavisNT: This is unrelated fix.

src/components/settings/Settings.h
```diff
@@ -308,6 +320,8 @@ namespace Pinetime {
         uint16_t shakeWakeThreshold = 150;
 
         Controllers::BrightnessController::Levels brightLevel = Controllers::BrightnessController::Levels::Medium;
+
+        DfuAndFsMode dfuAndFsMode = DfuAndFsMode::Disabled;
```
NeroBurner, line 324: If the mode EnabledTillReboot is saved into the settings, and restored on reboot then it is effectively always enabled. Please check my suspicion

src/components/ble/ImmediateAlertService.cpp
FintasticMan: I think it's to do with the fact that we don't use the notification buffer optimally (or even correctly really). I've proposed fixes for this in #1694 and #1695, but I haven't followed up on them.

src/components/settings/Settings.h
mark9064: See https://github.com/InfiniTimeOrg/InfiniTime/pull/1891/files#r1807283475

src/components/ble/ImmediateAlertService.cpp
DavisNT: @NeroBurner Sure! I will make a separate PR for this. 
@FintasticMan I think when copying a string using `memcpy()` the size always needs to be `strlen()` of the string plus 1 for the trailing zero-byte string terminator (unless the destination memory is guaranteed to be prefilled with zero-bytes).

src/components/settings/Settings.cpp
DavisNT: @mark9064 I would prefer to leave this as it is now. Currently it will fail-secure (set `settings.dfuAndFsMode` to `DfuAndFsMode::Disabled`) if something very unexpected happens. 

@NeroBurner This is the reason why `DfuAndFsMode::EnabledTillReboot` should never get restored from the saved settings on reboot.

src/components/settings/Settings.h
DavisNT: @NeroBurner The `DfuAndFsMode::EnabledTillReboot` should never be restored on reboot. Please check the link from @mark9064's comment above.

src/components/settings/Settings.cpp
mark9064: Sorry if it was unclear, I'm suggesting you do this
```cpp
  if (bufferSettings.version == settingsVersion) {
    if (bufferSettings.dfuAndFsMode == DfuAndFsMode::EnabledTillReboot) {
      bufferSettings.dfuAndFsMode = DfuAndFsMode::Disabled;
    }
    settings = bufferSettings;
  }
```
I believe this does not change how the code behaves at all?

DavisNT: This is exactly the thing I would like to keep as is (unless requested by the maintainers to change it).
If possible I would like to keep this code as it is and keep it at the very end of `Settings::LoadSettingsFromFile()`.

The change you suggest would not change how the code behaves.
However in case of refactoring it would increase the chances of the `if (bufferSettings.dfuAndFsMode == DfuAndFsMode::EnabledTillReboot) {` to be forgotten or moved to a code path that is not always executed.

Maybe I should add a code comment above `if (settings.dfuAndFsMode == DfuAndFsMode::EnabledTillReboot) {` telling that this (when needed) switches off DFU and FS access after reboot? 
Or, should I make the requested code change + a code comment?

mark9064: I think I'd personally prefer the suggested change with a comment explaining that this is needed for firmware+files security to work properly.

Maybe it'd be better to fix this in the data format itself thought. What about introducing a new boolean member of Settings.h that's not inside the Settings struct that stores whether DFU+FS is temporarily enabled.
Then `GetDfuAndFsMode` can compute the correct enum value using the setting+the member, and `SetDfuAndFsMode` can unpack the state into Disabled,Disabled+member set,Enabled. That way it will be impossible to break when refactoring and we don't need to mess with saving/loading settings

src/components/ble/ImmediateAlertService.cpp
DavisNT: @NeroBurner Created the PR #2159

src/components/settings/Settings.cpp
DavisNT: @mark9064 Thanks! Two bool variables (one in `settings` and the other one outside) sounds like a really good idea, I have implemented it (see the recent push).

DavisNT:
@mark9064 Building the simulator fails on `main` branch with exactly the same error messages.

The updated PR:
- is rebased
- the ImmediateAlertService fix has been moved to a separate PR #2159
- changing of setting value on loading has been replaced with two separate bool variables - one volatile and one persisted in the `settings` object (code from `Settings::LoadSettingsFromFile()` has been removed)

DavisNT:
@FintasticMan, @NeroBurner could it be possible to review this PR (and merge it, if it is OK)?

NeroBurner:
We are currently in feature freeze for 1.15. I've scheduled your PR to 1.16

DavisNT:
@NeroBurner I have rebased this PR and resolved the merge conflicts.
Can you please review and merge this into 1.16?

DavisNT:
Thank you very much!

PR #1901, opened 2023-10-30
Music app redesign

coffandro:
I found the stock music app unsatisfactory so therefor I decided to redesign it, on my watch I've found it to be more stable than stock and it's easier to see what buttons do and such
image


coffandro:
Thank you, I 100% agree, also wanted to somehow make the title more or less fill the entire  empty space but couldn't find a way to do it in LGVL

coffandro:
Newest version provides a better layout for the song name though it does have a tendency to overflow slightly which I'm working on

JustScott:
Loving this design! I notice that swiping up and down increases and decreases the volume, removing the ability to swipe down to exit the app. I'm not sure if this was intentional or a bug, but assuming it was intentional, I think the Infinitime developers prefer apps to not take over the swipe to exit ability (I think I saw someone say that in a PR, I may also be wrong). Anyways... I'll still be using it, thank you for the work you did on this!

JustScott:
@coffandro I noticed you changed the notification label color in this pull request, as that's unrelated to the music app it should probably be removed.

coffandro:
That is true, updating it in a day or so for 1.14 and ill remove it and recheck everything then

coffandro:
> Loving this design! I noticed that swiping up and down increases and decreases the volume, removing the ability to swipe down to exit the app. I'm not sure if this was intentional or a bug, but assuming it was intentional, I think the Infinitime developers prefer apps to not take over the swipe to exit ability (I think I saw someone say that in a PR, I may also be wrong). Anyways... I'll still be using it, thank you for the work you did on this!

Pretty sure the stock music app already lock up/down, i have just adapted it for my own use case, and the one i see as more practical, where the music app "takes over" control until you press the button, can remove this if a maintainer wants me to tho

JustScott:
> That is true, updating it in a day or so for 1.14 and ill remove it and recheck everything then

I just adapted it to 1.14 so i could merge it into my fork, you can use it if you'd like! https://github.com/JustScott/InfiniTime/tree/change_music_app_design

coffandro:
Remaking the changes

PR #1902, opened 2023-10-30
Colour change

coffandro:
This just makes most of the OS actually follow the colours found via the BG variable 
image
image


RageGamerBoi:
I agree that the colors aren’t perfect, but this a bit bland, especially the removal of the yellow notification header in the provided screenshot (and the likely removal of the colored icons in the settings menu).
Overall this is a step in the right direction, but in the future there should be a way to deal with something like the accents in a better way without removing it entirely.
Maybe some kind of accent color system in settings could work…

coffandro:
Yeah that's true and in my opinion we should add a system setting for that, either somehow in the companion apps or settings menu. I have a branch for this but haven't really made any progress yet

NitroNils:
Might I suggest an all-around "night-mode", i.e a possibility to det all colors to red-tones after dark (to preserve night vision and keeps melatonin levels up)?
Possibly in the settings where one chooses notifications and chimes?
I'm not a dev so I have no idea how this would fit in currently.

coffandro:
> Might I suggest an all-around "night-mode", i.e a possibility to det all colors to red-tones after dark (to preserve night vision and keeps melatonin levels up)? Possibly in the settings where one chooses notifications and chimes? I'm not a dev so I have no idea how this would fit in currently.
> 
> EDIT: I realize this should be its own feature request. Hopefully I'll compose it one day.

I would only be making a setting to change the colour and this one's just to unify the UI but if you do make that, please link it here

coffandro:
> I agree that the colors aren’t perfect, but this a bit bland, especially the removal of the yellow notification header in the provided screenshot (and the likely removal of the colored icons in the settings menu). Overall _**this is a step in the right direction**_, but in the future there should be a way to deal with something like these accents in a better way without removing them entirely. Maybe some kind of accent color system in settings could work…

I do agree so I'm thinking of maybe still having them but making them work under a variable still, i'm making a setting on another branch and then I'mma link it in here

ThePinkUnicorn6:
> Might I suggest an all-around "night-mode", i.e a possibility to det all colors to red-tones after dark (to preserve night vision and keeps melatonin levels up)? Possibly in the settings where one chooses notifications and chimes? I'm not a dev so I have no idea how this would fit in currently.
> 
> EDIT: I realize this should be its own feature request. Hopefully I'll compose it one day.

This has already been talked about in #1612 if you are interested in reviving interest in the idea

PR #1906, opened 2023-11-08
watchface: SlowTime

AzureOrange404:
Proposal for a new watch face:

I created a 24 hour clock analog watchface based on WatchFaceAnalog.
It displays a full 24 hour day and only features an hour and minute hand.
Battery, date etc. are still included as in original Analog watchface.

Tested in InfiniSim, but never loaded onto PineTime.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -433,6 +433,7 @@ list(APPEND SOURCE_FILES
         displayapp/screens/WatchFaceTerminal.cpp
         displayapp/screens/WatchFacePineTimeStyle.cpp
         displayapp/screens/WatchFaceCasioStyleG7710.cpp
+        displayapp/screens/WatchFaceSlowTime.cpp
 
         ##
 
--- a/src/displayapp/WatchFaces.h
+++ b/src/displayapp/WatchFaces.h
@@ -7,8 +7,9 @@ namespace Pinetime {
       Analog = 1,
       PineTimeStyle = 2,
       Terminal = 3,
-      Infineat = 4,
-      CasioStyleG7710 = 5,
+      SlowTime = 4,
+      Infineat = 5,
+      CasioStyleG7710 = 6,
     };
   }
 }
--- a/src/displayapp/screens/Clock.cpp
+++ b/src/displayapp/screens/Clock.cpp
@@ -13,6 +13,7 @@
 #include "displayapp/screens/WatchFaceAnalog.h"
 #include "displayapp/screens/WatchFacePineTimeStyle.h"
 #include "displayapp/screens/WatchFaceCasioStyleG7710.h"
+#include "displayapp/screens/WatchFaceSlowTime.h"
 
 using namespace Pinetime::Applications::Screens;
 using namespace Pinetime::Applications;
@@ -54,6 +55,9 @@ Clock::Clock(Controllers::DateTime& dateTimeController,
           break;
         case WatchFace::CasioStyleG7710:
           return WatchFaceCasioStyleG7710();
+          break;
+        case WatchFace::SlowTime:
+          return WatchFaceSlowTimeScreen();
           break;
       }
       return WatchFaceDigitalScreen();
@@ -130,4 +134,12 @@ std::unique_ptr<Screen> Clock::WatchFaceCasioStyleG7710() {
                                                              heartRateController,
                                                              motionController,
                                                              filesystem);
+}
+
+std::unique_ptr<Screen> Clock::WatchFaceSlowTimeScreen() {
+  return std::make_unique<Screens::WatchFaceSlowTime>(dateTimeController,
+                                                      batteryController,
+                                                      bleController,
+                                                      notificationManager,
+                                                      settingsController);
 }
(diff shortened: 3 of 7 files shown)


SteveAmor:
Hi, this looks like a great idea. As space will always be an issue, would it be possible to extend the existing analogue watch face based on the 12/24 hour toggle (which I see can be toggled when you have the analogue watch face selected but it's not used)?

mark9064:
I like the thinking but I'm not quite sure that would work how people expect, at least for me if I set 24h time I'd expect to have 24h when displayed digitally (be it other apps / menus or watchfaces) and 12h on an analogue clock as usual. Having to have 12h time on menus in order to have a normal analogue clock would annoy me - and there's also the change of the seconds hand disappearing

AzureOrange404:
> I like the thinking but I'm not quite sure that would work how people expect, at least for me if I set 24h time I'd expect to have 24h when displayed digitally (be it other apps / menus or watchfaces) and 12h on an analogue clock as usual. Having to have 12h time on menus in order to have a normal analogue clock would annoy me - and there's also the change of the seconds hand disappearing

Yes I see the struggle with this approach. Would annoy me too, so I don't think that's an option.

PineTimeStyle makes use of settings, maybe a settings tile could be included to the Analog watchface to switch between 12h and 24h style?

AzureOrange404:
@SteveAmor @mark9064 
I tried to integrate my SlowTime Watchface into the standard Analog Watchface by using the same settings mechanics as in the PineTime Watchface.

Like this digital clocks in other places are not affected by the change from 12h to 24h clock.
Also the second hand can be toggled in both the 12h and the 24h view.

12 hour clock
12 hour clock, no seconds
24 hour clock
24 hour clock, no seconds
settings menu

mark9064:
Testing on hardware now. Seems to work exactly as intended :)
What do the cyan markers on 24h mode represent? If it's meant to be every other hour I find it strange that it's so much bolder than the even number hour markers. Maybe cyan ticks along the border with the same size as the white ticks would work, or even just more white ticks
Remember to bump the settings version as well

AzureOrange404:
> Testing on hardware now. Seems to work exactly as intended :) 

Great, thank you :)

> What do the cyan markers on 24h mode represent? If it's meant to be every other hour I find it strange that it's so much bolder than the even number hour markers. Maybe cyan ticks along the border with the same size as the white ticks would work, or even just more white ticks 

They are meant to be the odd number markers, I had numbers for all the even hours, but it was to bulky.
The bold white markers are needed to divide minutes every fifth. If I make the cyan ticks the same, they end up between white ticks, I'm not shure if this isn't too noisy on the eye. But I could make them shorter and maybe add them for every hour (although this would mean more code as 12 and 24 needs to be skipped).

>Remember to bump the settings version as well

I'll look into it.

AzureOrange404:
@mark9064 
What do you think about this? 
24hour-marks-every-hour
Is it weard that the outer circle (minutes) is white and the inner (hours) is cyan?

mark9064:
https://github.com/InfiniTimeOrg/InfiniTime/blob/f3d4f04827fcfe65d6647271d84590d76ce2115b/src/components/settings/Settings.h#L277 this is the line to change for settings version. As you're adding a new field to the saved settings, it becomes incompatible layout-wise with anyone's previously saved settings; this is why the version needs changing.

I definitely prefer the revised hour markers, but I don't actually daily drive the analogue watch face so I feel like my opinion isn't worth listening to too much :P

mark9064:
Are you still interested in getting this merged? I'm happy to review this properly after you've rebased this if so

azureorangexyz:
> this is the line to change for settings version. As you're adding a new field to the saved settings, it becomes incompatible layout-wise with anyone's previously saved settings; this is why the version needs changing.

Okay, thank you very much, I'll look into it.
If I understand correctly I just have to increment it by one (0x0007)?

> I definitely prefer the revised hour markers, but I don't actually daily drive the analogue watch face so I feel like my opinion isn't worth listening to too much :P

Well, it's the only opinion I get, so I still value it. :P

> Edit: Just saw your edit. They're pretty even to me, but I slightly prefer the gapless variant. But as before I'm not the one to listen to

Thank you, I'll look into the design one more time in the next few days!
I thought about a little more "complicated" but nicer design, but that would increase line cound by a little bit I guess.
I don't know how much of a problem this would be and how much this adds to disk usage.

> Are you still interested in getting this merged? I'm happy to review this properly after you've rebased this if so

I'd be very interested in getting this merged! Have never contributed to a large project and would be very happy to! :smile:

mark9064:
Yep, the value needs to be incremented by one relative to the main branch. So rebase the branch first, and then change the value (it might've already been incremented since this PR began, I don't remember).

Making the design more complex should be OK, it doesn't use much extra disk space and memory is the main concern. But there is quite a lot of memory free for watchfaces, so as long as you don't go completely wild it should be fine.

Mention me once you're ready for review / have more questions and I can take a look :)
Just from a quick glance at the current changes you'll want to clean out commented out code etc. once you've settled on the final design

azureorangexyz:
@mark9064 I pushed my changes and they are ready for you to look at.

The settings version number is incremented by one.

- By long-pressing the watchface you may no change between the 12h and 24h style.
- You also can hide the seconds hand (for time passing even slower :P)

WatchFaceAnalog-12h-seconds-on WatchFaceAnalog-12h-seconds-off
WatchFaceAnalog-24h-seconds-on WatchFaceAnalog-24h-seconds-off
WatchFaceAnalog-settings

PR #1907, opened 2023-11-08
Fix brightness icon when swiping to close flashlight

### LINKED ISSUE #1751: Brightness icon changes to flashlight's brightness when flashlight is closed by swipe
### Verification

### What happened?

When the flashlight is closed by swiping, the quick settings brightness icon is set to the flashlight's brightness

The icon should represent the actual brightness

### Reproduction steps

1. Open the flashlight
2. Set it to a brightness which is not the current main brightness
3. Enable it
4. Swipe from the top edge to close it
5. The brightness icon shows the wrong brightness

### More details?

This only persists until the quick settings are closed and does not occur if the flashlight is closed with the side button. It also does not occur if the flashlight is disabled.

### Version

v1.12.0

### Companion app

_No response_

serosi:
module: DisplayApp.cpp

Fixed #1751: when swiping to close flashlight, the brightness icon now resets to the actual brightness instead of the flashlight's brightness.


JF002:
The original issue was fixed by a refactoring of the brightness introduced by the always on feature #1869. Closing this one.

Thanks for your contribution @serosi :1st_place_medal:

Issue #775, opened 2021-10-22
Labels overlap in Navigation app


JF002:
### 

- [X] I searched for similar bug reports and found none was relevant.

### What happened?

Labels (instruction + distance) overlap in Navigation app

### What should happen instead?

Labels should be positionned so they do not overlap

### Reproduction steps

Ashimokawa is working on adding support for the Navigation app in Gadgetbridge (using Osmand) : https://codeberg.org/Freeyourgadget/Gadgetbridge/issues/2387 (branch https://codeberg.org/Freeyourgadget/Gadgetbridge/src/branch/osmand-experiments)

Instruction from Osmand are quite long, so long that the label with the instruction overlaps with the label that displays the distance remaining (pics in French and German):

![osmand](https://user-images.githubusercontent.com/2261652/138475839-55b25654-92c4-4ef9-8b95-1d7a03f77323.jpg)
![gadgetbridge](https://user-images.githubusercontent.com/2261652/138475971-79598ee7-1886-45b4-b1b7-262bc2174d6c.jpg)

### More details?

The UI should probably be reworked so that the labels do not overlap:
 - crop the instruction if it's too long
 - vertically scroll the intruction
 - move the distance label on the side of the icon (either left or right)
 - ...

### Version

1.6.0

### Companion app

Gadgetbridge

Avamander:
I guess this could be fixed when navigation icons are moved to external flash.

PR #1908, opened 2023-11-08
Navigation UI tweaks

kieranc:
Now that navigation is usable on android I've been actually using it, and found that the interface needed a bit of tweaking - the text showing the road name could overflow when it was too long, and the distance to the next direction was too small to easily see when driving. This PR aims to fix these 2 issues.

Before:
InfiniSim_2023-11-08_205346

After:
InfiniSim_2023-11-08_205440


FintasticMan (approved):
I like this a lot!

JF002:
Great, thanks! I think this PR fixes #775 as well :+1:

JF002 (approved):

PR #1910, opened 2023-11-12
A binary watch face

Zetabite:
This adds a binary watch face, which displays the time in binary. The binary number is represented in equal sized dots, which are colored red if 1, and colored grey if 0. The LSB is right. The first row is hours, the second is minutes, and if enabled, the third row displays seconds. Additionally this watch face supports both 12h and 24h time format. In case of 12h, one hour dot is hidden and above the left most minute dot, text displays if its AM or PM, same as for the dots, red colored meaning this is the current time, while grayed means its not that time. I included some examples.
WatchFaceBinary_Seconds_12h
12h, with seconds enabled
WatchFaceBinary_Seconds_24h
24h, with seconds enabled
WatchFaceBinary_Without_Seconds_12h
12h, with seconds disabled
WatchFaceBinary_Without_Seconds_24h
24h, with seconds disabled


kieranc:
I like this, I have a binary clock and I would love the ability to have one on my wrist. The only problem is, my binary clock uses BCD rather than pure binary, and I'm wondering if it would be possible to make this watchface configurable between BCD and the current pure binary display?

Zetabite:
> I like this, I have a binary clock and I would love the ability to have one on my wrist. The only problem is, my binary clock uses BCD rather than pure binary, and I'm wondering if it would be possible to make this watchface configurable between BCD and the current pure binary display?

Excuse, I am not quite familiar with the term, BCD, could you elaborate?

Zetabite:
> > I like this, I have a binary clock and I would love the ability to have one on my wrist. The only problem is, my binary clock uses BCD rather than pure binary, and I'm wondering if it would be possible to make this watchface configurable between BCD and the current pure binary display?
> 
> Excuse, I am not quite familiar with the term, BCD, could you elaborate?
Ah, I assume this: https://codegolf.stackexchange.com/questions/1198/draw-a-bcd-binary-clock

Zetabite:
Ok, I looked at it, it works inherently different to my implementation, but I have no issue onto adding this watch face on a personal branch for now

Zetabite:
@kieranc Here is a working implementation: https://github.com/Zetabite/InfiniTime/tree/binary-watch-face-bcd-1.13.0

shaleh:
I have refactored @Zetabite 's branch to work with the current repo. I also made the colors set by defines in a config file because I found the Orange hard to read in the daylight. You can see my version at shaleh:binary-watchface. I can open a new PR if Zetabite is no longer around or interested.

PR #1919, opened 2023-11-27
Update VS Code config to C++20

mark9064:


JF002 (approved):
I don't use VS Code so I'll trust you on this ;-)

PR #1920, opened 2023-11-27
Fix calculation of number of app screens

mark9064:
Was testing an extra app (so 12 in total) and noticed the extra screen added
Also removed the old comment

There might be a nicer C++ way to do this than this macro I spotted lying around, but I can't find one

### DIFF
--- a/src/displayapp/screens/ApplicationList.h
+++ b/src/displayapp/screens/ApplicationList.h
@@ -38,8 +38,7 @@ namespace Pinetime {
 
         static constexpr int appsPerScreen = 6;
 
-        // Increment this when more space is needed
-        static constexpr int nScreens = (UserAppTypes::Count / appsPerScreen) + 1;
+        static constexpr int nScreens = CEIL_DIV(UserAppTypes::Count, appsPerScreen);
 
         ScreenList<nScreens> screens;
       };
--- a/src/displayapp/screens/ApplicationList.h
+++ b/src/displayapp/screens/ApplicationList.h
@@ -38,7 +38,7 @@ namespace Pinetime {
 
         static constexpr int appsPerScreen = 6;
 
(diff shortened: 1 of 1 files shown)


src/displayapp/screens/ApplicationList.h
```diff
@@ -38,8 +38,7 @@ namespace Pinetime {
 
         static constexpr int appsPerScreen = 6;
 
-        // Increment this when more space is needed
-        static constexpr int nScreens = (UserAppTypes::Count / appsPerScreen) + 1;
+        static constexpr int nScreens = CEIL_DIV(UserAppTypes::Count, appsPerScreen);
```
FintasticMan, line 41: I'm not sure where this macro comes from. I would prefer using `(UserAppTypes::Count - 1) / appsPerScreen + 1`. You should also remove the comment above as it no longer makes sense.

FintasticMan (changes requested):
Good find.

src/displayapp/screens/ApplicationList.h
mark9064: Beaten you to it with the comment, it's already gone :)
The macro is from the NRF SDK. Personally I think it's clearer semantically but I'll change it to what you suggest

FintasticMan: Thank you! While I do think it is clearer what it's doing with the macro, I don't think that it makes sense to use the nRF SDK here.

FintasticMan (approved):

src/displayapp/screens/ApplicationList.h
mark9064: That's a valid point, probably not a great dependency for this file

JF002 (approved):

JF002:
@mark9064  Good catch, thanks!

PR #1921, opened 2023-11-27
Add CodeQL Workflow for Code Security Analysis

b4yuan:
# Summary

This pull request introduces a CodeQL workflow to enhance the security analysis of this repository.

# What is CodeQL

CodeQL is a static analysis tool that helps identify and mitigate security vulnerabilities. It is primarily intra-function but does provide some support for inter-function analysis. By integrating CodeQL into a GitHub Actions workflow, it can proactively identify and address potential issues before they become security threats.

For more information on CodeQL and how to interpret its results, refer to the GitHub documentation and the CodeQL documentation (https://codeql.github.com/ and https://codeql.github.com/docs/).

# What this PR does

We added a new CodeQL workflow file (.github/workflows/codeql.yml) that
- Runs on every pull request (functionality to run on every push to main branches is included as a comment for convenience).
- Runs daily.
- Excludes queries with a high false positive rate or low-severity findings.
- Does not display results for git submodules, focusing only on our own codebase.

## Validation

To validate the functionality of this workflow, we have run several test scans on the codebase and reviewed the results. The workflow successfully compiles the project, identifies issues, and provides actionable insights while reducing noise by excluding certain queries and third-party code.

## Using the workflow results

If this pull request is merged, the CodeQL workflow will be automatically run on every push to the main branch and on every pull request to the main branch. To view the results of these code scans, follow these steps:
1. Under the repository name, click on the Security tab.
2. In the left sidebar, click Code scanning alerts.

# Is this a good idea?

We are researchers at Purdue University in the USA. We are studying the potential benefits and costs of using CodeQL on open-source repositories of embedded software.

We wrote up a report of our findings so far. The TL;DR is “CodeQL outperforms the other freely-available static analysis tools, with fairly low false positive rates and lots of real defects”. You can read about the report here: https://arxiv.org/abs/2310.00205

# Review of engineering hazards

License: see the license at https://github.com/github/codeql-cli-binaries/blob/main/LICENSE.md: 

```
Here's what you may also do with the Software, but only with an Open Source Codebase and subject to the License Restrictions provisions below:

Perform analysis on the Open Source Codebase.

If the Open Source Codebase is hosted and maintained on GitHub.com, generate CodeQL databases for or during automated analysis, CI, or CD.
```

False positives: We find that around 20% of errors are false positives, but that these FPs are polarized and only a few rules contribute to most FPs. We find that the top rules contributing to FPs are: cpp/uninitialized-local, cpp/missing-check-scanf, cpp/suspicious-pointer-scaling, cpp/unbounded-write, cpp/constant-comparison, and cpp/inconsistent-null-check. Adding a filter to filter out certain rules that contribute to a high FP rate can be done simply in the workflow file.


minacode:
This sounds very interesting, but you missed the obvious question 😉: What/how much issues where you able to find with this tool in InfiniTime?

b4yuan:
@minacode Currently on our fork of InfiniTime the CodeQL analysis shows 10 critical and high severity cases, and 2 errors:

image

Checking in any of the errors...

image

A few months ago during our data collection, we raised following: https://github.com/InfiniTimeOrg/InfiniTime/pull/1742. Devs mentioned overhead cost concerns, in which case you are able to dismiss the bugs found by CodeQL for the following reasons:

image

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,127 @@
...
+      with:
+        sarif_file: ${{ steps.step1.outputs.sarif-output }}
+        category: "/language:${{matrix.language}}"
+
+    - name: Archive CodeQL results
+      uses: actions/upload-artifact@v3
+      with:
+        name: codeql-results
+        path: ${{ steps.step1.outputs.sarif-output }}
+        retention-days: 5
```
FintasticMan, line 127: Missing newline

.github/workflows/fail_on_error.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python3
```
FintasticMan, line 1: When does this script get used? The workflow doesn't seem to run it.

.github/workflows/codeql-buildscript.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
```
FintasticMan, line 1: I would rather not add another build script, as that just makes it harder whenever we change something in the build procedure. Is there a way we can use the Docker image or somehow use the existing workflow for building the project?

FintasticMan (changes requested):
I think this could make things easier and safer. I can't see the code scanning alerts in your repository because I don't have access to it.

.github/workflows/fail_on_error.py
b4yuan: @FintasticMan Apologies, it must have not reflected for some reason. The initial motivation was to have a step that calls this script to force the workflow to fail if any errors are found. Is this something of interest to you or shall I remove it altogether?

.github/workflows/codeql-buildscript.sh
b4yuan: Updated to include build instructions in the existing `codeql.yml` workflow.

.github/workflows/codeql.yml
b4yuan: Updated

.github/workflows/codeql-buildscript.sh
FintasticMan: I think you misunderstood. I didn't mean to just move the code from the script to the workflow file. I'm not entirely sure what the point of building the firmware in the codeql workflow is. Is it to have the workflow fail if the firmware doesn't build? If so, isn't it possible to have the codeql workflow only run if the build-firmware workflow succeeded?

b4yuan: I see. Source code needs to be built: https://docs.github.com/en/code-security/code-scanning/troubleshooting-code-scanning/no-source-code-seen-during-build. If there's another workflow that builds this repo, unfortunately this workflow won't be able to analyze the build from that one. This is a limitation, but we don't believe it really hinders use.

FintasticMan: I see. Thanks for the link! Can you also use Docker to build it, or does that not work with CodeQL? I would prefer that to help with maintainability.

b4yuan: We have the option of adding the following lines to the wf:

```
# Specify the container in which actions will run
    container: 
      image: mcr.microsoft.com/cstsectools/codeql-container
      options: --user root
```

(from https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/running-codeql-code-scanning-in-a-container)

PR #1922, opened 2023-11-28
apps: Restore app list order

FintasticMan:
The build-time selection PR updated the order, this restores what it was before. I feel this has more frequently used apps on the first page, and less frequently used ones on the second. If someone has a better order, please tell me.


mark9064:
This was also the first thing I did on my testing branch xD. Feel the same as you re ordering

PR #1904, opened 2023-11-03
weather: Add support for conditions

vkareh:
Use OpenWeatherMap conditions to determine the correct condition icon. This is more accurate than attempting to calculate the condition based on precipitation and cloud cover, which often yields the wrong condition.

Companion PR to Gadgetbridge: https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/3407

PR #1924, opened 2023-12-09
SimpleWeather service : new weather implementation

JF002:
# Introduction

The Weather Service was released in InfiniTime 1.8 and, while it has been integrated by multiple companion apps (like ITD, Amazfish, Gadgetbridge,...), we only started using it with InfiniTime 1.13, when we implemented the weather display in the PTS watchface.

Unfortunately, it didn't take long before we received reports of [incorrect values being displayed](https://github.com/InfiniTimeOrg/InfiniTime/issues/1786), [strange font corruption](https://github.com/InfiniTimeOrg/InfiniTime/issues/1788) and crashes.

Multiple contributors analyzed these issues and tried to fix them ([here](https://github.com/InfiniTimeOrg/InfiniTime/pull/1904), [here](https://github.com/InfiniTimeOrg/InfiniTime/pull/1890), [here](https://github.com/InfiniTimeOrg/InfiniTime/pull/1882), [here](https://github.com/InfiniTimeOrg/InfiniTime/pull/1879), [here](https://github.com/InfiniTimeOrg/InfiniTime/pull/1860), and [here](https://github.com/InfiniTimeOrg/InfiniTime/pull/1822)) but unfortunately, we weren't able to find a satisfying solution to those issues.

So I decided to have a closer look at those bug reports to try to understand what was happening. I wrote my observations [here](https://github.com/InfiniTimeOrg/InfiniTime/pull/1822#issuecomment-1712842416). Here is a summary:

- The memory usage of the WeatherService is virtually infinite : older events are erased only when they expires. The memory usage grows if the companion app sends new data while the previous ones are still valid.
- The API does not provide any way for the companion apps to help manage the memory usage (they do not know what events were already sent by this or another companion app).
- Since multiple events of the same type can be stored in the timeline, InfiniTime does not know which value must be displayed as "current temperature", for example.
- The timeline implementation allows a lot of advanced use-cases, but is, IMHO, too complex to support simple use-cases like the one we implement right now (display the current weather on the watch face).

Like a few other contributors, I talked to developers who have already worked on the weather feature, I did many tests and experiments, I tried to find solutions to improve  the current implementation, but in the end, I couldn't find a satisfying solution to make the weather feature reliable, stable and easy to maintain.

I eventually came to the conclusion that the current implementation and API do not fit our needs, and that we will have to start designing the feature from scratch. It was not an easy decision to take : I know that many people spent a lot of their precious time working on the current implementation and on the integration of the API in companion apps. Starting from scratch seemed like a huge waste of time and energy but... I couldn't find any other solution.

To design this new implementation of the weather feature, I first listed the use-cases we want to support : display the current weather (temperature and weather condition like sun, rain, clouds,...) and forecast for the next few days.

I then talk with a few companion apps developers ([Adam](https://github.com/piggz/) and [Josef](https://github.com/jmlich/) from [Amazfish](https://github.com/piggz/harbour-amazfish), [Elara](https://gitea.elara.ws/Elara6331) from [ITD](https://gitea.elara.ws/Elara6331/itd)) so that I could design a BLE API that allows companion apps to provide the data needed by InfiniTime. Josef even provided me with a [branch in Amazfish that implements the new weather protocol](https://github.com/jmlich/harbour-amazfish/tree/weather) so that I could implement and test it in InfiniTime.

And finally, I implemented the API in InfiniTime and ensure that the data are correctly displayed in PTS.

image

image

# BLE API

The BLE API is documented in [doc/SimpleWeatherService.md](https://github.com/InfiniTimeOrg/InfiniTime/blob/simple-weather-service/doc/SimpleWeatherService.md). It's a very simple service that provide a single characteristic. The companion app uses this characteristic to write the current weather information and the forecast for the next 5 days.
Since the API uses a message type field and a version field, it can be extended in the future if we need to add more data to the weather information or implement new uses-cases.

# Implementation and integration in InfiniTime

The implementation (in [src/components/ble/SimpleWeatherService.h](https://github.com/InfiniTimeOrg/InfiniTime/blob/simple-weather-service/src/components/ble/SimpleWeatherService.h) and [src/components/ble/SimpleWeatherService.cpp](https://github.com/InfiniTimeOrg/InfiniTime/blob/simple-weather-service/src/components/ble/SimpleWeatherService.cpp)) of the API in InfiniTime first parses and validates the data from the BLE characteristic and then store those data. It does not store multiple versions of the data though : the new data always override the previous ones. SimpleWeatherService exposes those data as `std::optional` fields. `std::optional` is a tool from C++17 that manage a value that *may* or *may not* be present. This comes in handy to express that weather were received or not and if they are still valid or not.

I [integrated the SimpleWeatherService in PTS](https://github.com/InfiniTimeOrg/InfiniTime/blob/simple-weather-service/src/displayapp/screens/WatchFacePineTimeStyle.cpp#L555) to provide the same functionality than before.

# What's next

I'll leave this PR open for review for a few days to let develops provide feedback about it. I'm mostly looking for feedback from companion apps developers : what do you think of the API? Do you plan on integrating it in a future release of your app?


FintasticMan (changes requested):
I think this is quite a reasonable API given our current memory limitations and issues we've had with the current weather service.

FintasticMan:
We should also remove the old weather debug app, but that can be done in another PR.

doc/SimpleWeatherService.md
```diff
@@ -0,0 +1,68 @@
...
+ - [45] : icon ID 
+   - 0 = Sun, clear sky
+   - 1 = Few clouds
+   - 2 = Clouds
+   - 3 = Heavy clouds
+   - 4 = Clouds & rain
+   - 5 = Rain
+   - 6 = Thunderstorm
+   - 7 = snow
+   - 8 = mist, smog
```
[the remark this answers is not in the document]
JF002: Done!

src/components/ble/SimpleWeatherService.cpp
```diff
@@ -0,0 +1,160 @@
...
+namespace {
+  enum class MessageType { CurrentWeather, Forecast, Unknown };
+
+  SimpleWeatherService::CurrentWeather CreateCurrentWeather(const uint8_t* dataBuffer) {
+    char cityName[33];
+    std::memcpy(&cityName[0], &dataBuffer[13], 32);
+    cityName[32] = '\0';
+    return SimpleWeatherService::CurrentWeather {dataBuffer[2] + (dataBuffer[3] << 8) + (dataBuffer[4] << 16) + (dataBuffer[5] << 24) +
+                                                   ((uint64_t) dataBuffer[6] << 32) + ((uint64_t) dataBuffer[7] << 40) +
+                                                   ((uint64_t) dataBuffer[8] << 48) + ((uint64_t) dataBuffer[9] << 54),
```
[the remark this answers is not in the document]
JF002: Yes, you're right! I re-wrote the conversion by simply casting the data in the buffer since the bytes are already in the correct order.

src/components/ble/SimpleWeatherService.cpp
```diff
@@ -0,0 +1,160 @@
...
+        return MessageType::CurrentWeather;
+        break;
+      case 1:
+        return MessageType::Forecast;
+        break;
+      default:
+        return MessageType::Unknown;
+        break;
+    }
+  }
```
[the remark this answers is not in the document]
JF002: Yes, done! I just ensured that any other value will always be converted to MessageType::Unknown.

src/components/ble/SimpleWeatherService.h
```diff
@@ -0,0 +1,141 @@
...
+
+        uint64_t timestamp;
+        uint8_t temperature;
+        uint8_t minTemperature;
+        uint8_t maxTemperature;
+        Icons iconId;
+        char location[33]; // 32 char + \0 (end of string)
+
+        bool operator==(const CurrentWeather& other) const;
+        bool operator!=(const CurrentWeather& other) const;
```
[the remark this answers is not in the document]
JF002: Today, I learnt something, thanks!

src/displayapp/screens/StopWatch.cpp
```diff
@@ -5,6 +5,8 @@
 
 using namespace Pinetime::Applications::Screens;
 
+constexpr int Pinetime::Applications::Screens::StopWatch::maxLapCount;
+
```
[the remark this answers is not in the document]
JF002: Probably an artifacts of another development... I reverted it.

src/displayapp/screens/WatchFacePineTimeStyle.cpp
```diff
@@ -42,6 +42,41 @@ namespace {
...
+        break;
+      case Pinetime::Controllers::SimpleWeatherService::Icons::BrokenClouds:
+        return Symbols::cloud;
+        break; // TODO missing symbol
+      case Pinetime::Controllers::SimpleWeatherService::Icons::Thunderstorm:
+        return Symbols::cloud;
+        break; // TODO missing symbol
+      case Pinetime::Controllers::SimpleWeatherService::Icons::Snow:
+        return Symbols::cloud;
+        break; // TODO missing symbol
```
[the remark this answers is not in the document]
JF002: Done!

src/components/ble/SimpleWeatherService.cpp
```diff
@@ -0,0 +1,160 @@
...
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+#include <algorithm>
+#include "SimpleWeatherService.h"
+#include <cstring>
+#include <nrf_log.h>
+#include <array>
```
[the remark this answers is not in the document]
JF002: Done

src/components/ble/SimpleWeatherService.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+using namespace Pinetime::Controllers;
+
+namespace {
+  enum class MessageType { CurrentWeather, Forecast, Unknown };
```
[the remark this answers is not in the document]
JF002: Done

doc/SimpleWeatherService.md
```diff
@@ -0,0 +1,68 @@
...
+  - [16] Day 1 Icon ID
+  - [17] Day 2 Minimum temperature
+  - [18] Day 2 Maximum temperature
+  - [19] Day 2 Icon ID
+  - [20] Day 3 Minimum temperature
+  - [21] Day 3 Maximum temperature
+  - [22] Day 3 Icon ID
+  - [23] Day 4 Minimum temperature
+  - [24] Day 4 Maximum temperature
+  - [25] Day 4 Incon ID
```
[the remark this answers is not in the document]
JF002: Done

JF002:
> We should also remove the old weather debug app, but that can be done in another PR.

Thanks for the review! :+1: 

I think I addressed all the points you mentioned !

src/displayapp/screens/Symbols.h
```diff
@@ -45,6 +45,9 @@ namespace Pinetime {
         static constexpr const char* cloudShowersHeavy = "\xEF\x9D\x80";
         static constexpr const char* smog = "\xEF\x9D\x9F";
         static constexpr const char* cloud = "\xEF\x83\x82";
+        static constexpr const char* cloud_meatball = "\xEF\x9C\xBB";
```
FintasticMan, line 48: We use camelCase for variables.

```suggestion
        static constexpr const char* cloudMeatball = "\xEF\x9C\xBB";
```

FintasticMan (changes requested):

src/displayapp/screens/Symbols.h
JF002: Done, thanks again for the review!

FintasticMan (changes requested):
As I was looking to implement #1805 for this new service, I noticed that the temperatures are stored as unsigned integers, which doesn't allow for negative values! This is definitely an issue, because temperatures below 0 are very common.

I'm also wondering if we might want more precision than 1 degree. The way the previous weather service implemented it is that the temperatures were stored in int16s in degrees Celcius multiplied by 100. (This would also solve the issue where we aren't able to store temps above 127 degrees, but I don't think this is really an issue.)

JF002:
@FintasticMan You are right, temperatures should be stored as signed values! I'll fix this!

I figured that values from -128 to +127 was indeed a reasonable range (that's also a reason why the temperature are expressed in °C in the BLE protocol).
I don't have a strong opinion about the precision of the temperature. If you think we need more precision the 1 degree, we can indeed use `int16` instead of `int8`.

FintasticMan:
Yeah, I think it is best if we use `int16` for the temperature, as it allows a lot more freedom. The Fahrenheit value would also be off by 1 or 2 degrees if we only use 1 degree of Celsius precision.

src/components/ble/SimpleWeatherService.h
```diff
@@ -0,0 +1,140 @@
...
+      };
+
+      struct Forecast {
+        uint64_t timestamp;
+        uint8_t nbDays;
+
+        struct Day {
+          uint8_t minTemperature;
+          uint8_t maxTemperature;
+          uint8_t iconId;
```
FintasticMan, line 97: This should probably be an `Icons`.

```suggestion
          Icons iconId;
```

src/displayapp/screens/WatchFacePineTimeStyle.cpp
```diff
@@ -42,6 +42,41 @@ namespace {
     auto* screen = static_cast<WatchFacePineTimeStyle*>(obj->user_data);
     screen->UpdateSelected(obj, event);
   }
+
+  const char* GetIcon(const Pinetime::Controllers::SimpleWeatherService::Icons icon) {
```
FintasticMan, line 46: This is quite a universal function, so I think it'd be best to put this somewhere that other apps that want to display weather can access.

src/components/ble/SimpleWeatherService.cpp
```diff
@@ -0,0 +1,151 @@
...
+  enum class MessageType : uint8_t { CurrentWeather, Forecast, Unknown };
+
+  uint64_t ToUInt64(const uint8_t* data) {
+    return *(reinterpret_cast<const uint64_t*>(data));
+  }
+
+  SimpleWeatherService::CurrentWeather CreateCurrentWeather(const uint8_t* dataBuffer) {
+    char cityName[33];
+    std::memcpy(&cityName[0], &dataBuffer[13], 32);
+    cityName[32] = '\0';
```
FintasticMan, line 37: This seems unnecessary when the CurrentWeather initialiser already uses `std::memcpy` to copy out what we need.

src/components/ble/SimpleWeatherService.cpp
```diff
@@ -0,0 +1,151 @@
...
+    return SimpleWeatherService::CurrentWeather {ToUInt64(&dataBuffer[2]),
+                                                 dataBuffer[10],
+                                                 dataBuffer[11],
+                                                 dataBuffer[12],
+                                                 dataBuffer[13 + 32],
+                                                 cityName};
```
FintasticMan, line 43: It would be best to use parentheses to make it clear that we aren't using struct initialisation, but the constructor.

```suggestion
    return SimpleWeatherService::CurrentWeather(ToUInt64(&dataBuffer[2]),
                                                dataBuffer[10],
                                                dataBuffer[11],
                                                dataBuffer[12],
                                                dataBuffer[13 + 32],
                                                cityName);
```

src/components/ble/SimpleWeatherService.cpp
```diff
@@ -0,0 +1,151 @@
...
+      return this->forecast;
+    }
+  }
+  return {};
+}
+
+bool SimpleWeatherService::CurrentWeather::operator==(const SimpleWeatherService::CurrentWeather& other) const {
+  return this->iconId == other.iconId && this->temperature == other.temperature && this->timestamp == other.timestamp &&
+         this->maxTemperature == other.maxTemperature && this->minTemperature == other.maxTemperature;
+}
```
FintasticMan, line 151: Shouldn't this also do a strcmp for the location?

FintasticMan (changes requested):
A couple more things I noticed going through the code again :-)

src/components/ble/SimpleWeatherService.cpp
JF002: You're right! Now the memcpy is done once and the array is moved to the constructor of CurrentWeather.

src/components/ble/SimpleWeatherService.cpp
JF002: Done

src/components/ble/SimpleWeatherService.cpp
JF002: Good catch! Done!

src/displayapp/screens/WatchFacePineTimeStyle.cpp
JF002: Yes, I moved it to a new file!

src/components/ble/SimpleWeatherService.h
JF002: Done

JF002:
@FintasticMan Thanks again for this very deep review :)

I updated the protocol so that it uses int16_t instead of uint8_t to store the temperature. I'll also update the PR in Amazfish to reflect those changes.

doc/ble.md
```diff
@@ -92,7 +92,10 @@ The following custom services are implemented in InfiniTime:
 
 - Since InfiniTime 1.8:
 
-  - [Weather Service](/src/components/ble/weather/WeatherService.h): `00040000-78fc-48fe-8e23-433b3a1942d0`
+  - ~~Weather Service: `00040000-78fc-48fe-8e23-433b3a1942d0`~~ (replaced by Simple Weather Service in InfiniTime 1.14)
+
+- Since InfiniTime 1.14
+  - [Simple Weather Server](SimpleWeahterService.md) : `00050000-78fc-48fe-8e23-433b3a1942d0`
```
NeroBurner, line 98: Double Typo
```suggestion
  - [Simple Weather Service](SimpleWeatherService.md) : `00050000-78fc-48fe-8e23-433b3a1942d0`
```

NeroBurner (changes requested):

doc/ble.md
JF002: Good catch, fixed!

FintasticMan (approved):
I think this looks good now!

src/components/ble/SimpleWeatherService.cpp
```diff
@@ -0,0 +1,160 @@
...
+  enum class MessageType : uint8_t { CurrentWeather, Forecast, Unknown };
+
+  uint64_t ToUInt64(const uint8_t* data) {
+    return data[0] + (data[1] << 8) + (data[2] << 16) + (data[3] << 24) + (static_cast<uint64_t>(data[4]) << 32) +
+           (static_cast<uint64_t>(data[5]) << 48) + (static_cast<uint64_t>(data[6]) << 48) + (static_cast<uint64_t>(data[7]) << 56);
+  }
+
+  int16_t ToInt16(const uint8_t* data) {
+    return data[0] + (data[1] << 8);
+  }
```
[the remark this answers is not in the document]
JF002: I forgot to explain that! Well... The `reinterpret_cast<>()` would simply crash (go to the HardFault handler). I'm honestly not sure why tough, maybe an alignment issue? I don't know. 
Anyway, since c++ discourages reinterpret_cast and casting from unsigned to signed (and vice-versa), I figured that it would be safer and simpler to avoid sketchy optimizations on raw pointers :)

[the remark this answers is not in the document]
FintasticMan: It crashing is definitely good enough reason to not do it 😄!

doc/SimpleWeatherService.md
```diff
@@ -0,0 +1,68 @@
+# Simple Weather Service
+
+## Introduction
+
+The Simple Weather Service provide a simple and straightforward API to specify the current weather and the forecast for the next 5 days. It effectively replaces the original Weather Service (from InfiniTime 1.8) since InfiniTime 1.14
```
NeroBurner, line 5: one line per sentence for easier diffs in the future + "The .. Service provide**s**"

```suggestion
The Simple Weather Service provides a simple and straightforward API to specify the current weather and the forecast for the next 5 days.
It effectively replaces the original Weather Service (from InfiniTime 1.8) since InfiniTime 1.14.
```

src/components/ble/SimpleWeatherService.cpp
```diff
@@ -0,0 +1,160 @@
...
+#include <nrf_log.h>
+
+using namespace Pinetime::Controllers;
+
+namespace {
+  enum class MessageType : uint8_t { CurrentWeather, Forecast, Unknown };
+
+  uint64_t ToUInt64(const uint8_t* data) {
+    return data[0] + (data[1] << 8) + (data[2] << 16) + (data[3] << 24) + (static_cast<uint64_t>(data[4]) << 32) +
+           (static_cast<uint64_t>(data[5]) << 48) + (static_cast<uint64_t>(data[6]) << 48) + (static_cast<uint64_t>(data[7]) << 56);
```
NeroBurner, line 33: byte idx 5 and byte idx 6 are both shifted by 48 bits

NeroBurner (approved):
format and a small bug in uint64 conversion, afterwards all good (at least the parts I understand about the Bluetooth stack)

doc/SimpleWeatherService.md
JF002: Done

src/components/ble/SimpleWeatherService.cpp
JF002: Oof good catch! Thanks!

PR #1926, opened 2023-12-10
Watch face selection at build time

JF002:
This PR allows to select which watch faces will be built into the firmware at build time. This is done in a similar way than in #1894 : a list of watch face descriptions is generated at build time (using `constexpr` / `consteval` and template meta programming). This list is then used at run time to build the setting menu and to load the watchface when requested by the user.

Using this PR together with #1894 (which is already merged), it is now possible to build a bare minimum firmware with no user app and a single watchface. This reduces the image size from 377KB (79%) down to 311KB (65%) in flash memory.


src/displayapp/screens/settings/SettingWatchFace.h
```diff
@@ -29,26 +32,16 @@ namespace Pinetime {
         auto CreateScreenList() const;
         std::unique_ptr<Screen> CreateScreen(unsigned int screenNum) const;
 
+        static constexpr int settingsPerScreen = 4;
+        std::array<Screens::CheckboxList::Item, UserWatchFaceTypes::Count> watchfaceItems;
+        static constexpr int nScreens = UserWatchFaceTypes::Count > 0 ? (UserWatchFaceTypes ::Count - 1) / settingsPerScreen + 1 : 1;
```
FintasticMan, line 37: Given that we assert that `UserWatchFaceTypes::Count >= 1`, we don't need the check here. Of course, it might be good to keep it just in case.

```suggestion
        static constexpr int nScreens = (UserWatchFaceTypes ::Count - 1) / settingsPerScreen + 1;
```

src/displayapp/Apps.h
```diff
@@ -42,14 +43,31 @@ namespace Pinetime {
       Weather
     };
 
+    enum class WatchFace : uint8_t {
+      Digital = 0,
+      Analog = 1,
+      PineTimeStyle = 2,
+      Terminal = 3,
+      Infineat = 4,
+      CasioStyleG7710 = 5,
```
FintasticMan, line 52: It's unnecessary to assign these values.

```suggestion
      Digital,
      Analog,
      PineTimeStyle,
      Terminal,
      Infineat,
      CasioStyleG7710,
```

FintasticMan (changes requested):
This looks good! Just a couple of small details.

src/displayapp/screens/settings/SettingWatchFace.h
JF002: Yes, I agree but... I have the feeling that if I read this code in a few months, I'll have forgetten about that asssert and this that there's a bug :-D. Since this is a `constexpr` evaluation, it shouldn't hurt to keep it. What do you think?

src/displayapp/Apps.h
JF002: On the code perspective, you are right, it's not necessary to assign specific values to the field. However, since those specific value are part of the specification of the protocol, I think it's safer to make them explicit. This way, we'll think twice before changing them.

JF002: You are right, I removed those assignations.

src/displayapp/screens/settings/SettingWatchFace.h
FintasticMan: There's definitely no harm in keeping it of course. I think it's fine as it is, I just noticed this as I was reading the code :-)

FintasticMan (approved):

PR #1928, opened 2023-12-17
User Applications selection using CMake

JF002:
Use CMake's configure_file() functionality to generate the list of User Applications.

All the apps included in current versions of InfiniTime are enabled by default, but this can now be overridden by setting variables ENABLE_APP_XXX to True or False.

CMake CMP0140 is set to NEW to enable the return PROPAGATE functionality.


FintasticMan (changes requested):
This looks good! It allows a lot more freedom for building with whatever apps you want. I would say we should add this to the Apps.md documentation before merging, as you now need to register new apps in more places.

JF002:
> This looks good! It allows a lot more freedom for building with whatever apps you want. I would say we should add this to the Apps.md documentation before merging, as you now need to register new apps in more places.

Thanks! I've just updated the doc to add instructions related to CMake.

FintasticMan (approved):

FintasticMan:
It seems like InfiniSim will also need to include these same option declarations. @NeroBurner do you know if it would be possible to have InfiniSim get the options from InfiniTime?

CMakeLists.txt
```diff
@@ -35,6 +34,19 @@ endif()
 set(TARGET_DEVICE "PINETIME" CACHE STRING "Target device")
 set_property(CACHE TARGET_DEVICE PROPERTY STRINGS PINETIME MOY-TFK5 MOY-TIN5 MOY-TON5 MOY-UNK)
 
+option(ENABLE_APP_STOPWATCH "Enable the Stopwatch application" True)
```
NeroBurner, line 37: We can move those options and the Apps.h.in file into its own directory, create a CMake header only library target and link that to everywhere we need it. Because it is in its own folder I can easily include it in InfiniSim afterwards. Like we do for the fonts and images

CMakeLists.txt
```diff
@@ -70,5 +82,29 @@ set(VERSION_EDIT_WARNING "// Do not edit this file, it is automatically generate
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/src/Version.h.in ${CMAKE_CURRENT_BINARY_DIR}/src/Version.h)
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/docker/post_build.sh.in ${CMAKE_CURRENT_BINARY_DIR}/post_build.sh)
 
+function(AddToListIfEnabled list enabled type)
+  if(${enabled})
+    list(APPEND ${list} ${type})
+  endif ()
+  #return(PROPAGATE ${list})
```
NeroBurner, line 89: Not enabled, but comment in PR sais it is

FintasticMan: The version of CMake in the docker container is too old for this. See the second commit in this PR.

minacode:
I don't know much about CMake, but could we put all optional apps in a separate directory and then autogenerate the CMake options?

```
.../displayapp
    apps/
        music/
            music.cpp
            music.h
        twos/
            twos.cpp
            twos.h
        ...
```

FintasticMan:
I would imagine that that would be possible, and it would improve the dev experience quite a lot, not having to remember every single place to add a new app 😄

FintasticMan:
One thing I've just thought of is that this doesn't allow choosing the order the apps appear in the app list. I think that that is quite an important feature, so we should try to figure out a way to get that to work.

One simple way would just be to have a string CMake parameter that gets placed in UserAppTypes.

CMakeLists.txt
JF002: I'll have a look at this!

JF002:
@minacode 
> I don't know much about CMake, but could we put all optional apps in a separate directory and then autogenerate the CMake options?

It might be possible, yes. Actually, I would like InfiniTime to support User Applications coming from external repositories. This would allow maintainers of those user apps to maintain their apps the way they want, and to reduce the load on InfiniTime maintainers. I think we can do this in another future PR though :)

@FintasticMan
 > One thing I've just thought of is that this doesn't allow choosing the order the apps appear in the app list. I think that that is quite an important feature, so we should try to figure out a way to get that to work.

Good point! I'll try to find a solution for this!

FintasticMan (approved):

CMakeLists.txt
JF002: Done  in [c200816](https://github.com/InfiniTimeOrg/InfiniTime/pull/1928/commits/c200816416e09b890237625ab27d947904843eae) and integrated in InfiniSim here : https://github.com/InfiniTimeOrg/InfiniSim/pull/134

CMakeLists.txt
```diff
@@ -51,6 +45,12 @@ string(STRIP "${PROJECT_GIT_COMMIT_HASH}" PROJECT_GIT_COMMIT_HASH)
 
 message("PROJECT_GIT_COMMIT_HASH_SUCCESS? " ${PROJECT_GIT_COMMIT_HASH_SUCCESS})
 
+if(DEFINED ENABLE_USERAPPS)
```
NeroBurner, line 48: I'd move this variable definition into src/displayapp/apps/CMakeLists.txt. then we'll have one place to update the default apps list and won't need to change the default list in InfiniSim as well.

Otherwise I'd need to update InfiniSim everytime the app list is changed to match InfiniTimes default

JF002: Good point, done!

JF002:
@NeroBurner I'll apply the same change to the watchface selection when you confirm this is ok for you :)

mark9064:
What about apps that introduce settings pages specific to them? We don't actually have any right now (arguably steps), but some are proposed e.g background heart rate settings

minacode:
Handling such dependencies would be a next step. Apps also come with custom BLE interfaces (e.g. music).

JF002:
@mark9064 That's a good question! And I agree with @minacode, we can figure that out in a next step. In this current state, the worse that can happen is that settings will show a setting that no userapp will actually use, but the whole firmware will still be functional.

doc/code/Apps.md
```diff
@@ -59,7 +59,7 @@ and then the function `create` is called to create an instance of the app.
 
 The list of user applications is generated at build time by the `consteval` function `CreateAppDescriptions()`
 in `UserApps.h`. This method takes the list of applications that must be built into the firmware image.
-This list of applications is defined as a list `Apps` enum values named `UserAppTypes` in `Apps.h`.
+This list of applications is defined as a list `Apps` enum values named `UserAppTypes` in `Apps.h`.
```
NeroBurner, line 62: trailing space

```suggestion
This list of applications is defined as a list `Apps` enum values named `UserAppTypes` in `Apps.h`.
```

doc/code/Apps.md
```diff
@@ -159,6 +159,12 @@ If your application is a **user** application, you don't need to add anything in
 everything will be automatically generated for you.
 The user application will also be automatically be added to the app launcher menu.
 
+Since the list of **user** application is generated by CMake, you need to add the variable `ENABLE_USERAPPS` to the command line of CMake. This variable must be set with a string composed of an ordered list of the **user** applications that must be built into the firmware. The items of the list are fields from the enumeration `Apps`. Ex : build the firmware with 3 user application : Alarm, Timer and MyApp (the application will be listed in this specific order in the application menu).
```
NeroBurner, line 162: use one line per sentence, makes the diffs in future easier, and "user application**s**"

```suggestion
Since the list of **user** applications is generated by CMake, you need to add the variable `ENABLE_USERAPPS` to the command line of CMake.
This variable must be set with a string composed of an ordered list of the **user** applications that must be built into the firmware.
The items of the list are fields from the enumeration `Apps`.
Ex : build the firmware with 3 user applications : Alarm, Timer and MyApp (the applications will be listed in this specific order in the application menu).
```

src/displayapp/apps/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+else ()
+    set(USERAPP_TYPES "Apps::Navigation, Apps::StopWatch, Apps::Alarm, Apps::Timer, Apps::Steps, Apps::HeartRate, Apps::Music, Apps::Paint, Apps::Paddle, Apps::Twos, Apps::Metronome" CACHE STRING "List of user apps to build into the firmware")
+endif ()
+
+add_library(infinitime_apps INTERFACE)
+target_sources(infinitime_apps INTERFACE "${CMAKE_CURRENT_BINARY_DIR}/Apps.h")
+target_include_directories(infinitime_apps INTERFACE "${CMAKE_CURRENT_BINARY_DIR}/")
+
+# Generate the list of user apps to be compiled into the firmware
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/Apps.h.in ${CMAKE_CURRENT_BINARY_DIR}/Apps.h)
```
NeroBurner, line 12: add newline to end of file

```suggestion
configure_file(${CMAKE_CURRENT_SOURCE_DIR}/Apps.h.in ${CMAKE_CURRENT_BINARY_DIR}/Apps.h)

```

NeroBurner (approved):
small typos, afterwards good to go from my side :tada:

PR #1935, opened 2023-12-24
Fix git hash calculation with Docker build

### LINKED ISSUE #1842: Short ref doesn't appear in CI builds
### Verification

### What happened?

When checking the About or Firmware pages, if running a build from GitHub CI, there will be no short ref

There should be one

### Reproduction steps

1. Download and install a [release build](https://github.com/InfiniTimeOrg/InfiniTime/releases)
2. Go to the about page and see a commit reference next to short ref (for 1.13, this is `5f3acfad`)
3. Download and install a [CI build](https://github.com/InfiniTimeOrg/InfiniTime/actions/runs/5364794676) (this is the 1.13 commit)
4. Go to the about page and notice a distinct lack of a short ref

### Version

v1.13.0

### Companion app

Gadgetbridge v0.75.1

febrezo:
Fix error when compiling the Pinetime using the Docker image. If done with Docker, the container does not trust the /sources folder, leading to a blank response of the command that grabs the git commit `git rev-parse --short HEAD`.

```
fatal: detected dubious ownership in repository at '/sources'
To add an exception for this directory, call:

        git config --global --add safe.directory /sources
PROJECT_GIT_COMMIT_HASH_SUCCESS? 128

BUILD CONFIGURATION
-------------------
    * Mode : Release
    * Version : 1.3.0
    * Toolchain : /opt/gcc-arm-none-eabi-10.3-2021.10
    * GitRef(S) :
    * NRF52 SDK : /opt/nRF5_SDK_15.3.0_59ac345
    * Target device : PINETIME
    * Build DFU (using adafruit-nrfutil) : Enabled
    * Build resources : Enabled
```

If the `git config --global --add safe.directory /sources` is added to the Dockerfile, the problem is solved and the hash is added correctly.

Fixes #1842

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -64,6 +64,9 @@ RUN bash -c "source /opt/build.sh; GetMcuBoot;"
 
 # Add the infinitime user for connecting devcontainer
 RUN adduser infinitime
-    
+
+# Configure Git to accept the /sources directory as safe
+RUN git config --global --add safe.directory /sources
+
 ENV SOURCES_DIR /sources
 CMD ["/opt/build.sh"]


FintasticMan:
This is likely the cause for #1842. If possible I would prefer us to figure out why the ownership is dubious and fix that before just adding marking the source dir as safe.

NeroBurner (approved):
I think this is a valid fix for dev containers as described here: https://www.kenmuse.com/blog/avoiding-dubious-ownership-in-dev-containers/

image

JF002 (approved):

JF002:
> This is likely the cause for #1842. If possible I would prefer us to figure out why the ownership is dubious and fix that before just adding marking the source dir as safe.

Do you think we should block this PR until we figure that out, or can we merge it in the meantime?

FintasticMan (approved):
It's fine to merge it in the meantime. I've just updated this to the latest main.

PR #1937, opened 2023-12-25
Subpixel font rendering

0x0000ff:
I've enabled support for subpixel fonts in the lv_conf.h file and the generate.py script. The "subpixel" boolean option can be added to the options in the fonts.json file to generate subpixel fonts.

### DIFF
--- a/src/displayapp/fonts/generate.py
+++ b/src/displayapp/fonts/generate.py
@@ -18,8 +18,10 @@ class Source(object):
         self.symbols = d.get('symbols')
 
 
-def gen_lvconv_line(lv_font_conv: str, dest: str, size: int, bpp: int, sources: typing.List[Source], compress:bool=False):
+def gen_lvconv_line(lv_font_conv: str, dest: str, size: int, bpp: int, sources: typing.List[Source], compress:bool=False, subpixel:bool=False):
     args = [lv_font_conv, '--size', str(size), '--output', dest, '--bpp', str(bpp), '--format', 'lvgl']
+    if subpixel:
+        args.append('--lcd')
     if not compress:
         args.append('--no-compress')
     for source in sources:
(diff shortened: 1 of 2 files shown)


0x0000ff:
I thought I'd just add this as well for your future consideration.

I wanted to use this to get the text smaller whilst keeping it's readability, in order to get more test and more usable screen space.

I lowered the size of the main font to 15, with and without subpixels, and and there is a noticeable difference:
InfiniSim_2023-12-16_110531 InfiniSim_2023-12-16_110707

I then lowered the font to 12, just to see what would happen, and the differences are more notable with this enabled: 
InfiniSim_2023-12-16_111309 InfiniSim_2023-12-16_111110

I also tried using a different font. I changed it to Open_Sans_light and put the size back to 15:
InfiniSim_2023-12-25_125816 InfiniSim_2023-12-25_125901

The characters are very thin with it enabled, and they also don't seem to be vary visible in bright backgrounds:
InfiniSim_2023-12-25_125811 InfiniSim_2023-12-25_125854


So I used the medium version of OpenSans, which made the text readable again:
InfiniSim_2023-12-25_151240 InfiniSim_2023-12-25_151115
 InfiniSim_2023-12-25_151512 InfiniSim_2023-12-25_141206

src/displayapp/fonts/generate.py
FintasticMan, line 1: You probably want to apply these changes to `src/resources/font-generate.py` as well.

0x0000ff:
I should probably add a comparison between this and anti-aliasing at 2bpp
InfiniSim_2023-12-31_114613InfiniSim_2023-12-31_114449

escoand:
The first image looks much better, at least on my mobile display.

0x0000ff:
You'll likely see vivid colour banding on the fonts on some screens due to the nature of subpixles and how a display's pixels are arranged. On the watch's screen it is rendered correctly.

This is using the subpixel font
IMG_20231231_161706
...And this is one with anti-aliasing


They look similar to each other, but I think the subpixel fonts are more sharper and brighter.  It does increase the filesize and has some colours banding, but I think it would be good for makeing small fonts legible.

kieranc:
I like this, AA or similar for fonts has been something nice to have which we haven't had the space for for ages. 

I'm wondering how much additional space it consumes, both vs current setup and vs 2bpp AA, and also how jarring it'll be if only some fonts are smoothed and others not?

PR #1938, opened 2023-12-28
Fixes #1895: use just 2 bytes header according to GATT specification

### LINKED ISSUE #1895: AlertNotification Service/Client are not compliant to the spec
### Verification

### What happened?

AlertNotification Service/Client use a 3 bytes header, the spec defines a 2 bytes header

AlertNotification Service/Client should comply to the BLE spec.

### Reproduction steps

N/A

### More details?

This issue was reported by the [Amazfish community](https://github.com/piggz/harbour-amazfish/pull/286) : InfiniTime expects the header of all notifications is a 3 bytes header ([here](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/src/components/ble/AlertNotificationService.cpp#L50) and [here](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/src/components/ble/AlertNotificationClient.cpp#L147)).

However, the GATT spec [specifies a header of 2 bytes](https://github.com/oesmith/gatt-xml/blob/master/org.bluetooth.characteristic.new_alert.xml).

Now, even if InfiniTime is not 100% compliant, all companion apps have currently implemented this 3 bytes header, and changing this might break the compatibility with those companion apps so I'm not sure if we should fix this. 
If we do want to change the header size, we'll have to communicate with companion app developers to ensure smooth transition.

It might be possible to support both 2 and 3 bytes header by checking the value of the 3rd byte of the buffer : it can be ignored if it's equal to 0x00 or lower than 0x20.

The [documentation](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/doc/ble.md#new-alert) should however be completed to mention that this is not completely compliant to the spec of the AlertNotificationService.

### Version

<=1.13.0

### Companion app

All

jmlich:
In my opinion the implementation should be fixed to comply with specification. 
The extra byte is just skipped

### DIFF
--- a/src/components/ble/AlertNotificationService.cpp
+++ b/src/components/ble/AlertNotificationService.cpp
@@ -47,7 +47,7 @@ AlertNotificationService::AlertNotificationService(System::SystemTask& systemTas
 int AlertNotificationService::OnAlert(struct ble_gatt_access_ctxt* ctxt) {
   if (ctxt->op == BLE_GATT_ACCESS_OP_WRITE_CHR) {
     constexpr size_t stringTerminatorSize = 1; // end of string '\0'
-    constexpr size_t headerSize = 3;
+    constexpr size_t headerSize = 2;
     const auto maxMessageSize {NotificationManager::MaximumMessageSize()};
     const auto maxBufferSize {maxMessageSize + headerSize};
 


FintasticMan:
I don't think the 3rd byte is ignored though? It'll still be copied into the notification string, and thus the max message size is practically 1 char smaller when using a header of 3 bytes.

jmlich:
You are right. It seems the 3rd byte should be send only in case when Category CustomHuami is used:

https://github.com/Freeyourgadget/Gadgetbridge/blob/275deb4d06de6b64c0d7e73d13f6f2ad2879fe25/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/btle/profiles/alertnotification/AlertNotificationProfile.java#L106 

https://github.com/piggz/harbour-amazfish/blob/1462454cbc9489fead578ccad2ce030c35e0b329/daemon/src/services/alertnotificationservice.cpp#L20

I will update my pull request according to that.

PR #1942, opened 2023-12-29
PineTimeStyle: fix brief display of weather dummy data

liamcharger:
When navigating into the PineTimeStyle watch face and weather display is on, the weather icon is set as "cloudSunRain" until it is updated to the correct icon in the Refresh function. (This behavior is most prominent when navigating from the Quick Settings.) The "temperature" variable is not set until the Refresh function is called, so it displays "Text" until it is updated in the function. This PR aims to fix that by setting the default value of temperature to "--" and the weather icon to "ban".


FintasticMan (approved):
Nice catch!

JF002 (approved):

PR #1945, opened 2024-01-02
Simple Weather Service : Fix timestamp

JF002:
In the documentation, specify that the timestamp is expressed in seconds from epoch (instead of nanoseconds). SimpleWeatherService now uses "localtime" (GetCurrentDateTime()) instead of UTC time.

### DIFF
--- a/doc/SimpleWeatherService.md
+++ b/doc/SimpleWeatherService.md
@@ -28,7 +28,7 @@ The byte array must contain the following data:
 
  - [0] : Message type = `0`
  - [1] : Message version = `0`
- - [2][3][4][5][6][7][8][9] : Timestamp (64 bits UNIX timestamp, number of nanoseconds elapsed since 1 JAN 1970)
+ - [2][3][4][5][6][7][8][9] : Timestamp (64 bits UNIX timestamp, number of seconds elapsed since 1 JAN 1970)
  - [10, 11] : Current temperature (°C * 100)
  - [12, 13] : Minimum temperature (°C * 100)
  - [14, 15] : Maximum temperature (°C * 100)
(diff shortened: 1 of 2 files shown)


NeroBurner (approved):

FintasticMan (changes requested):
I think we should mention in the doc that we expect the timestamp to be in local time.

JF002:
> I think we should mention in the doc that we expect the timestamp to be in local time.

Yes, right! Done!

FintasticMan (approved):

doc/SimpleWeatherService.md
```diff
@@ -28,7 +28,7 @@ The byte array must contain the following data:
 
  - [0] : Message type = `0`
  - [1] : Message version = `0`
- - [2][3][4][5][6][7][8][9] : Timestamp (64 bits UNIX timestamp, number of nanoseconds elapsed since 1 JAN 1970)
+ - [2][3][4][5][6][7][8][9] : Timestamp (64 bits UNIX timestamp, number of seconds elapsed since 1 JAN 1970)  in local time (the same timezone than the one used to set the time)
```
NeroBurner, line 31:
```suggestion
 - [2][3][4][5][6][7][8][9] : Timestamp (64 bits UNIX timestamp, number of seconds elapsed since 1 JAN 1970)  in local time (the same timezone as the one used to set the time)
```

doc/SimpleWeatherService.md
```diff
@@ -50,7 +50,7 @@ The byte array must contain the following data:
 
   - [0] : Message type = `0`
   - [1] : Message version = `0`
-  - [2][3][4][5][6][7][8][9] : Timestamp (64 bits UNIX timestamp, number of nanoseconds elapsed since 1 JAN 1970)
+  - [2][3][4][5][6][7][8][9] : Timestamp (64 bits UNIX timestamp, number of seconds elapsed since 1 JAN 1970) in local time (the same timezone than the one used to set the time)
```
NeroBurner, line 53:
```suggestion
  - [2][3][4][5][6][7][8][9] : Timestamp (64 bits UNIX timestamp, number of seconds elapsed since 1 JAN 1970) in local time (the same timezone as the one used to set the time)
```

NeroBurner (approved):
small wording change, afterwards I'm good again :)

doc/SimpleWeatherService.md
JF002: Done

Issue #1949, opened 2024-01-04
BLE Motion Service


jones139:
### Verification

- [X] I searched for similar issues (including closed issues) and found none was relevant.

### Introduce the issue

I have been experimenting with the BLE Motion Service - I was very pleased to see you have implemented this, because I was about to look into the code to work out how to do it myself, thank you!
I have spotted a couple of things that I would like to fix, but wanted to check if these are intentional first:
  - I think the units of x, y, z acceleration are in raw accelerometer counts?  I wonder if we should report them in milli-g (which will fit into a 16 bit integer still) - I noticed this when I tried changing the accelerometer rage and the values changed.
  - I am only managing to receive data to my Android app at about 10 Hz, even though the accelerometer is running at 100Hz by default - I wonder if this is because we are only sending a single measurement at a time and there is a connection overhead?

### Preferred solution

I'd suggest that we:

- Convert the accelerometer signal into units of milli-g (as a 16 bit integer) - probably at the driver level as this is where the instrument range is set.
- Maintain a small buffer (18 byte) of accelerometer signals (x1, y1, z1, x2, y2, z2, x3, y3, z3) and only notify subscribers when that buffer is full so we download three data points at once rather than one - I am managing at least 25 Hz transfer rate on a BangleJS using this method.   I suggest 18 bytes because I think there is a 20 byte limit in some BLE implementations, and I would want to keep all the x,y,z values of a particular measurement together and not split them.  (25Hz is all that I want from the service - I would need to down-sample if we transferred at 100Hz).

But was it implemented in the current way on purpose for a reason I have missed?   Are many people using the BLE motion service, because changing the units might break their apps? (I could do the conversion to milli-g on the phone, but I am trying to have all of my data sources use the same unit).

### Version

1.13.0 (or rather my modified version that keeps the BLE services running when the watch screen switches off (https://github.com/OpenSeizureDetector/InfiniTimeSD)

jones139:
I have implemented the idea of buffering the accelerometer values within MotionService, and was a bit surprised that I am still only getting a 10Hz sample frequency.
I have delved back into the code a bit further, and I don't think it is working like I was expecting - Am I right that we are just polling the accelerometer within the SystemTask loop?
When I saw the accelerometer was running at 100Hz, I assumed it would be interrupt driven, or at least in a high speed task of its own.

So I think that to achieve a higher sample frequency we have a few options:

[...]

PR #1950, opened 2024-01-05
Change Acceleration Values to milli-g units rather than Raw Counts

jones139:
This change adds the required scaling factors to the Bma421 accelerometer driver so that the accelerometer readings can be converted to milli-g for any of the valid instrument ranges (the default 2g up to 16g).

This addresses part of issue #1949.

To end users the change should not be significant because the scaling factor is such that for the default 2g range, the raw values are very close to the milli-g values anyway - it is much more noticeable if you switch the device range as this change will mean the returned values are unchanged (previously the acceleration values halved if you selected the +/-4g range).

I have also updated the Motion screen to display the milli-g values to check it is working.

I have checked it on a physical pinetime device using both the +/-2g and +/-4g ranges to confirm that the static value returned is close to 1000 mg.


FintasticMan:
I like the idea of normalising the motion values so that the same value corresponds to the same force, regardless of the precision it's configured for. I do think it's a shame that conversion to milli-Gs from the raw values is a (slightly) lossy calculation. Probably not really an issue though.

I do know that the current raise to wake and lower to sleep implementations rely on 1G being 1024, and they would need to be adjusted slightly if it were 1000 instead (it would make the trig slightly harder as we use 32767 as the range for sin, which is just `1024 * 32 - 1`). I presume that things such as Sleep as Android (and maybe OpenTracks) also rely on that.

Maybe we can compromise and use mibi-Gs (a term I just came up with for a 1024th of a G ;p)?

jones139:
Strangely on my device I am seeing over 1000 milli-g so I think there is an offset (or calibration error) on the accelerometer in my watch.
I do not think that the 1000/1024 reduction in precision will really be significant unless you are trying to do something very precise, which is unlikely on  a watch?

You are right though that we do not want to make calculations any more complex because I am already concerned about the amount of calculations we are doing, in case that is causing the issue I am seeing with not being able to achieve 25Hz sampling via the BLE MotionService.

I could make it return raw values and provide a conversion function toMillig() that would convert the values in the structure to milli-g on demand if you prefer?

Graham.

FintasticMan:
I would say that normalizing the data so that 1024 == 1G makes sense, and providing a function that converts to milli-Gs. All the function would need to do is `val * 1000 / 1024`.

jones139:
Ok - I'll update it to return a 'binary milli-g' where 1G=1024 counts for all ranges - then we do not need to change anything else - it will be this evening (UK time) before I do it though.

jones139:
> The reason you aren't seeing the 25Hz on the BLE service is likely due to the fact that the SystemTask refresh is 10Hz, and updating the service only happens every SystemTask refresh.

Ah, that is it then - I do see about 10Hz.   I'll have a think about that because I definitely need 25Hz, but will be happy with returning several samples at each refresh, so if I can get buffering of the samples working the 10Hz refresh will be ok - thanks for pointing that out as I had missed it!

jones139:
Commit c261579 does the change to 'binary milli-g'

src/drivers/Bma421.cpp
```diff
@@ -24,6 +24,17 @@ namespace {
...
 
+// Scale factors to convert accelerometer counts to milli-g
+// from datasheet: https://files.pine64.org/doc/datasheet/pinetime/BST-BMA421-FL000.pdf
+// The array index to use is stored in accel_conf.range
+const short Bma421::accelScaleFactors[] = {
+  1024,  // LSB/g +/- 2g range
+  512,   // LSB/g +/- 4g range 
+  256,   // LSB/g +/- 8g range
+  128    // LSB/g +/- 16g range
+};
```
FintasticMan, line 35: This should be `constexpr` rather than `const`, so it can be evaluated at compile-time and `int16_t` instead of `short`. I would also move it into the anonymous namespace above so that it's only available in the current compilation unit. I would also recommend that you explicitly set the indices.

```suggestion
constexpr int16_t accelScaleFactors[] = {
  [BMA4_ACCEL_RANGE_2G] = 1024,  // LSB/g +/- 2g range
  [BMA4_ACCEL_RANGE_4G] = 512,   // LSB/g +/- 4g range 
  [BMA4_ACCEL_RANGE_8G] = 256,   // LSB/g +/- 8g range
  [BMA4_ACCEL_RANGE_16G] = 128   // LSB/g +/- 16g range
};
```

src/displayapp/apps/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 if(DEFINED ENABLE_USERAPPS)
     set(USERAPP_TYPES ${ENABLE_USERAPPS} CACHE STRING "List of user apps to build into the firmware")
 else ()
-    set(USERAPP_TYPES "Apps::Navigation, Apps::StopWatch, Apps::Alarm, Apps::Timer, Apps::Steps, Apps::HeartRate, Apps::Music, Apps::Paint, Apps::Paddle, Apps::Twos, Apps::Metronome" CACHE STRING "List of user apps to build into the firmware")
+    set(USERAPP_TYPES "Apps::Navigation, Apps::StopWatch, Apps::Alarm, Apps::Timer, Apps::Steps, Apps::Motion, Apps::HeartRate, Apps::Music, Apps::Paint, Apps::Paddle, Apps::Twos, Apps::Metronome" CACHE STRING "List of user apps to build into the firmware")
```
FintasticMan, line 4: You shouldn't add the Motion app to the default list. You can add it for individual builds with `-DUSERAPP_TYPES` on the cmake command-line.

FintasticMan (changes requested):
A couple of small things.

src/drivers/Bma421.cpp
jones139: Setting the indices explicitly is very neat - I didn't know you could do that, thanks (I don't know any of the C++ fancy bits - I'm just trying to treat it like C with classes!).

src/displayapp/apps/CMakeLists.txt
jones139: Yes, this one should not really have been part of the pull request.

src/drivers/Bma421.cpp
FintasticMan: Setting the indices is even possible in plain old C as well! As long as you're using C99 or later that is.

jones139: Ha!  C99 isn't 'plain old C' - I remember the debates about whether to use K&R or ANSI C coding style in a project!

jones139:
> > The reason you aren't seeing the 25Hz on the BLE service is likely due to the fact that the SystemTask refresh is 10Hz, and updating the service only happens every SystemTask refresh.
> 
> Ah, that is it then - I do see about 10Hz. I'll have a think about that because I definitely need 25Hz, but will be happy with returning several samples at each refresh, so if I can get buffering of the samples working the 10Hz refresh will be ok - thanks for pointing that out as I had missed it!

This is more complicated than I had hoped - buffering within MotionService did not help - I still only achieved 10Hz sample frequency (but had to read the characeristic less often).   I think we are polling the accelerometer from within SystemTask, rather than using the chip's FIFO and interrupt capabilities?   I have put a few thoughts here:  https://github.com/InfiniTimeOrg/InfiniTime/issues/1949#issuecomment-1880201731

I'd welcome any thoughts on the best way to implement higher frequency sampling without breaking other parts of InfiniTime!

FintasticMan:
I've just commented in the issue. If you can move the array out of the class, into the anonymous namespace and rebase to fix the conflicts, I'd be happy to approve this PR. Then we could figure out where to go from there afterwards.

src/displayapp/apps/CMakeLists.txt
jones139: I think I have pulled in the latest version from the upstream repository

jones139:
I think this is now aligned with the latest commits to the upstream repository - is that what you meant by 'rebase'?

FintasticMan (approved):

FintasticMan:
@jones139 Could you fix the formatting issues as shown by the test-format check?

jones139:
I could not work out how to find what the issue was - do I have to run lint
locally to get the list of problems?

FintasticMan:
If you go to the details page for the failing check, then click summary, there should be a file you can download called Patches. This is a zip file with patches you can apply to your code that will fix the formatting. It is probably better to install and use clang-format yourself though, as then you can format the code before you commit and push.

JF002 (approved):
Those changes look good to me. 
If I understand correctly, the values that are exposed on the BLE Motion service will also change, right? If so, I think we should also update [the documentation](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/doc/MotionService.md) to reflect that change.

FintasticMan:
In the end, we decided on a way of doing this in which the values aren't changed in the default configuration. 1G is still 1024. The change comes when you change the range of the accelerometer, as then 1G stays 1024. Before, 1G would be lower if the range was increased.

jones139:
I have updated the MotionService documentation to make it clear that the unit are "binary milli-g" because the units were not specified on that page previously.

PR #1958, opened 2024-01-10
Up to date calendar app

JustScott:
This calendar app was originally created by @thnikk in pull request #923, then was updated in June of 2023 by @Boteium to work with the most recent Infinitime version at that time. In the most recent version of Infinitime, 1.14 in January of 2024 (now), the way apps are handled has changed, so in this PR I've updated the code to work with those changes and to be able to be merged into version 1.14 of Infinitime. I've tested the pinetime-mcboot-app version on my sealed Pinetime and have yet to find and issue with it.

Also, I commented out the swipe up and down to increase and decrease the current calendar year feature as that interfered with the swipe down to exit ability users have come to expect when navigating apps... feel free to uncomment the code if you'd like that feature to stay.


JustScott:
I just saw the 'PRs are either rebase or squash merged' part in https://github.com/InfiniTimeOrg/InfiniTime/blob/ecf2f564f7a0884b6acdfdf5530abe2b98cb9aa9/doc/maintainer-guide.md#reviewing-prs.

I don't squash merges or rebase often so I'm not great at it, but I can try if this is required, just let me know!

vkareh:
@JustScott the easiest way is to run `git rebase -i HEAD~3` (where 3 is the number of commits that you want to squash).

In there it will give you instructions on how to use it:
- mark all but the first commit to be squashed
- write a proper commit message
- save
- run `git push -f` to update the PR

Note: You might have issues doing the above, since I think the branch you sourced this from has the actual calendar implementation buried deep in the git history. I would start over with a clean branch off of `main` and just cherry-pick the commit (or just copy/paste the code), then fix and squash.

JustScott:
@vkareh Okay I reset this PRs (local) branch to the latest upstream commit, then cherry picked code from @thnikk's & @Boteium's branches onto my branch (which helped me avoid some of the uneeded legacy code, thanks for the suggestion!).

Since I didn't bring along any of the prior commits from the other branches, and can just put all of my changes under a single commit now, doesn't that mean I don't need to do any squashing?

FintasticMan:
Thanks for making this new PR. Could you attribute @thnikk and @Boteium in the commit? You can do that by adding `Co-authored-by: <commit author>` to the commit message.

vkareh:
@JustScott
> Since I didn't bring along any of the prior commits from the other branches, and can just put all of my changes under a single commit now, doesn't that mean I don't need to do any squashing?

That's correct. The point of squashing is to merge multiple commits, which is no longer necessary in your case.

I also agree with @FintasticMan about adding attribution. You can change the commit message using the same `git rebase -i HEAD~1` command and choosing `reword` (or just `r`) as the command. Once you save and exit, it will open up the editor with the commit message to edit. Save again, exit, `git push -f`.

src/displayapp/screens/Calendar.cpp
```diff
@@ -0,0 +1,88 @@
...
+    today.year = static_cast<int>(dateTimeController.Year());
+    today.month = static_cast<int>(dateTimeController.Month());
+    today.day = static_cast<int>(dateTimeController.Day());
+
+    // Set today's date
+    lv_calendar_set_today_date(calendar, &today);
+    lv_calendar_set_showed_date(calendar, &today);
+
+    // Use today's date as a reference for which month to show if moved
+    current = today;
```
vkareh, line 40: Do we need different `current` and `today` variables? The app always opens with the current date in view, and the `lv_calendar_set_today_date` never gets overridden (and also there's no refresh call). I think it would work the same if you used a single `lv_calendar_date_t` variable, but you'd save some memory.

src/displayapp/screens/Calendar.h
```diff
@@ -0,0 +1,42 @@
...
+  namespace Applications {
+    namespace Screens {
+      class Calendar : public Screen {
+      public:
+        Calendar(Controllers::DateTime& dateTimeController);
+        ~Calendar() override;
+      private:
+        bool OnTouchEvent(TouchEvents event);
+        Controllers::DateTime& dateTimeController;
+        lv_obj_t* label_time;
```
vkareh, line 25: This label is not used anywhere

JustScott:
So what are the thoughts on the swipe up and down to increase and decrease the current year of the calendar? I personally don't like this as it interferes with the user's ability swipe down to exit the app, but I'll leave it in if others want this capability.

vkareh:
@JustScott 
> @vkareh I believe I addressed all the concerns above with this last commit. I also tested this on my pinetime and everything is working as expected.

Awesome! You can still squash the new commit you made if you want to clean things up further. Not sure if it's necessary though.

> I'm also wondering what the thoughts on the ability to swipe up and down to increase and decrease the current year of the calendar are? I personally don't like this as it interferes with the user's ability swipe down to exit the app, but I'll leave it in if others want this capability.

I saw that and thought the same thing. It would certainly be useful, but interfering with the global swipe mechanics feels wrong. You would need to use the button to exit, so it's a tradeoff between fuctionality and consistency. I would leave it up to the repo maintainers to determine that.

JustScott:
@vkareh I figured I'd wait until this was ready to be merged to the main branch to squash the commits so the maintainers can more clearly see what's changed. Also, I appreciate all the help you've given me with rebase commands above!

src/displayapp/screens/Calendar.cpp
```diff
@@ -0,0 +1,95 @@
...
+  // Set alignment
+  lv_obj_align(calendar, NULL, LV_ALIGN_IN_BOTTOM_MID, 0, -5);
+  // Disable clicks
+  lv_obj_set_click(calendar, false);
+
+  // Set style of today's date
+  lv_obj_set_style_local_text_color(calendar, LV_CALENDAR_PART_DATE, LV_STATE_FOCUSED, LV_COLOR_RED);
+
+  // Set style of inactive month's days
+  lv_obj_set_style_local_text_color(calendar, LV_CALENDAR_PART_DATE, LV_STATE_DISABLED, lv_color_hex(0x505050));
```
minacode, line 39: Can we use a color from the InfiniTime theme here?

JustScott: I can replace the red with `Colors::deepOrange`, but `Colors::lightGray` doesn't look quite right. Would it be alright if I added something like `static constexpr lv_color_t gray = LV_COLOR_MAKE(0x50, 0x50, 0x50);` to the `Colors` namespace?

minacode: Sounds good to me

ZekeSmith:
Very nice app, I am personally a fan of the original implementation from #923 in which the calendar was placed to the right of the clock. I have found as I add more applications to infinitime its really nice to be able to access the calendar so simply. I'd personally like it as a default, however it might be a better option for a future pull request to customize the location of the launcher, setting and notification gestures too.

https://github.com/InfiniTimeOrg/InfiniTime/assets/81902668/7b72d291-21ec-4a83-87c3-583d71ca2f88

I find it also feels a bit more natural for the vertical swiping with this setup as its closer to a traditional calendar.

src/displayapp/screens/Calendar.cpp
```diff
@@ -0,0 +1,96 @@
...
+        current.year++;
+        lv_calendar_set_showed_date(calendar, &current);
+        return true;
+    }
+    case TouchEvents::SwipeDown: {
+        current.year--;
+        lv_calendar_set_showed_date(calendar, &current);
+        return true;
+    }
+    */
```
mark9064, line 87: Can be removed

adrian4096:
Maybe not the time and place, but are there any plans to add event support from gadgetbridge?

Eve1374:
Hi, I like this calendar and rebased it on top of main. I pulled @JustScott 's merge_calendar_upstream branch and fixed an error (see below), now what would be the clean way to update this PR ?
The fix needed is adding following in the structure at the end of Calendar.h
`      static bool IsAvailable(Pinetime::Controllers::FS& /*filesystem*/) { ` 
`         return true; ` 
`       };` 

Thanks @JustScott

mark9064:
I don't know if the original PR author is still around, it's been a long time. Also not sure what the status of this PR is, is the code all done? In that case it's waiting for review

Eve1374:
@mark9064 I think the code is done (except for the rebase I made locally). I'm using it since more than 6 month, it works smoothly and does not crash. I don't know how to upload on this PR the rebase I made.

JustScott:
> Hi, I like this calendar and rebased it on top of main. I pulled @JustScott 's merge_calendar_upstream branch and fixed an error (see below), now what would be the clean way to update this PR ? The fix needed is adding following in the structure at the end of Calendar.h ` static bool IsAvailable(Pinetime::Controllers::FS& /*filesystem*/) {` ` return true;` ` };`
> 
> Thanks @JustScott

Everything built correctly and the simulator ran fine... so I'm assuming I added you're change correctly. Thanks for the fix @Eve1374!

JustScott:
> I don't know if the original PR author is still around, it's been a long time. Also not sure what the status of this PR is, is the code all done? In that case it's waiting for review

It's been a long time since I worked on this PR, but I think I applied the changes everyone suggested back then, and now with eve fixing this more recent issue, everything should be up to date. I'd say it definitely should be reviewed again, after sitting for such a long time.

Eve1374:
One other change that I would suggest is adding a setting to set first time of the week to either sunday or monday. Now in this branch it is sunday ; in my fork I changed to monday without bothering to make it a setting

AkiroPhi:
In the issue #2402 https://github.com/InfiniTimeOrg/InfiniTime/issues/2402#issuecomment-3720895202, it's seems that the 24h/12h settings does that. I don't know if the PR use any of the things discuss in there. And I don't know either if this app can be used to fix the issue.

mark9064:
Good points

It would be nice if it respected the first day of week preference, which currently is Sun for 12h and Mon for 24h

PR #1959, opened 2024-01-10
settings: Add global widget selection

vkareh:
Instead of each watch face implementing their own settings for which widgets to display, we can have a global selection of widgets. All watch faces can then determine whether it is enabled and so display it in whichever way makes sense for that face.

Current widgets supported are heart rate, step counter, and weather.

Widgets off:
InfiniSim_2024-01-10_161937
InfiniSim_2024-01-10_161850 InfiniSim_2024-01-10_161904 InfiniSim_2024-02-05_134208

Widgets on:
InfiniSim_2024-01-10_162123
InfiniSim_2024-02-05_134638 InfiniSim_2024-02-05_134518 InfiniSim_2024-02-05_134257


FintasticMan (approved):
I think this is very useful, thanks!

minacode:
I think you could use the `CheckboxList` class for the settings menu.

vkareh:
@minacode the badly-named `CheckboxList` class only deals with single-selection settings (radio buttons)

minacode:
Oh, ok. Maybe we should rename it to something like `RadioButtonGroup`.

Do you think you could still extract an abstract component from your implementation? I could imagine that an actual checkbox list is common enough. 

Maybe it doesn't make sense though. That would be ok. I am just asking.

vkareh:
> Oh, ok. Maybe we should rename it to something like `RadioButtonGroup`.

No, the name is based on the lvgl library, which uses checkbox as the naming. There is no radio component in lvgl, so the checkboxlist class emulates the radio behaviour by re-theming the checkboxes and overriding the select mechanism.

> Do you think you could still extract an abstract component from your implementation? I could imagine that an actual checkbox list is common enough.

I did an initial PoC, but it means that now CheckboxList needs to return arrays of settings, whether it's a checkbox or a radio. This and the additional code required to make that work makes the entire thing to complex/bloated.

> Maybe it doesn't make sense though. That would be ok. I am just asking.

I can poke at it some more at another time, but likely out of scope for this PR.

minacode:
Ok, nice. Thanks for the detailed answer. Given that, I think your PR is fine the way it is now :)

vkareh:
I've added new commits for configurable widgets on the Analog watch face. Attached screenshots to the description on top.

mark9064:
I think it would be really great if we could get this merged soon, would you be able to fix build (CI is currently failing)? I can review after

vkareh:
Fixed the build now.

src/displayapp/screens/WatchFacePineTimeStyle.cpp
```diff
@@ -527,34 +520,36 @@ void WatchFacePineTimeStyle::Refresh() {
...
+        lv_label_set_text(temperature, "--");
+        lv_label_set_text(weatherIcon, Symbols::ban);
       }
-      lv_label_set_text_fmt(temperature, "%d°", temp);
-      lv_label_set_text(weatherIcon, Symbols::GetSymbol(optCurrentWeather->iconId));
-    } else {
-      lv_label_set_text(temperature, "--");
-      lv_label_set_text(weatherIcon, Symbols::ban);
     }
-    lv_obj_realign(temperature);
```
mark9064, line 556: I'm not great at LVGL, but just want to check that removing these realigns is intentional

mark9064 (commented):
Looks great! Just a few things:
- In the future there may be some way to modify settings over bluetooth. At the moment changing the enabled widgets could crash a watchface. I think it would be better if the watchfaces store the enabled widgets when they initialise. Maybe as simple as assigning the lvgl object fields to nullptr if they haven't been drawn and then checking if they're not null in `Refresh()`?
- The settings version needs bumping too

shaleh:
I have been pondering a refactor of watch faces so there is a core "watch" and the watch faces load this. The face would pass in refresh handlers and init handlers for all of the mini widgets -- weather, steps, bluetooth, battery, etc etc. This would allow for a consistent behavior of the widgets and the faces to decide how and what they render. This PR is a solid step in that direction.

src/displayapp/screens/settings/SettingWidgets.cpp
```diff
@@ -0,0 +1,77 @@
...
+  void event_handler(lv_obj_t* obj, lv_event_t event) {
+    auto* screen = static_cast<SettingWidgets*>(obj->user_data);
+    if (event == LV_EVENT_VALUE_CHANGED) {
+      screen->UpdateSelected(obj);
+    }
+  }
+}
+
+SettingWidgets::SettingWidgets(Pinetime::Controllers::Settings& settingsController) : settingsController {settingsController} {
+  lv_obj_t* container1 = lv_cont_create(lv_scr_act(), nullptr);
```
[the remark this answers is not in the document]
vkareh: fixed.

src/displayapp/screens/WatchFacePineTimeStyle.cpp
vkareh: Good catch. I readded the align.

PR #1960, opened 2024-01-10
Updated docker build scripts

jones139:
Updated docker build scripts to provide a simple command line, and allow building of a single target (e.g. pinetime-mcuboot-app) without errors from post_build.sh

From the main InfiniTime repository folder I do:
`./docker/runBuild pinetime-mcuboot-app`
to create a dfu file for just the main InfiniTime firmware without re-building the other components.

I still haven't found out how to solve the issue with the version number not updating when you change it in CMakelists.txt - so might add that change to this Pull Request if I fix it tomorrow.

### DIFF
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -74,7 +74,7 @@ CmakeGenerate() {
 CmakeBuild() {
   local target="$1"
   [ -n "$target" ] && target="--target $target"
-  cmake --build "$BUILD_DIR" --config $BUILD_TYPE $target -- -j$(nproc)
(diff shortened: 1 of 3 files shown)


jones139:
I have added in a change to src/CMakelists.txt which forces SystemInfo.cpp to re-compile whenever Version.h changes (which will usually be because of a change to the version number in ./CMakelists.txt).
This fixes the issue with the DFU .zip file having one version number in the filename, but the version number displayed in the SystemInfo (About) screen being different (Discussion #1956)

PR #1963, opened 2024-01-13
WatchFaceAnalog: Use status icons class

vkareh:
Instead of recreating the status icons for BLE and battery, we can reuse the StatusIcon class to automatically display them.


vkareh:
This doesn't change any functionality, it just reuses the existing StatusIcon class rather than duplicate its code in the Analog face.

FintasticMan:
This has been discussed before, but I'm willing to discuss it further: #1336.

vkareh:
@FintasticMan ah, I see the point. I wasn't aware of that discussion, but it makes sense with regards to the display overlap. It feels like a useful widget, the status icons, but it's being used only in a single context, which is unfortunate.

PR #1964, opened 2024-01-13
BatteryIcon: Change color with charge percentage

vkareh:
Silly little change to dynamically color the battery based on the charge left.

battery

### DIFF
--- a/src/displayapp/screens/BatteryIcon.cpp
+++ b/src/displayapp/screens/BatteryIcon.cpp
@@ -29,15 +29,7 @@ void BatteryIcon::SetBatteryPercentage(uint8_t percentage) {
   lv_obj_set_height(batteryJuice, percentage * 14 / 100);
   lv_obj_realign(batteryJuice);
   if (colorOnLowBattery) {
-    static constexpr int lowBatteryThreshold = 15;
-    static constexpr int criticalBatteryThreshold = 5;
-    if (percentage > lowBatteryThreshold) {
-      SetColor(LV_COLOR_WHITE);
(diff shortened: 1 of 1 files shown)


minacode:
Nice idea, I like it a lot! 😊
I think you could improve the contrast a little by using HSV colors and mapping the percentage to the hue interval between red and green.
[Lvgl can do the conversion to RGB.](https://docs.lvgl.io/7.11/overview/display.html?highlight=hsv#_CPPv419lv_color_hsv_to_rgb8uint16_t7uint8_t7uint8_t)

vkareh:
@minacode That's a brilliant suggestion! I've updated the code and added a new screenshot - looks way better, thanks!

z3t0:
I love this change. Thank you.

mark9064:
Could I suggest a colour curve such as `-e^{-4x}+1` (where x is the battery between 0 and 1)? This way it only starts tinting yellow around 25% ish and is fully yellow around 15%

mark9064:
https://github.com/InfiniTimeOrg/InfiniTime/commit/c451d2ce9fae0c491a5e516a85f2faf1cf25273e

Forgot to ping you but the colour algorithm is in main already for the terminal watchface :)

I think it'd make sense to use it here too

vkareh:
> [c451d2c](https://github.com/InfiniTimeOrg/InfiniTime/commit/c451d2ce9fae0c491a5e516a85f2faf1cf25273e)
> 
> Forgot to ping you but the colour algorithm is in main already for the terminal watchface :)
> 
> I think it'd make sense to use it here too

Ha! That was easy, thanks :)

PR #1965, opened 2024-01-13
WatchFacePineTimeStyle: Fix conditional in weather display

vkareh:
Since returning a valid weather is always considered an updated value, if the current weather is empty, the face will attempt to display the temperature and icon as empty values, rather than clearing the labels.


FintasticMan (approved):
Thanks for finding this!

PR #1970, opened 2024-01-15
Fix motion service not updating when sleeping

### LINKED ISSUE #1843: Motion Controller issue when screen off
### Verification

### Introduce the issue

Hello, 

    I'm not sure since when, but i have an issue regarding Bluetooth availability of Motion stats with "Sleep as Android", when screen is on, no issue motion is reported correctly, but when screen go to off, it doesn't report any more. 

    Not sure, if this behaviour is intended, or is it just me.

   Thank you for all your work !

### Preferred solution

_No response_

### Version

v1.13.0

FintasticMan:
This PR fixes the motion service not being updated when the device is sleeping. This would only happen in the case that there is no motion-based wake-up mode enabled.

Thanks to @apilat for the implementation.

Fixes #1843.
Supersedes #1207.

### DIFF
--- a/src/components/ble/MotionService.cpp
+++ b/src/components/ble/MotionService.cpp
@@ -120,3 +120,7 @@ void MotionService::UnsubscribeNotification(uint16_t attributeHandle) {
   else if (attributeHandle == motionValuesHandle)
     motionValuesNoficationEnabled = false;
 }
+
+bool MotionService::IsMotionNotificationSubscribed() const {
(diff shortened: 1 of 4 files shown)


NeroBurner (approved):
The changes look good. I can only assess the new code as I don't know much about the BLE services and when the 'issubscribed' boolean is set.

Could you also open an MR on InfiniSim to update the motion service there as well?

PR #1971, opened 2024-01-15
timer: Add ringing and counter

vkareh:
The timer app currently issues a short buzz once and then disappears. There is no trace left that the timer finished or how long ago. This change makes the motor start ringing and presents a timer counter. The motor will continue the ringing pattern for 10 seconds and the timer counter will finally reset after 60 seconds.


mark9064:
Works great, huge usability improvement for the timer. Been running this for a while and it makes me a more consistent cook for sure!

I think it would be nice if the alarm and timer had different vibration patterns, but vibration needs to be overhauled anyway so this is good to go for now IMO

src/components/timer/Timer.cpp
```diff
@@ -12,11 +12,13 @@ void Timer::StartTimer(std::chrono::milliseconds duration) {
 }
 
 std::chrono::milliseconds Timer::GetTimeRemaining() {
+  TickType_t remainingTime = 0;
   if (IsRunning()) {
-    TickType_t remainingTime = xTimerGetExpiryTime(timer) - xTaskGetTickCount();
-    return std::chrono::milliseconds(remainingTime * 1000 / configTICK_RATE_HZ);
+    remainingTime = xTimerGetExpiryTime(timer) - xTaskGetTickCount();
+  } else {
+    remainingTime = xTaskGetTickCount() - xTimerGetExpiryTime(timer);
```
mark9064, line 19: If the timer is not running, `xTimerGetExpiryTime` is undefined according to the docs.
> If the timer is running then the time in ticks at which the timer will next expire is returned. If the timer is not running then the return value is undefined.

mark9064 (commented):
I see why you changed this (count up after ringing), but we probably want to avoid relying on UB

src/components/timer/Timer.cpp
vkareh: huh interesting. The reason it has worked for me so far is probably because RTOS hasn't allocated data to that space and so the subtraction yields the expected value.

All it takes is another xTask running to mess the timer count-up...

I see three possible paths forward then:
- Drop the count-up feature (and show some other text, etc.)
- Implement a counter at the application level to not rely on the xTimer
- Change the timer application so that the ringing is based on math, but the xTimer remains active until the user dismisses it. I think this is a no-go, as it adds a lot more unnecessary complexity.

I'm considering just dropping the count-up feature, since a count-up that resets at 60 seconds is only marginally useful (and bumping that to something like 10 minutes sounds unnecessary, but I may be wrong). But I'm open to implementing it at the application level if folks are interested in this.

Thoughts?

mark9064: Personally I'm a big fan of the counting up and I'd love to see it stay. Could record the tick at which the timer expires alongside the timer and then do `(current tick - expiry tick) / tick rate`?

rnwgnr:
Tested this today while cooking and this is indeed a great usability improvement for the timer.

Until now i missed a lot of the alarms, the vibration is just to short. Great work.

src/components/timer/Timer.cpp
vkareh: Finally got around to fixing this.

src/components/timer/Timer.cpp
```diff
@@ -12,11 +12,13 @@ void Timer::StartTimer(std::chrono::milliseconds duration) {
 }
 
 std::chrono::milliseconds Timer::GetTimeRemaining() {
+  TickType_t remainingTime = 0;
   if (IsRunning()) {
-    TickType_t remainingTime = xTimerGetExpiryTime(timer) - xTaskGetTickCount();
-    return std::chrono::milliseconds(remainingTime * 1000 / configTICK_RATE_HZ);
+    remainingTime = xTimerGetExpiryTime(timer) - xTaskGetTickCount();
```
mark9064, line 17: On further reflection, we shouldn't change the semantics of this method

The Timer component is meant to be a generic component that multiple applications can be use (it should be just a thin wrapper around FreeRTOS timers)

One idea:
- When setting a timer, the expiry tick time is stored in this class
- New method GetTimeSinceExpired which returns xtaskgettickcount - expiry, or 0 (if running or stopped)
- This method remains pretty much unchanged
- Inside the timer screen we calculate the value as GetTimeRemaining if the timer is running, else with GetTimeSinceExpired

The only part of this I don't really like is the semantics of GetTimeSinceExpired when it hasn't expired yet or has been stopped, and also GetTimeRemaining when stopped or expired

Another idea:
Method GetTimerStatus which returns a variant of
- Stopped
- Running: time to expiry
- Expired: time since expiry

This way semantics are always clear

Curious to know what you're thinking, any of these make sense? Not sure I've got any perfect solutions here, though I quite like the second one so far (haven't thought about either for too long)

DavisNT:
There are many cases when 10 seconds of vibration might not be very noticeable (e.g. when playing sports games, using some vehicles, doing woodworking etc.). Also attention is an important key factor (it is much easier to notice an alarm when all the attention is on testing that alarm; the same alarm might go unnoticed if attention is focused on something else).

Could it be possible to introduce a setting which would allow to select vibration length (e.g. 10 seconds, 60 seconds and infinite) for Timer and Alarm Clock?

Also, I think that timers, alarm clocks and similar reminders should be very reliable, noticeable and persistent, because consequences of reminders failing to go off or going off unnoticed are much worse than the consequences of reminders always needing manual action to silence them.
I.e. I think that timer and alarm clock should continue to vibrate until the button is pressed, otherwise there will be cases when people miss important things due to missing the alarms.

What do other users of InfiniTime think about this?

P.S. I am **not** InfiniTime maintainer, I am InfiniTime user.

marigoldfish:
I like the idea of a settings menu where we can choose between different vibration durations and "until a button is pressed." I'd probably set mine to something like 30 seconds. That being said, the 10-second ring is still a lot more useful to me than the single buzz. This PR is at the top of my wishlist, for sure

mark9064:
@vkareh this goes for a few of your other PRs as well: I see that the PRs are rebased regularly while review is pending, is there anything I could clarify with review feedback? Hope all's well, totally get it if you're busy :)

src/components/timer/Timer.cpp
vkareh: > The Timer component is meant to be a generic component that multiple applications can be use (it should be just a thin wrapper around FreeRTOS timers)

Is it, though? That's what the `xTimer` interface is meant to do. This wrapper into a `Timer` component is only used for the Timer app. That said, I think this `Timer` class could still work elsewhere in the code as is, if there was a use case for it. If a new use case shows up, and a change is needed, we can do it then - not sure creating a super generic class is useful without knowing what use cases it's meant to address.

That's just my opinion though, and I may be entirely wrong (and I'm just having a hard time finding time to work on this these days - which is why I'm hesitant to get started on a change I probably won't finish).

src/displayapp/DisplayApp.cpp
```diff
@@ -343,14 +343,17 @@ void DisplayApp::Refresh() {
         if (state != States::Running) {
           PushMessageToSystemTask(System::Messages::GoToRunning);
         }
+        // Load timer app if not loaded
+        if (currentApp != Apps::Timer) {
+          LoadNewScreen(Apps::Timer, DisplayApp::FullRefreshDirections::Up);
+        }
+        // Once loaded, set the timer to ringing mode
         if (currentApp == Apps::Timer) {
```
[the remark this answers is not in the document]
vkareh: Yes, this is redundant, but the builder fails when initializing the `*timer` screen as a direct clause of the `case`. Kind of baffled about this one, but I can only build successfully if I check that the current app is the Timer before I try to initialize the Timer screen. :shrug:

src/displayapp/screens/Timer.cpp
```diff
@@ -129,16 +142,31 @@ void Timer::SetTimerRunning() {
...
+  lv_obj_set_style_local_bg_color(btnPlayPause, LV_BTN_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_GREEN);
+}
+
+void Timer::SetTimerRinging() {
+  isRinging = true;
+  minuteCounter.HideControls();
+  secondCounter.HideControls();
+  lv_label_set_text_static(txtPlayPause, "Reset");
+  lv_obj_set_style_local_bg_color(btnPlayPause, LV_BTN_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_RED);
+  timer.SetExpiredTime();
```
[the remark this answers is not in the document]
vkareh: Without calling this, there's no count-up in the timer: it just stays at `00:00` while ringing. Not sure I want to skip this feature - I find it way too useful - and so you need to tell the Timer component what the expiry time is so that the `GetTimeRemaining` call returns the correct value (otherwise `expired` remains at 0 and nothing happens).

vkareh:
> Could it be possible to introduce a setting which would allow to select vibration length [...]

No, I feel that's unnecessary complexity. Let's find a good default and hardcode it.

> There are many cases when 10 seconds of vibration might not be very noticeable [...]

I'm willing to bump this to 15, 30, or 60 seconds if folks agree.

vkareh:
@mark9064 
> @vkareh this goes for a few of your other PRs as well: I see that the PRs are rebased regularly while review is pending, is there anything I could clarify with review feedback? Hope all's well, totally get it if you're busy :)

Thanks @mark9064 - I'm quite busy these days, so I can only keep up with rebases and minimal changes. That said, if there's clear direction on how to address some of the nuanced change requests, I'm willing to put some time, but I'm saving most of my brain capacity for work, so I can't see myself going through the FreeRTOS API and debug undefined behaviour or things like that :exploding_head: 

A lot of your review suggestions sound very reasonable, but everything I've tried/tested points to "this is the only way to make it work, apart from reimplementing the xTimer interface in-app". I use this timer every single day, multiple times a day, in many configurations, all of them real life scenarios, and I have not had a single instance of it behaving any different than what I intended in the code.

marigoldfish:
For what it's worth, this branch has been my daily driver for several months (thanks @vkareh for always keeping it rebased - you're the MVP). I've not run into any issues whatsoever, and I use the timer quite frequently for cooking, etc. 

10 sec of vibration is usually sufficient for my needs, but I would not be opposed to a longer default, like 30.

JustScott:
Loving this vkareh, great job! One issue I'm having is that the buzzing for the expired time will go forever if the display falls asleep before it reaches 10 seconds, along with it counting past 60 seconds if the displays asleep (I can see it quickly reset the timer once I wake the display. This would be an issue if I started a timer and took my watch off, leaving it to vibrate til I got back.

My guess is that the lv_task doesn't run Refresh while the display is sleeping. This isn't an issue while using the always on display.

JustScott:
@vkareh I fixed the infinite buzzing issue I raised above in a PR on your fork [here](https://github.com/vkareh/InfiniTime/pull/1). It just creates a wake lock while the exipred timer is less than 11 seconds and the motorController is buzzing, and releases it after.

src/displayapp/DisplayApp.cpp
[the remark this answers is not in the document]
vkareh: Fixed

src/displayapp/screens/Timer.cpp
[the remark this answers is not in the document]
vkareh: Fixed

src/components/timer/Timer.cpp
mark9064: OK, so I've discussed this with the team a bit and we definitely think it's better as a generic component to avoid coupling to FreeRTOS so tightly. It also makes implementing InfiniSim easier.

This was discussed before in #1650 - in fact the the timer class used to be a controller specific to the timer app, and it was made generic with that PR

So I think we need to change/extend the API provided here with generic usage in mind

Specifically the Timer as implemented in InfiniTime only needs to be a one-shot resettable timer that calls a callback specified when creating the timer.

I'm happy to extend the timer class to fit the use case of this PR if that works for you. I can send up a PR for that, it can get merged in quickly and then you can rebase onto it (we're trying to get 1.16 released soon and it'd be great to land this PR)

How does that sound?

mark9064:
It's a bit hidden so just in case I responded to the timer method semantics problem here: https://github.com/InfiniTimeOrg/InfiniTime/pull/1971/files#r2492311746

Would be super great to get this in for 1.16 which we're hoping to release in the next few days :)

src/components/timer/Timer.cpp
vkareh: @mark9064 
> I'm happy to extend the timer class to fit the use case of this PR if that works for you. I can send up a PR for that, it can get merged in quickly and then you can rebase onto it (we're trying to get 1.16 released soon and it'd be great to land this PR)
> 
> How does that sound?

That sounds good - I'm happy to rebase against that.

JF002:
I've just tested this PR and it looks very good, thanks @vkareh !
One thing I noticed : when the timer reaches 00:00, it looks like it stays on that value for 2 seconds before showing 00:01. It's not a big deal, but I think it makes the "elapsed for xx seconds" value off by one second.

See this video : https://github.com/user-attachments/assets/2d6ba5c3-43c0-42fa-92e8-1779267c42b2

mark9064:
I think this is the correct behaviour since the time always rounds to zero.

```
actual | shown
 5.00 | 5
 4.99 | 4
 4.00 | 4
...
 1.00 | 1
 0.99 | 0
 0.00 | 0
-0.99 | 0
-1.00 | 1
-1.99 | 1
-2.00 | 2
```
so you should expect it to show 0 for 2 seconds (from time +1 to -1)

JF002:
@mark9064 Good point, that makes sense! :+1:

mark9064:
@vkareh #2365 is in :) let me know if it all works ok when you get around to rebasing on to it

DavisNT:
Could it be possible to add an option to ring indefinitely (in many cases people will not notice the 10 seconds ringing, especially if the timer is used during physical activity/sports)?

mark9064:
I totally get where you're coming from but I think it'd be best to choose one sensible default here. At least how I've been using the timer 10s ringing has felt good to me, even if I'm moving about I definitely notice it before it stops. I'm not sure if you've had the chance to try this PR yet, but if not maybe give it a go and see how you find 10s? Maybe it's more noticeable than you'd think :)

vkareh:
@mark9064 - just rebased. Had to fix some conflicts, but it all led to basically dropping the last commit, as it became a no-op :tada:

src/displayapp/screens/Timer.cpp
```diff
@@ -130,16 +149,30 @@ void Timer::SetTimerRunning() {
...
 
 void Timer::SetTimerStopped() {
+  isRinging = false;
   minuteCounter.ShowControls();
   secondCounter.ShowControls();
   lv_label_set_text_static(txtPlayPause, "Start");
+  lv_obj_set_style_local_bg_color(btnPlayPause, LV_BTN_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_GREEN);
+}
+
+void Timer::SetTimerRinging() {
```
mark9064, line 172: Take the wakelock here. Once you've locked it, you only need to release it once (but it's fine to do it repeatedly if easier)

Set the motor controller ringing here

src/displayapp/screens/Timer.cpp
```diff
@@ -62,7 +63,9 @@ Timer::Timer(Controllers::Timer& timerController) : timer {timerController} {
   // Create the label as a child of the button so it stays centered by default
   txtPlayPause = lv_label_create(btnPlayPause, nullptr);
 
-  if (timer.IsRunning()) {
+  if (motorController.IsRinging()) {
```
mark9064: Check the timer, see if it's expired and if so if the duration is < 10s

src/displayapp/DisplayApp.cpp
```diff
@@ -368,19 +368,21 @@ void DisplayApp::Refresh() {
...
+        // Load timer app if not loaded
+        if (currentApp != Apps::Timer) {
           LoadNewScreen(Apps::Timer, DisplayApp::FullRefreshDirections::Up);
         }
-        motorController.RunForDuration(35);
+        // Set the timer to ringing mode
+        lv_disp_trig_activity(nullptr);
+        auto* timerScreen = static_cast<Screens::Timer*>(currentScreen.get());
+        timerScreen->SetTimerRinging();
+        motorController.StartRinging();
```
mark9064: Move this functionality into the timer app

I think the timer should manage this state itself - since it has to interact with the motor controller directly to stop ringing it may as well manage setting it ringing too, doing it in two places seems confusing

src/displayapp/screens/Timer.cpp
```diff
@@ -103,7 +106,23 @@ void Timer::UpdateMask() {
 }
 
 void Timer::Refresh() {
-  if (timer.IsRunning()) {
+  if (isRinging) {
+    DisplayTime();
+    if (motorController.IsRinging()) {
```
mark9064: This check wouldn't be needed anymore

vkareh:
@mark9064 
> I'd really like to see this use the `expired` field of the timer rather than tracking that state externally. I've pointed out a few consequences of doing that but there's definitely more changes that would follow

These are really good suggestions. I've implemented them and it made the code a lot cleaner. Thanks!

src/displayapp/DisplayApp.cpp
```diff
@@ -368,19 +368,20 @@ void DisplayApp::Refresh() {
...
-        } else {
+        // Load timer app if not loaded
+        if (currentApp != Apps::Timer) {
           LoadNewScreen(Apps::Timer, DisplayApp::FullRefreshDirections::Up);
         }
-        motorController.RunForDuration(35);
+        // Set the timer to ringing mode
+        lv_disp_trig_activity(nullptr);
+        auto* timerScreen = static_cast<Screens::Timer*>(currentScreen.get());
+        timerScreen->SetTimerRinging();
```
mark9064: If the timer app is loaded L376, `SetTimerRinging()` gets called twice right? (once in the constructor, once here)

src/displayapp/screens/Timer.cpp
```diff
@@ -73,6 +78,15 @@ Timer::Timer(Controllers::Timer& timerController) : timer {timerController} {
 
 Timer::~Timer() {
   lv_task_del(taskRefresh);
+
+  // If timer has expired, reset it when leaving the screen
+  auto timerStatus = timer.GetTimerState();
+  if (timerStatus && timerStatus->expired) {
+    motorController.StopRinging();
+    wakeLock.Release();
```
mark9064: Wakelock destructor automatically releases it so this can be omitted

mark9064 (approved):
Think that's all good from me :D thanks for sticking through the review

Haven't tested it on my watch yet but planning on doing so very soon

mark9064:
@JF002 still good to merge on your end?

JF002:
> @JF002 still good to merge on your end?

Yes, I'll merge it! Thanks @vkareh and @mark9064  for your work on this!

PR #1972, opened 2024-01-16
Fix bluetooth setting not actually changing bluetooth radio settings.

### LINKED ISSUE #1961: Bluetooth setting is behaving abnormally
### Verification

### What happened?

Opening bluetooth setting turns off bluetooth on ui only, toggling bluetooth off does not work when connected to companion app

Opening bluetooth setting should not turn off bluetooth, disabling bluetooth should disconnect from companion app

### Reproduction steps

The issue could be divided in two parts:
### 1. When visiting the bluetooth setting and leaving without changing anything results in UI reporting bluetooth not connected (check the icon on the top right corner)
I'm intentionally saying "UI reporting bluetooth not connected" because as you can see from the video the device is still connected as Gadgetbrige it's still able to trigger the "find lost watch" routine. 

https://github.com/InfiniTimeOrg/InfiniTime/assets/7011942/a30f46df-505b-4053-a4a1-1d59987fc433

### 2. Bluetooth does not turn off while the watch is connected to the companion app (phone)
When bluetooth is disabled the watch should disconnect from the paired device, but it does not as again Gadgetbridge is able to launch the "find lost watch" routine. It triggers off only once also Gadgetbridge also disconnects.

https://github.com/InfiniTimeOrg/InfiniTime/assets/7011942/06bcf135-3112-46a4-8f99-4d4c4a1cef45

P.S.: sorry for my frustrating mis-clicks, poor quality and overall shakiness of the videos, I'm only hoping to help :smile: 

### More details?

I'd gladly make a PR but my day job is taking all my time at the moment.

### Version

1.14.0 but, did not test in previous versions

### Companion app

Gadgetbridge

JustScott:
### Long Explanation
NimbleController.cpp was running bleController.Disconnect() each time NimbleController::EnableRadio() was called by SystemTask.cpp's case matching Messages::BleRadioEnableToggle. This meant that each time the bluetooth settings page was opened, the Messages::BleRadioEnableToggle case was matched and the setting variable bleRadioEnabled was set to false. This caused the conditionals for the checkboxList class call in SettingBluetooth.cpp to not allow toggling the bluetooth on and off, and caused the bluetooth status icon to dissappear in StatusIcons.cpp

### Short Explanation
The nimblecontroller was telling the bluetooth controller to disconnect when the radio was enabled, which was happening each time the bluetooth settings page was opened. This caused the toggles/checboxes in the bluetooth settings page to be un-changeable, and the bluetooth status icon to disappear.

(A large explanation for removing a single line of code... I know ;) )

This fixes #1961

### DIFF
--- a/src/components/ble/NimbleController.cpp
+++ b/src/components/ble/NimbleController.cpp
@@ -402,7 +402,6 @@ void NimbleController::NotifyBatteryLevel(uint8_t level) {
 
 void NimbleController::EnableRadio() {
   bleController.EnableRadio();
-  bleController.Disconnect();
   fastAdvCount = 0;
   StartAdvertising();
 }


rambonette:
Thanks @JustScott ! Thought it was something simple, but didn't have the time to investigate 👍

vkareh (approved):
Works. Thanks for fixing this (has been driving my crazy when testing settings on a new build).

JustScott:
#2037 created by @JF002 better addresses this issue at its core by only pushing the `BleRadioEnableToggle` Message when the toggle is actually changed, instead of when the bluetooth settings page is opened. 

If you're currently using my PR in your fork, please switch to using #2037 instead, as my solution could (and probably will) cause weird bugs in the future if anything is altered with `NimbleController` or `BleController`.

PR #1974, opened 2024-01-17
Show blue icon when connected, grey when disconnected and hide when disabled

riban-bw:
This PR changes behaviour of bluetooth icon display.

Previous behaviour was that the icon showed coloured white when enabled and connected.

New behaviou is that the icon shows blue when connected, grey when disconnected and is hidden when disabled. This provides more info in the same space and is consistent with many common implementations which may offer a more familiar workflow.


minacode:
How is the contrast of the blue icon on the black background? Alternatively white for "connected" would also be an option.

riban-bw:
Blue is okay but I did change it to white because it may be simpler and more stylish to have a monochromatic status. I will commit the change to this PR.

PR #1975, opened 2024-01-17
Rationalise use of status in faces.

riban-bw:
This PR adds StatusIcons to WatchFaceAnalog. This rationalises the use of StatusIcons that was previously used on many screens but not the analogue watchface. The core advantage is that we have common code that reduces undesirable or unexpected behaviour, e.g. changes to StatusIcons would not be evident in the analouge watch face.


FintasticMan:
This has been discussed before: #1336 and #1963.

riban-bw:
Okay - I will review those threads. I discovered the issue because the analogue face behaves different to other faces / apps which gives an inconsistent UI. Hopefully I will find some info in one of those threads that provides a roadmap for resolving this issue.

PR #1976, opened 2024-01-17
Add option to keep display on always

riban-bw:
The PR adds an option to keep the watch display always on. Previously there were options to keep the diplay on for:

- 5s
- 7s
- 10s
- 15s
- 20s
- 30s

This change offers options:

- 5s
- 10s
- 15s
- 30s
- 60s
- On (always on)

Display can still be turned off (sleep) with button press.

There may be apps or workflows that require the display to remain on until manually turned off or where the available durations are not sufficient.

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -212,11 +212,11 @@ void DisplayApp::Refresh() {
       }
       queueTimeout = lv_task_handler();
 
-      if (!systemTask->IsSleepDisabled() && IsPastDimTime()) {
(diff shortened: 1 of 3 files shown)


riban-bw:
Okay - thanks for the info about doNotSleep which I shall use in the app for which this will be desirable. I look forward to seeing https://github.com/InfiniTimeOrg/InfiniTime/pull/1869 integrated in a future release. (For now I shall use my patched fork.) Cheers!

riban-bw:
I think the setting is called `doNotGoToSleep` and is only for system tasks, not screens. Is there a way for a screen / app to disable sleeping?

FintasticMan:
Yes, an example is here: #681.

PR #1980, opened 2024-01-18
DisplayApp: Go to clock on wakeup if no app loaded

vkareh:
When waking the screen up, if there is no actual app loaded (i.e. we are still in the QuickSettings, Settings, or Launcher screens) we should just reload the Clock app directly.


vkareh:
This supercedes https://github.com/InfiniTimeOrg/InfiniTime/pull/1436 since that one has conflicts and uses outdated function calls.

mark9064:
Future compatibility thought: with always on display (#1869), it'd be preferred if the app was loaded on sleep entry rather than on wake. But the whole sleep state might end up being refactored anyway so if this is annoying to implement then don't worry

vkareh:
@mark9064 - that's a good point. I tried it on sleep and it works even better (the transition is less choppy). Thanks for the suggestion!

Regarding future refactor, I feel that this is a simple enough change and can be moved around as needed without much overhead.

mark9064:
Great to hear it's working even better. To clarify what I meant originally: "as part of 1869, we might have to refactor some sleep stuff so if it's difficult to change this PR to work on sleep entry at the moment then don't worry". But you've managed to work it out anyway, so it doesn't matter :+1:

src/displayapp/DisplayApp.cpp
```diff
@@ -246,6 +246,10 @@ void DisplayApp::Refresh() {
         lcd.Sleep();
         PushMessageToSystemTask(Pinetime::System::Messages::OnDisplayTaskSleeping);
         state = States::Idle;
+        // Since the active screen is not really an app, go back to Clock.
```
mark9064, line 249: Might want to move this to before the message is pushed to the system task, as the system task will turn off the SPI bus and external flash when it receives this message. This may cause some problems loading fonts for the watchface when you call LoadScreen()

This may also cause the watchface to flash briefly on sleep entry though. To resolve this you'd want to somehow turn off the display backlight earlier

vkareh: fixed. I don't think it made much of a difference, but you're right - it seems like a better location.

JustScott:
Finding this quite nice as I consistently forget to return to the watch face. I'm wondering what thoughts are on having this also apply to the notifications, as that isn't an app that really needs to stay open after you view it once.

vkareh:
@JustScott at first I assumed that this wouldn't be necessary for the Notifications screen as I usually just dismiss notifications fairly quickly after viewing them, but then realized I use https://github.com/InfiniTimeOrg/InfiniTime/pull/1716 as part of my daily driver branch, so it's a bit less relevant.

However, after testing it some more, I think it makes sense to add Notifications to the list. It works well with it, so I've pushed a new change.

JF002:
I noticed a small glitch when the watch wakes up and switches to the clock. Any idea was could be the cause of this?


https://github.com/user-attachments/assets/f7f4c4f8-d623-4c7b-80ba-62f0e02c124d

mark9064:
Could call the lvgl task handler in a loop until it has no more work to do after loading the clock screen. That way you can be sure it's fully rendered to the display before sleeping

vkareh:
@mark9064 sounds interesting, how do I accomplish that? I'm not sure I fully understand the LVGL task API

mark9064:
`while (!lv_task_handler()) {};`
`lv_task_handler` runs LVGL, it returns the number of ms to wait until running it again. If it returns 0 it would like to be run immediately i.e it still has more things to do

mark9064:
Yeah definitely don't block there, DisplayApp is the lowest priority task so anything else that wants to run will already preempt it

vkareh:
Just pushed the change with the `lv_task_handler` loop.
It loads the clock app and then waits until it's loaded before it turns off the LCD. The effect is much nicer when turning it back on.

vkareh:
I've put the clock loading code in between setting the brightness and setting the LCD state. This gets rid of the glitch entirely on both screen off and screen on.

JF002:
> I've put the clock loading code in between setting the brightness and setting the LCD state. This gets rid of the glitch entirely on both screen off and screen on.

Great, I prefer this solution rather than the infinite loop! I confirm that the glitch is not visible anymore :+1:

JF002 (approved):

Issue #1985, opened 2024-01-19
lv_img_conv.py doesn't handle 8-bit pixel mode


ekirchman:
### Verification

- [X] I searched for similar issues (including closed issues) and found none was relevant.

### Introduce the issue

Pillow supports a lot of different modes as seen [here](https://pillow.readthedocs.io/en/stable/handbook/concepts.html#modes)
'P' mode is an 8-bit pixel mode that not only is space efficient but also supports color and transparency. When trying save an an external resource using `CF_TRUE_COLOR_ALPHA`, the error `TypeError: 'int' object is not subscriptable` is printed because this mode does not have "channels".

Steps to reproduce:
1. Follow the [official docs](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/doc/ExternalResources.md) for external resources.
2. Use the following image to generate an lv compatible file:
![Enemy](https://github.com/InfiniTimeOrg/InfiniTime/assets/47666987/8068d818-f911-4786-890c-6e9dd2e5c533)
3. See error is reproduced.

A workaround for now is to use Pillow to convert the mode from 'P' to 'RGBA', but this increases the file size.

Full traceback:
```
Beginning conversion of /home/user/Documents/projects/pinetime/InfiniTime/src/resources/images/FEPT/Enemy.png
overwriting Enemy.bin
Traceback (most recent call last):
  File "/home/user/Documents/projects/pinetime/InfiniTime/src/resources/lv_img_conv.py", line 193, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/user/Documents/projects/pinetime/InfiniTime/src/resources/lv_img_conv.py", line 122, in main
    r_act = classify_pixel(pixel[0], 5)
                           ~~~~~^^^
TypeError: 'int' object is not subscriptable
Traceback (most recent call last):
  File "/home/user/Documents/projects/pinetime/InfiniTime/src/resources/generate-img.py", line 59, in <module>
    main()
  File "/home/user/Documents/projects/pinetime/InfiniTime/src/resources/generate-img.py", line 54, in main
    subprocess.check_call(line)
  File "/usr/lib64/python3.11/subprocess.py", line 413, in check_call
    raise CalledProcessError(retcode, cmd)
[...]

PR #1989, opened 2024-01-21
lv_img_conv: support other modes like 'P'

NeroBurner:
Support other image modes like `P`, which uses 8 bits per pixel and a color palette to save space.

Luckily the Pillow module can do the mode conversion for us.

Fixes: https://github.com/InfiniTimeOrg/InfiniTime/issues/1985

### DIFF
--- a/src/resources/lv_img_conv.py
+++ b/src/resources/lv_img_conv.py
@@ -101,6 +101,13 @@ def main():
     img = Image.open(img_path)
     img_height = img.height
     img_width = img.width
+    if args.color_format == "CF_TRUE_COLOR_ALPHA" and img.mode != "RGBA":
+        # support pictures stored in other formats like with a color palette 'P'
+        # see: https://pillow.readthedocs.io/en/stable/handbook/concepts.html#modes
+        img = img.convert(mode="RGBA")
+    elif args.color_format == "CF_INDEXED_1_BIT" and img.mode != "L":
+        # for CF_INDEXED_1_BIT we need just a grayscale value per pixel
+        img = img.convert(mode="L")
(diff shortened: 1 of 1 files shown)


NeroBurner:
@ekirchman please review

ekirchman:
Is there any reason we don't want to use try and handle each mode separately? GIMP and Gwenview can even can tell that these images have transparency to them. And at least before they're converted to RGBA, 'P' type images are smaller in file size than RGBA images

NeroBurner:
`CF_INDEXED_1_BIT` is a special case, that is essentially a `P` image with just one bit, so a color palette with two colors, in our case hard coded to black and white
https://github.com/InfiniTimeOrg/InfiniTime/blob/84e942555fbeb21a811920c83f355e70397a5071/src/resources/lv_img_conv.py#L141

Other than that I don't really know why we "only" have `ARGB8565_RBSWAP` and `CF_INDEXED_1_BIT`. I just converted the nodejs script to python because I was tired of the nodejs dependencies to break :sweat_smile: 

You may look at the original javascript based `lv_img_conv` script here: https://github.com/lvgl/lv_img_conv

My guess is, that any binary format this script can produce is in some way usable by LVGL. Maybe only a subset of those are usable by PineTime, but I honestly don't know

---

In any case, I hope your original issue is resolved with this. If so please approve the PR :)

FintasticMan:
The output is controlled entirely by the color_format and binary_format options, so converting the input shouldn't have any effect on the output. This means that you don't need to be concerned about the file size being bigger with the "RGBA" mode.

ekirchman (commented):
I agree with @fintasticman that due to the color format being the same the file size shouldn't different if at all

PR #1993, opened 2024-01-28
README: Make hex and UTF-8 code consistent for the chosen example

timteichmann:
This small change makes the example consistent regarding hex and UTF-8 code, which should make it easier to follow for a newbie like myself ;-)


FintasticMan (approved):
Looks good to me!

JF002 (approved):

PR #1995, opened 2024-01-31
weather: Add new app with forecast

vkareh:
Added a new weather app that shows the current weather and forecast. If forecast is not available, it just shows the current weather.

InfiniSim_2024-02-02_145908

Building the sim requires https://github.com/InfiniTimeOrg/InfiniSim/pull/138


FintasticMan:
Looks good! Is there enough space to add the ° symbol to the forecast? It looks a bit off without it IMO. It might also be nice to know the current min and max temp?

About the forecast not showing up with Breezy Weather + Gadgetbridge, could you check if it works with Tiny Weather Forecast Germany? The Gadgetbridge side should be working, so I suspect that Breezy Weather just doesn't provide that information.

src/displayapp/screens/Weather.h
```diff
@@ -0,0 +1,56 @@
...
+
+      class Weather : public Screen {
+      public:
+        Weather(Controllers::Settings& settingsController, Controllers::SimpleWeatherService& weatherService);
+        ~Weather() override;
+
+        void Refresh() override;
+
+      private:
+        const char* getCondition(const Pinetime::Controllers::SimpleWeatherService::Icons icon);
```
FintasticMan, line 29: This function could be moved to displayapp/screens/WeatherSymbols.{h,cpp}, as I imagine it might be useful to other apps as well. Also, we use PascalCase for functions.

FintasticMan (changes requested):
One small thing, for the rest this looks good!

vkareh:
Ha! Found the issue. The wrong variable was being referenced in Gadgetbridge and so was not sending forecast data. I've submitted a patch to fix it: https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/3551

@FintasticMan Thanks for reviewing, I'll start addressing these issues sometime today.

FintasticMan:
Oh oopsie, that was my code in GB :sweat_smile:

Thanks for finding and submitting a fix.

FintasticMan:
I think that looks good, but I'm not exactly the best at UI design. If some more people chime in, I think that this is good! (You do still need to fix some formatting FYI. You can use the git pre-commit hook to do this automatically.)

vkareh:
@FintasticMan thanks - formatting is fixed now. The sim build breaks due to a new operator override in the Forecast struct. Should be fixed by https://github.com/InfiniTimeOrg/InfiniSim/pull/138

vkareh:
I removed the location display. It was causing my watch to crash and reboot. I might debug it later, but easier to remove it for now...

InfiniSim_2024-02-01_112143

FintasticMan:
Do positive numbers above 100 display fine? That is possible with Fahrenheit I think...

vkareh:
Yep, they look fine. The problem would be if it gets to 100 below zero (e.g. `-100` has 4 characters), but we have far bigger problems if that's the case :laughing: 

Acceptable display range would then be `-99` to `999`, from a purely technical point of view.

InfiniSim_2024-02-01_113440

vkareh:
@minacode 
> One small thing: you could use InfiniTimes theme colors.

I'd be more than happy to take suggestions on how to improve the thing. Here are two options using the InfiniTimeTheme colors:

InfiniSim_2024-02-01_130349 InfiniSim_2024-02-01_130434

Any preference? Other color combinations? I haven't been able to figure out how to change the color of the text in the forecast table...

KaffeinatedKat:
that looks incredible, great work. Would it be possible to have the color change based on the temp? (eg. blue for cold and yellow/red for hot). That would really help bring it together imo

JustScott:
Wow this looks great! Do you think it would be possible to fit the date and/or week day above or below each forecast. Example: 'thu', 'fri', '2/1', '2/17'?

Oh and a quick suggestion for the colors: maybe have them dynamically change when they meet certain temp ranges, for example: blue when below freezing, orange when above freezing, and red when above some higher temp.

vkareh:
> that looks incredible, great work. Would it be possible to have the color change based on the temp? (eg. blue for cold and yellow/red for hot). That would really help bring it together imo

Thanks! I love this idea.

> Wow this looks great! Do you think it would be possible to fit the date and/or week day above or below each forecast. Example: 'thu', 'fri', '2/1', '2/17'?

I'll try, but I suspect there won't be much room for this. Also, this app is already annoyingly heavy (3388B as of last build), and calculating the days based on timestamps and then displaying that on a table will eat up a lot more, I think.

vkareh:
What are your thoughts on this? Anything 0°C or below is blue, 40°C or above is red, anything else is yellowish:
InfiniSim_2024-02-01_132859 InfiniSim_2024-02-01_132904 InfiniSim_2024-02-01_133258

JustScott:
> > Wow this looks great! Do you think it would be possible to fit the date and/or week day above or below each forecast. Example: 'thu', 'fri', '2/1', '2/17'?
> 
> I'll try, but I suspect there won't be much room for this. Also, this app is already annoyingly heavy (3388B as of last build), and calculating the days based on timestamps and then displaying that on a table will eat up a lot more, I think.

Perhaps just adding the weekday and date for today somewhere could work, like is done on the digital watch face where it's a less visible grey.

minacode:
Maybe increase the space between the big temperature and the description above?

vkareh:
How about now?
InfiniSim_2024-02-01_154859 InfiniSim_2024-02-01_155008

I tried putting the current day/time (last updated time, really) but it was hard to align everything. It was just too cluttered...

KaffeinatedKat:
that looks really awesome, we don't really need to show the current day/time because that's easily accessible elsewhere. Having the foretasted days shown is more important. Great work

vkareh:
@MoonlightWave-12 
> The maximum should be at the top, because it is higher.

Good idea. Done:
InfiniSim_2024-02-01_161858

With regards to temperature colors, anything between 0°C and 40°C starts to become really subjective. Where I live currently 5°C is a very warm winter day, whereas 30°C is a hot summer day. However, I grew up in a place where 90°F is a mild winter day.

I'm okay with your ranges, but would like some consensus before comitting to those numbers. I also propose these:

* 0 or lower: blue (freezing)
* 0 - 4: cyan (ice warning)
* 4 - 32: yellow (nothing special, but color white doesn't look great here)
* 32 or higher: red (hot)

The 32°C number is from [wikipedia](https://en.wikipedia.org/wiki/Heat_index#Effects_of_the_heat_index_(shade_values)) but we can also use 27°C

vkareh:
@MoonlightWave-12 
> Maybe also colour the forecast-values, so that one can see the temperature-ranges easily at a glance?

That'd be nice, but I haven't figured out how to apply color to the table cells, only the table border (which I made black to blend in with the background)

FintasticMan:
Wow this is really coming together! I would suggest using 27 for hot rather than 32 personally, but I don't live anywhere very hot 😄. I'd say it'd be best to capitalise the first letter of the weekdays, as they are proper nouns, so it looks strange to me in lower-case.

vkareh:
@FintasticMan: done!

InfiniSim_2024-02-01_172313 InfiniSim_2024-02-01_172333
InfiniSim_2024-02-01_172319 InfiniSim_2024-02-01_173145

vkareh:
I studied the lvgl table spec and was able to simplify the table code and re-color some table elements:
InfiniSim_2024-02-02_140145
I made them light gray for now but we can change them on an as-need basis.

KaffeinatedKat:
is it possible to color the table temps based on the same schema as the current? Or do they all have to be the same color. If you can change them all individually see what the color temps look like on the forecast. It'll either look great or not so great

vkareh:
@KaffeinatedKat - here's a selection with colorized forecast:
InfiniSim_2024-02-02_145908 InfiniSim_2024-02-02_145858
InfiniSim_2024-02-02_145905 InfiniSim_2024-02-02_145902
InfiniSim_2024-02-02_150153

KaffeinatedKat:
I know that this feature is probably outside the scope for this PR, but when this is merged this would be the perfect app for the swipe left action on the clock

vkareh:
@KaffeinatedKat 
> I know that this feature is probably outside the scope for this PR, but when this is merged this would be the perfect app for the swipe left action on the clock

Last night I was thinking about a different feature. A setting where you can select a few apps (in a checkbox, for example) and those will be "always open" and you can cycle through them by swiping left. Also adding an auto-cycle option.

Definitely out of scope here. We can start a discussion thread about it outside this PR.

vkareh:
Bad news, the last commit (which colorizes the forecast temperatures) crashes my watch when loading the app... will debug and maybe revert depending on what I find.

vkareh:
Fixed the crash.
It was happening due to the task refresh callback being called every 20ms, which is probably not enough time for the fully-colorized table to render correctly before the next callback happened. I bumped that to 1000ms and it fixed the issue - there's no need to refresh the weather data that often anyway.

minacode:
What do you think about aligning the numbers per day?

Examples:
```
1
2

 1
12

 1
-3

  0
-11
```

vkareh:
@minacode

> What do you think about aligning the numbers per day?

not sure I understand this comment and the accompanying example

minacode:
I'm sorry, I thought about right-aligning the temperatures in the table.

vkareh:
@minacode 
> I'm sorry, I thought about right-aligning the temperatures in the table.

Good idea, I just pushed a new version with right-aligned temperatures. I think it looks good!

InfiniSim_2024-02-05_092207

minacode:
Yes! It's a little easier to the eyes to grasp the structure.

FintasticMan:
Is it always all the way right aligned? I.e. what will it do when you have a day with a 1-char value and one with a 2-char value?

vkareh:
> Is it always all the way right aligned? I.e. what will it do when you have a day with a 1-char value and one with a 2-char value?

Nothing different. The whole thing is an LVGL table with invisible borders, so it behaves like any numeric table (spreadsheet, etc.):
```
        +-----+-----+-----+-----+-----+-----+
right:  |    1|   -1|   11|  -11|  111| -111|
        +-----+-----+-----+-----+-----+-----+
center: |  1  |  -1 |  11 | -11 | 111 | -111|
        +-----+-----+-----+-----+-----+-----+
left:   |1    |-1   |11   |-11  |111  |-111 |
        +-----+-----+-----+-----+-----+-----+
```
The issue with anything other than right-aligned is that things look wrong, which is why spreadsheets typically default to right-aligning numbers.

This is probably a good visualization:
```
|right|centr|left |
+-----+-----+-----+
|    1|  1  |1    |
+-----+-----+-----+
|   -1|  -1 |-1   |
+-----+-----+-----+
|   11| 11  |11   |
+-----+-----+-----+
|  -11| -11 |-11  |
+-----+-----+-----+
|  111| 111 |111  |
+-----+-----+-----+
| -111| -111|-111 |
+-----+-----+-----+
```
Notice how the right-aligned column grows progressively from the right, and unit numbers are always aligned, as are the tens and hundreds, etc. Also the negative sign doesn't shift the number from alignment, as it's always away from the alignment.

FintasticMan:
It would've been nice to have it align right as far as its column is wide, if that makes any sense. But as I understand it that's not possible with LVGL tables.

I was thinking something like this:
```plaintext
+---+  +----+  +---+  +---+
| 5 |  | -1 |  |  5|  | -2|
| 1 |  | -5 |  |-23|  |-30|
+---+  +----+  +---+  +---+
```

FintasticMan:
I'm still not sure about the alignment. I think that right-aligning the numbers makes them easier to read, but I think it looks wrong in this situation:
InfiniSim_2024-02-08_223348

Ideally, we'd have it look like this:
InfiniSim_2024-02-08_223436 InfiniSim_2024-02-08_223504_edited InfiniSim_2024-02-08_223608_edited

Is this possible? If not, we should decide whether right-aligned looks better in most cases or not.

minacode:
We could center the cells and pad them with spaces if necessary.

FintasticMan:
Good thinking!

vkareh:
Sounds like a good idea. I can try, but each cell will need to know how many digits are in the temperature above or below itself and then add padding based on that. Maybe just centering all of them is not that bad after all?

Anyway, I'll see what I can pull off next time I'm playing with this.

vkareh:
@FintasticMan @minacode - success! I managed to use character padding to align temperatures:
* With table borders (to visualize the alignment):
InfiniSim_2024-02-10_131558
* Without table borders:
InfiniSim_2024-02-10_132039
* With more realistic data:
InfiniSim_2024-02-10_132314  InfiniSim_2024-02-10_132253 InfiniSim_2024-02-10_132229

src/displayapp/screens/Weather.cpp
```diff
@@ -0,0 +1,203 @@
...
+          case -1:
+            maxPadding = " ";
+            break;
+          case 1:
+            minPadding = " ";
+            break;
+          case 2:
+            minPadding = "  ";
+            break;
+        }
```
FintasticMan, line 188: I don't really like using `std::string` here, as it can quite easily be done with a C array.

```suggestion
        // Pad cells based on the largest number of digits on each column
        char maxPadding[3] = "  ";
        char minPadding[3] = "  ";
        int maxSize = snprintf(nullptr, 0, "%d", maxTemp);
        int minSize = snprintf(nullptr, 0, "%d", minTemp);
        switch (maxSize - minSize) {
          case -1:
            maxPadding[1] = '\0';
            break;
          case 1:
            minPadding[1] = '\0';
            break;
        }
```

FintasticMan (changes requested):

src/displayapp/screens/Weather.h
```diff
@@ -0,0 +1,58 @@
...
+  namespace Controllers {
+    class Settings;
+  }
+
+  namespace Applications {
+    namespace Screens {
+
+      class Weather : public Screen {
+      public:
+        std::array<const char*, 7> WeekDays = {"Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"};
```
minacode, line 23: 1. There is something like this in the DatetimeController. Maybe you can use this instead so we don't have multiple definitions of the days names.
2. Why is this public?

src/displayapp/screens/Weather.cpp
minacode: This is missing
```cpp
  case 0:
    minPadding[0] = '\0';
    maxPadding[0] = '\0';
```

We also might want to handle longer numbers correctly, even though they are most likely wrong. This could be done by changing the `case 0` to `default`.

FintasticMan: Oh yeah! Forgot about that

FintasticMan: I've just updated it to handle larger numbers without having to add cases for them.

src/displayapp/screens/Weather.cpp
```diff
@@ -0,0 +1,203 @@
...
+#include "components/ble/SimpleWeatherService.h"
+#include "components/settings/Settings.h"
+#include "displayapp/DisplayApp.h"
+#include "displayapp/screens/WeatherSymbols.h"
+#include "displayapp/InfiniTimeTheme.h"
+
+using namespace Pinetime::Applications::Screens;
+
+namespace {
+  lv_color_t temperatureColor(int16_t temperature) {
```
JF002, line 12: Please use PascalCase for function names

src/displayapp/screens/Weather.cpp
```diff
@@ -0,0 +1,203 @@
...
+      return Colors::blue;
+    } else if (temperature <= 400) { // ice
+      return LV_COLOR_CYAN;
+    } else if (temperature >= 2700) { // hot
+      return Colors::deepOrange;
+    }
+    return Colors::orange; // normal
+  }
+
+  uint8_t temperatureStyle(int16_t temperature) {
```
JF002, line 23: Please use PascalCase for function names

src/displayapp/screens/Weather.cpp
```diff
@@ -0,0 +1,203 @@
...
+      lv_obj_set_style_local_text_color(temperature, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, temperatureColor(temp));
+      char tempUnit = 'C';
+      if (settingsController.GetWeatherFormat() == Controllers::Settings::WeatherFormat::Imperial) {
+        temp = Controllers::SimpleWeatherService::CelsiusToFahrenheit(temp);
+        minTemp = Controllers::SimpleWeatherService::CelsiusToFahrenheit(minTemp);
+        maxTemp = Controllers::SimpleWeatherService::CelsiusToFahrenheit(maxTemp);
+        tempUnit = 'F';
+      }
+      temp = temp / 100 + (temp % 100 >= 50 ? 1 : 0);
+      maxTemp = maxTemp / 100 + (maxTemp % 100 >= 50 ? 1 : 0);
```
JF002, line 130: This calculation is done multiple times (and it's not easy to understand what it does - rounding?). Consider moving it to a dedicated function. The name of this function will also help understanding the meaning of it.

JF002 (changes requested):
I'm very happy to see this new app since the protocol allows for forecast data, but we had no way of displaying those info so far!

This feature looks good overall and I guess we'll be able to merge it as soon as the remaining code change requests and review are addressed :)

src/displayapp/screens/Weather.cpp
```diff
@@ -0,0 +1,203 @@
...
+      lv_label_set_text(minTemperature, "");
+      lv_label_set_text(maxTemperature, "");
+    }
+  }
+
+  currentForecast = weatherService.GetForecast();
+  if (currentForecast.IsUpdated()) {
+    auto optCurrentForecast = currentForecast.Get();
+    if (optCurrentForecast) {
+      std::tm localTime = *std::localtime((const time_t*) &optCurrentForecast->timestamp);
```
FintasticMan, line 151: This cast should be done with `reinterpret_cast`.

```suggestion
      std::tm localTime = *std::localtime(reinterpret_cast<const time_t*>(&optCurrentForecast->timestamp));
```

FintasticMan (changes requested):

src/displayapp/screens/Weather.cpp
FintasticMan: Yes, this is so that it gets rounded properly. I added it to the PTS watch face, but maybe we should add a function to SimpleWeatherService that does this calculation.

FintasticMan: I've just created that function in #2015.

src/displayapp/screens/Weather.cpp
vkareh: I love this, thanks!

src/displayapp/screens/Weather.h
vkareh: Fixed.

src/displayapp/screens/Weather.cpp
vkareh: Fixed

src/displayapp/screens/Weather.cpp
vkareh: Fixed

src/displayapp/screens/Weather.cpp
vkareh: I added this function directly in the weather app to avoid having to push a new change to InfiniSim, but let me know if we should change this.

FintasticMan (approved):

src/displayapp/screens/Weather.cpp
FintasticMan: I think it would be better to have the function in the weather service tbh, but it doesn't matter so much.

liamcharger:
Awesome work, this will make a great addition to the weather service when it is merged!

JF002 (approved):
LGTM! Thanks for your work @vkareh !

I've just tested it with Tiny Weather Forecast Germany and Gadgetbridge : 


Reminder : the current version of Gadgetbridge (0.78) has a bug that prevents forecast from being sent to the watch (https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/3551). The fix is merged, so I guess it'll be available in 0.79 (which is already tagged but not available in f-droid).

PR #2000, opened 2024-02-02
notifications: add option to clear all notifications

KaffeinatedKat:
I have added an action to dismiss all notifications in the notification viewer. You swipe left and it moves the notification to reveal a button that dismisses all of the notifications. If the menu is open and you swipe right it closes.

The button looks kinda jank currently. I really like the idea of swiping to reveal a button, but I don't know how to make it look good

image image


JustScott:
@KaffeinatedKat What if you implemented a long press action that pops up your button in the middle of the screen?

KaffeinatedKat:
I thought about that, but I hate how inconsistent the long press action is detected, so I went with a swipe action because it's much more consistent

minacode:
If we ever get to implement iOS notifications, we will need both left and right swipe to interact with them. Not super relevant right now, but there might be a conflict in the (far) future.

KaffeinatedKat:
how come? How are iOS notifications different from android ones?

minacode:
Apple has its own BLE service 😄

PR #2001, opened 2024-02-03
Terminal Watchface Improvements and added Weather.

JustScott:
* Use @vkareh's new [weather forecast app](https://github.com/InfiniTimeOrg/InfiniTime/pull/1995) as a template to implement the weather into the terminal watch face. 
* Use InfintimeTheme Colors instead of hardcoded hex values. 
* Reorder code to match the order of the UI. 
* Implement @vkareh's [variable battery icon](https://github.com/InfiniTimeOrg/InfiniTime/pull/1964) color to the battery percentage text. 
* Added a new InfinitimeTheme color, gray, and turn certain values gray when they contain no data, like weather.

terminal_weather_misty
terminal_weather_clear
terminal_less_gray
terminal_all_gray

I don't personally use the terminal watch face, so I'm looking for suggestions on what colors you guys want, and if I should change certain text like the `steps#` in the step counter.

### DIFF
--- a/src/displayapp/screens/WatchFaceTerminal.cpp
+++ b/src/displayapp/screens/WatchFaceTerminal.cpp
@@ -117,7 +117,7 @@ void WatchFaceTerminal::Refresh() {
         hour = hour - 12;
         ampmChar[0] = 'P';
       }
-      lv_label_set_text_fmt(label_time, "[TIME] %02d:%02d:%02d %s#", hour, minute, second, ampmChar);
+      lv_label_set_text_fmt(label_time, "[TIME] %02d:%02d:%02d %s", hour, minute, second, ampmChar);
     } else {
       lv_label_set_text_fmt(label_time, "[TIME] %02d:%02d:%02d", hour, minute, second);
     }
@@ -127,7 +127,7 @@ void WatchFaceTerminal::Refresh() {
       uint16_t year = dateTimeController.Year();
       Controllers::DateTime::Months month = dateTimeController.Month();
       uint8_t day = dateTimeController.Day();
-      lv_label_set_text_fmt(label_date, "[DATE] %04d-%02d-%02d#", short(year), char(month), char(day));
+      lv_label_set_text_fmt(label_date, "[DATE] %04d-%02d-%02d", short(year), char(month), char(day));
     }
   }
   
@@ -170,7 +170,7 @@ void WatchFaceTerminal::Refresh() {
 
   stepCount = motionController.NbSteps();
   if (stepCount.IsUpdated()) {
-    lv_label_set_text_fmt(stepValue, "[STEP] %lu steps#", stepCount.Get());
+    lv_label_set_text_fmt(stepValue, "[STEP] %lu steps", stepCount.Get());
   }
 
 
@@ -180,9 +180,9 @@ void WatchFaceTerminal::Refresh() {
     if (heartbeatRunning.Get()) {
 
       lv_obj_set_style_local_text_color(heartbeatValue, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, Colors::deepOrange);
-      lv_label_set_text_fmt(heartbeatValue, "[L_HR] %d bpm#", heartbeat.Get());
+      lv_label_set_text_fmt(heartbeatValue, "[L_HR] %d bpm", heartbeat.Get());
     } else {
-      lv_label_set_text_static(heartbeatValue, "[L_HR] ---#");
+      lv_label_set_text_static(heartbeatValue, "[L_HR] ---");
       lv_obj_set_style_local_text_color(heartbeatValue, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, Colors::gray);
     }
   }
@@ -191,14 +191,14 @@ void WatchFaceTerminal::Refresh() {
   bleRadioEnabled = bleController.IsRadioEnabled();
   if (bleState.IsUpdated() || bleRadioEnabled.IsUpdated()) {
     if (!bleRadioEnabled.Get()) {
-      lv_label_set_text_static(connectState, "[STAT] Disabled#");
+      lv_label_set_text_static(connectState, "[STAT] Disabled");
       lv_obj_set_style_local_text_color(connectState, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, Colors::gray);
     } else {
       if (bleState.Get()) {
-        lv_label_set_text_static(connectState, "[STAT] Connected#");
+        lv_label_set_text_static(connectState, "[STAT] Connected");
         lv_obj_set_style_local_text_color(connectState, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, Colors::blue);
       } else {
-        lv_label_set_text_static(connectState, "[STAT] Disconnected#");
+        lv_label_set_text_static(connectState, "[STAT] Disconnected");
         lv_obj_set_style_local_text_color(connectState, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, Colors::gray);
       }
     }
--- a/src/displayapp/screens/WatchFaceTerminal.cpp
+++ b/src/displayapp/screens/WatchFaceTerminal.cpp
@@ -100,7 +100,7 @@ void WatchFaceTerminal::Refresh() {
   notificationState = notificationManager.AreNewNotificationsAvailable();
   if (notificationState.IsUpdated()) {
     if (notificationState.Get()) {
-      lv_label_set_text_static(notificationIcon, "You have mail.");
+      lv_label_set_text_static(notificationIcon, "[1]+ Notify");
     } else {
       lv_label_set_text_static(notificationIcon, "");
     }
@@ -144,11 +144,10 @@ void WatchFaceTerminal::Refresh() {
       char tempUnit = 'C';
       lv_obj_set_style_local_text_color(weather, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, temperatureColor(temp));
       if (settingsController.GetWeatherFormat() == Controllers::Settings::WeatherFormat::Imperial) {
-        condition = Symbols::GetSimpleCondition(optCurrentWeather->iconId);
         temp = Controllers::SimpleWeatherService::CelsiusToFahrenheit(temp);
         tempUnit = 'F';
       }
-      lv_label_set_text_fmt(weather, "#ffffff [WTHR]# %i°%c %s ", temp / 100, tempUnit, condition);
+      lv_label_set_text_fmt(weather, "#ffffff [WTHR]# %i°%c %s ", temp / 100, tempUnit, Symbols::GetSimpleCondition(optCurrentWeather->iconId));
     } else {
       lv_label_set_text(weather, "#ffffff [WTHR]# ---");
       lv_obj_set_style_local_text_color(weather, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, Colors::gray);
@@ -165,7 +164,7 @@ void WatchFaceTerminal::Refresh() {
     uint8_t hue = batteryPercentRemaining.Get() * 120 / 100;
     lv_obj_set_style_local_text_color(batteryValue, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hsv_to_rgb(hue, 100, 100));
     lv_label_set_text_fmt(batteryValue, "#ffffff [BATT]# %d%%", batteryPercentRemaining.Get());
-    if (batteryController.IsPowerPresent()) {
+    if (batteryController.IsCharging()) {
       lv_label_ins_text(batteryValue, LV_LABEL_POS_LAST, " Charging");
     }
   }
--- a/src/displayapp/screens/WatchFaceTerminal.cpp
+++ b/src/displayapp/screens/WatchFaceTerminal.cpp
@@ -12,7 +12,6 @@
 #include "displayapp/screens/WeatherSymbols.h"
 #include "displayapp/InfiniTimeTheme.h"
 
-
 using namespace Pinetime::Applications::Screens;
 
 namespace {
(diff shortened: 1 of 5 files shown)


vkareh:
This is really cool, I like the graying out of n/a data!

How does it truncate if you have longer weather condition names, like Scattered Clouds or something?

Also, there seems to be a `#` sign after some of the lines, is that on purpose?

One suggestion though: the PS1 could be, say, lightGray so that it doesn't overpower the rest of the data, maybe then the time can be white so that it's easier to see at a quick glance?

I haven't really used this face a lot, but this PR might tip the scale for me.

src/displayapp/screens/WatchFaceTerminal.cpp
```diff
@@ -2,65 +2,84 @@
...
 #include "components/motion/MotionController.h"
 #include "components/settings/Settings.h"
+#include "displayapp/screens/WeatherSymbols.h"
+#include "displayapp/InfiniTimeTheme.h"
+
 
 using namespace Pinetime::Applications::Screens;
 
+namespace {
+  lv_color_t temperatureColor(int16_t temperature) {
```
vkareh, line 19: This makes me think that we can make this part of the InfiniTimeTheme...

src/displayapp/screens/WatchFaceTerminal.cpp
```diff
@@ -120,32 +117,90 @@ void WatchFaceTerminal::Refresh() {
...
+      char tempUnit = 'C';
+      lv_obj_set_style_local_text_color(weather, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, temperatureColor(temp));
+      if (settingsController.GetWeatherFormat() == Controllers::Settings::WeatherFormat::Imperial) {
+        condition = Symbols::GetSimpleCondition(optCurrentWeather->iconId);
+        temp = Controllers::SimpleWeatherService::CelsiusToFahrenheit(temp);
+        tempUnit = 'F';
+      }
+      lv_label_set_text_fmt(weather, "[WTHR] %i°%c %s ", temp/100, tempUnit, condition);
+    } else {
+      lv_label_set_text(weather, "[WTHR] ---°");
```
vkareh, line 149: nitpick: I'd remove the degree symbol here

JustScott:
@vkareh I'm confused about the #'s at the end of the lines as well, they were already there so i didn't want to remove them incase there was a reason. As far as the trunacation, I modified your `GetCondition` function to `GetSimpleCondition`, which just uses a single word, like `cloudy` instead of `Scattered Clouds`... it misses out on some detail, but I just don't see a way of fitting more than one word.

src/displayapp/screens/WatchFaceTerminal.cpp
```diff
@@ -120,32 +117,90 @@ void WatchFaceTerminal::Refresh() {
...
 
+
+  powerPresent = batteryController.IsPowerPresent();
+  batteryPercentRemaining = batteryController.PercentRemaining();
+  if (batteryPercentRemaining.IsUpdated() || powerPresent.IsUpdated()) {
+    // HSV color model has red at 0° and green at 120°.
+    // We lock satuation and brightness at 100% and traverse the cilinder
+    // between red and green, thus avoiding the darker RGB on medium battery
+    // charges and giving us a much nicer color range.
+    uint8_t hue = batteryPercentRemaining.Get() * 120 / 100;
```
vkareh, line 162: Any chance this can live elsewhere and we can reuse it here? Not super important, since my color PR may not ever be merged though 😛

src/displayapp/screens/WatchFaceTerminal.cpp
JustScott: @AlbeyAmakiir asked for it [in this discussion](https://github.com/InfiniTimeOrg/InfiniTime/discussions/1999)... you don't like it?

src/displayapp/screens/WatchFaceTerminal.h
```diff
@@ -30,13 +31,15 @@ namespace Pinetime {
...
-                          Controllers::MotionController& motionController);
+                          Controllers::MotionController& motionController,
+                          Controllers::SimpleWeatherService& weatherService);
         ~WatchFaceTerminal() override;
 
         void Refresh() override;
 
       private:
         Utility::DirtyValue<int> batteryPercentRemaining {};
+        Utility::DirtyValue<uint8_t> hue {};
```
vkareh, line 42: I think this can be a local variable inside the function, no need to track its state at the class level

src/displayapp/screens/WeatherSymbols.cpp
```diff
@@ -34,3 +34,28 @@ const char* Pinetime::Applications::Screens::Symbols::GetSymbol(const Pinetime::
...
+const char* Pinetime::Applications::Screens::Symbols::GetSimpleCondition(const Pinetime::Controllers::SimpleWeatherService::Icons icon) {
+  switch (icon) {
+    case Pinetime::Controllers::SimpleWeatherService::Icons::Sun:
+      return "Clear";
+    case Pinetime::Controllers::SimpleWeatherService::Icons::CloudsSun:
+      return "Cloudy";
+    case Pinetime::Controllers::SimpleWeatherService::Icons::Clouds:
+      return "Cloudy";
+    case Pinetime::Controllers::SimpleWeatherService::Icons::BrokenClouds:
+      return "Cloudy";
```
vkareh, line 47: I see what you did here. Clever.
It will conflict with the weather app PR, so we'll need to find a good middle ground if both make it into `main`

vkareh: Okay, I see this is a different, simplified, function. Makes sense now! 👍

src/displayapp/screens/WatchFaceTerminal.h
JustScott: Fixed.

JustScott:
This is with the time and date fonts white. I think the prompts look better in the first image with the lightGray, instead of the second image with gray. Thoughts?

terminal_lightgray_prompt
terminal_gray_prompt

vkareh:
> I think the prompts look better in the first image with the lightGray, instead of the second image with gray. Thoughts?

In fully agree. Light gray looks better

AlbeyAmakiir:
I really like the prompts being light gray. Good idea.

I'm a fan of each item being a different colour the way it currently is, and especially the `[LABEL]` part being the base terminal colour, as the colour helps direct my eye, but I do like the idea of the battery changing colour and the empty items being grayed out.

I dunno where the `#` has come from. It's not on mine, and I use the watchface all the time. Here's what mine looks like, at version 1.14.0:
IMG_20240204_133547

src/displayapp/screens/WatchFaceTerminal.cpp
AlbeyAmakiir: I think the degree symbol makes it easier to read at a glance, though I know it's a less standard symbol than basic terminals normally show. Was there another reason to remove it?

vkareh: Uh, wait, the line I highlighted in my comment is the `---°`. The actual temperature should have the symbol, of course, I'm referring to the empty value. Apologies for causing confusion here.

FintasticMan:
I agree with Albey above that I prefer the labels being white. The # symbols come from the way that the text was being coloured. It was done with `#012345 text#`, and I presume the final # wasn't removed at first. It has been fixed now though, I see.

I definitely prefer lightGray to gray, but perhaps if the labels are the same colour white works too?

Is there a reason that the same function for the weather description from the forecast app can't be used here? Are some words too long to fit?

I quite liked the colour variety of the current watch face. Maybe we can experiment with the colours a bit?

JustScott:
@FintasticMan Ah that's where the # came from, and yes more than a one word weather description overflows. Here's a version of the face with the labels white (#ffffff), the time and date using their original color value, and the prompts still using the lightGray value.

terminal_grayed
terminal_active

src/displayapp/screens/WatchFaceTerminal.cpp
minacode: Maybe in the theme too?

AlbeyAmakiir:
The extra line is nice and obvious. Only changing colour wouldn't communicate to new users of the face why it has done so. Moreso, communicating any information through colour alone makes it inaccessible to colourblind users.

That said, I'm not opposed to adding that colour to the mail text. That could make it stand out even better than it does now, I think.

Was there another reason you suggested changing the colour instead of having the line?

JustScott:
@AlbeyAmakiir I just thought it was a bit out of place for a terminal watch face to have random text above the first prompt appear. But you're right, just changing the color wouldn't work, I hadn't taken into account color blind users. 

I think having a more realistic terminal message could fit the lore a little better, I made an example of a background 'notify' job completing in the picture below, but if anyone has a better one let me know. (I'm also completely fine with leaving it as it was if that's how people prefer it).

terminal_notification

AlbeyAmakiir:
Hmm, I'm not sure. Having it at the top does have the advantage that it's right next to the direction in which you need to swipe for notifications, which I only noticed because your alternative is at the other extreme. But you're right it's a little off-lore. I also considered a `[MAIL]` entry in the output, but that doesn't have the benefit of being entirely invisible until there's a notification, which seems like it might be important for grabbing attention.

So it might be best to leave it as is, but I'm chill if others feel differently. I'm not sure how to weigh up the pros and cons of all these options.

src/displayapp/screens/WatchFaceTerminal.cpp
```diff
@@ -120,32 +122,86 @@ void WatchFaceTerminal::Refresh() {
...
+
+  currentWeather = weatherService.Current();
+  if (currentWeather.IsUpdated()) {
+    auto optCurrentWeather = currentWeather.Get();
+    if (optCurrentWeather) {
+      int16_t temp = optCurrentWeather->temperature;
+      char tempUnit = 'C';
+      lv_obj_set_style_local_text_color(weather, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, temperatureColor(temp));
+      if (settingsController.GetWeatherFormat() == Controllers::Settings::WeatherFormat::Imperial) {
+        condition = Symbols::GetSimpleCondition(optCurrentWeather->iconId);
```
vkareh, line 147: This is in the wrong scope. It also doesn't need DirtyValue tracking.

I suggest just calling it in the `lv_label_set_text_fmt`:
```c
lv_label_set_text_fmt(weather, "#ffffff [WTHR]# %i°%c %s ", temp / 100, tempUnit, Symbols::GetSimpleCondition(optCurrentWeather->iconId));
```
Then get rid of the `condition` variable.

src/displayapp/screens/WeatherSymbols.cpp
```diff
@@ -34,3 +34,28 @@ const char* Pinetime::Applications::Screens::Symbols::GetSymbol(const Pinetime::
...
+    case Pinetime::Controllers::SimpleWeatherService::Icons::CloudShowerHeavy:
+      return "Rainy";
+    case Pinetime::Controllers::SimpleWeatherService::Icons::CloudSunRain:
+      return "Rainy";
+    case Pinetime::Controllers::SimpleWeatherService::Icons::Thunderstorm:
+      return "Stormy";
+    case Pinetime::Controllers::SimpleWeatherService::Icons::Snow:
+      return "Snowy";
+    case Pinetime::Controllers::SimpleWeatherService::Icons::Smog:
+     return "Misty";
```
vkareh, line 57: This line is misaligned

vkareh, line 57: I think these should follow the OWM main group names more closely:
- Clear
- Clouds
- Clouds
- Clouds
- Drizzle
- Rain
- Thunder(storm) <--- Thunderstorm doesn't fit, but Thunder does
- Snow
- Mist <--- This one doesn't have a single name for the main group, but I've seen several apps defaulting to Mist, since it's the first one in the atmosphere condition group.

terminal-weather

src/displayapp/screens/WatchFaceTerminal.cpp
```diff
@@ -120,32 +122,86 @@ void WatchFaceTerminal::Refresh() {
...
+  batteryPercentRemaining = batteryController.PercentRemaining();
+  if (batteryPercentRemaining.IsUpdated() || powerPresent.IsUpdated()) {
+    // HSV color model has red at 0° and green at 120°.
+    // We lock satuation and brightness at 100% and traverse the cilinder
+    // between red and green, thus avoiding the darker RGB on medium battery
+    // charges and giving us a much nicer color range.
+    uint8_t hue = batteryPercentRemaining.Get() * 120 / 100;
+    lv_obj_set_style_local_text_color(batteryValue, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hsv_to_rgb(hue, 100, 100));
+    lv_label_set_text_fmt(batteryValue, "#ffffff [BATT]# %d%%", batteryPercentRemaining.Get());
+    if (batteryController.IsPowerPresent()) {
```
vkareh, line 168: I think this should call `IsCharging()` instead, as once it reaches 100% the controller will stop charging once the battery is full. Also, displaying `100% Charging` looks odd.

vkareh:
@JustScott 
> What are your guys thoughts on replacing the 'you have mail' text for new notifications, with just turning the the prompts orange (or some other color)?

I agree with @AlbeyAmakiir with regards to accessibility. Color-coding on a notification is not great UX. The `[1]+  Done` line at the bottom looks really cool, but the implication is that a background job is finished. The `Done` message is bash's rendering of a POSIX signal (I think SIGHUP) and so the text can be anything (bash has built-ins for `Done`, `Stopped`, `Exit`, etc...). I think the message can very well be `[1]+  Notify`, a reasonable rendering of a made-up `SIGNOTIFY` POSIX signal (we're all playing make believe here anyway, right?)

Playing with a terminal, I recreated what the flow would look like this (and yes, I modified my `PS1` and created a `now` alias just for this test :stuck_out_tongue_closed_eyes:):
```
user@watch:~ $ now
[TIME] 12:28:51
[DATE] 02/05/24
user@watch:~ $ echo "You have mail." &
[1] 71270
You have mail.
[1]+  Done                    echo "You have mail."
user@watch:~ $ now
[TIME] 12:28:54
[DATE] 02/05/24
user@watch:~ $ notify-send "You have mail." &
[1] 71281
user@watch:~ $ now
[TIME] 12:28:59
[DATE] 02/05/24
[1]+  Done                    notify-send "You have mail."
user@watch:~ $ 
```

Based on this, the signal message can be either before the top prompt (in the case of using `echo &`) or before the bottom prompt (in the case of using `notify-send &`).
In both cases it shows up before the prompt, and never after it.

Translating this to the watch face, I think we have two options that could work and satisfy the lore:
InfiniSim_2024-02-05_123151 InfiniSim_2024-02-05_123308

I personally prefer the first one.

vkareh:
Not sure if this is desired, but continuing on the theme/lore aspect, I recolored the watchface to use a 3-bit color space (8 colors). Maybe better to just use LVGL's pre-defined colors (which _more or less_ simulate a 4-bit color space)
InfiniSim_2024-02-05_142017

Not sure, maybe it's too much...

src/displayapp/screens/WatchFaceTerminal.cpp
JustScott: I agree.

AlbeyAmakiir:
I won't know for sure until I see it on the actual watch screen, but on my computer the 3-bit colours look a bit too harsh. Maybe it'll be fine irl, though? I'm not sure.

I definitely prefer the notify text being at the top over the bottom. (Also, I really appreciate that you actually tested that on a terminal. Fantastic. XD )

FintasticMan:
I would say it would be best to break the weather display out of this PR into its own, so that we can discuss that separately from changing the UI.

Pbone3:
I much prefer the white prompts. It fits the terminal theme better (at least in my experience) and using less colours makes the face as a whole less busy.

JustScott:
> I would say it would be best to break the weather display out of this PR into its own, so that we can discuss that separately from changing the UI.

I created a new PR without the weather changes: #2135, I also created a PR for the simpler weather condition change: #2134

L3br4nd:
@JF002 I would really love to have the weather information in the Terminal watch face as well. What needs to be done to make this PR part of the next milestone?

JustScott:
I've split the features of this PR into there own separate PRs so that one can be merged without the others.

- #2204 
- #2135
- #2134

PR #2002, opened 2024-02-03
settings: Add sleep/wake notification times

chmeeedalf:
Add configure screens for sleep and wake notification times, so the watch can automatically set notifications to sleep or on as applicable.

This could probably be done in a much better way, but I didn't want to refactor the whole date/time UI to add this.  An alternative, doing that, is to create a widget base class for all time-like setting widgets, and derive from that for each of the various interfaces.

### DIFF
--- a/src/displayapp/screens/settings/SettingSetSleepTime.cpp
+++ b/src/displayapp/screens/settings/SettingSetSleepTime.cpp
@@ -1,0 +1,103 @@
+#include "displayapp/screens/settings/SettingSetSleepTime.h"
+#include <lvgl/lvgl.h>
+#include <nrf_log.h>
+#include "displayapp/DisplayApp.h"
+#include "displayapp/screens/Symbols.h"
+#include "components/settings/Settings.h"
+#include "displayapp/InfiniTimeTheme.h"
+
+using namespace Pinetime::Applications::Screens;
+
+namespace {
+  constexpr int16_t POS_Y_TEXT = -7;
+
+  void SetTimeEventHandler(lv_obj_t* obj, lv_event_t event) {
+    auto* screen = static_cast<SettingSetSleepTime*>(obj->user_data);
+    if (event == LV_EVENT_CLICKED) {
+      screen->SetTime();
+    }
+  }
+
+  void ValueChangedHandler(void* userData) {
+    auto* screen = static_cast<SettingSetSleepTime*>(userData);
+    screen->UpdateScreen();
+  }
+}
+
+SettingSetSleepTime::SettingSetSleepTime(Pinetime::Controllers::DateTime& dateTimeController,
+                               Pinetime::Controllers::Settings& settingsController,
+                               Pinetime::Applications::Screens::SettingSetSleepWakeTime& settingSetSleepWakeTime)
+  : dateTimeController {dateTimeController}, settingsController {settingsController}, settingSetSleepWakeTime {settingSetSleepWakeTime} {
+
+  uint8_t h, m;
+
+  lv_obj_t* title = lv_label_create(lv_scr_act(), nullptr);
+  lv_label_set_text_static(title, "Set sleep time");
+  lv_label_set_align(title, LV_LABEL_ALIGN_CENTER);
+  lv_obj_align(title, lv_scr_act(), LV_ALIGN_IN_TOP_MID, 15, 15);
+
+  lv_obj_t* icon = lv_label_create(lv_scr_act(), nullptr);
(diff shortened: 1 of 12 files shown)


JF002:
Hi and thanks for your first contribution to the InfiniTime project !
Could you explain into more detail what those changes do? Maybe add a few screenshots of the new UI?

Oh and also do not forget to check that those changes are consistent with the [InfiniTime vision](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/doc/InfiniTimeVision.md).

FintasticMan:
I think this is similar to #1461, right? I think this is a very useful feature, let's decide on a good implementation together. It might be good if you could upload pictures from InfiniSim to see the UI.

chmeeedalf:
@FintasticMan yes, it's exactly like that, I didn't see it before.  Here are a couple screenshots:

image

image

image

(I noticed I forgot to change "current time" in my code, until I looked closer to get the screenshot).

Setting sleep and wake times to the same value disables the operation.

src/displayapp/screens/settings/SettingSetWakeTime.cpp
```diff
@@ -0,0 +1,103 @@
...
+      lv_label_set_text_static(lblampm, "AM");
+    }
+  }
+}
+
+void SettingSetWakeTime::SetTime() {
+  const int hoursValue = hourCounter.GetValue();
+  const int minutesValue = minuteCounter.GetValue();
+  NRF_LOG_INFO("Setting wakeup time to %02d:%02d:00", hoursValue, minutesValue);
+  settingsController.SetSleepTime(static_cast<uint8_t>(hoursValue),
```
Itai-Nelken, line 100: I believe this is setting the sleep time instead of the wake time.

src/displayapp/screens/settings/SettingSetSleepTime.cpp
```diff
@@ -0,0 +1,103 @@
...
+  }
+}
+
+void SettingSetSleepTime::SetTime() {
+  const int hoursValue = hourCounter.GetValue();
+  const int minutesValue = minuteCounter.GetValue();
+  NRF_LOG_INFO("Setting sleep time to %02d:%02d:00", hoursValue, minutesValue);
+  settingsController.SetSleepTime(static_cast<uint8_t>(hoursValue),
+                                  static_cast<uint8_t>(minutesValue));
+  settingSetSleepWakeTime.Quit();
```
Itai-Nelken, line 102: This will make pressing the "set" button advance to the set wake time screen (like what happens with the set date & time setting).
```suggestion
  settingSetSleepWakeTime.Advance();
```

src/displayapp/screens/settings/SettingSetWakeTime.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+SettingSetWakeTime::SettingSetWakeTime(Pinetime::Controllers::DateTime& dateTimeController,
+                               Pinetime::Controllers::Settings& settingsController,
+                               Pinetime::Applications::Screens::SettingSetSleepWakeTime& settingSetSleepWakeTime)
+  : dateTimeController {dateTimeController}, settingsController {settingsController}, settingSetSleepWakeTime {settingSetSleepWakeTime} {
+
+  uint8_t h, m;
+
+  lv_obj_t* title = lv_label_create(lv_scr_act(), nullptr);
+  lv_label_set_text_static(title, "Set current time");
```
Itai-Nelken, line 35: I believe you said you changed this? maybe you forgot to push the change?
```suggestion
  lv_label_set_text_static(title, "Set wake time");
```

chmeeedalf: I must've only changed it in my InfiniSim checkout.  Will update, along with all your suggestions.

Itai-Nelken:
I have been testing this pr (with my suggestions above) for the past few days, and for some reason, sleep mode only automatically starts if the notification settings is set to ring (and not do not disturb). I don't see any code that does that, so I think its a weird bug?

chmeeedalf:
> I have been testing this pr (with my suggestions above) for the past few days, and for some reason, sleep mode only automatically starts if the notification settings is set to ring (and not do not disturb). I don't see any code that does that, so I think its a weird bug?

It's intentional.  SetNotificationStatusByTime() first checks if the notification status is "Off" (looks like DnD to me), and if so it doesn't change it.  The idea being that you would explicitly set/unset DnD, and the watch changing it itself would be a surprise, and go against your desire to not be disturbed.

PR #2004, opened 2024-02-06
MusicService: Autostart music app on music start

vkareh:
When starting music on a paired device, the music service will automatically trigger the music app to open.

Fixes https://github.com/InfiniTimeOrg/InfiniTime/discussions/1619


mark9064:
I like the sound of this feature, but I think it should be optional. I have my phone and computers all connected with KDE connect, so any audio I play on any device syncs to my phone which sends it to InifiniTime. This is great, but making it switch to the music app every time any audio is played anywhere would drive me insane as it'd be switching all the time

FintasticMan:
I personally don't think that this would be a very nice user experience. As Mark says, I play media on my phone quite a lot, which will have it start and stop quite frequently. If that causes it to switch to the music app each time, it would get quite annoying. I think that this would be better implemented as a quick-switch action, where you can configure an app to get to quickly.

vkareh:
I've added the ability to configure which apps auto-open based on certain events:
- battery: when charging the phone (code directly from #1876)
- music: when playing music on paired device (tested with Gadgetbridge / Spotify Lite)
- navigation: when starting map navigation on paired device (tested with Gadgetbridge / OsmAnd~)

InfiniSim_2024-02-09_142706

mark9064:
I like this generic solution a lot, my preferred watch face has battery percentage on it so being able to turn off auto battery show would be nice

src/components/ble/NavigationService.h
```diff
@@ -27,11 +27,16 @@
...
+  namespace System {
+    class SystemTask;
+  }
+
   namespace Controllers {
 
     class NavigationService {
     public:
-      NavigationService();
+      NavigationService(Pinetime::System::SystemTask& systemTask);
```
FintasticMan, line 39: This should be marked explicit so that it can't accidentally be used for casts.

FintasticMan (changes requested):
I like it!

src/components/ble/NavigationService.h
vkareh: Yes, good point! Fixed.

FintasticMan (approved):

JF002:
I understand that such a feature might be useful in some cases but... I would like to challenge it for the following reasons:

 - UX : when enabled, the Navigation/Music/Battery app would be automatically loaded when a event (new direction, new music data, new charging state) is received. It means that the state of the application that is currently running would be erased. Let's say I'm playing a game, changing some settings or using the chrono -> I lost everything I was doing before the event. That would be kind of infuriating to me. 
 
 - Code : I don't think this implementation is compatible with the new compile time configuration of applications and watch faces : what happens in the Navigation or Music app are not built into the firmware?

 - [InfiniTime Vision](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/doc/InfiniTimeVision.md#core-principles) : In my opinion, this feature is not consistent with a few core principles of the InfiniTime project. For example, we avoid adding new settings whenever possible and we favor implementing a single solid default behavior instead.

As I said, I understand why some users request this kind of feature, but I'm not sure this is the best way to implemented. feel free to share your opinions :)

vkareh:
@JF002 
> UX : [...] It means that the state of the application that is currently running would be erased [...]

I updated the patchset to only load the app if there's no app loaded currently.

> Code : I don't think this implementation is compatible with the new compile time configuration of applications and watch faces : what happens in the Navigation or Music app are not built into the firmware?

Great question. Testing this now to see what happens.

> [...] we avoid adding new settings whenever possible and we favor implementing a single solid default behavior instead.

Fair point. I think this concern is the only one we cannot address in code and would require a decision from the InfiniTime core team (InfiniTeam? :stuck_out_tongue: )

vkareh:
@JF002 
> * Code : I don't think this implementation is compatible with the new compile time configuration of applications and watch faces : what happens in the Navigation or Music app are not built into the firmware?

I've accounted for this now. The DisplayApp verifies if the app is available before attempting to perform the action.

@MoonlightWave-12 
> It should only automatically open apps while the smartwatch is not used for something else; so, basically only while it is on the watch-face.

This is now fixed.

> [...] have a way for a companion-app to request opening any app on the smartwatch [...]

This could be interesting... Gadgetbridge has device-specific settings. InfiniTime could implement a service that allows it to interact with apps.

> [...] or tell the user how to make it available [...]

General users won't be able to decide which app gets installed on their watch.

> Automatically opening an app can make the usage-flow much easier.

I agree. This feature's goal is to increase usability. When I play music/podcasts, I expect my watch to know so that I can see/interact. Same with navigation. My companion app (Gadgetbridge) is smart enough that it doesn't just randomly sends any sound/notification as though it were media, and limits it only to actual media playing on my phone (Spotify Lite / AntennaPod). Things like movies from streaming services do _not_ get sent to the watch (which is what I expect anyway), so the behaviour here is really what I would expect of a connected device.

tituscmd:
I was just thinking... what if the app doesn't open right up, but gets placed to the right of the watchface?

mark9064:
After thinking about this for a while I'm in a similar space to JF. I think this isn't a good fit in terms of the vision and that either
1) integrations (e.g current track playing) should be a feature of watchfaces themselves
2) if an app should load by default (e.g the firmware update app), there need be no configuration option

vkareh:
This does not align with InfiniTime's vision (and to be honest, I've stopped maintaining this branch since there are conflicts, as I don't use this feature myself). Closing.

PR #2005, opened 2024-02-06
Missing character for navigation font

kieranc:
Whichever idiot touched the navigation app UI last managed to think about imperial and metric distance units and the characters necessary to represent them, but not that some distances might need the . (period/full stop) character.

The result is that currently, 3.1km displays as 31km, and is somewhat confusing.
I am the idiot. I forgot the . character. This PR fixes my mistake.


FintasticMan (approved):
There are a couple of extra commits from the last PR, but that doesn't matter if we squash merge.

NeroBurner (approved):

PR #2009, opened 2024-02-09
ApplicationList: Reset app menu screen when loading watch face

### LINKED ISSUE #2006: App menu stays at second screen
### Verification

### Introduce the issue

1. Go to the app menu.
2. Go to the second screen. 
3. Press the button to return to the watch face.
4. Swipe to enter the app menu. 
5. You are still on the second screen.

### Preferred solution

I would expect the app menu to reset to the first screen when returning to the watch face.

### Version

1.14 (custom, but I doubt it matters)

vkareh:
This prevents the application list from loading in the last used screen and instead goes back to the first screen whenever the watch face is loaded.

Fixes #2006

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -446,6 +446,7 @@ void DisplayApp::LoadScreen(Apps app, DisplayApp::FullRefreshDirections directio
       else {
         currentScreen.reset(userWatchFaces[0].create(controllers));
       }
+      settingsController.SetAppMenu(0);
     } break;
     case Apps::Error:
       currentScreen = std::make_unique<Screens::Error>(bootError);


JF002:
Have you checked why the page number of the launcher is actually stored? I wouldn't want to apply a change that conflicts with the original intend of the code.

I think it would also be a good idea to check that the behavior reported in #2006 does not happen for other apps as well, like settings for example (that you can open with a double press on the button iirc)? Maybe the fix should be more generic than it currently is?

vkareh:
> Have you checked why the page number of the launcher is actually stored? I wouldn't want to apply a change that conflicts with the original intend of the code.

Yes. It was added as part of #217 with the description:
> * Go back to the same menu page when we leave an app

This behaviour is preserved, as we only reset the page when we're not in an app.

> I think it would also be a good idea to check that the behavior reported in https://github.com/InfiniTimeOrg/InfiniTime/issues/2006 does not happen for other apps as well, like settings for example (that you can open with a double press on the button iirc)? Maybe the fix should be more generic than it currently is?

This issue doesn't happen in the settings, since we don't store the current screen there.

FintasticMan:
Git bisect says the bug was introduced in 39bc166e549e8ccae75468aa2dd3613d51f54e27. There was no call to SetAppMenu removed in that commit... Why didn't the bug occur before?

JF002:
> Git bisect says the bug was introduced in https://github.com/InfiniTimeOrg/InfiniTime/commit/39bc166e549e8ccae75468aa2dd3613d51f54e27. There was no call to SetAppMenu removed in that commit... Why didn't the bug occur before?

The call to Settings::SetAppMenu() was done in the ctor of the `Clock` class, which was removed in this https://github.com/InfiniTimeOrg/InfiniTime/commit/39bc166e549e8ccae75468aa2dd3613d51f54e27 commit!

And I think the @vkareh found a better place for this in `DisplayApp` so that watch faces do not have to manage this state.

JF002 (approved):

Issue #1991, opened 2024-01-26
Restore timer duration when finished


Rathmox:
### Verification

- [X] I searched for similar issues (including closed issues) and found none was relevant.

### Introduce the issue

I had to use the timer multiple times for the same duration, but I had to reset it everytime to 30 seconds which is annoying

### Preferred solution

restore the previously selected duration when timer finished

### Version

1.14.0

PR #2013, opened 2024-02-11
Timer: Remember last timer setting

tausen:
Remember last duration used in Timer app. Reset to zero with long press on start button.

Fixes https://github.com/InfiniTimeOrg/InfiniTime/issues/1991

Tested in InfiniSim and on hardware, seems to work as desired:

InfiniSim_2024-02-11_202413

A few notes to consider:

- There is no difference between restarting a paused timer and starting a new one, so briefly pausing will also update the suggested new duration
- Storing minutes and seconds in two simple ints rather than a `std::chrono::seconds` is probably lighter (less memory, fewer divisions/modulo 60), but this is probably more readable? I have no strong opinions here.
- Probably requires some changes once https://github.com/InfiniTimeOrg/InfiniTime/pull/1971 is merged


src/displayapp/screens/Timer.cpp
```diff
@@ -6,6 +6,8 @@
 
 using namespace Pinetime::Applications::Screens;
 
+static std::chrono::seconds lastTimer;
```
JF002, line 9: Is the global static variable needed here? Couldn't we add this variable as a member variable of the class?

src/displayapp/screens/Timer.cpp
```diff
@@ -131,23 +133,36 @@ void Timer::SetTimerStopped() {
   lv_label_set_text_static(txtPlayPause, "Start");
 }
 
+void Timer::SetCounters(std::chrono::seconds& duration) {
```
JF002, line 136: `duration` should be `const`

JF002 (commented):
Thanks for this 1st contribution @tausen !

> There is no difference between restarting a paused timer and starting a new one, so briefly pausing will also update the suggested new duration

I'm wondering if this would be the expected behavior by the users. Shouldn't the timer be reset to the last settings instead of the last paused value? This is an open question, I'm not really sure about this.

tausen:
Thanks for the feedback! PR updated accordingly.

> Shouldn't the timer be reset to the last settings instead of the last paused value? 

After some contemplation I've updated the PR to handle pause and stop differently. This ended up a bit more intrusive, but ultimately for the better IMO :)

I couldn't really find a good way to fit two buttons in the bottom like in the stopwatch app without things getting tight, so I've re-used the hold action on the start/stop/pause button: 

- Short press to start/pause
- Long press while paused to stop, reverting to previously used timer setting
- Long press while stopped to reset to 00:00

InfiniSim_2024-04-28_150611

I also had to nudge `configTOTAL_HEAP_SIZE` from 40 to 39 with these changes. Unsure whether I did something silly or things were just on the edge before?

tausen:
Pushed one more commit (to be squashed or discarded) -- this saves the last timer setting as soon as it's changed instead of when the timer expires or is stopped. The user can now enter the timer app, configure it for some duration, exit the app again (for example to read a notification), then go back to the timer app and not have to re-configure it.

src/displayapp/screens/Timer.cpp
tausen: Had resolved, but latest commit re-introduces this -- if `lastTimer` is a member of the Timer app, it'll be destroyed when closing the app... Open for suggestions :)

thun11o:
> Thanks for this 1st contribution @tausen !
> 
> > There is no difference between restarting a paused timer and starting a new one, so briefly pausing will also update the suggested new duration
> 
> I'm wondering if this would be the expected behavior by the users. Shouldn't the timer be reset to the last settings instead of the last paused value? This is an open question, I'm not really sure about this.

New to this project and just got my first PineTime, I'm super happy with it!

I had a Samsung Watch Active before and it resets to the initial value once the timer expires which to me seems correct.

tausen:
Thanks for the input, @thun11o, I believe the behavior with my changes should match what you're used to then :)
I'm using the timer a few times every day and usually at the same setting +/- a minute and it works quite well :+1: 

I've rebased this branch onto the latest main.

tausen:
I've started work on a version based on https://github.com/InfiniTimeOrg/InfiniTime/pull/1971 over here: https://github.com/tausen/InfiniTime/tree/timer-ringing-memory
A few things I'd change before promoting it to a PR, but seems to work well so far. With longer vibration time, count up when done and memory of the last setting, it behaves much like the timers I'm used to from other devices :)

PR #2014, opened 2024-02-12
lowersleep: Improve algorithm by checking wrist angle

FintasticMan:
Inspired by @darkdragon-001 https://github.com/InfiniTimeOrg/InfiniTime/pull/827#issuecomment-1881580414.

Checks if the user's wrist is pointing towards the floor, rather than only the amount of roll. It will turn off the display if the watch is angled at 60 degrees or more, and rotated at least 30 degrees within the last ~600ms.

### DIFF
--- a/src/components/motion/MotionController.cpp
+++ b/src/components/motion/MotionController.cpp
@@ -40,15 +40,15 @@ void MotionController::Update(int16_t x, int16_t y, int16_t z, uint32_t nbSteps)
     service->OnNewStepCountValue(nbSteps);
   }
 
-  if (service != nullptr && (this->x != x || yHistory[0] != y || zHistory[0] != z)) {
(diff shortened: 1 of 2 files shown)


JF002 (approved):

tituscmd:
Just for my understanding, is it actually optimal to check if the watch is facing the floor?
I would imagine the angle most people rest their watch at would be facing to their left, assuming they're wearing the watch on their left wrist.

FintasticMan:
Sorry for the confusion, this doesn't check if the screen is facing the floor, it checks if your wrist is pointing at the floor (or the ceiling, so that it also works for people who wear their watch on their right wrist).

tituscmd:
I see, that sounds good. Will definitely try it out soon!

PR #2015, opened 2024-02-12
weather: Add function for converting to units

FintasticMan:
Handles rounding correctly.


src/components/ble/SimpleWeatherService.h
```diff
@@ -112,6 +112,10 @@ namespace Pinetime {
         return celsius * 9 / 5 + 3200;
       }
 
+      static int16_t TempToUnits(int16_t temp) {
```
vkareh, line 115: I think this name is not quite representative, since it doesn't include units. Maybe something like "RenderTemp" or "DisplayTemp" or "RoundTemp"? Not entirely sure, naming is hard :/

Other than that, I really like what you're doing centralizing this function.

minacode: What is the meaning of the input and output?

FintasticMan: The input is a temperature in units multiplied by 100, and the output is temperature in the actual unit. I wasn't sure of the naming either, it's a surprisingly difficult operation to give a name to.

FintasticMan: `CentiToUnit`? As in centicelsius to celsius or centifahrenheit to fahrenheit.

JF002: This method could simply be called `Convert()` if we used strong types for arguments. Ex: 
```c++
static DisplayTemperature Convert(const SimpleWeatherService::RawTemperature temperature)
```

Using stronger types for
 - temperature coming from the weather service (in "centi celsius degrees")
 - temperature in °C
 - temperature in °F

Would probably help with code readability.

minacode:
You could use structs of one int to make the types really safe.

JF002:
I'm ok with the current state. I however thought about another option.

SimpleWeatherService would define a Temperature type : `PineTime::Controllers::SimpleWeatherService::Temperature`. It's the internal representation of a temperature in the SompleWeatherService expressed in "°C * 100"

```c++
struct Temperature {
  int16_t temperature;
}
```

Somewhere under DisplayApp, we would define another Temperature type : `Pinetime::Applications::Temperature`. It represent the temperature for the UI side of InfiniTime as an `std::variant` (TODO : check that the overhead of std::variant is negligible)

```c++
struct TemperatureCelsius {
  int16_t temperature;
}

struct TemperatureFahrenheit {
  int16_t temperature;
}

using Temperature = std::variant<TemperatureCelsius, TemperatureFahrenheit temperature
```

Then, also under DisplayApp, we define a function that converts from SimpleWeatherService Temperature to "display" temperature : 

```c++
Pinetime::Applications::Temperature Convert(PineTime::Controllers::SimpleWeatherService::Temperature, Controllers::Settings::WeatherFormat format);
```

This way, all the "temperature" types are strongly typed and it's impossible to implicitly convert from one to another. Also SimpleWeatherService does not have any code related to the UI anymore.

src/components/ble/SimpleWeatherService.h
```diff
@@ -61,13 +61,42 @@ namespace Pinetime {
         Unknown = 255
       };
 
+      class Temperature {
+      public:
+        explicit Temperature(int16_t raw = 0) : raw {raw} {
```
FintasticMan, line 66: I'm not a big fan of having the default value be 0 here, but otherwise the compiler complains about there being no default value for the array of `Day`s in the `Forecast`.

FintasticMan (commented):
@JF002 I've just refactored this a bit, what do you think? I wasn't able to use `std::variant`, as it is very strict about knowing which variant to use at compile-time. That meant that we needed 2 completely separate code paths for Celsius and Fahrenheit.

src/components/ble/SimpleWeatherService.h
minacode: I think it would be conceptually better (maybe not memory-wise) to make the days array of type:
`std::array<std::optional<Day>, nbOfDays>`. Then you would not need to construct `Day`s without values.

minacode:
I like the current approach with the raw value and the two representations! :)

src/components/ble/SimpleWeatherService.h
L3br4nd: Would defining Temperature in Kelvin internally and converting to DisplayTemperature in °C or °F depending on the user setting for displaying be helpful?
Sorry, I'm not into object based programming that much...

minacode: I don't think that is the problem here. The internal representation is never shown directly and can therefore be whatever. The problem is, that right now we need a temperature value in a place where no value should be.

FintasticMan: I think `std::optional` does make sense, I'm going to see how implementing it like that looks.

FintasticMan: I think this works quite well, what do you think?

src/displayapp/screens/Weather.cpp
```diff
@@ -152,24 +153,22 @@ void Weather::Refresh() {
     if (optCurrentForecast) {
       std::tm localTime = *std::localtime(reinterpret_cast<const time_t*>(&optCurrentForecast->timestamp));
 
-      for (int i = 0; i < Controllers::SimpleWeatherService::MaxNbForecastDays; i++) {
-        int16_t maxTemp = optCurrentForecast->days[i].maxTemperature;
-        int16_t minTemp = optCurrentForecast->days[i].minTemperature;
-        lv_table_set_cell_type(forecast, 2, i, TemperatureStyle(maxTemp));
-        lv_table_set_cell_type(forecast, 3, i, TemperatureStyle(minTemp));
+      for (int i = 0; i < optCurrentForecast->nbDays; i++) {
```
FintasticMan, line 156: It should be correct to use nbDays rather than MaxNbForecastDays here, but we should test it before merging.

JF002 (approved):
LGTM!

PR #2016, opened 2024-02-12
chrono: Resolve TODOs related to C++20 chrono feats

FintasticMan:
There are a couple more TODOs related to using `std::chrono::year`, which was added in C++20, but not supported by GCC 10 it seems.


JF002 (approved):
LGTM!

> [...] but not supported by GCC 10 it seems.

We might want to update our version of GCC to fix those issues and benefit from more recent compiler :)

PR #2020, opened 2024-02-13
Fix OOB flash write when DFU interrupted

mark9064:
If a DFU is restarted, the write indices aren't reset causing the image to be written out of bounds. The CRC check prevents the faulty image from booting but LittleFS still gets nuked.

There are other issues with DFU if the central is acting maliciously, but this resolves a corruption encountered in normal usage. If a DFU fails for whatever reason (bad bluetooth signal, whatever), the user is likely to retry the update which will fail if they do not reboot first due to this issue.


FintasticMan (approved):
LGTM!

JF002 (approved):

PR #2027, opened 2024-02-22
Fix PineTime Link in README.md

dcoolArne:


dcoolArne:
There is already another pull request

PR #2037, opened 2024-03-14
Fix bluetooth setting : emit the radio toggle message only if the value of the setting changed

### LINKED ISSUE #1961: Bluetooth setting is behaving abnormally

JF002:
Emit the message BleRadioEnableToggle to DisplayApp only if the enable state of the radio has actually changed.

This fixes an issue where the BLE connected logo would disappear when opening and closing the BLE setting (without changing it) while InfiniTime was already connected to a companion app.

This fix is based on the work done by @JustScott in #1972 and also thanks to the very detailed description of the issue by @rambonette in #1961.

I think this fix makes more sense since we do not actually need to call `NimbleController::EnableRadio()` again if the radio was previously enabled and need to stay enabled. 

This change also removed the need to send the message in `SettingBluetooth::~SettingBluetooth()` by capturing this in the lambda (which is called in `CheckboxList::~CheckboxList()` which is destroyed before `SettingBluetooth`.

What do think about this @JustScott?

Fixes #1961.

### DIFF
--- a/src/displayapp/screens/settings/SettingBluetooth.cpp
+++ b/src/displayapp/screens/settings/SettingBluetooth.cpp
@@ -36,17 +36,19 @@ namespace {
 
 SettingBluetooth::SettingBluetooth(Pinetime::Applications::DisplayApp* app, Pinetime::Controllers::Settings& settingsController)
   : app {app},
+    settings {settingsController},
     checkboxList(
       0,
       1,
       "Bluetooth",
       Symbols::bluetooth,
       settingsController.GetBleRadioEnabled() ? 0 : 1,
-      [&settings = settingsController](uint32_t index) {
+      [this](uint32_t index) {
         const bool priorMode = settings.GetBleRadioEnabled();
         const bool newMode = options[index].radioEnabled;
         if (newMode != priorMode) {
           settings.SetBleRadioEnabled(newMode);
+          this->app->PushMessage(Pinetime::Applications::Display::Messages::BleRadioEnableToggle);
         }
       },
       CreateOptionArray()) {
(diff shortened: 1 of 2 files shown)


rambonette:
Glad to help! Wish I had more time to tackle these problems myself 😵‍💫

JustScott:
Glad I could help @JF002! Your solution address the issue at its core and works great on my pinetime!

JustScott (approved):

FintasticMan (approved):

PR #2038, opened 2024-03-14
Restore brightness level when closing the Flashlight application

### LINKED ISSUE #1751: Brightness icon changes to flashlight's brightness when flashlight is closed by swipe

JF002:
The brightness level is stored when entering the app, and restored when closing it. Based on another fixed suggested by @serosi in #1907.

Fixes #1751.

### DIFF
--- a/src/displayapp/screens/FlashLight.cpp
+++ b/src/displayapp/screens/FlashLight.cpp
@@ -15,7 +15,7 @@ namespace {
 }
 
 FlashLight::FlashLight(System::SystemTask& systemTask, Controllers::BrightnessController& brightnessController)
-  : systemTask {systemTask}, brightnessController {brightnessController} {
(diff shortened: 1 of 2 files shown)


mark9064:
This is fixed as part of the brightness refactors for AOD, but merging this now would be fine, no problem rebasing

JF002:
Closing this PR since the original issue was fixed by #1869

PR #2043, opened 2024-03-22
LCD delay improvements

mark9064:
- No longer spins the CPU when the display task is waiting
- Unnecessary waits removed
- Sleep in - sleep out minimum interval now respected

This all assumes the datasheet is correct :P It could definitely be inconsistent in places
I'd appreciate if it at least one reviewer could check that they interpret the datasheet the same way
https://wiki.pine64.org/images/5/54/ST7789V_v1.6.pdf
See pages 48, 49, 163, 182, 184

Split from #1869

### DIFF
--- a/src/drivers/St7789.cpp
+++ b/src/drivers/St7789.cpp
@@ -1,6 +1,5 @@
 #include "drivers/St7789.h"
 #include <hal/nrf_gpio.h>
-#include <libraries/delay/nrf_delay.h>
 #include <nrfx_log.h>
 #include "drivers/Spi.h"
 
@@ -45,7 +44,7 @@ void St7789::WriteSpi(const uint8_t* data, size_t size) {
 
 void St7789::SoftwareReset() {
   WriteCommand(static_cast<uint8_t>(Commands::SoftwareReset));
-  nrf_delay_ms(150);
+  vTaskDelay(pdMS_TO_TICKS(150));
 }
 
 void St7789::SleepOut() {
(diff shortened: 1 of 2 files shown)


src/drivers/St7789.cpp
```diff
@@ -158,8 +181,13 @@ void St7789::DrawBuffer(uint16_t x, uint16_t y, uint16_t width, uint16_t height,
 
 void St7789::HardwareReset() {
   nrf_gpio_pin_clear(pinReset);
-  nrf_delay_ms(10);
+  vTaskDelay(pdMS_TO_TICKS(1));
```
FintasticMan, line 184: Wouldn't this delay for 0 ticks?

src/drivers/St7789.h
```diff
@@ -2,6 +2,10 @@
 #include <cstddef>
 #include <cstdint>
 
+#include <hal/nrf_gpio.h>
+#include <nrfx_log.h>
+#include "drivers/Spi.h"
+
```
FintasticMan, line 8: Why move thse to the header? Also, if we do move them, we should remove the Spi forward declaration.

mark9064: Honestly not totally sure, I wrote this a while ago. Will move them back I think unless I rediscover why...

mark9064:
(InfiniSim PR now ready)

JF002 (approved):
LGTM!
I can't remember why I added those delays in `St7789::ColMod()`, `St7789::DisplayInversionOn()`, `St7789::NormalModeOn()`  and a few other methods. It seems to work fine without them (and I trust you made those changes according to the data sheet anyway :) ).

Issue #1022, opened 2022-03-07
Inverted colors after dead battery


jgonyea:
### Verification

- [X] I searched for similar bug reports and found none was relevant.

### What happened?

Infinitime OS colors are inverted after running battery completely out and recharging.

### What should happen instead?

Colors should remain as normal

### Reproduction steps

I'm currently on Infinitime 1.8.0

I ran out of battery power one day when I forgot to recharge the watch.  After placing the watch on the charger, the watch rebooted, but the colors appear to be inverted (see attached photo below).  I can attach a video

The colors are correct during boot (the pine logo shows the green progress loading indicator), but as soon as Infinitime completes its boot, the colors flash the correct colors briefly, then invert.

### More details?

Attached image of inverted screen:
![IMG_20220306_214426](https://user-images.githubusercontent.com/1057441/156959302-ed27b3e0-ca67-4d9c-8b07-36ded9d6bdb9.jpg)

Attempts to fix (all fail):
* Reboot
* Reflash 1.8.0

The only fix I've found for the last time this happened is to let the watch battery completely die again, then recharge again.

### Version

1.8.0 (adc7909c)

### Companion app

Gadgetbridge 0.65.0

jgonyea:
Follow up, after letting the battery run out, this time it did not return to normal colors. The face is still inverted.

jgonyea:
Another followup:
I downgraded to 1.7.1 and the colors were correct again.  I then restarted (to roll back to 1.8.0) and colors remained normal.

jgonyea:
What's odd is that the colors are correct during the logo boot process (show a green or red pine logo), but then inverted after the full OS is booted.

geekbozu:
Could be a bad theme setting somehow... not sure off hand!

Colt45ws:
I have this problem with the PineTimeStyle watch face. One time it did this, I rebooted and it went back to normal. After the 2nd time I havent been able to fix it. I switched to Analog watch face it its normal. The menus are not affected on mine, only that one watch face.

koolfy:
Menus are affected for me when this happens, I fixed it just yesterday by charging the battery to 100%
This is all very confusing, I'm not sure I correctly describe what actually addressed the situation. 

It has happenned 3-4 times for me as well, all on 1.8.0 and I am using pinetimestyle as well (but the inverted colors happen in all menus and at a low level. I also notice boot screen is not inverted)

koolfy:
[...]

PR #2046, opened 2024-03-25
SPI transaction hooks

### LINKED ISSUE #1984: Colors invert randomly
### Verification

### What happened?

Colors inverted for no apparent reason

Colors should stay the same

### Reproduction steps

?

### More details?

Looks similar to #1022 but did not happen when charging
Happened on 2 different Infinitime watches
IMG_20240119_193153_564
IMG_20240119_193202_507

### Version

1.14.0

### Companion app

_No response_

mark9064:
This replaces the original technique for protecting the LCD data command pin (task notifications), which didn't work super well.

This is replaced with SPI transaction hooks, where when an SPI write request (transaction) is started, you can specify a hook that runs before the write and after it finishes. It's not a super elegant solution as the hook function has to be static (if someone with C++ knowledge wants to go nuts on that be my guest!) but it's pretty straightforward and simple at least. So for the LCD, the hook just sets the data command pin to the correct setting before the transaction starts, and does nothing on transaction end. It also allows us to remove task notifications completely which saves a bit of memory on each task.

I didn't bother hooking this up for reading as nothing needing hooks reads at the moment. But the idea would be identical to writes. Also, this PR only uses the pre-transaction action, so it could also be simplified to only support that.

As the hook function has to be static, the LCD driver is changed to reference the hardware pinmap directly. Not as nice an abstraction but we are only ever going to have one LCD so it's not too bad IMO.

Performance also seems to be better than waiting for notifications.

Closes #1984 (probably also #1022). Also fixes occasional temporary garbling of small parts of the display.
Split from #1869

### DIFF
--- a/src/drivers/St7789.cpp
+++ b/src/drivers/St7789.cpp
@@ -29,27 +29,27 @@ void St7789::Init() {
   DisplayOn();
 }
 
-void St7789::EnableDataMode(bool isStart) {
-  if (isStart) {
-    nrf_gpio_pin_set(pinDataCommand);
-  }
-}
-
-void St7789::EnableCommandMode(bool isStart) {
-  if (isStart) {
-    nrf_gpio_pin_clear(pinDataCommand);
-  }
-}
-
 void St7789::WriteData(uint8_t data) {
-  WriteSpi(&data, 1, [this](bool isStart) {
-    EnableDataMode(isStart);
+  WriteData(&data, 1);
+}
+
+void St7789::WriteData(const uint8_t* data, size_t size) {
+  WriteSpi(data, size, [pinDataCommand = pinDataCommand](bool isStart) {
+    if (isStart) {
+      nrf_gpio_pin_set(pinDataCommand);
+    }
   });
 }
 
-void St7789::WriteCommand(uint8_t cmd) {
-  WriteSpi(&cmd, 1, [this](bool isStart) {
-    EnableCommandMode(isStart);
+void St7789::WriteCommand(uint8_t data) {
+  WriteCommand(&data, 1);
+}
+
+void St7789::WriteCommand(const uint8_t* data, size_t size) {
+  WriteSpi(data, size, [pinDataCommand = pinDataCommand](bool isStart) {
+    if (isStart) {
+      nrf_gpio_pin_clear(pinDataCommand);
+    }
   });
 }
 
@@ -138,9 +138,7 @@ void St7789::SetAddrWindow(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1) {
 
 void St7789::WriteToRam(const uint8_t* data, size_t size) {
   WriteCommand(static_cast<uint8_t>(Commands::WriteToRam));
-  WriteSpi(data, size, [this](bool isStart) {
-    EnableDataMode(isStart);
-  });
+  WriteData(data, size);
 }
 
 void St7789::SetVdv() {
--- a/src/drivers/St7789.cpp
+++ b/src/drivers/St7789.cpp
@@ -34,10 +34,8 @@ void St7789::WriteData(uint8_t data) {
 }
 
 void St7789::WriteData(const uint8_t* data, size_t size) {
-  WriteSpi(data, size, [pinDataCommand = pinDataCommand](bool isStart) {
-    if (isStart) {
-      nrf_gpio_pin_set(pinDataCommand);
-    }
+  WriteSpi(data, size, [pinDataCommand = pinDataCommand]() {
+    nrf_gpio_pin_set(pinDataCommand);
   });
 }
 
@@ -46,15 +44,13 @@ void St7789::WriteCommand(uint8_t data) {
 }
 
 void St7789::WriteCommand(const uint8_t* data, size_t size) {
-  WriteSpi(data, size, [pinDataCommand = pinDataCommand](bool isStart) {
-    if (isStart) {
-      nrf_gpio_pin_clear(pinDataCommand);
-    }
+  WriteSpi(data, size, [pinDataCommand = pinDataCommand]() {
+    nrf_gpio_pin_clear(pinDataCommand);
   });
 }
 
-void St7789::WriteSpi(const uint8_t* data, size_t size, std::function<void(bool)> TransactionHook) {
-  spi.Write(data, size, TransactionHook);
+void St7789::WriteSpi(const uint8_t* data, size_t size, const std::function<void()>& transactionHook) {
+  spi.Write(data, size, transactionHook);
 }
 
 void St7789::SoftwareReset() {
--- a/src/drivers/St7789.cpp
+++ b/src/drivers/St7789.cpp
@@ -49,8 +49,8 @@ void St7789::WriteCommand(const uint8_t* data, size_t size) {
   });
 }
 
-void St7789::WriteSpi(const uint8_t* data, size_t size, const std::function<void()>& transactionHook) {
-  spi.Write(data, size, transactionHook);
+void St7789::WriteSpi(const uint8_t* data, size_t size, const std::function<void()>& preTransactionHook) {
+  spi.Write(data, size, preTransactionHook);
 }
 
 void St7789::SoftwareReset() {
--- a/src/drivers/St7789.cpp
+++ b/src/drivers/St7789.cpp
@@ -131,12 +131,11 @@ void St7789::SetAddrWindow(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1) {
   WriteData(y0 & 0xff);
   WriteData(y1 >> 8);
   WriteData(y1 & 0xff);
-
-  WriteToRam();
 }
 
-void St7789::WriteToRam() {
+void St7789::WriteToRam(const uint8_t* data, size_t size) {
   WriteCommand(static_cast<uint8_t>(Commands::WriteToRam));
+  WriteSpi(data, size, EnableDataMode);
 }
 
 void St7789::SetVdv() {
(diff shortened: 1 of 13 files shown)


mark9064:
(InfiniSim PR now ready)

JF002:
I've never been satisfied with the synchronization mechanism in the ST7789/SPI driver but I haven't tried to find a better design/implementation since I wrote it a while ago, so I'm happy someone is actually working on it.

The current implementation based on FreeRTOS notifications reduces the coupling between the drivers and the application layer, but add a dependency to FreeRTOS in the low-level driver, which is something I would like to avoid. It also adds a bit of overhead : once the write operation (aka Transaction) is done, a task notification is pushed to FreeRTOS. This notification will be taken care of when the scheduler decides to give the display task some CPU task and the notification.

We need this synchronization mechanism because of the Data/Command pin of the ST7789 LCD driver : we cannot change its state while a transaction is ongoing. Since this pin is not part of the SPI bus, it's managed by the ST7789 driver class. 
The data transfer to the SPI Bus is asynchronous (thanks to DMA). This is great for performance : it's faster and it allows the CPU to do other things while the transfer is ongoing. The SPI driver uses a mutex to ensure that the previous transfer is finished before starting a new one.
If the ST7789 class would simply change the state of the D/C pin before calling Spi::Write(), it could change the state of the pin while a transfer is still ongoing. That's why we wait for a notification from the SPI bus before changing the pin state and starting a new transfer.

The implementation in this PR reduces the overhead a lot : simply call a function (via a function ptr) that clears/sets the D/C pin before/after the transaction.

I totally agree with the idea, but the implementation has a few disadvantages that you've already mentioned : the callback needs to be static, so it doesn't know the ST7789 instance and the pin number that must be toggled.

Using `PinMap` is a nice workaround but I would rather like the low level drivers to be as self-contained as possible (that's why I think removing the dependency to FreeRTOS is a good idea). 

On a side note, I'm working a new design that will allow InfiniTime to build and run on other platform and hardware (see [my demo from FOSDEM](https://codeberg.org/JF002/InfiniTime-multi-target)) and reducing the dependence between modules and layers will definitely help for such an goal!

So how can we improve this? Maybe we can try with lambas and `std::function`  instead of plain function pointers? `std::function` might add indirection and dynamic memory allocation so we have to be very careful though.
But, all we want to store in the lambda/std::function is a pin number, maybe we can avoid the dynamic memory allocation, and the indirection while calling the function might still be more efficient than the notification? I guess all we have to do is test and measure :)

> Performance also seems to be better than waiting for notifications.

Out of curiosity, do you have any numbers to share about this?

mark9064:
No numbers unfortunately - just based off how it seems. I suspect as going through the RTOS adds some waiting due to task switching etc. But equally this could be entirely inside my head :)

There's a lot of work to do to abstract InfiniTime over different hardware. I'd love to see more on how you are approaching hardware abstraction in the future so I can write code that fits with that as a contributor - at the moment I have no clue what plans are. I'll take some time to look through the multi target repository, I missed the mention of that at FOSDEM!

Adding code that will cause known problems later makes no sense though, so I'm totally in favour of re-abstracting this in that case.

A simple extension would be to add an extra argument to the hook/spi Write that's a user data pointer. Then the hook can cast that pointer back and execute a method after. This is the LVGL/FreeRTOS way but isn't very pretty. As you've said though there'll be a nicer C++ abstraction for this so I'll experiment a little

JF002:
> No numbers unfortunately - just based off how it seems. I suspect as going through the RTOS adds some waiting due to task switching etc. But equally this could be entirely inside my head :)

That's fine. The more I work on display related stuff the more I realize that "it looks better to my eyes" is a valid metric :-)

My experiments with the multi-platform design is still early days, I'm trying multiple designs and implementations to find "the best ones". I mentioned it in the [last Pine64 community update](https://pine64.org/2024/03/17/march-update-making-waves/#pinetime) : it shows the same app running on the Pinephone, the PineTime and a BL618 (RISC-V) dev board.




Basically, the idea is built around C++ concepts. The "application" (DisplayTask, SystemTask, user applications,...) is common to all platform. Anything that can change from a platform/hardware to another is abstracted by C++20 concepts. For example, there's a `is_display` concept that defines a `Init()` function and a `DrawBuffer()` function. That's all the application knows about a Display.

Then, each platform has to provide an implementation of a "Display" that implements the contract `is_display`. I currently have 2 implementations : the ST7789 and the one based on SDL for the simulator.

The ST7789 takes an object that implements the contact `is_data_command_serial_bus`.  The actual implementation must take care of the D/C pin, the CS pin and the SPI bus. The implementation for the PineTime is a class `DisplaySpi` that drives the D/C pin and the SPI bus of the NRF52. The implementation for the BL618 uses the integrated "DBI" peripheral which handle everything (D/C, CS and SPI) in hardware.

If the take the example of this new hook, it definitively makes sense for the NRF implementation of the DataCommandSerialBus, but it will not be used for the BL618 implementation.

Anyway, this is still work in progress and might change in the future ;-)

> A simple extension would be to add an extra argument to the hook/spi Write that's a user data pointer. Then the hook can cast that pointer back and execute a method after. This is the LVGL/FreeRTOS way but isn't very pretty. As you've said though there'll be a nicer C++ abstraction for this 

Yes, that would work, but it looks like old C code style :-) We'll hopefully find something that looks a bit more like modern C++ with more type safety .

> so I'll experiment a little

Keep us up to date :-) I'll definitively have a look at those lambdas and std::function.

mark9064:
OK, re-abstracted to use lambdas now. If we decide to go with this approach I'll rebase the branch to be more sensible after

mark9064:
I've been doing some further reading on lambdas and std::function. It looks like that if the capture list is small enough, the captured variables fit within the std::function instance and so the heap is not used. Because the capture list is only a reference to an object, the heap should not be used in this case. So I think this avoids dynamic allocation. As for the cost, I believe it will add another layer of indirection as inlining will not be possible. But effectively, it boils down to a callable class that stores a pointer to the member function and a pointer to the instance to call the member function on, all of which fit on the stack. So a bit less efficient that the C style function pointer/data pointer approach, but only one function calls worth so perfectly acceptable IMO.

Could be worth analysing what the compiler actually does here. Doing this statically looks quite challenging though

mark9064:
Just refactored the write data/command interface to be a bit nicer. The lambdas also only capture one integer now and reference static functions only which should hopefully give the compiler an easier time optimising/inlining

src/drivers/St7789.cpp
```diff
@@ -29,18 +29,32 @@ void St7789::Init() {
...
+}
+
+void St7789::WriteCommand(uint8_t data) {
+  WriteCommand(&data, 1);
 }
 
-void St7789::WriteSpi(const uint8_t* data, size_t size) {
-  spi.Write(data, size);
+void St7789::WriteCommand(const uint8_t* data, size_t size) {
+  WriteSpi(data, size, [pinDataCommand = pinDataCommand](bool isStart) {
```
mark9064, line 49: To optimise more aggressively this hook could set the pin when `isStart` is false, and then the write data hook can be a no-op (nullptr)

JF002:
> It looks like that if the capture list is small enough, the captured variables fit within the std::function instance and so the heap is not used

Yes, that's also what I figured out. However, this behavior is not part of the specification : it's up to the implementation to decide to use the heap or not. 
However, I added a breakpoint in  `malloc()` and `pvPortMalloc()` and it wouldn't break when creating/executing the callback, so I guess it confirms that the heap is not use in this specific case.

> As for the cost, I believe it will add another layer of indirection as inlining will not be possible

I think this additional indirection is an acceptable tradeoff for a better design :)

> Could be worth analysing what the compiler actually does here. Doing this statically looks quite challenging though

I had a look at the disassembly code in CLion, but it just shows that it branches to the function in the STL, but does not show the ASM code of this function. I didn't try to figure that out from the complete ASM code of the project.

So, yeah, I think we can use this implementation based on std::function, it seems to work fine and improves the code at the "application" level (we do not have to manually wait for the end of the SPI transfer in LittleVGL.cpp).

I still have a few comments to do on the code, if you don't mind:
 - Since we do not have anything to do when `isStart` is false, I suggest we just remove this argument and rename the hook to something like "startOfTransactionHook" or "TransactionStartHook". 
 - Please remember to apply the camelCase [naming convention](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/doc/coding-convention.md#follow-these-guidelines-while-writing-code) for function parameters and class fields.
 - The std::function of the transaction hook could be moved from `St7789`  to `Spi` and then to `SpiMaster`. This will ensure that the object is not copied at each function call. For this, simply declare the parameter as a rvalue and call `std::move` when passing it to the next function. Ex:

```
void St7789::WriteSpi(const uint8_t* data, size_t size, std::function<void(bool)>&& TransactionHook) {
  spi.Write(data, size, std::move(TransactionHook));
}

bool Spi::Write(const uint8_t* data, size_t size, std::function<void(bool)>&& TransactionHook) {
  return spiMaster.Write(pinCsn, data, size, std::move(TransactionHook));
}

bool SpiMaster::Write(uint8_t pinCsn, const uint8_t* data, size_t size, std::function<void(bool)>&& transactionHook) {
  [...]
  this->TransactionHook = std::move(transactionHook);
```

Thanks!

JF002:
@mark9064 btw, I can apply all those small changes if you wish :)

mark9064:
OK, I think we were both overcomplicating it! As we don't need to store the lambda anymore, we can just pass in a reference. This way it's nice and simple

mark9064:
Noticed I forgot to apply your naming suggestion now the semantics of the hook has changed. Corrected now

JF002:
> As we don't need to store the lambda anymore, we can just pass in a reference.

Yes, you're right. Passing the reference works fine and does the job, but it, in this case, it works because the lambda is executed "synchronously" during the call to Spi.Write(). It wouldn't work for a post transaction hook, for example. 
In this regard, I think the move semantic is a bit better : the caller does not have to worry about the life time of the function pointer.

On the other hand, your implementation is simple and works so we can use it and extend it later on if we ever need it ;-)

JF002 (approved):

mark9064:
Sounds good. Let's switch to the move system if we ever need post-hooks. For now it's as simple and efficient as it can be

PR #2048, opened 2024-03-31
Stopwatch: allow sleep

liamcharger:
I often use a stopwatch for things like running, and noticed that the watch won’t sleep when the stopwatch is running. 

This PR gives the stopwatch the ability to run while the display is off.

### DIFF
--- a/src/displayapp/screens/StopWatch.cpp
+++ b/src/displayapp/screens/StopWatch.cpp
@@ -3,6 +3,8 @@
 #include "displayapp/screens/Symbols.h"
 #include "displayapp/InfiniTimeTheme.h"
 
+#include "systemtask/SystemTask.h"
+
 using namespace Pinetime::Applications::Screens;
 
 namespace {
@@ -34,7 +36,7 @@ namespace {
   constexpr TickType_t blinkInterval = pdMS_TO_TICKS(1000);
 }
 
-StopWatch::StopWatch(System::SystemTask& systemTask) : systemTask {systemTask} {
+StopWatch::StopWatch() {
   static constexpr uint8_t btnWidth = 115;
   static constexpr uint8_t btnHeight = 80;
   btnPlayPause = lv_btn_create(lv_scr_act(), nullptr);
@@ -79,7 +81,6 @@ StopWatch::StopWatch(System::SystemTask& systemTask) : systemTask {systemTask} {
 
 StopWatch::~StopWatch() {
   lv_task_del(taskRefresh);
-  systemTask.PushMessage(Pinetime::System::Messages::EnableSleeping);
   lv_obj_clean(lv_scr_act());
 }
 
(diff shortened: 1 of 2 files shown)


mark9064:
So keep-awake was added in #432

I'm guessing it was so the device doesn't go to sleep right as you want to stop the timer / record a lap. So for people using the stopwatch to time things exactly / shorter durations, this PR is a regression. But when using the stopwatch for longer durations, keeping the screen on is a waste of energy and is annoying (e.g accidental touchscreen clicks while running from clothes)

Maybe there's a solution that keeps both parties happy?

liamcharger:
That's true...

Having a setting for the stopwatch specifically would be a waste of resources--maybe implementing a setting for all the apps to enable or disable sleep could be nice, I don't know if the benefit would be great enough. I can see it possibly working if the user wanted to stop the watch from sleeping when using the navigation app, or maybe when playing a game?

mark9064:
Hmmm, I like the idea but I don't think it'll be convenient in practice as you may want different settings for each app, and also you need to remember to change it before you start the stopwatch. I think a keep awake toggle in the watch app itself (maybe hidden as a long press or a new button) is the only thing that would work, but I don't know how it'd be implemented in a way that keeps to the InfiniTime vision

liamcharger:
> I don't think it'll be convenient in practice as you may want different settings for each app

I meant something like a list of checkboxes in the settings app for each user app, and selecting one would enable/disable sleep for that specific app. 

> I don't know how it'd be implemented in a way that keeps to the InfiniTime vision

What part of the vision would it be violating?

mark9064:
Especially with #2141, I'm leaning towards allowing sleep by default as well. There's just not that many situations where I want a screen on stopwatch compared to one where I want to allow idling. Especially as keeping the screen on means ghost touch inputs are very common

liamcharger:
Agreed! I just updated the branch and fixed merge conflicts, so feel free to review.

src/displayapp/screens/StopWatch.cpp
```diff
@@ -3,6 +3,8 @@
 #include "displayapp/screens/Symbols.h"
 #include "displayapp/InfiniTimeTheme.h"
 
+#include "systemtask/SystemTask.h"
```
mark9064, line 6: Is this include still needed?

liamcharger: I don't believe so... Removed.

mark9064:
Looks like you might also need `#include <algorithm>` and perhaps some others (I guess SystemTask includes them so previously you were getting them with that)

liamcharger:
Yep, noticed the build failure. I'll get to it as soon as I'm on my laptop.

mark9064:
Could you rebase this? Happy to approve after, sorry I missed this PR last time

liamcharger:
Just used the merge button on github.com

Hopefully that's alright :)

mark9064 (approved):
LGTM :)

IMO keep awake with the stopwatch makes it less usable due to accidental screen touches pausing/resetting it, so it's better without (even though the screen turning off might be annoying sometimes too)

PR #2049, opened 2024-03-31
Alarm: improve preview UX

liamcharger:
This PR improves the user experience when viewing the time until the alarm goes off. 

Previously, it was a navy blue rectangle with no indication of how to dismiss the dialog. It’s not hard to figure out how to dismiss it, but as the InfiniTime vision states, it should be predictable and easy to use. 

I think the updated UI can be improved, so any feedback is welcome.

### DIFF
--- a/src/displayapp/screens/Alarm.cpp
+++ b/src/displayapp/screens/Alarm.cpp
@@ -148,7 +148,7 @@ void Alarm::OnButtonEvent(lv_obj_t* obj, lv_event_t event) {
       ShowInfo();
       return;
     }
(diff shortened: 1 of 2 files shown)


liamcharger:
Sure, I've just updated my original comment!

minacode:
Nice idea to make it fit better to the rest :) 
I think it would look good if you align the hours, minutes, ... vertically like they are in the previous version. That makes it easier to understand.

tituscmd:
I'd propose the idea of putting the X on the info screen at the same location as the i on the alarm screen.

PR #2067, opened 2024-05-28
settings: use 2024 as default year

vladopajic:
having 2024 as default year will make it less annoying to configure year

### DIFF
--- a/src/displayapp/screens/settings/SettingSetDate.h
+++ b/src/displayapp/screens/settings/SettingSetDate.h
@@ -29,7 +29,7 @@ namespace Pinetime {
 
         Widgets::Counter dayCounter = Widgets::Counter(1, 31, jetbrains_mono_bold_20);
         Widgets::Counter monthCounter = Widgets::Counter(1, 12, jetbrains_mono_bold_20);
-        Widgets::Counter yearCounter = Widgets::Counter(1970, 9999, jetbrains_mono_bold_20);
+        Widgets::Counter yearCounter = Widgets::Counter(2024, 9999, jetbrains_mono_bold_20);
       };
     }
   }


NeroBurner (approved):
We probably have to update the variable year by year then, but I think it makes sense to have the default year close the the release date of the firmware

mark9064:
One option would be using the compiler to insert the current year at time of compilation, I'm sure that's possible as the build date page in settings already exists

mark9064:
Also this is the wrong place to introduce a default year. By changing the counter parameter instead the minimum year the spinbox can show is adjusted. The spinboxes otherwise display the current time. The correct way to do this is to initialise the default time on boot to 1 Jan on the year of the compile time. This means that the time can still be adjusted freely. I would suspect that main.cpp where the time is loaded from the persistent RAM is the correct place to set the time as a fallback.

The compile time could be used as the default date and time, but I think this is confusing whereas setting to 1 Jan YEAR 00:00 is unambiguously a default time.

mark9064 (changes requested):
This is the wrong place to introduce a default year. By changing the counter parameter instead the minimum year the spinbox can show is adjusted. The spinboxes otherwise display the current time. A better and more correct way to do this is to initialise the default time on boot to 1 Jan on the year of the compile time. This means that the time can still be adjusted freely. I would suspect that main.cpp where the time is loaded from the persistent RAM is the correct place to set the time as a fallback.

The compile time could be used as the default date and time, but I think this is confusing whereas setting to 1 Jan YEAR 00:00 is unambiguously a default time.

FintasticMan:
I like the idea of using the `__DATE__` macro to use the current year, but something bugs me about using `atoi` to do the conversion as that'll run at runtime rather than compile-time. That's why it could be fun (although completely unnecessary) to write a simple `constexpr` `atoi`-like function so that all of it can be done at compile time.

mark9064:
This attempt at constexpr atoi looks pretty bleak https://stackoverflow.com/a/59537554 😅
Maybe we only need part of it though

FintasticMan:
```cxx
constexpr int isdigit_constexpr(int c) {
    return c >= '0' && c <= '9';
}

constexpr int atoi_constexpr(char const *str) {
    int result = 0;
    while (isdigit_constexpr(*str)) {
        result = result * 10 + *str - '0';
        ++str;
    }
    return result;
}
```

I have no idea why that implementation looks so difficult, this is all it takes.

mark9064:
That looks much more reasonable and does the job here :+1:

vladopajic:
@mark9064 thanks, good suggestions. 

however i would not be able to make all those changes. my reasoning was to to make this PR because it is vary easy to change constant, instead of opening an issue about this.

please feel free to take over this PR or close it and make another one.

mark9064:
With #2133 this is integrated now and can be closed

FintasticMan:
@vladopajic Thank you for the idea of setting the year by default! And for the initial implementation, it was very helpful for working out the final solution.

vladopajic:
> @vladopajic Thank you for the idea of setting the year by default! And for the initial implementation, it was very helpful for working out the final solution.

🙏

kudos to you guys!

PR #2070, opened 2024-06-02
Fix MotorController::Init call order

pipe01:
The call to `LoadNewScreen` [here](https://github.com/InfiniTimeOrg/InfiniTime/blob/0dcfb2edb7ac071aa0a22c01609122d577d4c05d/src/displayapp/DisplayApp.cpp#L130) (and the one below it) calls `MotorController::StopRinging()` [here](https://github.com/InfiniTimeOrg/InfiniTime/blob/0dcfb2edb7ac071aa0a22c01609122d577d4c05d/src/displayapp/DisplayApp.cpp#L416), which in turn calls `xTimerStop` to stop the ringing timer. However, this timer only gets created when [`DisplayApp::InitHw`](https://github.com/InfiniTimeOrg/InfiniTime/blob/0dcfb2edb7ac071aa0a22c01609122d577d4c05d/src/displayapp/DisplayApp.cpp#L150) is called, which currently happens in [`DisplayApp::Process`](https://github.com/InfiniTimeOrg/InfiniTime/blob/0dcfb2edb7ac071aa0a22c01609122d577d4c05d/src/displayapp/DisplayApp.cpp#L143) after the initial call to `LoadNewScreen`. This means that `xTimerStop` is called on uninitialized data, which has undefined behaviour.

This PR makes it so that the call to `MotorController::Init` happens before `LoadNewScreen`.

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -125,6 +125,7 @@ void DisplayApp::Start(System::BootErrors error) {
   bootError = error;
 
   lvgl.Init();
+  motorController.Init();
 
   if (error == System::BootErrors::TouchController) {
     LoadNewScreen(Apps::Error, DisplayApp::FullRefreshDirections::None);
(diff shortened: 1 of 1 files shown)


JF002 (approved):
Good catch, thank you!

I'm wondering if those calls to `LoadNewScreen` shouldn't go in `InitHw()` or in `Process()` to make things more tidy, but I think your fix is valid anyway :)

PR #2072, opened 2024-06-03
watchfaces: add LCARS

w4tsn:
This adds an LCARS watchface, a UI known from the Star Trek series.

The following issues are still open and I might need a hand with these:

- the background image is around 170 kB. I suspect because of it's size (240x240) and the embedded text. This could be reduced by a lot I guess if converted to point-rendering in lvgl similar to Infineat
- there are alignment issues currently only solvable through calling the lv_obj_align function (again) after calling the local label_make function, which I added to reduce duplicate code
- coming from the settings screen the watchface loads quite slowly, probably because of the size of the background image
- needs rebase to latest development branch, currently tracks 1.14.0
- the app is currently fully integrated (as default in the build) because I did not catch how to make it a "UserApp" that can be selected when invoking cmake

Screenshot from 2024-06-03 18-34-32

image

[Screencast from 2024-06-03 18-34-02.webm](https://github.com/InfiniTimeOrg/InfiniTime/assets/2271467/b2baef41-6065-42e8-bcad-afbcfbf379e1)

EDIT: here the pre-build OTA files for the firmware and resources.

[pinetime-mcuboot-app-dfu-1.14.0.zip](https://github.com/user-attachments/files/15944699/pinetime-mcuboot-app-dfu-1.14.0.zip)
[infinitime-resources-1.14.0.zip](https://github.com/user-attachments/files/15944701/infinitime-resources-1.14.0.zip)


ljahn:
I'm more of a TNG fan :wink: but this looks nice :+1: 

Having the background as an image really is using a lot of storage space. Alas, you can not render images from the SPI flash, so making this fit is going to be difficult. Have you been able to run it on the device, not the simulator? 
If you decide to convert it to be rendered "vector graphics style" with lvgl, watch out for the heap memory usage. If you go crazy detailed (like I did once with my watch face), you can easily fill the heap completely and get weird behaviour that does not show on the sim.
For an immediate size decrease of the background picture, you could also use one of lvgl's indexed color modes. If you convert to 4 or 8 bit indexed color, you can only have 16 or 265 colors, respectively (16 will certainly make your curves look less anti-aliased), but the storage use would decrease to 29kB or 58kB.

You alignment issue is a funny one that definitely never happened to anyone I know :wink:.
You pass the value `-5` to an argument of type `uint8_t`.

ljahn:
> Alas, you can not render images from the SPI flash

This is wrong. I got mixed up while thinking about how I went to optimize the storage needs of my development. 
I tried to save space by re-using duplicate parts, but I could not get the rotation features of lvgl to work.
It turned out it is not possible to _rotate_ images that are displayed from the SPI flash.

You clearly can render images from SPI flash (actually you do it). But I would also think this is the reason the watch face loads slowly.
Nevertheless indexed color will probably speed things up by giving you size reduction.

L3br4nd:
@w4tsn I really like this watchface. Do you think it is possible to additionally add the current weather information? Probably the most useful way would be to make it configurable. E.g. I don't use the heart rate measurement because it drains the battery so fast but I like to know the (rough) temperature. An option to configure the available information like in the PineTimeStyle watchface were awesome but I understand if this takes some time since it increases the complexity of the watch face considerably.

jimfim:
Great watchface, I started using it yesterday. After some time though. Usually after a notification the screen goes all funky and after  a few minutes the firmware rolls back

jimfim:
You can replicate the issue by just toggling back and forwards between the watch face and the notifications. You don't need any notifications. After about 20~ish transitions you see what I see above in my pictures. 
Checking the minfree value between transitions you can see the value reduce to 0 as you  perform more transitions and eventually you will get alloc errors and the watch will reset.

romibi:
The memory leak is in `set_label_text_from_uint`. I unrolled the loop and did not call that function and the issue is gone.
Maybe the function could be fixed but i though unrolling the loop and hard coding the labels was easier …

I also have made a version that does not use the image as a background but draws it with circles and rectangles and labels.
It draws faster but the remaining memory is veeery tight and the code is not nice yet … nothing i can publicly commit yet …

src/displayapp/screens/WatchFaceLCARS.cpp
```diff
@@ -0,0 +1,345 @@
...
+      ampmChar[0] = 'P';
+    } else if (hour > 12) {
+      hour = hour - 12;
+      ampmChar[0] = 'P';
+    }
+    lv_label_set_text(labelTimeAmPm, ampmChar);
+    weekNumberFormat = "%V";
+  } else {
+    weekNumberFormat = "%U";
+  }
```
romibi, line 224: Whats the reasoning for 12h → Week starts at Monday (%V), 24h → Week starts at Sunday (%U)?
I personally prefer 24h format with a Week that starts at Monday …

romibi:
See [My copy of this branch](https://github.com/romibi/InfiniTime/commits/watchface-lcars/) where I:
- Rebased onto main
- Fixed the memory leak
Still not sure I fixing the function would be nicer instead of unrolling the loops.
But I'm not too used to cpp to be certain how/when to properly free the temporary variables.
- Fixed the alignment issues
Main culprit was providing negativ numbers to uint8_t
And for one situation providing the wrong "container"
- Fixed the week number format mixup mentioned above
- Implemented drawing the bg with circles, rectangles and labels
Not completely happy with the font for the bg labels
- Added weather info for when no heartrate is available

@w4tsn You can cherry pick from it or reset to it to update this pull request …

Todo: 
- remove png again (and maybe svg … or keep as reference?)
- not happy with memory usage

Pictures:
image

image

Build results:
[pinetime-mcuboot-app-dfu-1.15.0.zip](https://github.com/user-attachments/files/18416907/pinetime-mcuboot-app-dfu-1.15.0.zip)
[infinitime-resources-1.15.0.zip](https://github.com/user-attachments/files/18416908/infinitime-resources-1.15.0.zip)

L3br4nd:
@romibi nice, thank you 😊

romibi:
While using this watchface (my updated version) for the past two weeks I was quite happy how it is.
But one thing I was missing was some indication on if I missed any notifications.
So I added notification icons/count to [my branch](https://github.com/romibi/InfiniTime/commits/watchface-lcars/) of this.
(Todos from above unchanged)

Notification count if any:
image

Icon for not yet displayed Notifications (When notifications off/sleep mode):
image

Build results:
[pinetime-mcuboot-app-dfu-1.15.0.zip](https://github.com/user-attachments/files/18586531/pinetime-mcuboot-app-dfu-1.15.0.zip)
[infinitime-resources-1.15.0.zip](https://github.com/user-attachments/files/18586533/infinitime-resources-1.15.0.zip)

av1d:
That was a difficult task because of the ecosystem in general but I finally got this installed. It's the best face I've seen yet, love all the features and the aesthetic. Thank you!

PR #2076, opened 2024-06-13
README: Change links from Pine wiki to new documentation

1Maxnet1:
As the Wiki was replaced by a new documentation website and is read-only, we should update the links, to point to the most up-to-date documentation.


JF002 (approved):

JF002:
Thanks :+1:

Issue #1275, opened 2022-08-11
Failed to build InfiniTime on Docker Container under Windows 10


A4EVA:
### Verification

- [X] I searched for similar bug reports and found none was relevant.

### What happened?

Docker container do not build the firmware under Windows 10

### What should happen instead?

The Build process should have been completed

### Reproduction steps

Just clone and initialize the repository, after that pull the docker image with `docker pull infinitime/infinitime-build` considering that we are under Windows to avoid problems instead of using `docker run --rm -it -v $(pwd):/sources infinitime/infinitime-build` use `docker run --rm -it -v C:\...FULL-PATH\Projects\InfiniTime:/sources infinitime/infinitime-build` the process will start but once reached 36 % of the build this error will come up : 

`[ 36%] Linking C static library libnimble.a
[ 36%] Built target nimble
gmake: *** [Makefile:91: all] Error 2
+ return 1
+ BUILD_RESULT=1
+ '[' '' '!=' true -a 1 == 0 ']'`

once inside the container with: 
`docker run --rm -it -v C:\...FULL-PATH\Projects\InfiniTime:/sources infinitime/infinitime-build bin/bash` 
I have launched the build process manually with make and a bit more exact error can now be seen:

`[ 36%] Generating jetbrains_mono_bold_20.c
(Stripping trailing CRs from patch; use --binary to disable.)
patching file jetbrains_mono_bold_20.c
patching file jetbrains_mono_bold_20.c
patch: **** malformed patch at line 3:

Traceback (most recent call last):
  File "/sources/src/displayapp/fonts/generate.py", line 75, in <module>
    main()
  File "/sources/src/displayapp/fonts/generate.py", line 70, in main
    subprocess.check_call(['/usr/bin/env', 'patch', name+'.c', patch])
  File "/usr/lib/python3.10/subprocess.py", line 369, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['/usr/bin/env', 'patch', 'jetbrains_mono_bold_20.c', 'jetbrains_mono_bold_20.c_M.patch']' returned non-zero exit status 2.
make[2]: *** [src/displayapp/fonts/CMakeFiles/infinitime_fonts_jetbrains_mono_bold_20.dir/build.make:73: src/displayapp/fonts/jetbrains_mono_bold_20.c] Error 1
make[2]: *** Deleting file 'src/displayapp/fonts/jetbrains_mono_bold_20.c'
[...]

PR #2077, opened 2024-06-14
Fix EOL related build issues on Windows

madmini:
Proper fix for #1275 that does no longer require manual intervention. 

On Windows, git has `core.autocrlf=true` set at _system_ level, hence any Windows user who has not changed their _global_ git config will checkout all text files with CRLF line endings. For some files this seems to cause build issues.

With this change, I can build using docker on Windows 11 as follows:

```pwsh
git clone https://github.com/madmini/InfiniTime.git --branch fix/eol-on-windows
cd InfiniTime
git submodule update --init
docker run --rm -it -v ${PWD}:/sources infinitime/infinitime-build
```

Without the change, the build fails with an error that matches the one described in #1275:

```
...
[ 34%] Linking C static library libnimble.a
[ 34%] Built target nimble
gmake: *** [Makefile:91: all] Error 2
```

<hr />

Note: when changing between branches with varying `.gitattributes` with `checkout` instead of directly cloning with `--branch`, resetting the line endings is more complicated:

```pwsh
git clone https://github.com/madmini/InfiniTime.git
# CRLF line endings
git checkout fix/eol-on-windows
# still CRLF line endings
git rm --cached -r .
git reset --hard
# LF line endings
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,5 @@
 # Set the default behavior, in case people don't have core.autocrlf set.
-* text=auto
+* text=auto eol=lf
 
 # Explicitly declare text files you want to always be normalized and converted
 # to native line endings on checkout.
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,8 +1,9 @@
 # Set the default behavior, in case people don't have core.autocrlf set.
+# Prevent build errors on non lf systems (like Windows), we need files with lf as newlines.
 * text=auto eol=lf
 
 # Explicitly declare text files you want to always be normalized and converted
-# to native line endings on checkout.
+# to lf line endings on checkout.
 *.c text
 *.cpp text
 *.h text


.gitattributes
```diff
@@ -1,5 +1,5 @@
 # Set the default behavior, in case people don't have core.autocrlf set.
-* text=auto
+* text=auto eol=lf
```
NeroBurner, line 2: add a comment on why `lf` is set

```suggestion
# Prevent build errors on non lf systems (like Windows), we need files with lf as newlines.
* text=auto eol=lf
```

.gitattributes
```diff
@@ -1,5 +1,5 @@
...
 
 # Explicitly declare text files you want to always be normalized and converted
 # to native line endings on checkout.
```
NeroBurner, line 5: Update the comment such that it reflects that we don't want the native line endings but uniformly `lf`

```suggestion
# Explicitly declare text files you want to always be normalized and converted
# to lf line endings on checkout.
```

NeroBurner (approved):
looking good. I'm scheduling for 1.16.0 as 1.15.0 feels close

mark9064:
If you rebase this onto main the CI should work now #2166 is merged

madmini:
> If you rebase this onto main the CI should work now #2166 is merged

done

mark9064 (approved):

PR #2078, opened 2024-06-14
Remove defunct .gitconfig file

madmini:
The `.gitconfig` is not actually used by git unless manually configured to do so. The repo-specific git config is located in `.git/config`. The only `.gitconfig` file read by git is the one in the user home directory. See [`git config` docs](https://git-scm.com/docs/git-config) and the answers to [this stackoverflow question](https://stackoverflow.com/q/18329621).

I believe it is misleading to have this file in the repo as it implies the config is automatically applied. I for one have wasted a lot of time trying to fix the `.gitconfig` file before I realized this and came up with #2077 instead.

If I am mistaken and the `.gitconfig` file represents recommendations for contributors or similar then I strongly recommend to have this reflected somewhere in the docs, or noted as a comment in the defunct `.gitconfig` file.

### DIFF
--- a/.gitconfig
+++ b/.gitconfig
@@ -1,5 +1,0 @@
-[core]
-    whitespace = blank-at-eol,blank-at-eof,space-before-tab
-    autocrlf = input
-[apply]
-    whitespace = fix


madmini:
The options, as outlined in #191 should probably be added to the docs.
#2077 should take care of line endings.

mark9064:
Agreed, it makes sense to move it into the docs. I think it'd make sense to move it and remove the file at the same time (otherwise it might get forgotten about), would you be able to move it into the docs as part of this PR? If you don't have the time that's no problem, I'll try get round to it at some point.

Sorry for the glacial review :(

madmini:
I have added `git config set ...` to "Clone the repo" steps in the docs. Not sure if this is enough and/or the right place.

I have considered adding a bullet point to the ["PR checklist" in CONTRIBUTING.md](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/CONTRIBUTING.md#pr-checklist):
> - Configure Git to automatically fix whitespace issues:
>  
>   ```sh
>   git config set core.whitespace trailing-space,space-before-tab
>   git config set apply.whitespace fix
>   ```
but this seems a bit out of place.

mark9064 (commented):
Actually on further thought, I think it might just be best to remove it:
`whitespace = blank-at-eol,blank-at-eof,space-before-tab` is the default, so this line does nothing unless you modified your global git config (and you better know what you're doing if you have)
And bad whitespace in general will be caught by clang-format, so CI will fail. I suspect most people are using editors which strip whitespace anyway: I don't think anybody really knows about the .gitconfig and I haven't seen any whitespace issues when reviewing

What do you think?

madmini:
> `whitespace = blank-at-eol,blank-at-eof,space-before-tab` is the default

ah, I missed this. for some reason I believed that one of them was not default.

> bad whitespace in general will be caught by clang-format, so CI will fail

I assume clang-format will only format C files? Then python scripts and other stuff require extra attention. But yes, I think this is the way to go.

> I suspect most people are using editors which strip whitespace anyway

yes

> I don't think anybody really knows about the .gitconfig

certainly not me before googling for this PR 😅

> What do you think?

I have not yet seen many projects that make recommendations for `git config` and I think my current way to put it in the docs (add `git config ...` to any `git clone ...` instructions) is weird and prone to be ignored. If you want standards to be enforced, then check them in CI, as you do. So as the settings are the default anyway AND for C code its checked in CI AND its weird, I'm with you on just removing it.

I'll add a revert commit, just squash it ^^

mark9064 (approved):
Fully agree, thanks for the comments and patch :)

JF002 (approved):

PR #2085, opened 2024-06-30
Improve timer vibration

DavisNT:
This is a super-simple fix for #391.
When timer ends the watch will vibrate until the side button is pressed or a sliding-down movement is made on the screen (maybe there are some more gestures).


mark9064:
For reference: alternative to #1971

DavisNT:
@NeroBurner, @mark9064 Could it be possible to review and schedule this super-simple bugfix for 1.16.0 (or maybe even include it in 1.15.0)?

mark9064:
1.15 is in feature freeze, and I suspect by the time 1.16 rolls around #1971 will probably be merged? If it doesn't move forward though I agree that ringing would be a good improvement

PR #897, opened 2021-12-30
BLE: announce heart rate service in ADV packet.

dennisguse:
Fixes #895.

Not tested due to lack of device.
Did not test compiling locally :)

PR #2088, opened 2024-07-04
Advertise HR service

### LINKED ISSUE #895: Expose the HeartRate service UUID in BLE ADV packet

mark9064:
Builds on top of #897
The original problem was that an extra 128bit uuid was too big for the advertising packet

Works with OpenTracks (shows with the advertising filter on), NRF Connect shows both the DFU and HR service

Closes #895

### DIFF
--- a/src/components/ble/DfuService.h
+++ b/src/components/ble/DfuService.h
@@ -77,6 +77,10 @@ namespace Pinetime {
         uint16_t ComputeCrc(uint8_t const* p_data, uint32_t size, uint16_t const* p_crc);
       };
 
+      static constexpr ble_uuid128_t serviceUuid {
+        .u {.type = BLE_UUID_TYPE_128},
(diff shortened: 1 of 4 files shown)


florisre:
Works for me to, I can now connect my PineTime as a HR sensor to my Garmin Edge Explore 2!

Also using PineTime as a motion sensor with Sleep as Android still works, same as before.

mark9064:
Forgot to mention, confirmed working with BLE gym equipment also

JF002 (approved):

PR #2099, opened 2024-08-04
Watchface: add weather option to Casio style watchface

Ghoelian:
Adds the option to turn one of the segments on the Casio watchface into weather.
InfiniSim_2024-08-04_230833
InfiniSim_2024-08-04_230835
InfiniSim_2024-08-04_230942


rnwgnr:
Das PR does not apply anymore due to the recent changes in the weather subsystem - would be cool to have this updated against the current state of `main`.

Ghoelian:
Ah alright, I'll look into getting it updated when I get the time.

Ghoelian:
I've updated my branch, seems to be working on my watch.

tituscmd:
Would it be possible to choose which segment is replaced by weather?
I made a hardcoded version of your idea as well a while back, but I replaced WeekNumber instead. If you could implement a way to choose which segment is replaced, we could essentially merge those ideas into one big PR.
image

Ghoelian:
> Would it be possible to choose which segment is replaced by weather? I made a hardcoded version of your idea as well a while back, but I replaced WeekNumber instead. If you could implement a way to choose which segment is replaced, we could essentially merge those ideas into one big PR. image

Yeah I think I can do that, doesn't sound too difficult.

tituscmd:
For reference, I put the weather icon to align top left at 10, 22 and the temperature to align top left 45, 22
Feel free to just copy these or take them as a reference point for improving the looks of it.

Ghoelian:
InfiniSim_2024-11-25_203125
InfiniSim_2024-11-25_203132

Added these two options.

tituscmd:
Oh I might even like that more than my approach! Good job, will definitely be using this, at least once 1.15 rolls out 😁

tituscmd:
For the segment that replaces the weeknumber, you should center the temperate and weather icon based on the weekday and offset them vertically (in the negative direction of course, since you're going up on the display).

Also, you merged something about notif vibration strength. I don't think you meant to put this into the PR.

Other than that, I'd love to have this in 1.16.0!

rnwgnr:
> Oops sorry yeah didn't mean to commit the vibration stuff to this branch, I'll fix that when I get the time.

I came across this "accidental addition" and i really like it. Would be cool if you'd provide an additional PR for this.

Ghoelian:
Sorry it took so long, but I've finally removed the unrelated commits. I'll get to fixing the alignment later!

Ghoelian:
The merge box thingy is telling me one of the reasons merging is blocked because the branch "must not contain merge commits". Is this a real requirement for getting this merged?

Ghoelian:
@rnwgnr created #2231 for the vibration settings

tituscmd:
> Sorry it took so long, but I've finally removed the unrelated commits. I'll get to fixing the alignment later!

About the alignment, you can check my branch with the extremely long name of "1.15_ANCS_HR_CASIO_DGTL-W" where I think I nailed the alignment perfectly.

Here's a few pics:
image
image
image

Ghoelian:
I've been having some issues getting the simulator running on my setup (trying to run from a dev container), but I'll hopefully fix that and then this PR this weekend.

Ghoelian:
@tituscmd I've fixed the alignment, looks a lot better now!

tituscmd:
Awesome! Glad to have been of help here 😄

PR #2102, opened 2024-08-23
Size optimise NRF SDK build

mark9064:
Currently the NRF SDK is built with -O3 rather than -Os like with everything else

This saves a bit of flash space and performance seems unchanged to me. Thoughts on this change? I could definitely see the performance improvement being justified if it exists, but I don't have a profiler or anything.


JF002:
I can't remember why we specified -O3 specifically for that target. The history is not clear about that, it was changed from -Os back and forth to -O3 multiple times with many merges... :/

I guess we can merge this if you didn't notice any unwanted side-effects :)

mark9064:
I think if we wanted to O3 a specific part it would make the most sense to do that to LVGL as that's the biggest CPU cruncher. But I haven't had time to try that yet

PR #2104, opened 2024-08-23
Clean unused DisplayApp messages

mark9064:
There are some old left over messages from before `RestoreBrightness` became `NotifyDeviceActivity`, and also some other ones that are well before my time. Since screens pull the time directly from the time controller, removing `UpdateDateTime` seems justified. And I can't think of anything that'd want to send `DimScreen` to DisplayApp, so I removed this too (the message is currently unused)


mark9064:
Simplified `RestoreBrightness()` away too

JF002 (approved):

PR #2107, opened 2024-08-23
Refactor doNotGoToSleep to a wakelock counter

mark9064:
Rather than tracking whether sleeping is disabled with a boolean, this moves it to a counter which records the number of requests to keep the device awake. When this counter is zero, sleeping is possible. This prevents one task from enabling sleeping when another task still wants sleeping disabled.

### DIFF
--- a/src/systemtask/WakeLock.cpp
+++ b/src/systemtask/WakeLock.cpp
@@ -1,0 +1,27 @@
+#include "systemtask/WakeLock.h"
+
+using namespace Pinetime::System;
+
+WakeLock::WakeLock(SystemTask& systemTask) : systemTask {systemTask} {
+  lockHeld = false;
+}
+
+WakeLock::~WakeLock() {
+  Release();
+}
+
+void WakeLock::Lock() {
+  if (lockHeld) {
+    return;
+  }
+  systemTask.PushMessage(Messages::DisableSleeping);
+  lockHeld = true;
+}
+
+void WakeLock::Release() {
+  if (!lockHeld) {
+    return;
+  }
+  systemTask.PushMessage(Messages::EnableSleeping);
+  lockHeld = false;
+}
--- a/src/systemtask/WakeLock.h
+++ b/src/systemtask/WakeLock.h
@@ -1,0 +1,19 @@
+#pragma once
+
+#include "systemtask/SystemTask.h"
+
+namespace Pinetime {
+  namespace System {
+    class WakeLock {
+    public:
+      WakeLock(SystemTask& systemTask);
+      ~WakeLock();
+      void Lock();
+      void Release();
+
+    private:
+      bool lockHeld;
+      SystemTask& systemTask;
+    };
+  }
+}
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -478,6 +478,7 @@ list(APPEND SOURCE_FILES
 
         systemtask/SystemTask.cpp
         systemtask/SystemMonitor.cpp
+        systemtask/WakeLock.cpp
         drivers/TwiMaster.cpp
 
         heartratetask/HeartRateTask.cpp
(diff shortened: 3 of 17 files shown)


mark9064:
Couple rebase mistakes breaking build I'll patch them this evening

mark9064:
Rebased to resolve conflicts

src/systemtask/WakeLock.cpp
```diff
@@ -0,0 +1,23 @@
+#include "systemtask/WakeLock.h"
+
+using namespace Pinetime::System;
+
+WakeLock::WakeLock(SystemTask& systemTask) : systemTask {systemTask} {
```
NeroBurner, line 5: should we add a destructor to release the lock if held?

src/systemtask/WakeLock.h
```diff
@@ -0,0 +1,22 @@
...
+      ~WakeLock() {
+        Release();
+      }
+
+    private:
+      bool lockHeld;
+      SystemTask& systemTask;
+    };
+  }
+}
```
NeroBurner, line 22: add newline at end of file to prevent stop symbol of GitHub 
image

```suggestion
}

```

NeroBurner (changes requested):

src/systemtask/WakeLock.cpp
mark9064: Destructor is in the header file, I could move it if that's better. As it's a 1 liner I just left it in there

src/systemtask/WakeLock.h
mark9064: Good spot, I'll fix this after feedback on the other suggestion thread

Should (can?) we change the autoformatter to enforce this? I thought it caught this kind of thing already

JF002 (approved):

src/systemtask/WakeLock.cpp
NeroBurner: I overlooked that the destructor is already here. Bummer, didn't find a bug :grin: 

I always try to put the implementations in the cpp file as leaving the implementations in the `h` or `hpp` files the one definition rule sometimes is violated and breaks compiles. But as this isn't the case here, I have no objections

src/systemtask/WakeLock.h
NeroBurner: If it is possible with `clang-format` then yes, I'd love it if it is enforced. But as `clang-format` is a black magic box for me I don't know how. I'm always happy for projects where somebody did put in the work to set-up `clang-format`

NeroBurner (changes requested):
I've updated the InfiniSim to use `WakeLock.cpp` if it is available, fixing this PR build pipeline
https://github.com/InfiniTimeOrg/InfiniSim/pull/150

Just the little newline and we're good to go from my side

src/systemtask/WakeLock.h
FintasticMan: I've just checked, and the option to insert trailing newlines has been added in clang-format 16. Last time I updated it was 14, so it wasn't included in that. I'll do a upgrade to a newer version soon.

What version do you think is reasonable to update to? Unfortunately, I don't think that clang-format ignores unrecognised options, so it'll be a minimum version for anyone using it.

NeroBurner:
merge conflict after https://github.com/InfiniTimeOrg/InfiniTime/pull/2109 got merged. Please resolve them when doing the little newline

src/systemtask/WakeLock.cpp
mark9064: I don't really know best C++ practices, if I should move it I'm very happy to

NeroBurner: no, leave it. If it ever becomes a problem the compiler will loudly and accurately complain about it

mark9064:
I also moved the destructor as IMO code should be minimally surprising, and now I think about it you don't really expect to find code in the header (except maybe const getters)

mark9064:
I always miss those formatting errors when rebasing...

NeroBurner (approved):
I also mess up formating on rebases :grin: That's why I love the CI job checking it for me ^^

LGTM

src/systemtask/WakeLock.h
NeroBurner: Opened https://github.com/InfiniTimeOrg/InfiniTime/issues/2127 to continue this conversation

Issue #2012, opened 2024-02-11
Watch freezes after some notifications


everypizza1:
### Verification

- [X] I searched for similar bug reports (including closed issues) and found none was relevant.

### What happened?

I received a notification and the touch screen froze for about 1 minute

### Reproduction steps

It doesn't consistently happen, but some things I've noticed:
* Only happens on notifications with emoji or with text in non-English languages.
* Only affects touchscreen, notification still scrolls.

### More details?

Only seems to affect notifications from YouTube for live streams and foreign language notifications from apps like holoplus.foregin

### Version

v1.14

### Companion app

Gadgetbridge 0.78.0

minacode:
InfiniTime does not support all characters. Gadgetbridge might handle this somehow, but I don't know how for sure. 

1. Do you have an example of a notification that causes the bug? 
2. Is there a way to find out what text Gadgetbridge is actually sending to the watch?

everypizza1:
One almost certain way that it's triggered is when someone I have subscribed to does a YouTube livestream. Gadgetbridge seems to be sending the notifications from YouTube with :red_circle: to replace the 🔴 in the notification. I don't have any notifications currently on the watch, but this is one from my notification history that caused it:
[Image]

(I did this through email so it might not work)


________________________________

sbe-arg:
I'm noticing this as well. Specially with emoji notifications from whatsapp and signal.

Also using gadgetbridge.

minacode:
Have you tried other companion apps?

FintasticMan:
Could you try finding a reproducable case by trying it out in GadgetBridge's Debug menu? That allows you to send test notifications.

sbe-arg:
> Could you try finding a reproducable case by trying it out in GadgetBridge's Debug menu? That allows you to send test notifications.

Been trying to reproduce it without any luck a message that might lag when coming from the app not necessarily lags when triggered via debug.

But ill keep checking

everypizza1:
I also can't reproduce it in Gadgetbridge.


________________________________

everypizza1:
I also can't reproduce this in Gadgetbridge.
________________________________

everypizza1:
Just happened again with this notification:
![Screenshot_20240216-070543.jpg](https://github.com/InfiniTimeOrg/InfiniTime/assets/82726593/aa94084b-345b-408d-9666-b786be4456f2)

minacode:
This was with Gadgedbridge? 
[...]

Issue #1790, opened 2023-07-02
Battery rapidly draining


segaretroboy:
### Verification

- [X] I searched for similar bug reports and found none was relevant.

### What happened?

I went to bed with the watch near half charged, woke up with no charge.

### What should happen instead?

I was expecting the watch to make it through the night.

### Reproduction steps

I have no idea how this happened.

### More details?

Gadgetbridge shows the battery dropped rapidly at around 5:30AM

![Screenshot_2023-07-02-16-35-36-50_4c96f65cafca1cafb769ce6b564b6184](https://github.com/InfiniTimeOrg/InfiniTime/assets/20261168/e646bb18-cb2e-4ba9-a7ed-e787ff9b8ff2)
![Screenshot_2023-07-02-16-35-31-66_4c96f65cafca1cafb769ce6b564b6184](https://github.com/InfiniTimeOrg/InfiniTime/assets/20261168/760de062-207e-4f82-a644-f07e36b6d6ea)

### Version

V1.13

### Companion app

GadgetBridge 0.74.0

steinhmor:
Same problem here. With the Update to 1.13 my battery only last half a day compared to 3,5 days before the update.
![Screenshot_20230704-213946](https://github.com/InfiniTimeOrg/InfiniTime/assets/11630531/48b8b4b7-f00a-437e-b98c-5779bf75b42d)

FintasticMan:
I think these are different issues. I think @segaretroboy's issue is somehow caused by a bug causing some component (like the display) to stay on when it should turn off.

I think that @steinhmor's issue is something else, but I'm not sure what. Could you tell me what features you've enabled (such as heart rate, raise to wake etc.) and what your display timeout is set to? I personally got just under a week of usage before the changes in 1.13, and get over 2 weeks with those changes, so it's really surprising to see such short battery life.

segaretroboy:
@FintasticMan I think you are correct on my issue, that was my suspicion but unfortunately I couldn't tell which it was since it happened overnight, the watch was acting normal the night prior. 
When I took it off the screen kept trying to turn on and I had to fiddle with the button, but it didn't come back on before I walked away from it after setting it down so that was probably unrelated. The watch has been acting normal since this as well.
[...]

PR #2109, opened 2024-08-23
SystemTask state refactor

mark9064:
Currently SystemTask is somewhat fragile to duplicate state transitions and unusual state transition message orders. Most notably this can cause it to get stuck in `GoingToSleep` in some scenarios.

This PR introduces GoToSleep() and changes GoToRunning() as immediately executable functions. The guarantee that state transitions happen immediately reduces complexity when it comes to ensuring the device is awake for events such as Chimes 

Fixes part of #1790 #2012 (there may be other issues contributing to these, but this is probably part of it)

### DIFF
--- a/src/components/ble/DfuService.cpp
+++ b/src/components/ble/DfuService.cpp
@@ -124,9 +124,11 @@ int DfuService::WritePacketHandler(uint16_t connectionHandle, os_mbuf* om) {
                    bootloaderSize,
                    applicationSize);
 
-      // wait until SystemTask has finished waking up all devices
-      while (systemTask.IsSleeping()) {
-        vTaskDelay(50); // 50ms
+      // Wait until SystemTask has disabled sleeping
+      // This isn't quite correct, as we don't actually know
+      // if BleFirmwareUpdateStarted has been received yet
+      while (!systemTask.IsSleepDisabled()) {
+        vTaskDelay(pdMS_TO_TICKS(5));
       }
 
       dfuImage.Erase();
--- a/src/components/ble/NimbleController.cpp
+++ b/src/components/ble/NimbleController.cpp
@@ -454,9 +454,15 @@ void NimbleController::PersistBond(struct ble_gap_conn_desc& desc) {
     /* Wakeup Spi and SpiNorFlash before accessing the file system
      * This should be fixed in the FS driver
      */
-    systemTask.PushMessage(Pinetime::System::Messages::GoToRunning);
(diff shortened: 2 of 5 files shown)


src/components/ble/DfuService.cpp
```diff
@@ -124,9 +124,11 @@ int DfuService::WritePacketHandler(uint16_t connectionHandle, os_mbuf* om) {
                    bootloaderSize,
                    applicationSize);
 
-      // wait until SystemTask has finished waking up all devices
-      while (systemTask.IsSleeping()) {
-        vTaskDelay(50); // 50ms
+      // Wait until SystemTask has disabled sleeping
+      // This isn't quite correct, as we don't actually know
```
JF002, line 128: I'm a bit worried about this comment. 
IIRC this "wait" loop is needed to ensure that the peripherals needed for DFU (the SPI flash, mainly) are awake and ready. If this is not the case, the DFU could fail which would prevent users from upgrading their watch.

JF002 (commented):
A documentation of the states of SystemTask (a state diagram for example) would probably be helpful to understand this PR and for future us. This might be not blocking for this PR, but this would be a really nice addition for developers and contributors ;-)

src/components/ble/DfuService.cpp
mark9064: It's equivalent to the current implementation in terms of correctness, actually a little better as it checks specifically whether sleeping is disabled rather than whether the device is awake. The comment exists because it is possible for something else to disable sleeping, pass this check in DfuService, something else to enable sleeping and sleep being entered before SystemTask actually processes the DisableSleeping message from DFUService. It's an extremely unlikely race condition, and there is pretty much nothing we can do about it unless we have some way to check with SystemTask that the DisableSleeping message from DFU has been received (rather than from any other task)

I'm open to some kind of "transactional wakelock" where you can wait for the system task to approve your wakelock request before continuing. But the implementation sounds tricky

JF002: I understand that this is not "worse" than before, so that's fine for me ;-)
Yeah, this "disable sleep" feature is much more complex than I originally expected!

mark9064:
I'm pretty terrible at graphics but I'll give it a go!

JF002 (approved):

NeroBurner:
Updating InfiniSim to support the newly used function `uxQueueMessagesWaiting` https://github.com/InfiniTimeOrg/InfiniSim/pull/151

NeroBurner (approved):
Everything looking sound

PR #2120, opened 2024-09-08
Step counter history added

Hunman:
I changed the MotionController to remember the past N (currently 2) day's step counter history after the midnight reset. And I also changed the Steps screen to display (and update) yesterday's step count in addition to today's.

I didn't feel comfortable changing the return type of `MotionService::NbSteps()` yet without feedback, instead I wrote a new method to return the whole history.

**I've tested this only using InfiniSim, testing with devkits would be welcome!**

In the future I would like to increase the history size from 2 to 7 and transfer the whole history to the companion applications, but I haven't really looked into how bluetooth communication or "API"s work, so feedback would be most appreciated!


minacode:
Nice! You should be able to test this on your sealed watch without problems. As long as you don't verify the firmware, a reboot will be enough to get it back to working. And if you do verify it, a force reboot into the backup firmware still enables you to install another firmware.

src/components/motion/MotionController.cpp
```diff
@@ -35,8 +36,13 @@ namespace {
   }
 }
 
-void MotionController::Update(int16_t x, int16_t y, int16_t z, uint32_t nbSteps) {
-  if (this->nbSteps != nbSteps && service != nullptr) {
+void MotionController::AdvanceDay() {
+  --nbSteps;           // Higher index = further in the past
+  nbSteps[today] = -1; // Ensure that the `this->nbSteps[today] != nbSteps` condition in `Update()` will be FALSE
```
mark9064, line 41: I think it'd be better to assign zero here and do service->OnNewStepCount() inside this function. Reasoning about -1 applied to an unsigned int is complex and so is reasoning about the delta steps calculation

src/components/motion/MotionController.h
```diff
@@ -18,7 +19,17 @@ namespace Pinetime {
         BMA425,
       };
 
-      void Update(int16_t x, int16_t y, int16_t z, uint32_t nbSteps);
+      enum Days {
+        today = 0,
```
mark9064, line 23: Enum members should be capitalised

src/components/motion/MotionController.h
```diff
@@ -18,7 +19,17 @@ namespace Pinetime {
...
+        yesterday = 1,
+      };
+
+      using step_t = uint32_t;
```
mark9064, line 27: I'm really split on this one. On one hand it's nice to have this as a type. On the other, I can't see this type ever being changed (nobody is ever gonna exceed 4 billion steps in a day :P) and it is slower to read and more verbose than `uint32_t`. Also IIRC C++ still will allow using `step_t` and `uint32_t` interchangeably right? (and perhaps any larger/smaller integer type depending on the context). If that's true then we don't get any safety benefits either

src/components/motion/MotionController.h
```diff
@@ -32,15 +43,19 @@ namespace Pinetime {
         return zHistory[0];
       }
 
-      uint32_t NbSteps() const {
+      step_t NbSteps() const {
```
mark9064, line 46: I think it'd be better if NbSteps took a member of Days as an argument  (as you suggest actually)
The step counter history method is still needed though if you plan on exposing this over BLE

src/systemtask/SystemTask.cpp
```diff
@@ -440,6 +440,7 @@ void SystemTask::UpdateMotion() {
 
   if (stepCounterMustBeReset) {
     motionSensor.ResetStepCounter();
+    motionController.AdvanceDay();
```
mark9064, line 443: Can I suggest removing `stepCounterMustBeReset` and the associated message as they aren't needed anymore (I2C is never slept)? This would cause the step count handling to be done immediately when a new day starts. It's a little out of scope for this PR technically but it needs to be refactored anyway and I reckon now is a sensible time to do it

mark9064 (commented):
Cool feature, looking good :) haven't tried it yet on my PT yet but hopefully will soon

src/components/motion/MotionController.cpp
Hunman: Done

src/components/motion/MotionController.h
Hunman: Done

src/components/motion/MotionController.h
Hunman: Done

I've set `Today` as default parameter.

Should I revert most of the changes to `Steps.cpp`, where I'm storing a const reference to `nbSteps`?

src/components/motion/MotionController.h
Hunman: I was torn when I wrote this too :D

Yeah, it is unlikely to change, since the hardware returns a `uint32_t` through `bma423_step_counter_output()` and since it's an alias, `step_t` is completely interchangeable with `uint32_t`, so having this does not make the code any safer or faster.

But on the other hand, it is more verbose, potentially making the code easier to read (at least for me) that this function returns or expects a type in which steps are stored (even though the function/parameter name should be enough), much like how `size_t` is used for most if not all functions where a container size is used.

At least that was my rationale when I wrote this, but I can change it back to `uint32_t` if needed (or move it out of the `MotionController` class).

src/systemtask/SystemTask.cpp
Hunman: It indeed sounds a bit out of scope

I can open a pull request only with this after this one if that's alright

mark9064: No worries, I'll get round to it

src/components/motion/MotionController.h
mark9064: I would personally as I think calling NbSteps() is a bit clearer than using the buffer reference

src/displayapp/screens/Steps.h
```diff
@@ -32,12 +32,15 @@ namespace Pinetime {
...
 
         lv_obj_t* lSteps;
+        lv_obj_t* lStepsYesterday;
         lv_obj_t* stepsArc;
         lv_obj_t* resetBtn;
         lv_obj_t* resetButtonLabel;
         lv_obj_t* tripLabel;
 
-        uint32_t stepsCount;
+        const char* yesterdayStr;
```
mark9064, line 41: Could yesterdayStr be a constant?
If that's fine, you could stick it in an anonymous namespace in the implementation as `constexpr const char *` (I think!) and then it will live in just flash rather than flash and RAM

src/components/motion/MotionController.h
mark9064: I think when it come verboseness the advantage I see is instead of having to reference `Controllers::MotionController::step_t` instead of `uint32_t`. To me that's easier to read, with the operations available on the type being clearer rather than having to look it up and go "oh, it's just an unsigned int, right". But your point about the semantics of the variable being clearer is fair.

I don't know, maybe someone else will have an opinion? I don't strongly oppose it either way

src/displayapp/screens/Steps.h
Hunman: In theory that's how it should work right now too, because I'm not copying the string, just taking the address of a string literal (that's why I don't need to free it either)

Unless I'm misunderstanding something

mark9064: It doesn't matter here too much as it's heap memory rather than stack, but you can avoid storing the pointer at all by effectively inlining the constant

src/components/motion/MotionController.h
minacode: It's just a count, right? Maybe keep it simple. I don't see, what problem would be solved, or is there any critical code, where confusion must not happen (e.g. temperature, where we handle values with different units)?

Hunman: Yeah, it's just a counter, I just didn't feel important to know the real underlying type everywhere in the code.

Even **if** `step_t` makes sense, it definitely shouldn't be defined within `MotionController`, so I'm removing it

mark9064:
Sorry for the seemingly endless review! It's looking really good now, almost there

src/components/motion/MotionController.h
```diff
@@ -18,6 +18,16 @@ namespace Pinetime {
         BMA425,
       };
 
+      enum class Days : uint8_t {
+        Today = 0,
+        Yesterday,
+        Last,
+      };
+
+      static constexpr size_t stepHistorySize = static_cast<std::underlying_type_t<Days>>(Days::Last); // Store this many day's step counter
```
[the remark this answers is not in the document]
Hunman: Yeah, theoretically out of bounds. In practice the CircularBuffer handles it internally by doing modulo (`%`) operation on the size, so `Days::Last` would be just index 0 again.

While it **is** nice to refer to the days with `Today` and `Yesterday`, maybe having to name 5 more enum values is not that nice.

One option is that the `Days` enum can be removed altogether. Or alternatively it can remain with only `Today` and `Yesterday` values, and `NbSteps()` function could get an overload that expects a `uint8_t`.

Hunman:
It's alright, I'm actually enjoying this opportunity to "nerd out" with people on a project 😄

src/components/motion/MotionController.h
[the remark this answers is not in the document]
mark9064: That second option sounds good to me, that way we have a nice human readable enum while also allowing specifying a number of days ago for more flexible usage

tituscmd:
Very nice! Do you have any InfiniSim screenshots or pictures of it on a real watch?

Hunman:
Yes, I have some pictures both in InfiniSim and on the real watch

Nothing fancy, so far. If there's demand, I'm thinking about storing the past 7 or so days and displaying it in a table (like the tasks in the Settings > About menu's 4th screen).

Eventually I'd also like to tell companion applications the step history count, but I'll need to look more into how BLE communication works for that.

image
SAM_1147

src/components/motion/MotionController.h
```diff
@@ -67,9 +76,13 @@ namespace Pinetime {
       }
 
     private:
-      uint32_t nbSteps = 0;
+      Utility::CircularBuffer<uint32_t, stepHistorySize> nbSteps = {0};
       uint32_t currentTripSteps = 0;
 
+      uint32_t& NbStepsRef(Days day = Days::Today) {
```
mark9064: Having thought about this for a bit, I think this API of returning a mutable reference is a bit unexpected.

I think I'd prefer something more explicit along the lines of `SetSteps(Days day)`

WhyNotHugo:
Been testing this a while, work fine so far.

The halo which fills up as one approaches the goal is an unusual pick, especially for a square device.

The menu to check the amount of steps taken and configure the step goals seem very disconnected from one another—but I think that can be refined later and doesn't need to be a blocker for this patch.

mark9064 (commented):
Sorry it's taken ages for me to look at this

Looking really good :)

mark9064:
> Been testing this a while, work fine so far.
> 
> The halo which fills up as one approaches the goal is an unusual pick, especially for a square device.
> 
> The menu to check the amount of steps taken and configure the step goals seem very disconnected from one another—but I think that can be refined later and doesn't need to be a blocker for this patch.

Circular progress dial looks nice to me. Do you think a square dial would be better then?

Agreed on the app configuration. This is a broader problem in InfiniTime where app settings and apps themselves live in different places, and worse the settings pages are not modular while the apps are. More in #2314 (and probably other tracking issues I've forgotten)

This patch only concerns step history though, so I think your points are probably better addressed elsewhere :) Please do feel free to open new issues to track them (if there isn't a ticket already)

mark9064:
If you rebase this onto main and resolve the conflicts, I can trigger CI

src/displayapp/screens/Steps.h
```diff
@@ -32,11 +32,14 @@ namespace Pinetime {
         uint32_t currentTripSteps = 0;
 
         lv_obj_t* lSteps;
+        lv_obj_t* lStepsYesterday;
         lv_obj_t* stepsArc;
         lv_obj_t* resetBtn;
         lv_obj_t* resetButtonLabel;
         lv_obj_t* tripLabel;
 
+        static constexpr const char* yesterdayStr = "Yest: %5lu";
```
mark9064: This can move into the .cpp (use an anonymous namespace)

src/components/motion/MotionController.h
```diff
@@ -66,9 +75,17 @@ namespace Pinetime {
...
     private:
-      uint32_t nbSteps = 0;
+      Utility::CircularBuffer<uint32_t, stepHistorySize> nbSteps = {0};
       uint32_t currentTripSteps = 0;
 
+      void SetSteps(Days day, uint32_t steps) {
+        nbSteps[static_cast<std::underlying_type_t<Days>>(day)] = steps;
+      }
+
+      uint32_t GetSteps(Days day) const {
```
mark9064: Can this method be removed since `NbSteps` is the same?

Or maybe we should rename `NbSteps` to `GetSteps` instead?

Hunman: Yeah, those two really were the same, I removed `GetSteps`

Renaming `NbSteps` would have broken pretty much every single pull requests that involve watchfaces

src/displayapp/screens/Steps.h
Hunman: It has been banished to the cpp file into an anonymous namespace

src/displayapp/screens/Steps.cpp
```diff
@@ -33,13 +39,19 @@ Steps::Steps(Controllers::MotionController& motionController, Controllers::Setti
   lSteps = lv_label_create(lv_scr_act(), nullptr);
   lv_obj_set_style_local_text_color(lSteps, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_LIME);
   lv_obj_set_style_local_text_font(lSteps, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, &jetbrains_mono_42);
-  lv_label_set_text_fmt(lSteps, "%li", stepsCount);
+  lv_label_set_text_fmt(lSteps, "%lu", stepsCount);
```
NeroBurner, line 42: nice one finding a wrong formatter. it has always been unsigned. now it is right!

PR #2123, opened 2024-09-17
SettingWakeUp: use Checkboxlist.h

NeroBurner:
Update `CheckboxList.h` to optionally be able to get checkboxes (instead of the default of radio-buttons) and be able to have multi-selection through a new callback.

Use these updated capabilities of `CheckboxList.h` to replace the nearly identical code of `SettingWakeUp.h`


minacode:
I like the idea and the implementation via the `IsChecked` function.

I am only reading the code right now, but think we can simplify the interface. If I understand it correctly, the function is provided for checkboxes and the original value is provided for radio buttons? Can we infer the style from that and omit the boolean flag? And maybe make the original value optional, if it is not always needed?

minacode (commented):
See my comment above.

PR #2125, opened 2024-09-18
Remove unused pointer to DisplayApp member variables

NeroBurner:
In the screens that use `DisplayApp *app` and pass it to a child item, or use the reference just in the constructor. Afterwards the `app` member is not used. So remove it from the private member variables.

~~`SettingDisplay.h` actually never uses `DisplayApp *app` and the parameter could be removed from the constructor. I'd be happy about feedback whether or not to remove it.~~
Completely remove `app` parameter from `SettingDisplay` constructor as it is unused.


NeroBurner:
build error because of unused variable, removing `app` from constructor

```
 /__w/InfiniTime/InfiniTime/src/displayapp/screens/settings/SettingDisplay.cpp:27:68: error: unused parameter 'app' [-Werror=unused-parameter]
   27 | SettingDisplay::SettingDisplay(Pinetime::Applications::DisplayApp* app, Pinetime::Controllers::Settings& settingsController)
      |                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~
cc1plus: all warnings being treated as errors
```

mark9064 (approved):

FintasticMan (approved):
LGTM!

PR #2130, opened 2024-09-26
MusicService: add missing includes for TickType_t and xTaskGetTickCount

NeroBurner:
Add includes for the directly used data type `TickType_t` in the header and the function `xTaskGetTickCount` from FreeRTOS's `task.h`

I found out while slimming down InfiniSim and trying to use `MusicService.h` from InfiniTime directly

### DIFF
--- a/src/components/ble/MusicService.cpp
+++ b/src/components/ble/MusicService.cpp
@@ -18,6 +18,7 @@
 #include "components/ble/MusicService.h"
 #include "components/ble/NimbleController.h"
 #include <cstring>
+#include <task.h>
 
 namespace {
   // 0000yyxx-78fc-48fe-8e23-433b3a1942d0
--- a/src/components/ble/MusicService.cpp
+++ b/src/components/ble/MusicService.cpp
@@ -18,6 +18,7 @@
 #include "components/ble/MusicService.h"
 #include "components/ble/NimbleController.h"
 #include <cstring>
+#include <FreeRTOS.h>
 #include <task.h>
 
 namespace {
--- a/src/components/ble/MusicService.h
+++ b/src/components/ble/MusicService.h
@@ -25,7 +25,7 @@
 #include <host/ble_uuid.h>
 #undef max
 #undef min
-#include "portmacro_cmsis.h"
+#include <FreeRTOS.h>
 
 namespace Pinetime {
   namespace Controllers {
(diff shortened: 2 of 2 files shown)


mark9064 (approved):

src/components/ble/MusicService.h
```diff
@@ -25,6 +25,7 @@
 #include <host/ble_uuid.h>
 #undef max
 #undef min
+#include "portmacro_cmsis.h"
```
FintasticMan, line 28: `portmacro_cmsis.h` is a header internal to FreeRTOS. It would be better to just include `FreeRTOS.h`. Also, we use `<>` for external libraries like this rather than `""`

src/components/ble/MusicService.cpp
```diff
@@ -18,6 +18,7 @@
 #include "components/ble/MusicService.h"
 #include "components/ble/NimbleController.h"
 #include <cstring>
+#include <task.h>
```
FintasticMan, line 21: You need to include `FreeRTOS.h` before `task.h`.

FintasticMan (changes requested):

src/components/ble/MusicService.cpp
mark9064: It's being provided by nimblecontroller.h (at least by fs.h), but I suppose it's better to avoid the indirect dependency?

FintasticMan: Yeah. I'm of the opinion that all of our headers should work regardless of how they're included in cpp files.

mark9064: I think that does make sense, that way it's tolerant to things being moved around later. This probably does mean that a lot of headers are missing across the project though

FintasticMan: I don't think we need to fix it for existing code, but going forward we should aim for it on a best-effort basis.

NeroBurner: added FreeRTOS.h include before tasks.h now

src/components/ble/MusicService.h
NeroBurner: thanks for the review. Indeed `<>` for external libraries. Review are IMPORTANT. Thanks for the time and suggestion to make it right

FintasticMan (approved):
Always nice to fix issues like these!

PR #1782, opened 2023-06-26
Update Date&Time settings label for "Date & Time"

ajira86:
Proposal to update Date&Time label for issue #1746

image

PR #2133, opened 2024-10-02
datetime: Set the default year to the year during compile

FintasticMan:
Supersedes #2067.

### DIFF
--- a/src/components/datetime/DateTimeController.cpp
+++ b/src/components/datetime/DateTimeController.cpp
@@ -7,16 +7,29 @@
 using namespace Pinetime::Controllers;
 
 namespace {
-  char const* DaysStringShort[] = {"--", "MON", "TUE", "WED", "THU", "FRI", "SAT", "SUN"};
-  char const* DaysStringShortLow[] = {"--", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat", "Sun"};
-  char const* MonthsString[] = {"--", "JAN", "FEB", "MAR", "APR", "MAY", "JUN", "JUL", "AUG", "SEP", "OCT", "NOV", "DEC"};
-  char const* MonthsStringLow[] = {"--", "Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"};
+  constexpr const char* const DaysStringShort[] = {"--", "MON", "TUE", "WED", "THU", "FRI", "SAT", "SUN"};
+  constexpr const char* const DaysStringShortLow[] = {"--", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat", "Sun"};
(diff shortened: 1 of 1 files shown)


tituscmd:
While were at it changing stuff related to Date & Time, we could also consider this PR:

https://github.com/InfiniTimeOrg/InfiniTime/issues/1782

NeroBurner (approved):
Code looks good. When tested on device good to go

src/components/datetime/DateTimeController.cpp
```diff
@@ -7,16 +7,28 @@
...
+    return result;
+  }
 }
 
 DateTime::DateTime(Controllers::Settings& settingsController) : settingsController {settingsController} {
   mutex = xSemaphoreCreateMutex();
   ASSERT(mutex != nullptr);
   xSemaphoreGive(mutex);
+
+  SetTime(compileTimeAtoi(&__DATE__[7]), 0, 0, 0, 0, 0);
```
mark9064, line 31: Maybe add a comment explaining char 7->end encodes the year? Not a huge deal though

mark9064 (approved):
LGTM. Haven't tested on HW

PS: Took me about 5 minutes to understand the const correctness of the arrays, but I think you've got the type right. Well done 😅

FintasticMan:
Yeah, const-correctness in C++ isn't very obvious haha. I'm not by my dev kit so I can't debug this, but it really doesn't seem to like calling `SetTime` in the constructor, as doing so seems to result in a non-booting firmware.

mark9064:
Ah, that would be as system task doesn't exist yet so it hits a null ptr deref when it tries to push OnNewTime

FintasticMan:
Good thinking! Adding a NULL check should indeed fix it.

FintasticMan:
It boots now, but I haven't been able to get the time to not be read from `.noinit` :sweat_smile:, so I can't test if it actually works. I'm sure if the battery runs out it'll reset the time (isn't it annoying having such good battery life!)

NeroBurner:
Is it possible to make a Firmware that always reads the noinit data I instead of the filesystem?

mark9064:
image
So I tried @pipe01's InfiniEmu and it looks like we're off by a little bit. I'd guess the month and day are meant to be ones

FintasticMan:
That makes a lot of sense haha.

FintasticMan:
My watch just ran out of battery, so I can confirm: this does work on actual hardware!

src/components/datetime/DateTimeController.cpp
```diff
@@ -46,7 +59,9 @@ void DateTime::SetTime(uint16_t year, uint8_t month, uint8_t day, uint8_t hour,
   UpdateTime(previousSystickCounter, true);
   xSemaphoreGive(mutex);
 
-  systemTask->PushMessage(System::Messages::OnNewTime);
+  if (systemTask) {
```
mark9064, line 62: The other checks in the file do `systemTask != nullptr`, maybe worth being consistent? Very minor though

mark9064 (approved):

JF002 (approved):

PR #2135, opened 2024-10-03
Improved the Terminal Watchfaces UI

JustScott:
These changes were originally proposed in PR #2001 along with the addition of weather data on it's own line, but it was suggested they be seperated. This PR is just the UI improvements without the weather, and #2001 will add the weather.

+ Reorder code to match the widgets order in the UI.

+ Use InfintimeTheme Colors instead of hardcoded hex values

+ Added a new InfinitimeTheme color: gray, using it to turn certain values gray when they contain no data

+ Implement @vkareh's [variable battery icon](https://github.com/InfiniTimeOrg/InfiniTime/pull/1964) color to the battery percentage text.

+ Replaced the 'You have mail.' notification message with the message '[1]+ Notify' to better fit the terminal lore.

InfiniSim_2024-10-03_155938
InfiniSim_2024-10-03_155950


src/displayapp/screens/WatchFaceTerminal.cpp
```diff
@@ -120,32 +101,67 @@ void WatchFaceTerminal::Refresh() {
...
+
+  powerPresent = batteryController.IsPowerPresent();
+  batteryPercentRemaining = batteryController.PercentRemaining();
+  if (batteryPercentRemaining.IsUpdated() || powerPresent.IsUpdated()) {
+    // HSV color model has red at 0° and green at 120°.
+    // We lock satuation and brightness at 100% and traverse the cilinder
+    // between red and green, thus avoiding the darker RGB on medium battery
+    // charges and giving us a much nicer color range.
+    uint8_t hue = batteryPercentRemaining.Get() * 120 / 100;
+    lv_obj_set_style_local_text_color(batteryValue, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, lv_color_hsv_to_rgb(hue, 100, 100));
```
mark9064, line 126: I think colours here are nice, but this shows yellow at 50% right? I think it would make more sense to follow a curve closer to the current battery icon (yellow at 15, red at 5). Not sure what function would represent this well but coming up with something in Desmos/similar would be easy

Also if this is to be used in other places as well, it should probably be moved out so other screens can use it. I don't know if that's the goal here though

src/displayapp/screens/WatchFaceTerminal.cpp
```diff
@@ -120,32 +101,67 @@ void WatchFaceTerminal::Refresh() {
...
+  if (stepCount.IsUpdated()) {
+    lv_label_set_text_fmt(stepValue, "#ffffff [STEP]# %lu steps", stepCount.Get());
+  }
+
   heartbeat = heartRateController.HeartRate();
   heartbeatRunning = heartRateController.State() != Controllers::HeartRateController::States::Stopped;
   if (heartbeat.IsUpdated() || heartbeatRunning.IsUpdated()) {
     if (heartbeatRunning.Get()) {
-      lv_label_set_text_fmt(heartbeatValue, "[L_HR]#ee3311 %d bpm#", heartbeat.Get());
+
```
mark9064, line 142: Newline

mark9064 (commented):
I don't use terminal so no strong opinion on visual style but looks nice to me

Maybe we should use this opportunity to fix the badly named fields (they should be camelCase)? But that could also be another PR, definitely not a blocker

src/displayapp/screens/WatchFaceTerminal.cpp
mark9064: I see this is from #1964 , so I think it definitely makes sense to move the colour calculation algorithm to somewhere where other apps can use it

pbone64:
Love the new steps colour and notification indicator :) though, if we're leaning into the terminal watch face "lore", shouldn't the prompts stay white? It also looks a little busy to have two different colours at the starts of lines, especially when `[1]+ Notify` is visible.

AlbeyAmakiir:
I think the non-info lines being dimmer might be nice, even if it doesn't quite match lore. Tho tbh I've never used full white text in a terminal anyway.

JustScott:
Moved the objects (label, rows, idk what to call them) into a container to avoid having to position each of them individually. This also allows easily adding and removing objects in the future without having to modify multiple position values. As far as I can tell there isn't a visual difference from before:

without_container
moved_to_container

src/displayapp/screens/WatchFaceTerminal.cpp
kieranc: Hi @JustScott any chance you can fix up this formatting so we can get this merged? :-)

mark9064:
Adding the colour code here LGTM
I think we should integrate this while we're here: https://github.com/InfiniTimeOrg/InfiniTime/pull/1964#issuecomment-2923352824 should be a quick formula tweak (using floating point is fine IMO we have a very capable FPU)

Rest looks good to me but only have had a quick glance, will review properly ASAP

src/displayapp/screens/WatchFaceTerminal.cpp
```diff
@@ -27,40 +27,43 @@ WatchFaceTerminal::WatchFaceTerminal(Controllers::DateTime& dateTimeController,
...
+  lv_obj_set_style_local_bg_opa(container, LV_CONT_PART_MAIN, LV_STATE_DEFAULT, LV_OPA_TRANSP);
 
-  label_date = lv_label_create(lv_scr_act(), nullptr);
-  lv_label_set_recolor(label_date, true);
-  lv_obj_align(label_date, lv_scr_act(), LV_ALIGN_IN_LEFT_MID, 0, -40);
+  notificationIcon = lv_label_create(container, nullptr);
 
-  label_prompt_1 = lv_label_create(lv_scr_act(), nullptr);
-  lv_obj_align(label_prompt_1, lv_scr_act(), LV_ALIGN_IN_LEFT_MID, 0, -80);
+  label_prompt_1 = lv_label_create(container, nullptr);
```
mark9064: All these members should be `camelCase` but not a blocker

src/displayapp/screens/BatteryIcon.cpp
```diff
@@ -51,3 +52,12 @@ const char* BatteryIcon::GetPlugIcon(bool isCharging) {
   else
     return "";
 }
+
+lv_color_t BatteryIcon::ColorFromPercentage(int batteryPercent) {
+  // HSV color model has red at 0° and green at 120°.
+  // We lock saturation and brightness at 100% and traverse the cylinder
+  // between red and green, thus avoiding the darker RGB on medium battery
+  // charges and giving us a much nicer color range.
+  const uint8_t hue = batteryPercent * 120 / 100;
```
mark9064: I'd prefer this to follow https://github.com/InfiniTimeOrg/InfiniTime/pull/1964#issuecomment-2923352824 (or a similar curve) so it doesn't start going yellow until the battery is low

JF002: Is this what it should look like ?

```
const float normalizedPercent = batteryPercent / 100.0f;
const uint8_t hue = (-expf(-4 * normalizedPercent) + 1.0f) * 120.0f / 100.0f;
return lv_color_hsv_to_rgb(hue, 100, 100);
```

src/displayapp/screens/WatchFaceTerminal.cpp
JF002: Done :-)

src/displayapp/screens/BatteryIcon.cpp
mark9064: That sounds good to me :)

I think just `const uint8_t hue = (-expf(-4 * normalizedPercent) + 1.0f) * 120.0f;` would be correct though, as the formula should output 0-1 in this range
Can also use `std::exp` to be more modern if you fancy

JF002: Yes, you're right! (checked with compiler explorer :)).

mark9064 (approved):
LGTM

Needs a hardware test (InfiniEmu aborts, maybe floating point simulation issues? not sure)
Commits have a minor typo, should be `WatchFaceTerminal` rather than `WhatFaceTerminal` 
Maybe worth a comment on `ColorFromPercentage` that the maths makes sure it is yellow around 15% and red around 5%

mark9064:
Same typo in #2204 actually ^^

mark9064:
Rebased onto main (so now has weather)
Applied above suggestion and fixed typo
Needs test on HW

@JF002 all looks good to me, merge when you're ready

mark9064:
Tested on hardware

Re-centred the the items to handle the new weather line
Fixed weather formatting (rebase mistake earlier I think)

Looks great IMO and ready to go :)

PR #2138, opened 2024-10-06
Remove unused submodule QCBOR

NeroBurner:
The submodule isn't used anymore. Remove the submodule reference completely.


mark9064 (approved):
Sounds good

mcuboot imgtool still relies on a python cbor library - not sure if this is used in practice either (it doesn't work without though and would need some code changes if it can be dropped)

FintasticMan (approved):
Looks good! Since switching from the old weather service to the simple weather service we don't need this.

Issue #2127, opened 2024-09-21
Improve clang-format to check newlines at end of file


NeroBurner:
### Verification

- [X] I searched for similar issues (including closed issues) and found none was relevant.

### Introduce the issue

I regularly criticize missing newline at end of file resulting in stop symbol of GitHub 
![stop symbol of GitHub at end of files](https://github.com/user-attachments/assets/0195b16f-7b5b-458f-92b4-5b76d82ed75a)

@mark9064  suggested to let `clang-format` do this check https://github.com/InfiniTimeOrg/InfiniTime/pull/2107#discussion_r1765677368

@FintasticMan adds info on which `clang-format` version we could use https://github.com/InfiniTimeOrg/InfiniTime/pull/2107#discussion_r1769634701

> I've just checked, and the option to insert trailing newlines has been added in clang-format 16. Last time I updated it was 14, so it wasn't included in that. I'll do a upgrade to a newer version soon.
> 
> What version do you think is reasonable to update to? Unfortunately, I don't think that clang-format ignores unrecognised options, so it'll be a minimum version for anyone using it.

### Preferred solution

`clang-format` checks the missing newline at end of file in CI

### Version

main

FintasticMan:
The latest version of clang-format is 19, the latest version in the Ubuntu 24.04 repositories (which we use for CI) is 18, and the latest version in the Debian stable repos is 16 (just for a more conservative comparison). I'd be quite happy to go straight for 18, but I really don't know which versions other developers are using.

If anyone is using a version of clang-format below 18, could you let me know here, along with which OS and version? Thanks!

NeroBurner:
Ubuntu 22.04 has clang-format-14

https://packages.ubuntu.com/jammy/clang-format

Clang-format-16 is available since Ubuntu 23.10 
https://packages.ubuntu.com/search?keywords=clang-format-16

Just a data point, although I'm on arch so I have newer packages available

mark9064:
In other words, the option we want is supported by Debian stable, Ubuntu LTS and anything newer right? So in that case I think we're safe to enable it

Do we even specify a minimum clang format version anywhere? If so I think it shouldn't be bumped past Debian stable (16) unless we want some newer feature

FintasticMan:
We do specify a minimum version in the clang-format commit hook, so I'll just bump to 16 for now.

PR #2139, opened 2024-10-09
clang-format: Update to 16

FintasticMan:
As discussed in #2127.
I've also updated the clang-format options to be compatible, and sorted the options.


FintasticMan:
Sorry for the force push spam :sweat_smile:. GitHub doesn't have an easy way to test CI locally, so I just did it in this PR. The failing clang-tidy test is because the infinitime-build container still uses Ubuntu 22.04. I think the reason we use the docker container in that check is to generate the `compile_commands.json` file that clang-tidy uses. Should I also update the docker container in this PR?

mark9064:
Yeah may as well do it all in one go I think, it makes sense to do the upgrade atomically

FintasticMan:
I've actually just thought that we don't necessarily need to update clang-tidy at the same time as clang-format, they don't interact at all. I might just revert the clang-tidy version bump and keep this PR to only updating clang-format.

tests/test-tidy.sh
```diff
@@ -17,7 +17,7 @@ do
   src/libs/*|src/FreeRTOS/*) continue ;;
   *.cpp|*.h)
     echo "::group::$file"
-    clang-tidy-14 -p build "$file" || true
+    clang-tidy-16 -p build "$file" || true
```
mark9064, line 20: If clang-tidy isn't being updated, this should be undone right?

mark9064 (commented):
LGTM. It makes sense to keep the clang tooling at the same version to me though - tools get installed at the same time so the versions will be the same on real systems. In what scenario would you have clang-format from clang 16 installed but clang-tidy from 14 - it seems like we'd be using a very strange configuration here

tests/test-tidy.sh
FintasticMan: Oh yeah, oops

FintasticMan:
I agree that clang-tidy 14 with clang-format 16 is a bit of a strange setup, but I think that it's fine for the interim while I get clang-tidy 16 working.

mark9064:
Oh, what's up with clang-tidy 16?

mark9064:
Now I look at this, isn't clang-tidy just completely broken

mark9064:
A working clang-tidy invocation: `clang-tidy -p build $($ARM_NONE_EABI_TOOLCHAIN_PATH/bin/arm-none-eabi-g++ -E -Wp,-v -xc++ /dev/null 1>/dev/null 2>&1 | grep ' /' | sed 's/ \//\//' | xargs realpath | sed 's/^/-extra-arg=-isystem/' | tr '\n' ' ') src/main.cpp`

mark9064:
Should probably link an explanation for that: https://github.com/llvm/llvm-project/issues/53468

Issue #303, opened 2021-04-20
Stopwatch resets when leaving the app.


PizzaLovingNerd:
Hello, I was using the stopwatch, and I noticed that whenever I closed the app, the stop watch would reset back to 0. Here is a video demonstrating the issue: https://www.youtube.com/watch?v=AcRQ5m9-PVQ

The expected behavior is having the stop watch still be running or atleast paused when the app is relaunched.

qbinary:
I think the problem is that the event which initializes the app is also called to reset the clock application after pressing the stop button. So that could be fixed by adding an additional method which gets invoked by the initialize event if the state of the stopwatch is not set. Otherwise the old state is kept and the app just continues running with the current state.

Soundtoxin:
I got hit by this badly earlier. I started downloading the new LineageOS Nightly OTA to flash and then started a stopwatch. It notified me about the download being done a few seconds after the stopwatch had started which just cleared it.

Itai-Nelken:
#783 implements persistence.

Eve1374:
Hello, 
I noticed that when receiving a notification while stopwatch is running, it also resets the stopwatch. I think issues are related and so I post it there instead of making a new issue. Tested on commit 00769625881e318e6fe29e926c7bd944c5f049bd

mark9064:
You're right about it being related, #2141 should resolve this and is looking to be in really good shape to me

PR #2141, opened 2024-10-20
StopWatch: add persistence

codingjourney:
This PR solves Issue #303 and is based on previous work by desttinghim (PR #783 ) and pptime02 (PR #1410 ). Well, it's actually just a re-base of #1410  onto the current main branch. I did my best not to disrupt functionality implemented on main in the past 2 years. Any feedback is appreciated.


src/displayapp/screens/StopWatch.cpp
```diff
@@ -4,37 +4,40 @@
 #include "displayapp/InfiniTimeTheme.h"
 
 using namespace Pinetime::Applications::Screens;
+using namespace Pinetime::Controllers;
 
 namespace {
-  TimeSeparated_t convertTicksToTimeSegments(const TickType_t timeElapsed) {
+  TimeSeparated convertTicksToTimeSegments(const TickType_t timeElapsed) {
```
mark9064, line 10: Since you're refactoring already, could you fix the names of these functions here to be PascalCase as per the style guidelines?

src/components/stopwatch/StopWatchController.h
```diff
@@ -0,0 +1,66 @@
+#pragma once
+
+#include <FreeRTOS.h>
+#include <timers.h>
+
+#define LAP_CAPACITY 2
```
mark9064, line 6: This should be a constexpr constant

src/components/stopwatch/StopWatchController.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+LapInfo* StopWatchController::LastLap(int lap) {
+  if (lap >= LAP_CAPACITY || lap > lapCount || lapCount == 0) {
+    // Return "empty" LapInfo_t
+    return &emptyLapInfo;
+  }
+  // Index backwards
+  int index = Wrap(lapHead - lap);
+  return &laps[index];
+}
```
mark9064, line 80: Could [CircularBuffer](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/src/utility/CircularBuffer.h) be used here? Maybe I'm misreading but it seems like this is implementing a similar thing

src/components/stopwatch/StopWatchController.h
```diff
@@ -0,0 +1,66 @@
...
+      /// Returns actual count of stored laps
+      int GetLapNum();
+
+      /// Returns lapCount
+      int GetLapCount();
+
+      /// Indexes into lap history, with 0 being the latest lap.
+      /// If the lap is unavailable, count and time will be 0. If there is a
+      /// real value, count should be above 0
+      LapInfo* LastLap(int lap = 0);
```
mark9064, line 46: Would it make more sense to return `std::optional<LapInfo>` rather than having to serve dummy data if there is nothing

mark9064 (commented):
Haven't tested on hardware, but this is looking really nice overall. Thanks for pulling together the previous PRs :)

src/components/stopwatch/StopWatchController.cpp
```diff
@@ -0,0 +1,101 @@
+#include "components/stopwatch/StopWatchController.h"
+
+using namespace Pinetime::Controllers;
+
+namespace {
+  TickType_t CalculateDelta(const TickType_t startTime, const TickType_t currentTime) {
```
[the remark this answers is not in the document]
codingjourney: Yes, TickType_t is indeed unsigned so the [modulo 2^n semantics](https://en.cppreference.com/w/cpp/language/operator_arithmetic#Overflows) apply. Method removed.

src/displayapp/screens/StopWatch.cpp
codingjourney: Done.

src/components/stopwatch/StopWatchController.h
codingjourney: Done, hopefully I placed it correctly (I'm not a C++ native).

src/components/stopwatch/StopWatchController.cpp
codingjourney: It could certainly be used but I would leave that change for a PR that deals with lap times specifically. Lap time handling in StopWatch is quite basic right now, I suppose it will see more work in the future.

src/components/stopwatch/StopWatchController.h
codingjourney: Done.

src/components/stopwatch/StopWatchController.cpp
mark9064: Hmm OK, my thought process is that if we're implementing a brand new controller we may as well do it right rather than refactor it later? I haven't seen any push for change with lap handling, so I doubt this will see future work. If you don't fancy doing it yourself I wouldn't mind having a go at it, perhaps my intuition about it being suitable here is wrong anyway

src/components/stopwatch/StopWatchController.h
```diff
@@ -0,0 +1,63 @@
...
+      int GetLapCount();
+
+      /// Indexes into lap history, with 0 being the latest lap.
+      std::optional<LapInfo> LastLap(int lap = 0);
+
+      bool IsRunning();
+      bool IsCleared();
+      bool IsPaused();
+
+    // private:
```
mark9064, line 48: !

src/components/stopwatch/StopWatchController.h
```diff
@@ -0,0 +1,63 @@
...
+  namespace Controllers {
+
+    enum class StopWatchStates { Cleared, Running, Paused };
+
+    struct LapInfo {
+      int count = 0;       // Used to label the lap
+      TickType_t time = 0; // Delta time from beginning of stopwatch
+    };
+
+    constexpr int lapCapacity = 2;
```
mark9064, line 20: I think this needs to go inside StopWatchController as it should be namespaced to it? You could have it as a private member with static constexpr

src/components/stopwatch/StopWatchController.cpp
```diff
@@ -51,32 +37,25 @@ void StopWatchController::PushLap() {
...
+  lapHead = lapCount % lapCapacity;
 }
 
 int StopWatchController::GetLapCount() {
   return lapCount;
 }
 
 int Wrap(int index) {
-  return ((index % LAP_CAPACITY) + LAP_CAPACITY) % LAP_CAPACITY;
+  return ((index % lapCapacity) + lapCapacity) % lapCapacity;
```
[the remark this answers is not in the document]
codingjourney: Not when index < 0 (which was legitimate at some point). The method is now gone as laps are stored in a CircularBuffer.

src/components/stopwatch/StopWatchController.cpp
codingjourney: I looked at CircularBuffer, it wasn't that much work and it made the code much simpler. Thanks!

src/components/stopwatch/StopWatchController.h
```diff
@@ -0,0 +1,64 @@
...
+namespace Pinetime {
+  namespace System {
+    class SystemTask;
+  }
+  namespace Controllers {
+
+    enum class StopWatchStates { Cleared, Running, Paused };
+
+    struct LapInfo {
+      int count = 0;       // Used to label the lap
```
mark9064, line 17: I think the default values can be removed as they're always set in `Clear()`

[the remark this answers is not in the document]
codingjourney: Well, if we're talking naming, the entire concept of "laps" is a bit iffy here. Let's say I'm timing a hamster racing around a table-top track and it runs each lap in about 5 seconds. I'd expect to see something like
```
#1   00:05.02
#2   00:04.97
#3   00:04.99
```
but StopWatch shows me
```
#1   00:05.02
#2   00:09.99
#3   00:14.98
```
Rather than lap times, these are _split times_ (known from downhill skiing, for example). It's actually smarter to store split times: while L(n) = S(n) - S(n-1) hence O(1), S(n) = sum(L(1)..L(n)) hence O(n).

What's the proper naming, then? We could go abstract:
```
struct Measurement {
  int index = 0;
  TickType_t value = 0;
}
...
Utility::CircularBuffer<Measurement, histSize> history;
```
This would also be somewhat consistent with `MotionController`, the other user of `CircularBuffer`. Related fields and methods in StopWatch would have to adopt this terminology as well, of course.

I understand naming is important and hard to get right, so if you have better ideas I'm all ears. I know I have a bias towards abstract names but this is honestly the best I can come up with at the moment.

src/components/stopwatch/StopWatchController.h
```diff
@@ -0,0 +1,64 @@
...
+      TickType_t startTime = 0;
+      // How much time was elapsed before current duration
+      TickType_t timeElapsedPreviously = 0;
+
+      // Number of stored laps
+      static constexpr int lapCapacity = 2;
+      // Lap storage
+      Utility::CircularBuffer<LapInfo, lapCapacity> laps;
+      // Total lap count; may exceed lapCapacity
+      int lapCount = 0;
```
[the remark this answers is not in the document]
codingjourney: See above.

src/components/stopwatch/StopWatchController.h
codingjourney: Initializers in declarations provide additional information in a convenient place, so I'm usually OK with them as long as they're not too expensive. Is there a particular reason in C++ or in InfiniTime to prefer imperative intialization? I know we need a `Clear()` method anyway, so are we going for code size savings here?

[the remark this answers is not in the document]
codingjourney: I tried it out and abstract terminology is not a good fit for the public methods. How about this?
```
struct LapInfo {
  int number; // Used for labeling
  TickType_t timeSinceStart; // Excluding pauses
}
...
void AddLapToHistory();
int getMaxLapNumber();
std::optional<LapInfo> getLapFromHistory(int index); // 0 is most recent
...
static constexpr int histSize = 2;
Utility::CircularBuffer<LapInfo, histSize> history;
int maxLapNumber;
```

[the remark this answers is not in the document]
mark9064: Looks perfect to me, like that naming a lot 👍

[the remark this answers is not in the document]
codingjourney: Applied.

codingjourney: Applied.

mark9064: On reflection you make a good point, there's no harm in default values and the performance/space cost is very small (if any)

mark9064:
@FintasticMan Can you trigger CI?

codingjourney:
Added a fix for a minor issue: after exiting StopWatch in the `Paused` state and then reopening it, the displayed time would shift by 4, sometimes 5 hundredths of a second. I think this is the delta between the last rendering and the moment we switch state to `Paused` and set `timeElapsedPreviously`. (The delta is consistent with the ~24 Hz refresh frequency I've observed.) The fix simply renders time once more once we are `Paused`.

src/displayapp/screens/StopWatch.cpp
```diff
@@ -123,96 +144,89 @@ void StopWatch::SetInterfaceStopped() {
...
   }
 }
 
 bool StopWatch::OnButtonPushed() {
-  if (currentState == States::Running) {
-    Pause();
+  if (stopWatchController.IsRunning()) {
+    stopWatchController.Pause();
+    DisplayPaused();
+    wakeLock.Release();
```
mark9064, line 229: Is RenderTime needed here (like the button handler)?

src/displayapp/screens/StopWatch.cpp
```diff
@@ -123,96 +144,89 @@ void StopWatch::SetInterfaceStopped() {
...
+void StopWatch::PlayPauseBtnEventHandler() {
+  if (stopWatchController.IsCleared() || stopWatchController.IsPaused()) {
+    stopWatchController.Start();
+    DisplayStarted();
+    wakeLock.Lock();
+  } else if (stopWatchController.IsRunning()) {
+    stopWatchController.Pause();
+    blinkTime = xTaskGetTickCount() + blinkInterval;
+    DisplayPaused();
+    RenderTime();
```
mark9064, line 209: Maybe rendertime before displaypaused? Rendering the buttons takes a while and there is a noticeable delay for the time to update on my watch

src/displayapp/screens/StopWatch.cpp
mark9064: Actually more broadly is there any different for the pause button handler and side button handler to be different? Maybe they could be unified

codingjourney: In InfiniTime 1.14, the on-screen button is a play/pause toggle while the side button is not: when pressed in the `Paused` state, it exits the screen. This PR preserves that behavior.

I do find the side button slightly counter-intuitive and occasionally exit the screen without meaning to. Having both buttons work the same way would improve UI consistency.

Having said that, I just noticed that pausing with the side button is much slower than with the on-screen button (like 300-400 vs. 150 ms) so the consitency is debatable. And, of course, some people may be used to exiting via the side button; I can't estimate how much of an issue that is.

On the balance, I'm in favor of keeping existing side button behavior. You are right, however, that the transition to the `Paused` state should be the same in both cases so I've extracted it into a common method.

src/displayapp/screens/StopWatch.cpp
codingjourney: Applied in `OnPause()`.

src/displayapp/screens/StopWatch.cpp
mark9064: Yeah the real kicker with the side button is that it triggers on button release rather than press I think - improving this is definitely out of scope for this PR though as you say

src/displayapp/screens/StopWatch.cpp
```diff
@@ -123,97 +144,92 @@ void StopWatch::SetInterfaceStopped() {
...
   return false;
 }
+
+void StopWatch::OnPause() {
+  stopWatchController.Pause();
+  blinkTime = xTaskGetTickCount() + blinkInterval;
+  RenderTime(); // make sure displayed time is not stale
+  DisplayPaused();
+  wakeLock.Release();
+}
```
mark9064, line 235: Missing newline

mark9064 (approved):
Looks great! Tested on hardware and works flawlessly

Thanks for sticking through the many rounds of feedback 😅

src/displayapp/screens/StopWatch.cpp
codingjourney: fixed

codingjourney:
By the way, I've noticed this PR breaks InfiniSim. Does that have to be adapted before things get merged in InfiniTime? I don't mind looking into it, it's just that the build setup seems a bit more involved (I went with [the Docker method](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/doc/buildWithDocker.md) for InfinTime and that's super smooth).

mark9064:
Yeah it's good practice to get an InfiniSim PR up and waiting if a PR here will break it (annoyingly the CI has given up again so I can't easily see). I haven't used the docker method so I can't comment on that but provided you're on Linux getting a build environment up and running shouldn't be too bad.

codingjourney:
I've created PR [#163](https://github.com/InfiniTimeOrg/InfiniSim/pull/163) for InfiniSim. Necessary changes were trivial. Functionality in InfiniSim seems to work as intended.

mark9064:
Great :)

src/displayapp/screens/StopWatch.cpp
```diff
@@ -4,49 +4,51 @@
...
+  TimeSeparated ConvertTicksToTimeSegments(const TickType_t timeElapsed) {
     // Centiseconds
     const int timeElapsedCentis = timeElapsed * 100 / configTICK_RATE_HZ;
 
     const int hundredths = (timeElapsedCentis % 100);
     const int secs = (timeElapsedCentis / 100) % 60;
     const int mins = ((timeElapsedCentis / 100) / 60) % 60;
     const int hours = ((timeElapsedCentis / 100) / 60) / 60;
-    return TimeSeparated_t {hours, mins, secs, hundredths};
+    return TimeSeparated {hours, mins, secs, hundredths};
```
mark9064, line 18: One thing I've noticed: if the stop watch is left running for a long time these calculations overflow. Could this be refactored to avoid anything exceeding a uint32? The actual tick count overflows every `2^32 / 2^10 = 2^22s ~= 48days`, but dividing that by 100 gives just 11.65h (as currently the tick count is multiplied by 100, so overflow happens 100x earlier). I don't think we need to worry about a 48 day overflow, but more than 11h would be nice.

src/displayapp/screens/StopWatch.cpp
```diff
@@ -123,97 +144,92 @@ void StopWatch::SetInterfaceStopped() {
...
-  if (currentState == States::Running) {
-    laps[lapsDone] = oldTimeElapsed + xTaskGetTickCount() - startTime;
-
-    TimeSeparated_t currentTimeSeparated = convertTicksToTimeSegments(laps[lapsDone]);
-    if (currentTimeSeparated.hours == 0) {
-      lv_label_set_text_fmt(time, "%02d:%02d", currentTimeSeparated.mins, currentTimeSeparated.secs);
+    if (lap) {
+      TimeSeparated laptime = ConvertTicksToTimeSegments(lap->timeSinceStart);
+      char buffer[16];
+      sprintf(buffer, "#%2d   %2d:%02d.%02d\n", lap->number, laptime.mins, laptime.secs, laptime.hundredths);
```
mark9064, line 184: Maybe use `snprintf` with the buffersize fixed here? Since the number of laps can now exceed two digits the length is now undefined and could overflow the buffer

Also, do you think the hour could be fit in the lap time here? I'm not sure how it'll look, but I think there might be enough horizontal space

codingjourney: I tried to fix this, fell down a rabbit hole and got stuck :-)

I first made the change as you suggested. Then I found I didn't like the lap number being unlimited in the first place, so I capped it at 999. That made `snprintf` technically not necessary but I left it in anyway (buffer overflows are a Bad Thing).

I tested a bit and discovered that elapsed time could now reach over 1000 hours (thanks to the integer overflow fix from a couple weeks ago). That didn't fit the display anymore, so I decided to cap elapsed time as well, at 999:59:59.99 (the watch resets to 0:00.00 afterwards and continues running).

Then I noticed that lap times didn't include hours so I added them (but only when time.hours > 0). I generally cleaned up the lap list format so that it remains aligned when the lengths of lap number or lap time change.

I thought I would expand the lap list because there is clearly room for 3 entries (even 4 when the time label switches to a smaller font at 1:00:00). This required aligning the time label to the top of the display rather than stacking it on top of the other labels as was the case before.

It all involved lots of iteration so I switched to InfiniSim for testing. I got everything looking really good after some effort. When I tried compiling for PineTime, however, I hit a snag: "region RAM overflowed with stack", with a Used Size of 64520 B.

I tried reverting various changes I'd made and eventually found that the problem arises when I raise `histSize` in `StopWatchController.h` to more than 2. `histSize` is the size of the circular buffer for `LapInfo` instances, so it must be at least equal to the maximum number of displayed laps, i.e. 4.

If I calculate correctly, `sizeof(LapInfo)` is 8 so raising `histSize` from 2 to 4 should be adding 16 bytes somewhere and this obviously breaches some limit close to 64K. I have no clue what to do about this so I would appreciate any hints or advice. The branch, as it stands now, will not build.

mark9064: Yeah this is expected, InfiniTime is pretty near the static memory limit. You can increase the memory available by shrinking the heap. To do this open `FreeRTOSConfig.h` and lower the heap size multiplier from 40 to 39. In the future this will ideally happen automatically with #2172 . It should build OK after

The other changes sound great! I will try to take a look soon

codingjourney: Heap shrunk, formatting issues fixed.

src/components/stopwatch/StopWatchController.h
```diff
@@ -0,0 +1,68 @@
...
+      /// Indexes into lap history, with 0 being the latest lap.
+      std::optional<LapInfo> GetLapFromHistory(int index);
+
+      bool IsRunning();
+      bool IsCleared();
+      bool IsPaused();
+
+    private:
+      // Time at which stopwatch wraps around to zero (1000 hours)
+      TickType_t elapsedTimeBoundary = configTICK_RATE_HZ * 60 * 60 * 1000;
```
mark9064, line 50: Can this be constexpr?

src/displayapp/screens/StopWatch.cpp
```diff
@@ -123,97 +143,113 @@ void StopWatch::SetInterfaceStopped() {
...
-  lapsDone = 0;
+void StopWatch::RenderTime() {
+  TimeSeparated elapsedTime = ConvertTicksToTimeSegments(stopWatchController.GetElapsedTime());
+  SetHoursVisible(elapsedTime.hours != 0);
+  if (!hoursVisible) {
+    lv_label_set_text_fmt(time, "%02d:%02d", elapsedTime.mins, elapsedTime.secs);
+  } else {
+    lv_label_set_text_fmt(time, "%02d:%02d:%02d", elapsedTime.hours, elapsedTime.mins, elapsedTime.secs);
+  }
+  lv_label_set_text_fmt(msecTime, "%02d", elapsedTime.hundredths);
```
mark9064, line 154: Out of scope for this PR probably, but for future reference DirtyValue (just pass in seconds) could be used to avoid redrawing all of the digits every frame (so the hundreths/tenths update faster)

mark9064 (commented):
Testing now, not been running long enough to see hours but looking good so far
I'm not sure how much I like the feel of the time being pushed right to the top? But that's probably just me being used to how it was before, so I'll see after living with it for a bit :+1:

src/components/stopwatch/StopWatchController.h
codingjourney: Fixed. Declaring the expression as `static constexpr` caused it to be interpreted as signed, resulting in an integer overflow error. I could either cast it or change `configTICK_RATE_HZ` from `1024` to `1024u`. The latter would have been technically correct but I didn't know what side effects it could cause so I chose the cast.

src/displayapp/screens/StopWatch.cpp
codingjourney: Thanks for pointing this out. I'd somehow thought the widget was smart enough to avoid redundant rendering but that's clearly not the case. So yeah, it's probably out of scope but I applied it anyway. Saving cycles in the refresh task can't hurt.

src/components/stopwatch/StopWatchController.h
mark9064: Interestingly my IDE says it becomes an unsigned value, but perhaps the negative integer result is then getting reinterpreted as unsigned? Not sure, but I agree that casting explicitly is a good idea nonetheless

Could you change this to use `static_cast` rather than a C-style cast?

src/components/stopwatch/StopWatchController.h
```diff
@@ -0,0 +1,68 @@
...
+  namespace System {
+    class SystemTask;
+  }
+
+  namespace Controllers {
+
+    enum class StopWatchStates { Cleared, Running, Paused };
+
+    struct LapInfo {
+      int number = 0;                // Used to label the lap
```
mark9064, line 18: I think it makes no difference here due to alignment, but could you shrink the integer types to the smallest applicable? So for here `uint16_t` should be good
Same goes for other member variables / constants, it can save some memory. I wouldn't get too bothered about this, personally I like to size everything but it only really matters for static variables

src/displayapp/screens/StopWatch.cpp
```diff
@@ -123,97 +143,116 @@ void StopWatch::SetInterfaceStopped() {
...
-    const TickType_t currentTime = xTaskGetTickCount();
-    if (currentTime > blinkTime) {
-      blinkTime = currentTime + blinkInterval;
-      if (lv_obj_get_state(time, LV_LABEL_PART_MAIN) == LV_STATE_DEFAULT) {
-        lv_obj_set_state(time, LV_STATE_DISABLED);
-        lv_obj_set_state(msecTime, LV_STATE_DISABLED);
+    if (lap) {
+      TimeSeparated laptime = ConvertTicksToTimeSegments(lap->timeSinceStart);
+      char buffer[19];
+      if (laptime.hours == 0) {
```
mark9064, line 182: I think hours should be shown if the total runtime is >= 1 hour, rather than lap time >=1 hour. Having a leading zero in the hours column is nicely unambiguous. What do you think?

mark9064:
Looking superb. Just needs a rebase for CI to run (and the last few bits of feedback) and then it's all good from me

src/displayapp/screens/StopWatch.cpp
mark9064: Another style thought: since the time digits above having leading zeroes for minutes, I kinda feel like the lap numbers should too? But again it's a taste thing, I'll leave it up to you I'm happy either way

src/components/stopwatch/StopWatchController.h
codingjourney: Done.

src/components/stopwatch/StopWatchController.h
codingjourney: Done.

src/displayapp/screens/StopWatch.cpp
codingjourney: Clicking around Wikipedia suggests lap times are generally written without leading zeroes (see [Nürburgring lap times](https://en.wikipedia.org/wiki/List_of_N%C3%BCrburgring_Nordschleife_lap_times#Lap_times) or [400-m sprint all-time bests](https://en.wikipedia.org/wiki/400_metres#All-time_top_25), for example). StopWatch was inconsistent about it, displaying minutes even for sub-minute laps but hiding hours until needed. I changed rendering to adhere to the convention and it now feels clean and tidy to me.

codingjourney:
> I'm not sure how much I like the feel of the time being pushed right to the top? But that's probably just me being used to how it was before, so I'll see after living with it for a bit 👍

Regarding vertical positioning of the time label, my thinking goes like this:
- the top margin is about half as wide as the side margins - not good
- on PineTime, the margin blends with the watch bezel so it's OK - good
- there's too much empty room above the buttons when no laps are listed - not good
- there's enough room to list 3 laps - good

It would be nice if the time was vertically centered between the top and the buttons and then float upwards as laps get added to the list, eventually ending up where it is now. I don't find the slight improvement is worth the effort, though.

mark9064 (commented):
Will test this latest version on HW soon, good to go IMO

src/displayapp/screens/StopWatch.h
```diff
@@ -3,72 +3,66 @@
...
-    struct AppTraits<Apps::StopWatch> {
-      static constexpr Apps app = Apps::StopWatch;
-      static constexpr const char* icon = Screens::Symbols::stopWatch;
+      Pinetime::System::WakeLock wakeLock;
+      Controllers::StopWatchController& stopWatchController;
+      TickType_t blinkTime = 0;
+      uint8_t displayedLaps = 3;
+      lv_obj_t *time, *msecTime, *btnPlayPause, *btnStopLap, *txtPlayPause, *txtStopLap;
+      lv_obj_t* lapText;
+      Utility::DirtyValue<TickType_t> renderedSeconds;
```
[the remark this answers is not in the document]
codingjourney: Fixed.

src/displayapp/screens/StopWatch.cpp
```diff
@@ -123,97 +143,118 @@ void StopWatch::SetInterfaceStopped() {
...
-  SetInterfacePaused();
-  startTime = 0;
-  // Store the current time elapsed in cache
-  oldTimeElapsed = laps[lapsDone];
-  blinkTime = xTaskGetTickCount() + blinkInterval;
-  currentState = States::Halted;
-  wakeLock.Release();
+void StopWatch::RenderPause() {
+  const TickType_t currentTime = xTaskGetTickCount();
+  if (currentTime > blinkTime) {
```
mark9064, line 162: There's a tiny bug here when the tick count wraps around, but it wasn't introduced by this PR

Just before the count wraps around, the blink time in the future will always be smaller than the current time
Instead, tracking the last blink time and then doing `currentTime - lastBlinkTime > blinkInterval` is correct (since `currentTime - lastBlinkTime` is always the number of ticks since `lastBlinkTime`, regardless of wraparound)

mark9064 (approved):
Lap formatting changes LGTM

PR #2143, opened 2024-10-23
cmake: fix python path with use multi-version python

xz-dev:
If I use python3.13 as my system python version, but I use python3.12 to create venv, cmake will find /usr/bin/python3.13 but not venv

Because cmake will use the latest version.

From https://cmake.org/cmake/help/latest/module/FindPython3.html
> Note
> This hint has the lowest priority of all hints, so even if, for example, you specify IronPython first and CPython in second, a python product based on CPython can be selected because, for example with Python3_FIND_STRATEGY=LOCATION, each location will be search first for IronPython and second for CPython. 

Refer: https://discourse.cmake.org/t/find-package-python3-is-not-finding-the-correct-python/10563


FintasticMan (approved):
Looks good to me!

mark9064 (approved):

PR #2145, opened 2024-10-23
ble: advertise battery charging state

liamcharger:
This PR allows companion apps to check if the battery is charging.

I created this because I am working on an enhancement in InfiniLink that depends on knowing if the battery is charging or not.

### DIFF
--- a/src/components/ble/BatteryInformationService.h
+++ b/src/components/ble/BatteryInformationService.h
@@ -30,7 +30,7 @@ namespace Pinetime {
 
       static constexpr ble_uuid16_t batteryInformationServiceUuid {.u {.type = BLE_UUID_TYPE_16}, .value = batteryInformationServiceId};
       static constexpr ble_uuid16_t batteryLevelUuid {.u {.type = BLE_UUID_TYPE_16}, .value = batteryLevelId};
-      static constexpr ble_uuid16_t batteryLevelStatusUuid {.u {.type = BLE_UUID_TYPE_16}, .value = batteryLevelStatusId}; // Renamed to batteryLevelStatusUuid
+      static constexpr ble_uuid16_t batteryLevelStatusUuid {.u {.type = BLE_UUID_TYPE_16}, .value = batteryLevelStatusId};
 
       struct ble_gatt_chr_def characteristicDefinition[3];
       struct ble_gatt_svc_def serviceDefinition[2];
--- a/src/components/ble/BatteryInformationService.h
+++ b/src/components/ble/BatteryInformationService.h
@@ -20,25 +20,23 @@ namespace Pinetime {
 
       int OnBatteryServiceRequested(uint16_t attributeHandle, ble_gatt_access_ctxt* context);
       void NotifyBatteryLevel(uint16_t connectionHandle, uint8_t level);
-      void NotifyBatteryCharging(uint16_t connectionHandle, bool isCharging);
+      void NotifyBatteryLevelStatus(uint16_t connectionHandle, bool isCharging);
 
     private:
       Controllers::Battery& batteryController;
       static constexpr uint16_t batteryInformationServiceId {0x180F};
       static constexpr uint16_t batteryLevelId {0x2A19};
-      static constexpr uint16_t isChargingId {0x2BED};
+      static constexpr uint16_t batteryLevelStatusId {0x2BED};
 
       static constexpr ble_uuid16_t batteryInformationServiceUuid {.u {.type = BLE_UUID_TYPE_16}, .value = batteryInformationServiceId};
-
       static constexpr ble_uuid16_t batteryLevelUuid {.u {.type = BLE_UUID_TYPE_16}, .value = batteryLevelId};
-
-      static constexpr ble_uuid16_t isChargingUuid {.u {.type = BLE_UUID_TYPE_16}, .value = isChargingId};
+      static constexpr ble_uuid16_t batteryLevelStatusUuid {.u {.type = BLE_UUID_TYPE_16}, .value = batteryLevelStatusId}; // Renamed to batteryLevelStatusUuid
 
       struct ble_gatt_chr_def characteristicDefinition[3];
       struct ble_gatt_svc_def serviceDefinition[2];
 
       uint16_t batteryLevelHandle;
-      uint16_t isChargingHandle;
+      uint16_t batteryLevelStatusHandle;
     };
   }
 }
--- a/src/components/ble/BatteryInformationService.h
+++ b/src/components/ble/BatteryInformationService.h
@@ -26,7 +26,7 @@ namespace Pinetime {
       Controllers::Battery& batteryController;
       static constexpr uint16_t batteryInformationServiceId {0x180F};
       static constexpr uint16_t batteryLevelId {0x2A19};
-      static constexpr uint16_t isChargingId {0x2A1B};
+      static constexpr uint16_t isChargingId {0x2BED};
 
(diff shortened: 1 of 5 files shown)


src/components/ble/BatteryInformationService.h
```diff
@@ -20,20 +20,25 @@ namespace Pinetime {
 
       int OnBatteryServiceRequested(uint16_t attributeHandle, ble_gatt_access_ctxt* context);
       void NotifyBatteryLevel(uint16_t connectionHandle, uint8_t level);
+      void NotifyBatteryCharging(uint16_t connectionHandle, bool isCharging);
 
     private:
       Controllers::Battery& batteryController;
       static constexpr uint16_t batteryInformationServiceId {0x180F};
       static constexpr uint16_t batteryLevelId {0x2A19};
+      static constexpr uint16_t isChargingId {0x2A1B};
```
mark9064, line 29: I can't find this number assigned in https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Assigned_Numbers/out/en/Assigned_Numbers.pdf

I could totally be missing it (I tried CTRL-F!), could you point me to it

liamcharger: That was my fault; I screwed up the number. Should be good now.

mark9064:
https://btprodspecificationrefs.blob.core.windows.net/gatt-specification-supplement/GATT_Specification_Supplement.pdf#3.29.3

As far as I can see this doesn't currently follow the structure required
We can leave the flags clear as we only need to transmit power state
In particular in power state we can leave the wired/wireless fields empty, and just fill battery present and battery charge state (i.e set bit 0 to 1, set bits 5-6 depending on state, and leave all other bits zero)

mark9064:
Someone pointed out that I should explain how the BLE specs actually work, forgive me for assuming so!

<details>
<summary>Length warning!!</summary>

## Explanation of BLE

Bluetooth low energy (BLE) GATT (Generic Attribute Profile) has many different services, which contain groups of characteristics. In InfiniTime, each service gets its own file, for example the Device Information Service (DIS) lives in DeviceInformationService.cpp. The group contains relevant characteristics, for example the DIS contains things like model number, manufacturer name, and software version. The specification for each service defines which characteristics are allowed inside the service, some will be mandatory and some optional.

Characteristics are how information is actually transferred. Characteristics can support multiple commands like Read, Write, Notify and Indicate (there are some additional less commonly used ones). Some commands are mandatory (e.g you must implement Read for Battery Level), and some are optional (you can also implement Notify for Battery Level, this allows the connected device to receive updates every time the battery level changes rather than having to check the level with Read periodically).  The data format which the characteristic communicates with is well defined in the BLE GATT specification supplement. For example, if we look at Battery Level, the supplement defines that the battery level must be transmitted as a percentage in the range 0-100 as a 1 byte unsigned integer.

So in this specific case we want to report the charging status of the device. If we look at the [Battery Service](https://www.bluetooth.com/specifications/specs/battery-service/), we can see there in Table 3.1 that the Battery Level characteristic is mandatory and that there are also many more optional characteristics we can provide with this service. If we read the the list of characteristics (3.1->3.12), we can see that Battery Level Status allows us to provide the charging state of the battery. It details all of the fields provided in the Battery Level Status. Once we know that we want to implement this characteristic, we need to know 1) the UUID of this characteristic 2) which commands we must support and 3) the exact structure the data within the characteristic takes.

For 1) we refer to the [Assigned Numbers](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Assigned_Numbers/out/en/Assigned_Numbers.pdf) document, which tells us that it is `0x2BED`. For 2) we can see in Table 3.1 that we must implement Read and Notify. For Notify, we can see in section 3.2.1 the conditions for when we must, may and must not Notify on the characteristic.

For 3) we refer to the [GATT Specification Supplement](https://btprodspecificationrefs.blob.core.windows.net/gatt-specification-supplement/GATT_Specification_Supplement.pdf). Section 3.29 details Battery Level Status.

Firstly there is a flags field, allowing to specify whether we will include the following fields 1) Identifier 2) Battery Level and 3) Additional Status. We do not need an identifier as we only have one battery, we can choose to report the Battery Level again if we want (there is not much reason to), and we do not need any of the functionality provided by Additional Status.

This means that the format of the data we are sending contains the fields "Flags" and "Power State" only. 3.29.1 specifies the format of those flags, and since we are not including any of the optional fields we set all of the flags to zero (RFU (Reserved for Future Use) bits are always left as zero).

The Power State field is 16 bits or 2 bytes in length. The meaning assigned to each bit is specified in 3.29.2, and Battery Service 3.2.1.2 gives more information on what the different options mean.
- Bit 0 Battery present: this will always be Yes
- Bit 1-2 Wired External Power Source Connected: this is Yes if the device is charging, otherwise No
- Bit 3-4 Wireless External Power Source Connected: always No
- Bit 5-6 Battery Charge State: Charging if the device is charging, otherwise Discharging with the battery active
- Bit 7-8 Battery Charge Level:  you could choose to make a choice based on the percentage, or choose to just leave it is as Unknown (I would probably leave it Unknown to avoid having to define thresholds for Low and Critical battery, which have their own special meaning in the specification)
- Bit 9-11 Charging Type: always Unknown (technically we know a little from the charger pins and datasheet but there's no need to complicate things by including this)
- Bit 12-14 Charging Fault Reason: always 0 for all bits (no fault) as we have no fault tracking
- Bit 15 RFU (Reserved for Future Use): these reserved bits are always left as 0

To actually create a 16 bit value with the right bits set, you'll want to OR together all of the bits you need and store the result in a `uint16_t`. Suppose we want to express the Battery Charge State bits (5-6) to signify Discharging. We take the value (2) and left shift (`<<`) it by 5 bits like so `2 << 5`. Of course in your code there will be a variable that has the value 2, so you might do `chargeState = ChargeState::Discharging; powerState = 1 | wiredExternalConnected << 1 | chargeState << 5 etc` or similar. Remember all bits will be zero by default, so you only need to include fields that can take values other than zero in your expression. Since the flags field is zero, you can just assign zero to a `uint8_t` and be done.

Finally in order to send the result you can create the `mbuf` as you have already, add the flags followed by the power state and you're done!! I haven't read over it in detail yet, but I think your current algorithm for choosing when to send notifications is already specification compliant (or very close to it!).  

---

The most important part to remember is that many things you might want to communicate over BLE already have a service defined, and even if you need to make your own service very often all of the characteristics you need are defined. Using these defined services and characteristics allows devices to be interoperable. For example, if you see that your phone can read the battery level of some generic BLE headphones, this is very likely because the headphones support the Battery Service. If every device used their own custom services and characteristics you'd have to write a custom driver for every device!

</details>

If you survive reading this, let me know 😅. Of course feel free to ask any questions

liamcharger:
Thank you! Updated to (hopefully) follows the spec. Still super new to this.

mark9064:
You're still a way off. It might be worth reading through what I wrote again (it will take a few reads to digest everything) and checking over all of the linked resources - those specifications linked are required reading to implement services correctly. In particular the GATT Specification Supplement shows you the structure of the data you need send for each characteristic. Again if any part isn't clear please ask, it isn't embarrassing to be confused by the mountain of documentation that is the BLE spec! I have to refer back to the documents constantly...

liamcharger:
Ok. I'll probably just close this for now until I learn everything I need to and have a solid understanding of it.

mark9064:
No rush man, why not set it as a work-in-progress/draft PR for now (IDK if you can since it's closed now actually). But take your time 👍

PR #2146, opened 2024-10-24
Maze watchface

Feksaaargh:
Adds a watchface integrated into a maze.

### Features
- A basic watchface
- 12/24 hour format support
- Real mazes
- Legible™ Text
- Battery and bluetooth connection indicators
- Amogus et al.

### Known issues
- When rebooting the watch, if this watchface is selected, the screen will be black. Refresh the screen by opening any other menu or long tapping the screen. I think this issue is because I'm not using LVGL and drawing straight to the screen, and I'm drawing too early. Any help on this is appreciated.

### Images
Watchface showing 12 hour format
Watchface showing 24 hour format
Two dripped out Blahaj holding hands

### Extra info
The maze generation is unbiased, and it uses Wilson's Algorithm for generation. You can see a very nice demonstration and explanation of it if you scroll down a bit on this page: https://professor-l.github.io/mazes.
In my usage, the seed is the numbers themselves (and the space allotted for the status indicators). However, since I have multiple seeds, it effectively generates multiple disconnected mazes. To fix this, I go through and search for borders between these disconnected mazes and poke holes to make them all connected again. I'm pretty sure that still makes an unbiased maze, but have no idea how to prove it.

Every time the minute rolls over, the RNG used to generate the maze is reseeded. So if you have two watches, they will both display the same maze. You can generate a new maze by long clicking. You can also generate a blank maze by long clicking twice in a row (press the button or long click again to go back to the regular watchface).
Also on this blank screen, you can swipe to enter codes to show silly images. I leave it an exercise to the users to discover what these display.
- Up Up Down Down Left Right Left Right
- Right Left Right Left Down Down Up Up
- Up Right Down Left
- Down Right Down Left Up Right Up
- Right Left Right Left Right Left Right Left (then tap it!)
- Up Left Down Right Up Left Down Right

<sub>Or if you don't like fun, here's an Imgur post containing the codes and resultant images: https://imgur.com/a/BOaVYQi</sub>

### DIFF
--- a/src/displayapp/screens/WatchFaceMaze.cpp
+++ b/src/displayapp/screens/WatchFaceMaze.cpp
@@ -199,14 +199,14 @@ WatchFaceMaze::WatchFaceMaze(Pinetime::Components::LittleVgl& lvgl,
                              Controllers::MotorController& motor,
                              const Controllers::Battery& batteryController,
                              const Controllers::Ble& bleController)
-  : dateTimeController{dateTimeController},
-    settingsController{settingsController},
-    motor{motor},
-    batteryController{batteryController},
-    bleController{bleController},
-    lvgl{lvgl},
-    maze{Maze()},
-    prng{MazeRNG()} {
+  : dateTimeController {dateTimeController},
+    settingsController {settingsController},
+    motor {motor},
+    batteryController {batteryController},
+    bleController {bleController},
+    lvgl {lvgl},
+    maze {Maze()},
+    prng {MazeRNG()} {
 
   // set it to be in the past so it always takes two clicks to go to the secret, even if you're fast
   lastLongClickTime = xTaskGetTickCount() - doubleDoubleClickDelay;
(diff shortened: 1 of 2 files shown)


Feksaaargh:
Not sure how to add labels, so apologies for not doing that. My first Github PR and all that...

mark9064:
Alright this a monster PR so review is going to be slow. It looks super cool though!

To start, could you run the autoformatter and make sure your code follows the [style guidelines](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/doc/coding-convention.md)? Actually I'm not sure why it isn't written there, but all functions (and enum members) should be PascalCase

Also, I see a few notes in the code where you're working around bugs (in InfiniSim or otherwise). Please remove the workarounds and report the bugs! It'll be much better for the project to fix the issues

Regarding time: use CurrentDateTime() for wall-clock time. For precise timing of things like inputs, animations etc, use `xTaskGetTickCount()` from FreeRTOS. The tick rate is defined by `configTICK_RATE_HZ`, and there's also a macro `pdMS_TO_TICKS` to convert a number of milliseconds to ticks - to convert seconds just multiply by the tick rate. Note that the tick count is an unsigned 32-bit integer that wraps around so you may have to account for this overflow - comparing if an absolute tick count is bigger/smaller than another will break on wraparound.

PR #2148, opened 2024-10-26
SimpleWeatherService #include <array>

lmamane:
not <vector> as that is what is actually used.
Fixes build failure
InfiniTime/src/components/ble/SimpleWeatherService.h:86:18: error: field ‘location’ has incomplete type ‘Pinetime::Controllers::SimpleWeatherService::Location’ {aka ‘std::array<char, 33>’

with g++ on Debian 12.2.0-14 and libstdc++-12-dev 12.2.0-14

### DIFF
--- a/src/components/ble/SimpleWeatherService.h
+++ b/src/components/ble/SimpleWeatherService.h
@@ -19,7 +19,7 @@
 
 #include <cstdint>
 #include <string>
-#include <vector>
+#include <array>
 #include <memory>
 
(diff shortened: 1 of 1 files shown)


mark9064 (approved):
InfiniTime officially only supports ARM GCC 10.3

This is a sound change though

FintasticMan (approved):
I thought I'd already reviewed this! Always good to fix these small things.

PR #2149, opened 2024-10-27
Correct tick overflow handling for timer reset

mark9064:
Imagine the tick count wraps around at 1000, the hold time is set to 150, and the current time is 900. With the current code the comparison becomes `900 > 900 + 150` which is `900 > 50` as (1050 % 1000 = 50) so the condition will immediately be true.

With the new code, the comparison becomes `0  > 150` so there is no problem.

### DIFF
--- a/src/displayapp/screens/Timer.cpp
+++ b/src/displayapp/screens/Timer.cpp
@@ -105,7 +105,7 @@ void Timer::UpdateMask() {
 void Timer::Refresh() {
   if (timer.IsRunning()) {
     DisplayTime();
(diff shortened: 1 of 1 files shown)


minacode:
Don't you get the same problem now, just the other way around?

1050 gets 50 and then you check 50 - 900 > 150 which can never be true.
Or is 50 - 900 negatively overflowing to some large positive number as well?

mark9064:
Yeah you get that negative wraparound. So 50 - 900 is 150

Crucially, t2 - t1 will always give you the correct duration between them in ticks (provided t2 is in the future from t1, and less than one wraparound of ticks has elapsed). When you know the duration, you can compare that with a constant without issue

PR #2150, opened 2024-10-28
Add "Run on InfiniEmu" link to pull requests, update actions/upload-artifact

pipe01:
Using nightly.link and corsproxy.io we can generate a link that, when clicked, will direct users to a build of the latest commit of a pull request running on InfiniEmu.

It's also worth noting that GitHub actions artifacts expire after 90 days.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -205,7 +205,7 @@ jobs:
         | data    | ${{ needs.build-firmware.outputs.data_size }}B | ${{ steps.output-sizes-diff.outputs.data_diff }}B |
         | bss     | ${{ needs.build-firmware.outputs.bss_size }}B  | ${{ steps.output-sizes-diff.outputs.bss_diff }}B  |
 
-        [Run in InfiniEmu](https://infiniemu.pipe01.net/?firmware=https://corsproxy.io/?https://nightly.link/${{ github.repository }}/actions/artifacts/${{ needs.build-firmware.outputs.firmware_artifact }}.zip&resources=https://corsproxy.io/?https://nightly.link/${{ github.repository }}/actions/artifacts/${{ needs.build-firmware.outputs.resources_artifact }}.zip)
+        [Run in InfiniEmu](https://infiniemu.pipe01.net/?firmware=artifact://${{ github.repository }}/${{ needs.build-firmware.outputs.firmware_artifact }}&resources=artifact://${{ github.repository }}/${{ needs.build-firmware.outputs.resources_artifact }})
         EOF
 
     - name: Upload comment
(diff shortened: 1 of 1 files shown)


liamcharger:
Would really like to see this merged, this makes testing PRs much easier.

.github/workflows/main.yml
```diff
@@ -200,10 +204,12 @@ jobs:
         | text    | ${{ needs.build-firmware.outputs.text_size }}B | ${{ steps.output-sizes-diff.outputs.text_diff }}B |
         | data    | ${{ needs.build-firmware.outputs.data_size }}B | ${{ steps.output-sizes-diff.outputs.data_diff }}B |
         | bss     | ${{ needs.build-firmware.outputs.bss_size }}B  | ${{ steps.output-sizes-diff.outputs.bss_diff }}B  |
+
+        [Run on InfiniEmu](https://infiniemu.pipe01.net/?firmware=https://corsproxy.io/?https://nightly.link/${{ github.repository }}/actions/artifacts/${{ needs.build-firmware.outputs.firmware_artifact }}.zip&resources=https://corsproxy.io/?https://nightly.link/${{ github.repository }}/actions/artifacts/${{ needs.build-firmware.outputs.resources_artifact }}.zip)
```
FintasticMan, line 208: Would it make more sense to say "Run in InfiniEmu"?

FintasticMan (approved):
Love it!

.github/workflows/main.yml
pipe01: Probably, I'm never sure which one to use haha

mark9064:
Is using an open CORS proxy wise? It feels dangerous to trust it with user traffic - if it started acting maliciously (got hacked, owner decided to cash in etc) it could return whatever it wanted to user's browsers

pipe01:
Maybe not, but the worst that could happen is that they serve either an invalid ZIP file which would fail to load, or they could serve a different firmware file which would just be annoying.

mark9064:
I'm trying to think if there's another way to load it, could InfiniEmu itself fetch the firmware somehow?

pipe01:
If you mean fetching from GitHub directly it's not really possible because there's no official way to download artifacts without being logged in or using an API key, both of which aren't really feasible in this case. That's what nightly.link solves, however it unfortunately doesn't allow CORS.

mark9064:
Hmm, could the InfiniEmu server fetch it and serve the artefact to the user?

pipe01:
There's no server per se right now, it's hosted on GitHub pages.

I've made a slight improvement by not hardcoding the CORS proxy into the URL in the comment and instead letting InfiniEmu handle that so that it's easier to change in the future if needed.

mark9064:
👍 makes sense
Excuse my ignorance, but would requiring being logged in to github remove the need for the proxy? Since anyone who's going to comment on the PR has to be logged in anyway, I don't think it would be an issue

mark9064:
Ah, I thought the github cookies for that might be allowed to be samesite, but I guess it makes sense that they're not and there'd have to be a full Oauth flow (which isn't possible on GH pages I guess)

pipe01:
@mark9064 I think this is ready to merge :pray: Would be nice to have soon to test PRs as they come in

mark9064 (approved):
Still not a fan of using a CORS proxy but that's now up to Infiniemu itself which could be refactored, so I guess that problem is now pushed out of scope

PR #2153, opened 2024-11-01
do not pollute Pinetime::Applications::Screens namespace

lmamane:
with datatypes used only in/by StopWatch class

### DIFF
--- a/src/displayapp/screens/StopWatch.h
+++ b/src/displayapp/screens/StopWatch.h
@@ -16,17 +16,17 @@ namespace Pinetime {
   namespace Applications {
     namespace Screens {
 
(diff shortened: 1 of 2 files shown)


src/displayapp/screens/StopWatch.h
```diff
@@ -16,17 +16,17 @@ namespace Pinetime {
...
-      struct TimeSeparated_t {
-        int hours;
-        int mins;
-        int secs;
-        int hundredths;
-      };
-
       class StopWatch : public Screen {
       public:
+        enum class States { Init, Running, Halted };
```
mark9064, line 21: This can be private

PR #2154, opened 2024-11-02
port to Font Awesome v6.6.0

lmamane:
Le me know whether that change is welcome.

It started because I wanted some icons from v6 for an app I'm writing, and I thought it would be cleaner to port to v6 rather than keep both v5 and v6 in the source tree. If for some reason you prefer to keep v5, no problem, we/I can ship both v5 and v6 in the source tree and pick some symbols from v5 and other symbols from v6. That would make the source tree larger, but as far as I understand not the built firmware.

- I dropped 0xf00c (single checkmark) because it seemed to be unused. The symbol named "check" is actually a single checkmark.
- The codepoint for tachymeter-alt in v5 corresponds to a quite different icon in v6; I changed to the codepoint in v6 for a very similar icon named gauge-high
- the home codepoint in v5 has a chimney but doesn't have in v6 (named house in v6), so I changed to the v6 codepoint for house-chimney to keep the icon as similar as possible; also the roof is not separated by a blank from the body of the house, that is thus a small graphical change.


lmamane:
Darn, I forgot the weather icons... building them must be disabled in my build, because it builds on my computer... Coming back soon with a revised one.

mark9064:
Need to test this but looks good on principle, agree that there is no need to keep v5 around so a migration to v6 makes sense

tituscmd:
Friendly little nudge :)

Just tried this on my devkit and I haven't had any issues either.

tituscmd:
> I dropped 0xf00c (single checkmark) because it seemed to be unused. The symbol named "check" is actually a single checkmark.

This icon is actually used for `LV_SYMBOL_OK`, used in the square checkboxes right here:
image
Without it, it just shows a green box without any checkmark, so it should be kept in the code.

tituscmd:
> Any chance you could do a run with InfiniSim and upload some screenshots of the difference? If not I can make some, but I don't have the time today

Since I'm already working with it right now, here's some notable differences - Left side is before, right side is after:
_(This is a custom fork of mine. The settings menus might be renamed and reordered, but the icons are the default ones)_


Screenshot 2025-06-11 at 17-58-01 InfiniEmu Screenshot 2025-06-11 at 17-56-23 InfiniEmu
---
Screenshot 2025-06-11 at 17-59-16 InfiniEmu Screenshot 2025-06-11 at 18-00-07 InfiniEmu
---
Screenshot 2025-06-11 at 17-58-24 InfiniEmu Screenshot 2025-06-11 at 17-57-04 InfiniEmu

tituscmd:
I felt like the new version of the calculator icon looked weird in InfiniEmu, but turns out it looks fine on-device:
image

lmamane:
> Since I'm already working with it right now, here's some notable differences - Left side is before

Yes, the house and tachometer things were described in my initial message.

tituscmd:
> > Since I'm already working with it right now, here's some notable differences - Left side is before
> 
> Yes, the house and tachometer things were described in my initial message.

Yep! I put them in there so others can see exactly what you're talking about 😄

lmamane:
> This icon is actually used for `LV_SYMBOL_OK`

I don't find the bit of code that does that, but indeed. I've been running with this modification for months but didn't cotton on there was _supposed_ to be a checkmark. Added it back.

lmamane:
rebased on current main, resolved merge conflicts

mark9064 (approved):
Haven't checked weather icons but rest looks good to me. Always good to keep on the latest versions

tituscmd:
> Haven't checked weather icons but rest looks good to me. Always good to keep on the latest versions

Fwiw, I'm running this on my main device now and every weather symbol so far has been represented without issue.

tituscmd:
Since Font Awesome v7 has been released already, we might as well just update to that. Was pretty straight forward for me to go to v7 from v6

Also, we get this neat symbol for free in v7 :)
image

PR #2157, opened 2024-11-03
documentation make cmake work out of the box

lmamane:
since the previous command done when one follows the documentation is "cd build", one needs to point cmake the root directory of the sources. Add "-S .." for that.

### DIFF
--- a/doc/buildAndProgram.md
+++ b/doc/buildAndProgram.md
@@ -66,7 +66,7 @@ DFU files are the files you'll need to install your build of InfiniTime using OT
 #### CMake command 
 
 ```
-cmake -DARM_NONE_EABI_TOOLCHAIN_PATH=... -DNRF5_SDK_PATH=...
+cmake -DARM_NONE_EABI_TOOLCHAIN_PATH=... -DNRF5_SDK_PATH=... -S ..
 ```
 
 ### Build the project


doc/buildAndProgram.md
```diff
@@ -66,7 +66,7 @@ DFU files are the files you'll need to install your build of InfiniTime using OT
 #### CMake command 
 
 ```
-cmake -DARM_NONE_EABI_TOOLCHAIN_PATH=... -DNRF5_SDK_PATH=...
+cmake -DARM_NONE_EABI_TOOLCHAIN_PATH=... -DNRF5_SDK_PATH=... -S ..
```
mark9064, line 69:
```suggestion
cmake -DARM_NONE_EABI_TOOLCHAIN_PATH=... -DNRF5_SDK_PATH=... ..
```
The `-S` is redundant here right? I could be reading the docs wrong, but from what I can see the last parameter to cmake is the source directory

lmamane: When a directory is passed as argument naked without `-S` nor `-B`, `cmake` will auto-sniff whether it is a source directory or an existing build directory and act accordingly as if `-S` or `-B` had been passed, respectively. So yes, in this case, in the context of the preceding commands of the documentation, removing the `-S` will will work the same.

My taste is rather to be explicit (explicit `-S`) for better robustness, and I think having the `-S` is more pedagogical for readers that may be exposed to `cmake` for the first time trying to build InfiniTime.

This being said, I would agree with either version (with or without `-S`) being merged.

mark9064: Ah, thanks for the explanation- I didn't know these nuances of cmake. I agree that `-S` is more correct

mark9064 (approved):

JF002 (approved):

PR #2158, opened 2024-11-03
work around g++ version 12, 13 and 14 spurious warning

lmamane:
which is an error since we compile with -Werror

### DIFF
--- a/src/displayapp/screens/SystemInfo.cpp
+++ b/src/displayapp/screens/SystemInfo.cpp
@@ -240,7 +240,12 @@ std::unique_ptr<Screen> SystemInfo::CreateScreen4() {
   lv_table_set_col_width(infoTask, 3, 90);
 
   auto nb = uxTaskGetSystemState(tasksStatus, maxTaskCount, nullptr);
+// g++ emits a spurious warning (and thus error because we compile with -Werror)
+// due to the way std::sort is implemented
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Warray-bounds"
   std::sort(tasksStatus, tasksStatus + nb, sortById);
+#pragma GCC diagnostic pop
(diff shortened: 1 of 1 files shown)


TheGreatMcPain:
I ran into this one as well.

```
/home/myswamp/git-repos/pinetime/InfiniTime/src/displayapp/screens/Navigation.cpp:194:15: error: possibly dangling reference to a temporary [-Werror=dangling-reference]
  194 |   const auto& image = GetIcon("flag");
      |               ^~~~~
```

Had to add this.

```
diff --git a/src/displayapp/screens/Navigation.cpp b/src/displayapp/screens/Navigation.cpp
index ee9f2a00..01e14209 100644
--- a/src/displayapp/screens/Navigation.cpp
+++ b/src/displayapp/screens/Navigation.cpp
@@ -191,7 +191,10 @@ namespace {
  *
  */
 Navigation::Navigation(Pinetime::Controllers::NavigationService& nav) : navService(nav) {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdangling-reference"
   const auto& image = GetIcon("flag");
+#pragma GCC diagnostic pop
   imgFlag = lv_img_create(lv_scr_act(), nullptr);
   lv_img_set_auto_size(imgFlag, false);
   lv_obj_set_size(imgFlag, 80, 80);
```

It's not exactly a "spurious warning", but it does happen in GCC14 for me.

mark9064:
That one can be resolved by modernising the type of icon
```diff
-  Icon GetIcon(const std::string& icon) {
+  Icon GetIcon(std::string_view icon) {
```
Nonetheless it is a spurious warning (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=114220)

mark9064:
Also, the bounds warning bug disappears with LTO enabled. Intriguing

NeroBurner (approved):

PR #2159, opened 2024-11-03
ImmediateAlertService: fix latent bug

DavisNT:
Include null terminator in the bytes copied.
Set `notif.size` like it is done in AlertNotificationService.cpp and AlertNotificationClient.cpp.

### DIFF
--- a/src/components/ble/ImmediateAlertService.cpp
+++ b/src/components/ble/ImmediateAlertService.cpp
@@ -63,7 +63,8 @@ int ImmediateAlertService::OnAlertLevelChanged(uint16_t attributeHandle, ble_gat
       auto* alertString = ToString(alertLevel);
 
       NotificationManager::Notification notif;
-      std::memcpy(notif.message.data(), alertString, strlen(alertString));
+      std::memcpy(notif.message.data(), alertString, strlen(alertString) + 1);
+      notif.size = strlen(alertString) + 1;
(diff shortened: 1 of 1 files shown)


FintasticMan (approved):
I'm fine with this implementation, but I think that it might be good to also use `std::min` to avoid overflowing the buffer. (Of course that isn't possible with the current notification strings, but it makes the code more clear if we do that)

Because there are some inconsistencies and incorrect handlings of the notification buffer elsewhere in the code, I'm thinking of adding a constructor to the struct that makes sure it's written correctly. In the meantime, this is a good fix.

FintasticMan:
I was planning on doing it, but if you want to, go ahead!

NeroBurner (approved):

mark9064 (approved):
At a glance seems fine to me - haven't looked at the code around this though

Refactoring this problem away would be better but that can be done another time

DavisNT:
@mark9064, @NeroBurner Should I do the refactoring (implement a constructor for `NotificationManager::Notification` that checks for buffer overflows)?

mark9064:
That would be great :)

DavisNT:
I have made a new PR with the refactoring: #2347 
It also requires corresponding changes in InfiniSim: https://github.com/InfiniTimeOrg/InfiniSim/pull/181

PR #2161, opened 2024-11-06
refuse to (fail to) calibrate shake sensitivity while in sleep mode

lmamane:


mark9064:
This all looks good, but I think the calibrating variable should be refactored into a state enum instead as it's currently somewhat unintuitive and the `255` value doesn't help. Perhaps something like `States {Idle, PrepareToCalibrate, Calibrating, CalibrationDisabled}`? These probably aren't the best names, I haven't thought about them much. Hopefully you get the idea though

dariusarnold:
This also has the issue that the text stays on the screen behind the other menu items if you leave the shake calibration screen.

lmamane:
Pull request #2268 enhances and replaces this one.

PR #2121, opened 2024-09-11
feat: Introduce Flake for development and builds

shymega:
This PR introduces a Nix flake, allowing for InfiniTime to be built as a Flake, including a FHS development environment.

It's derived from #1850 and
https://github.com/icewind1991/infinitime-builder/commit/c57c57f3c6a34c86eb816981d9b0ea56af4a09e5.

We also introduce `flake-compat`, allowing for non-Flake Nix mahcines to use the project as-is, both for building (`default.nix`), and development (`shell.nix`).

Additionally, we introduce `.envrc`, meaning that with `direnv`, the Nix Flake is activated automatically.

Fixes #1850.

PR #2166, opened 2024-11-14
ci: Normalise slash-containing `github.head_ref` values

shymega:
This fixes CI on #2121.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,25 +46,30 @@ jobs:
       # Unzip the package because Upload Artifact will zip up the files
       - name: Unzip DFU package
         run: unzip ./build/output/pinetime-mcuboot-app-dfu-*.zip -d ./build/output/pinetime-mcuboot-app-dfu
+      - name: Set ref_name, but replace slashes with dashes.
+        shell: bash
+        env:
+          ref_name: ${{ github.head_ref || github.ref_name }}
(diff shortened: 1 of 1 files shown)


shymega:
@mark9064 If you could run the workflow one more time - I was testing in zsh locally, not bash - tested in bash with the recent force-push, and it works locally. Should be fine now.

mark9064:
No worries, CI development is the worst because testing in prod is basically the only way :P

shymega:
One more time - then if that fails, I'll take a look either before or after working hours today.

FintasticMan:
I think you might need to escape the forward slash in the substitution. I.e. `${ref_name//\//-}`

shymega:
On 15.11.2024 06:30, FintasticMan wrote:
>I think you might need to escape the forward slash in the substitution. I.e. `${ref_name//\//-}`

The latest push should have the above implemented.

Best wishes,
--
Dom Rodriguez

FintasticMan (changes requested):
You've replaced `github.head_ref` with `github.ref_name`, meaning that the output is different.

FintasticMan (approved):
Looks good to me now!

shymega:
Excellent! Once we get this landed, I'll rebase #2121 - and then we can merge that, too!

JF002 (approved):

JF002:
Thanks for your contribution @shymega

PR #2168, opened 2024-11-15
Weather: fix min and max temperature being switched

lmamane:


FintasticMan (approved):
Oops, seems like I typo'd this...

mark9064 (approved):

PR #2169, opened 2024-11-23
LTO

mark9064:
Based off #867, a compiler upgrade isn't needed anymore and I also simplified the changeset to the minimal needed.

I'm testing it locally now and all seems good so far, but I've only been running it for a few hours.

The original in discussion in #867 regarding why the memory usage has changed is still to be answered - I haven't looked into this yet

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -853,6 +853,10 @@ if (${CMAKE_BUILD_TYPE} STREQUAL "Debug")
   # add_definitions(-DMYNEWT_VAL_BLE_HS_LOG_LVL=0)
 endif()
 
+if (CMAKE_BUILD_TYPE MATCHES Release)
+    set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
+endif()
+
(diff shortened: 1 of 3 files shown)


mark9064:
Since more functions get inlined, the stack has to accommodate more local variables at once. Given the increased stack usage of LTO, I think this is may be not such a good idea, as InfiniTime is very RAM limited - moreso than it is flash limited.

This PR works perfectly fine though, so keeping it around as a draft

mark9064:
With #2331 systemtask stack pressure is reduced now. Still, it needs some extra compared to before. I'm using GCC 14 locally, but I've never seen it go below 48 minimum with 400 size so that should be sufficient

mark9064:
One thing to do is make the CI build with LTO and without LTO (if it doesn't already), since some compiler errors will only appear with LTO off

mark9064:
I've also been running this for a long time, and I think it's well worth merging. Memory requirements for the SystemTask still do rise a bit, but 26K of flash back is huge and gives us plenty of breathing room for things like #2399

PR #2172, opened 2024-11-24
Use all free RAM for FreeRTOS heap

pipe01:
With this change you no longer need to fiddle with FreeRTOS's `configTOTAL_HEAP_SIZE` to make it fit into the RAM and instead it will always automatically use all available RAM for the heap.

This is achieved by using two symbols defined by the linker script: `__HeapLimit` and `__StackLimit`, which indicate where the used heap ends and where the stack "begins" (really ends) in memory respectively. **It's not yet ready for production** since I'm still not 100% sure there aren't any side effects (the emulator works fine with this build but if anyone with a devkit can run this for a while that'd be great), and also it currently only uses about 39.5 KB for the heap for reasons I'm still not quite sure about.

Here's a rough representation of the RAM:

```
|---------------|
|  0x20010000   |
|   <stack>     |
|   <stack>     |
|   <stack>     |
| __StackLimit  |
| <free space>  |
| <free space>  |
| <free space>  |
| <free space>  |
| <free space>  |
| __HeapLimit   |
|   <heap>      |
|   <heap>      |
| __HeapBase    |
| <global vars> |
| <global vars> |
| <global vars> |
|  0x20000000   |
|---------------|
```

*Note that in this case `<heap>` doesn't refer to FreeRTOS's heap but rather GCC's heap which is 0 bytes long in our case, so `__HeapLimit == __HeapBase`.*

Before this change, FreeRTOS's heap was statically allocated in the `<global vars>` region and the goal was to manually increment the heap size in order to make the `<free space>` region as small as possible, since that region is essentially wasted RAM space. Now we instead put the FreeRTOS heap right where the `<heap>` region ends and make it stretch right until the top of the stack (`__StackLimit`), which ensures we waste no RAM.

---

I did have to overwrite the `_sbrk` method because apparently there's a `printf` definition somewhere that uses a different implementation of `malloc` that doesn't use the FreeRTOS heap and instead allocates around 1 KB of RAM for itself, this is probably worth looking into just by itself.


mark9064:
This has been a wishlist item for me for ages! Awesome to see it realised

Could you explain the `_sbrk/printf` stuff a bit more? How did you find this out?

pipe01:
I found out about it when testing using InfiniEmu, it crashed because something was overwriting the FreeRTOS heap and it turned out to be that `printf` was `malloc`ing without using the FreeRTOS heap, I'm not sure how or why it's using its own `printf` and `malloc` implementations though.

image

pipe01:
Now LFS is using the `NRF_LOG` functions instead of `printf` so there's no longer an issue with any `malloc`s, however the solution is kind of ugly since we have to copy the entire `lfs_util.h` over and the CRC implementation, but it seems like there's no better way to customize littlefs other than forking it.

mark9064:
Why not define `LFS_TRACE/DEBUG/WARN/ERROR`? That way printf can never get included

pipe01:
That's what I'm doing [here](https://github.com/InfiniTimeOrg/InfiniTime/blob/30162b6588e303f8e9aa1b22285284784821f525/src/lfs_config.h#L5), the issue is that in order to define those macros before littlefs uses them we need to define `LFS_CONFIG` in cmake which will make littlefs load our file instead of its own `lfs_util.h` as seen [here](https://github.com/littlefs-project/littlefs/blob/ead50807f1ca3fdf2da00b77a0ce02651ded2d13/lfs_util.h#L16), but that means that we need to copy over the entire file to our project and also the CRC function for some reason.

mark9064:
Ah yeah I see the issue now, defining the macro at the right time is annoying.
I guess the one alternative is declaring them in FS.h before LFS is included for the first time? Not super pretty either though

pipe01:
Yeah that's what my first idea was but unfortunately it doesn't work because it wouldn't be included when compiling the `lfs.c`.

mark9064:
Ah, you're right :)
Probably would've been too hacky anyway

mark9064:
Any chance you could take a look at the build failure BTW? Also would I be right in saying that the sbrk implementation can be removed now with littlefs' printf removed?

mark9064:
Yeah I'm liking the sound of a hard fault there, nothing should ever be calling it. And same, I'm not sure malloc'ing there is correct, looking at the manpage shows some things that don't line up with it just being malloc (particularly the behaviour when decreasing the program break or checking it). I don't know if we could add some linker error to prevent the sbrk even being linked in? - it might be theoretically required by something anyway.

pipe01:
Sounds good. It does seem like sbrk is being used my something else, more research is needed.

pipe01:
It seems like there's no way to make `printf` and its variants use our `malloc` since it forcefully calls `_malloc_r` directly with no way to change it since we don't compile that code and instead it comes from the newlib binaries. I've written a `_sbrk` implementation that I'm pretty sure works how it's supposed to but I think it might be best to make it fault as discussed. I'm leaving the code here for posterity, just in case.

<details>
<summary>_sbrk implementation</summary>

```c
static int brk_size = 0;
static void *brk = NULL;

void* _sbrk(int size) {
  if (size == 0) {
    return brk;
  }
  brk_size += size;
  brk = pvPortRealloc(brk, brk_size);
  if (brk == NULL) {
    return (void*)-1;
  }
  return brk;
}
```

</details>

mark9064:
Very interesting reading on the topic https://nadler.com/embedded/newlibAndFreeRTOS.html
Looks like there are lots of approaches including wrapping malloc, linker tricks to not compile etc.
I need to find some time to understand them all, but maybe you'll find these useful?

pipe01:
That's very useful indeed, thanks! I'll try the ideas from the article and report back.

src/stdlib.c
```diff
@@ -25,3 +25,8 @@ void *pvPortRealloc(void *ptr, size_t xWantedSize);
 void* realloc( void *ptr, size_t newSize) {
   return pvPortRealloc(ptr, newSize);
 }
+
+void* _sbrk(int size) {
+  (void)size;
+  return *(void **)0; // Force fault
```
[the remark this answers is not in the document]
pipe01: I wasn't sure either but GCC seems to emit a `udf` instruction which generates an exception. I'm no longer using this though so it doesn't matter too much.

pipe01:
Seems to be working well now, I'm getting around 41KB of heap.

image

src/stdlib.c
```diff
@@ -10,23 +10,40 @@ void* malloc(size_t size) {
...
 }
 
+void __wrap_free(void* ptr) {
+  free(ptr);
+}
+
 void* calloc(size_t num, size_t size) {
-  (void)(num);
-  (void)(size);
+  (void) num;
```
mark9064, line 31: A random thought: any reason we don't implement this? Isn't it just `malloc(num * size); memset(ptr, 0, num*size)`? Perhaps I'm missing some alignment subtleties or size overflow or something

mark9064:
Looks great, lines up nicely with current heap size of 40960!

src/stdlib.c
pipe01: No reason, I just left as it was. I'll go ahead and implement it, I guess the only concern would be overflow as you mentioned but I don't think it's a big issue.

mark9064:
Hmm I see the lfs config breaking the formatting check. I haven't thought too much about where it belongs, but would it make sense to move it to the libs directory along with lv_conf.h (everything in libs is not format checked)? There might well be good reasons not to so I'm not sure

mark9064:
ohhhh nice yeah that undefine trick is good. Probably needs some comments to explain it. I'm all good to move the lfs_config.h to libs then

mark9064:
Booted fine on my watch. Will daily drive for the forseeable future and report back any issues

mark9064 (commented):
LGTM

No further comments for now, I'll approve after some wrist time (all looking good so far)

Edit: Might want to draft up an InfiniSim PR at some point, no rush though

mark9064 (approved):
Been working flawlessly. Is this ready in your eyes?

src/libs/lfs_config.h
```diff
@@ -0,0 +1,234 @@
+#pragma once
+
+#include <libraries/log/nrf_log.h>
+
+#define LFS_TRACE(fmt, ...) NRF_LOG_DEBUG("[LFS] %s:%d:trace: " fmt "\n", __FILE__, __LINE__, ##__VA_ARGS__)
+#define LFS_DEBUG(fmt, ...) NRF_LOG_DEBUG("[LFS] %s:%d:debug: " fmt "\n", __FILE__, __LINE__, ##__VA_ARGS__)
+#define LFS_WARN(fmt, ...) NRF_LOG_WARNING("[LFS] %s:%d:warn: " fmt "\n", __FILE__, __LINE__, ##__VA_ARGS__)
+#define LFS_ERROR(fmt, ...) NRF_LOG_ERROR("[LFS] %s:%d:error: " fmt "\n", __FILE__, __LINE__, ##__VA_ARGS__)
```
FintasticMan, line 8: I'm not sure I like the `##` here. It's a GNU extension, and we generally try to avoid those. We could use the trick used below with the empty string for a good alternative.

pipe01:
Been working fine on my end too, I think it's good to go!

src/libs/lfs_config.h
```diff
@@ -0,0 +1,234 @@
...
+
+// This is required in order for the CRC implementation in littlefs/lfs_util.c to be compiled
+#undef LFS_CONFIG
+
+// ====== Everything below this line is copied from lfs_util.h ======
```
FintasticMan, line 13: I'm not sure if I fully understood why we can't just add an include here to directly include the original file.

FintasticMan (commented):
Looks good to me too!

pipe01:
@FintasticMan made me realize that there's a better way to do the `lfs_config.h` stuff, I think it should work the same way.

FintasticMan (approved):
This looks good now!

JF002:
I'm a bit late to the party, but I would like to thank @pipe01 for these nice additions and improvements! They allow a better and more efficient usage of our precious RAM memory :1st_place_medal: 

I'm wondering if we could improve it just a bit more: the size reported by the compiler does not take the FreeRTOS heap into account anymore : 

```
           FLASH:      372944 B     474632 B     78.58%
             RAM:       23488 B        64 KB     35.84%
 372944	    948	  22536	 396428	  60c8c	pinetime-app-1.15.0.out
```

As you can see, the RAM region shows 35% utilization (by statically allocated memory), but does not show the FreeRTOS heap. I'm wondering if we could add a new region that would display the size of the heap? 

Also, I think it could be useful to  have some kind of warning if the size of this heap is reduced too much, or is allocated below a threshold? This would help us diagnose issues where a big chunk of memory is allocated by error, leaving too few memory for the application to run properly. This info could be added to the size comparison check:
image

I don't know how to implement those changes and if they are even possible, but I think it would be nice to think about those :)

PR #2201, opened 2024-12-16
Pride flag watchface

Aperture32GLaDOS:
Adds a new watch face with the trans flag as a background
# Features
* A new watch face, featuring the mlm, trans, bi, and wlw flags
* Bluetooth indicator in the top right
* Long day names
# Images
image
image
image
image


marigoldfish:
Love it! 🏳️‍⚧️😁 You are a very cool person, I'm going to be using this for a while!

IMG_20241229_213604426~2

src/displayapp/screens/WatchFaceTrans.h
```diff
@@ -0,0 +1,90 @@
...
+                       const Controllers::Ble& bleController,
+                       Controllers::NotificationManager& notificationManager,
+                       Controllers::Settings& settingsController,
+                       Controllers::MotionController& motionController);
+        ~WatchFaceTrans() override;
+
+        void Refresh() override;
+
+      private:
+        Utility::DirtyValue<int> batteryPercentRemaining {};
```
mark9064, line 39: `uint8_t` is sufficient here

mark9064 (commented):

Aperture32GLaDOS:
Thanks for the quick feedback! I've implemented those changes, hopefully everything's good now

mark9064:
All looks good to me! You couldn't have made review easier :)

I haven't had the chance to test on hardware yet and am away for now, but I should be able to test within a couple weeks time

marigoldfish:
I have been using this daily, and I really like it, but I have a few small suggestions:

I adore the "You have mail." for notifications, but I dislike that it overlaps so severely with the battery percentage:

IMG_20241231_082654043~2

I think it might look better to have "mail." on its own line below, and move "You have" up a little bit so they're both centered vertically in the blue stripe. 

(The 90's kid in me really wants it to read "You've got mail," like AOL, but that's not a request for a change, just a nostalgic observation)

My other suggestion is to name the watch face more specifically as "Trans Flag" instead of just "Trans."

IMG_20250103_220544316_HDR~2

Like I said, very very small suggestions. The other 99% of this watchface is grand!

Aperture32GLaDOS:
I definitely agree with the notification and the name. Fortunately, these are just one-line fixes so it's not that hard to change.
image

marigoldfish (commented):
If I can nitpick, I think "Mail" should be lowercase. But this looks really good 😁

Aperture32GLaDOS:
Good spot, I'll fix that

marigoldfish:
You are an incredibly awesome and speedy person! Thanks for implementing my suggestions, and so fast! I'll keep testing this daily for a while and let you know if I come across anything else 😁

marigoldfish:
Hello again, I noticed another little thing after using this watch face for a while. 

Usually I keep my watch in 24 hour time, but I had reason to use 12 hour time today, and I noticed that both AM and PM show up as "M."

(I manually set the watch to 2am to get the second picture, I didn't stay up late 😜)

IMG_20250117_184147295~2
IMG_20250117_184236513~3

Aperture32GLaDOS:
I think what's happening here is that the font used for the time doesn't have all the alphabet, if I change the font to something else it seems to work fine:
image
I've moved the AM/PM display to the day, since that uses a font which has all the alphabet
image

marigoldfish:
That's an elegant fix! I like it! 

Unfortunately, in addition to praise I also come bearing news of a new bug: it is displaying 24-hour time when I have 12-hour time enabled.

IMG_20250119_203002810~2

Aperture32GLaDOS:
Aah, well at least that's an easy fix; it *should* all be working now
image

NeroBurner:
could you update to/rebase onto the latest `main` branch. The toolchain won't run as the `artifacts v3` job was deprecated by GitHub, and that is fixen in the current `main` branch

rdub1a4:
Seems like this would wait for side loading. It may not fit the vision of "Only a minimal feature set in the flashed firmware".  But, no biggy.

mark9064:
Because it doesn't depend on any other PRs and was easy to review mainly

Things get added to the milestone when we want to target including them this cycle. If a PR prevents on lots of other refactoring for example, or has outstanding bugs that are taking time to resolve, then it usually won't be queued up for the next cycle until it's ready. InfiniTime generally lacks reviewers though, so what gets queued and what doesn't depends on whether someone's had time to look at it properly though

mark9064 (approved):
All LGTM code wise :)
Could you rename the watch face class/code/enum entry to (WatchFace)TransFlag? I think it makes sense for the code to match the displayed name

Haven't tested on HW yet

Aperture32GLaDOS:
Yeah, that makes sense - I've renamed all the WatchFaceTrans to WatchFaceTransFlag

joseph58tech:
Wait why don't we use this as a general LGBTQ flag watchface? It would make the most sense to add all of the flags into the design if we already have the one. You can choose which one you want the same way you change the colors on the Infinineat watchface.

Aperture32GLaDOS:
I really like that idea! I've implemented four pride flags, (the classic rainbow, the existing trans flag design, the bi flag and the lesbian flag), and the switching seems to work fine

Aperture32GLaDOS:
image
image
image
image

Aperture32GLaDOS:
Oops

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -0,0 +1,348 @@
...
+
+bool WatchFacePrideFlag::OnTouchEvent(Pinetime::Applications::TouchEvents event) {
+  if ((event == Pinetime::Applications::TouchEvents::LongTap) && lv_obj_get_hidden(btnClose)) {
+    lv_obj_set_hidden(btnPrevFlag, false);
+    lv_obj_set_hidden(btnNextFlag, false);
+    lv_obj_set_hidden(btnClose, false);
+    savedTick = lv_tick_get();
+    return true;
+  }
+  if ((event == Pinetime::Applications::TouchEvents::DoubleTap) && (lv_obj_get_hidden(btnClose) == false)) {
```
mark9064:
```suggestion
  if ((event == Pinetime::Applications::TouchEvents::DoubleTap) && !lv_obj_get_hidden(btnClose)) {
```

marigoldfish:
I have been using these new flag watchfaces for a while and have gotten several compliments from queer colleagues! I don't know if I have successfully convinced anyone to buy a Pinetime, but multiple people thought it was cool and wanted me to pass along their compliments. :rainbow_flag: 

I do have some minor readability feedback after test driving these for a month. As usual for my feedback they are EXTREMELY minor, and mostly come down to my taste, so take 'em or leave 'em as you wish :stuck_out_tongue: 

- I feel like the bi flag would have better contrast if the time was white and the date/steps/battery were black.

- I use the rainbow face the least, because the time is the least legible to me. I know this would make it inconsistent with the other faces, but  it would be easier to see if the time were centered on the yellow stripe.

- The time seems off-center vertically by a couple pixels or so. I wouldn't normally be nitpicky about something so small, but it's noticeable in the lesbian flag. Because the middle stripe is so narrow, the time aligns with the bottom but not the top.

Aperture32GLaDOS:
Thanks for the feedback! I definitely agree with your points - I've changed the time label's positioning (it's still not exactly center - it's an off by about half a pixel, but it's not really visibly wrong, even on the lesbian flag), and I've made it so that the bi flag has switched black and white text parts. I tried your suggestion with the rainbow flag, but I feel like what you gain in readability is lost in the formatting - I also tried using different colours for the time, but that didn't really help either.
image
image
image

mark9064:
When this is ready for review let me know ^^
All faces look finished to me visually other than the rainbow flag, the difference in lightness between the yellow and green definitely makes it hard... not sure I have any useful suggestions there
Also, it'd be worth checking how the watchfaces look with AOD, if they looks odd it might be worth tweaking some colours a little to make the AOD presentation nicer

marigoldfish:
I have continued to "field test" these faces and I don't have any more feedback on appearance or alignment or whatnot. 👍🏻 

I personally like the rainbow flag having the time entered on the yellow stripe, but I am willing to defer on matters of opinion. I have a rainbow watch strap so I'm more likely to use the trans or lesbian flag anyway.

That's a good point re AOD - I took a quick look while I was on the bus and they are quite different with it enabled (and I discovered a potential-bug in AOD that I'm going to investigate and write up separately if I can reproduce it). They show the time legibly, but the pride flags lose some of their distinction.

Aperture32GLaDOS:
I think for now it's easier to just remove the rainbow version - maybe in future I can figure out a way of making it look good. For now, I've replaced it with the mlm flag (it was also bugging me, I think if we have the wlw one we've gotta have the mlm one as well).
image

mark9064:
Sure, removing it for now until a better design is settled on makes sense.

For the mlm flag, should the text other than the time be white? Especially at the bottom the contrast looks quite poor

Aperture32GLaDOS:
image
I think I'm gonna have to have the text on top be black and the text on bottom be white

Aperture32GLaDOS:
image
Yeah, that works quite well I think

src/displayapp/screens/WatchFacePrideFlag.h
```diff
@@ -0,0 +1,126 @@
...
+        Utility::DirtyValue<uint32_t> stepCount {};
+        Utility::DirtyValue<bool> notificationState {};
+        Utility::DirtyValue<std::chrono::time_point<std::chrono::system_clock, std::chrono::days>> currentDate;
+        // Must be wrapped in a dirty value, since it is displayed in the day but is updated twice a day
+        Utility::DirtyValue<const char*> ampmChar {"AM"};
+
+        static Pinetime::Controllers::Settings::PrideFlag GetNext(Controllers::Settings::PrideFlag prideFlag);
+        static Pinetime::Controllers::Settings::PrideFlag GetPrevious(Controllers::Settings::PrideFlag prideFlag);
+
+        uint32_t savedTick = 0;
```
mark9064: This changes to FreeRTOS `TickType_t` also

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -0,0 +1,368 @@
...
+#include "displayapp/screens/Symbols.h"
+#include "components/battery/BatteryController.h"
+#include "components/ble/BleController.h"
+#include "components/motion/MotionController.h"
+#include "components/settings/Settings.h"
+
+using namespace Pinetime::Applications::Screens;
+
+namespace {
+  void event_handler(lv_obj_t* obj, lv_event_t event) {
```
mark9064: Should be `PascalCase`

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -0,0 +1,368 @@
...
+  lv_obj_clean(lv_scr_act());
+  delete[] backgroundSections;
+}
+
+bool WatchFacePrideFlag::OnTouchEvent(Pinetime::Applications::TouchEvents event) {
+  if ((event == Pinetime::Applications::TouchEvents::LongTap) && lv_obj_get_hidden(btnClose)) {
+    lv_obj_set_hidden(btnPrevFlag, false);
+    lv_obj_set_hidden(btnNextFlag, false);
+    lv_obj_set_hidden(btnClose, false);
+    savedTick = lv_tick_get();
```
mark9064: Since #2264 we are now using `xTaskGetTickCount()` everywhere for consistency

src/displayapp/screens/WatchFacePrideFlag.h
```diff
@@ -0,0 +1,126 @@
...
+        bool initialized = false;
+
+        lv_obj_t** backgroundSections;
+        char labelTimeColour[8] = "#000000";
+        char defaultTopLabelColour[8] = "#ffffff";
+        char defaultBottomLabelColour[8] = "#ffffff";
```
mark9064: Storing an `lv_color_t` for each would be more compact, but I haven't checked if it's possible to do it this way here. I think it should be though?

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -0,0 +1,368 @@
...
+  themeChanged = true;
+  auto prideFlagAsInt = static_cast<uint8_t>(prideFlag);
+  if (initialized) {
+    for (int i = 0; i < numBackgrounds; i++) {
+      lv_obj_del(backgroundSections[i]);
+    }
+    delete[] backgroundSections;
+  }
+  initialized = true;
+  switch (prideFlagAsInt) {
```
mark9064: It would be more readable here if you switched on the actual enum member instead, as then the names of each flag can be used in the case statements

NeroBurner:
Plesse increase the `settingsVersion` number. we do this everytime a new setting is added

src/displayapp/screens/WatchFacePrideFlag.h
```diff
@@ -0,0 +1,126 @@
+#pragma once
+
+#include <lvgl/src/lv_core/lv_obj.h>
+#include <chrono>
+#include <cstdint>
+#include <memory>
+#include <displayapp/Controllers.h>
```
NeroBurner: please use "" for this incluse. controllers.h is one of our headers

src/displayapp/screens/WatchFacePrideFlag.h
```diff
@@ -0,0 +1,126 @@
...
+
+        static Pinetime::Controllers::Settings::PrideFlag GetNext(Controllers::Settings::PrideFlag prideFlag);
+        static Pinetime::Controllers::Settings::PrideFlag GetPrevious(Controllers::Settings::PrideFlag prideFlag);
+
+        uint32_t savedTick = 0;
+        bool initialized = false;
+
+        lv_obj_t** backgroundSections;
+        char labelTimeColour[8] = "#000000";
+        char defaultTopLabelColour[8] = "#ffffff";
```
NeroBurner: can we move the colors and default values from the header to a private namespace in the CPP file? I'd like to keep the header as clean and minimal as possible

src/displayapp/screens/WatchFacePrideFlag.h
```diff
@@ -0,0 +1,126 @@
...
+
+#include <lvgl/src/lv_core/lv_obj.h>
+#include <chrono>
+#include <cstdint>
+#include <memory>
+#include <displayapp/Controllers.h>
+#include "displayapp/screens/Screen.h"
+#include "displayapp/widgets/StatusIcons.h"
+#include "components/datetime/DateTimeController.h"
+#include "components/ble/BleController.h"
```
NeroBurner: nitpick: I think it is possible to remove the includes for ble and other components as we forward declare them. please try :)

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -0,0 +1,368 @@
+#include <lvgl/lvgl.h>
+#include "displayapp/screens/WatchFacePrideFlag.h"
```
NeroBurner, line 1: please make the watchface header the first include, just feels better to have the corresponding header first

src/displayapp/screens/WatchFacePrideFlag.h
```diff
@@ -0,0 +1,126 @@
...
+        // Must be wrapped in a dirty value, since it is displayed in the day but is updated twice a day
+        Utility::DirtyValue<const char*> ampmChar {"AM"};
+
+        static Pinetime::Controllers::Settings::PrideFlag GetNext(Controllers::Settings::PrideFlag prideFlag);
+        static Pinetime::Controllers::Settings::PrideFlag GetPrevious(Controllers::Settings::PrideFlag prideFlag);
+
+        uint32_t savedTick = 0;
+        bool initialized = false;
+
+        lv_obj_t** backgroundSections;
```
NeroBurner: I don't like manual memory management. It's much too error prone to create a memory leak. please replace with a std::vector or similar

NeroBurner (changes requested):

Aperture32GLaDOS:
Yup, all makes sense. I've implemented the suggested changes, and it still compiles and works

Aperture32GLaDOS:
Oof, I'll sort that out in a bit

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -15,6 +9,24 @@ namespace {
...
+  static constexpr lv_color_t orange = LV_COLOR_MAKE(0xef, 0x76, 0x27);
+  static constexpr lv_color_t lightOrange = LV_COLOR_MAKE(0xff, 0x9b, 0x55);
+  static constexpr lv_color_t lightPurple = LV_COLOR_MAKE(0xd4, 0x61, 0xa6);
+  static constexpr lv_color_t darkPurple = LV_COLOR_MAKE(0xb5, 0x56, 0x90);
+  static constexpr lv_color_t magenta = LV_COLOR_MAKE(0xa5, 0x00, 0x62);
+  static constexpr lv_color_t darkGreen = LV_COLOR_MAKE(0x07, 0x8d, 0x70);
+  static constexpr lv_color_t cyan = LV_COLOR_MAKE(0x26, 0xce, 0xaa);
+  static constexpr lv_color_t lightGreen = LV_COLOR_MAKE(0x98, 0xe8, 0xc1);
+  static constexpr lv_color_t indigo = LV_COLOR_MAKE(0x50, 0x49, 0xcc);
+  static constexpr lv_color_t steelBlue = LV_COLOR_MAKE(0x3d, 0x1a, 0x78);
```
NeroBurner: we're now in a namespace, you can drop the `static` here :)

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -0,0 +1,368 @@
...
+  auto prideFlagAsInt = static_cast<uint8_t>(prideFlag);
+  if (initialized) {
+    for (int i = 0; i < numBackgrounds; i++) {
+      lv_obj_del(backgroundSections[i]);
+    }
+    delete[] backgroundSections;
+  }
+  initialized = true;
+  switch (prideFlagAsInt) {
+    case 0:
```
mark9064: There's a lot of code duplication here between different flags. I'd prefer if it each flag had their key values stored in a data structure somewhere (number of sections, colour, text alignment, text colour) and then there could be one code path implementing the actual logic

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -238,19 +249,16 @@ bool WatchFacePrideFlag::OnButtonPushed() {
 void WatchFacePrideFlag::UpdateScreen(Pinetime::Controllers::Settings::PrideFlag prideFlag) {
   themeChanged = true;
   auto prideFlagAsInt = static_cast<uint8_t>(prideFlag);
-  if (initialized) {
-    for (int i = 0; i < numBackgrounds; i++) {
-      lv_obj_del(backgroundSections[i]);
-    }
-    delete[] backgroundSections;
+  for (int i = 0; i < backgroundSections.size(); i++) {
+    lv_obj_del(backgroundSections[i]);
```
NeroBurner: with the use of `std::vector` we can now get rid of the indexed access and switch to a for-each loop (please check my syntax, didn't compile it :sweat_smile: )


```suggestion
  for (lv_obj_t* backgroundSection: backgroundSections) {
    lv_obj_del(backgroundSection);
```

edit: changed from auto to `lv_obj_t*` to be explicit with the pointer

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -0,0 +1,376 @@
...
+      strncpy(defaultTopLabelColour, "#ffffff", 7);
+      strncpy(defaultBottomLabelColour, "#ffffff", 7);
+      break;
+  }
+}
+
+Pinetime::Controllers::Settings::PrideFlag WatchFacePrideFlag::GetNext(Pinetime::Controllers::Settings::PrideFlag prideFlag) {
+  auto prideFlagAsInt = static_cast<uint8_t>(prideFlag);
+  Pinetime::Controllers::Settings::PrideFlag nextFlag;
+  if (prideFlagAsInt < 3) {
```
mark9064: Use something like `std::ranges::max` to automatically calculate the upper bound

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -238,19 +249,16 @@ bool WatchFacePrideFlag::OnButtonPushed() {
...
+  backgroundSections.clear();
   switch (prideFlagAsInt) {
     case 0:
       numBackgrounds = 7;
-      backgroundSections = new lv_obj_t*[numBackgrounds];
+      backgroundSections.reserve(numBackgrounds);
       for (int i = 0; i < numBackgrounds; i++) {
-        backgroundSections[i] = lv_obj_create(lv_scr_act(), nullptr);
+        backgroundSections.push_back(lv_obj_create(lv_scr_act(), nullptr));
         lv_obj_set_size(backgroundSections[i], LV_HOR_RES, (LV_VER_RES / numBackgrounds) + 1);
```
NeroBurner: the background sections creation has much code duplication. Can we maybe restructure it in a way such that we have a data structure with the colors per case stored as constexpr and use `std::transform` to create the background sections from it

NeroBurner (changes requested):
coming along nicely. Some code duplication I'd like to get rid of remains

NeroBurner:
totally optional: could you update the initial PR description to contain updated images for the currently implemented pride flags?

src/displayapp/screens/WatchFacePrideFlag.h
```diff
@@ -0,0 +1,103 @@
...
+
+        void UpdateScreen(Pinetime::Controllers::Settings::PrideFlag);
+
+      private:
+        Utility::DirtyValue<uint8_t> batteryPercentRemaining {};
+        Utility::DirtyValue<bool> powerPresent {};
+        Utility::DirtyValue<bool> bleState {};
+        Utility::DirtyValue<std::chrono::time_point<std::chrono::system_clock, std::chrono::seconds>> currentDateTime {};
+        Utility::DirtyValue<uint32_t> stepCount {};
+        Utility::DirtyValue<bool> notificationState {};
```
mark9064: These initialisers are redundant, as they are equivalent to the default initialiser

mark9064 (commented):
Looking good! Sounds like my thoughts are pretty similar to @NeroBurner regarding duplication

src/displayapp/screens/WatchFacePrideFlag.h
NeroBurner: maaaaaybe. Do you mean in `static Screens::Screen* Create(AppControllers& controllers) {`? Otherwise they are not used in a function so I thought it should be alright. Maybe a learning opportunity for me here :sweat_smile:

src/displayapp/screens/WatchFacePrideFlag.cpp
NeroBurner: how would that look like? Wouldn't we need to explicitly list all enum values for that to work? or is it really just `std::ranges::max(Pinetime::Controllers::Settings::PrideFlag)`?

src/displayapp/screens/WatchFacePrideFlag.h
mark9064: Looks like you're right that they can be removed here, but they will need to be moved into the .cpp file as their layout has to be known there

Also the BLE controller isn't stored as a reference or pointer, so that's an issue currently

src/displayapp/screens/WatchFacePrideFlag.cpp
mark9064: Yep that works I think, just need to static cast the result

Edit: wait maybe not, might be a bit more involved, the example I was looking at was subtly different. I think it's still doable without too much pain though

src/displayapp/screens/WatchFacePrideFlag.h
```diff
@@ -0,0 +1,103 @@
...
+        lv_obj_t* batteryValue;
+        lv_obj_t* stepValue;
+        lv_obj_t* notificationIcon;
+        lv_obj_t* btnClose;
+        lv_obj_t* btnNextFlag;
+        lv_obj_t* btnPrevFlag;
+
+        Controllers::DateTime& dateTimeController;
+        const Controllers::Battery& batteryController;
+        const Controllers::Ble bleController;
```
NeroBurner: this should be a reference

```suggestion
        const Controllers::Ble& bleController;
```

src/displayapp/screens/WatchFacePrideFlag.h
```diff
@@ -0,0 +1,103 @@
+#pragma once
+
+#include <lvgl/src/lv_core/lv_obj.h>
+#include <chrono>
+#include <vector>
+#include <cstdint>
+#include "displayapp/Controllers.h"
```
NeroBurner: we need the settingscontroller.h as `Controllers::Settings::PrideFlag` is used in the header file as value, so we can't forward declare it here. Just re-add the include for `SettingsController.h`

```suggestion
#include "components/settings/Settings.h"
#include "displayapp/Controllers.h"
```

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -0,0 +1,376 @@
+#include "displayapp/screens/WatchFacePrideFlag.h"
+#include <lvgl/lvgl.h>
+#include "displayapp/screens/Symbols.h"
```
NeroBurner, line 5: includes removed from the header are missing

```suggestion
#include "components/battery/BatteryController.h"
#include "components/ble/BleController.h"
#include "displayapp/screens/Symbols.h"
```

NeroBurner (changes requested):
suggestions to make the build work again

src/displayapp/screens/WatchFacePrideFlag.h
NeroBurner: code suggestions to fix those (and thus the build) in my latest review :)

src/displayapp/screens/WatchFacePrideFlag.cpp
mark9064: OK, while this is possible it requires a lot of metaprogramming currently. C++26 might change this

An alternative would be to add an END enum member, but that's kind of ugly and causes issues with that being a valid entry.

Since my suggestion looks to be (sensibly) unimplementable, I'm resolving this thread

NeroBurner: as far as I can see (or search for) we'd need `Boost.describe` or similar to make this work. C++20 alone doesn't seem to be enough
https://stackoverflow.com/questions/73679033/c20-range-from-enum

mark9064: https://taylorconor.com/blog/enum-reflection/ it does seem to be possible as the number of members in the enum is all we need to know. But it's not pretty :P

NeroBurner: let's wait for C++26 reflection instead :grin:

mark9064: definitely :D

src/displayapp/screens/WatchFacePrideFlag.h
```diff
@@ -30,30 +33,56 @@ namespace Pinetime {
...
+
         bool OnTouchEvent(TouchEvents event) override;
         bool OnButtonPushed() override;
 
         void Refresh() override;
 
         void UpdateSelected(lv_obj_t* object, lv_event_t event);
 
+        template <std::size_t N>
+        void UseFlagData(PrideFlagData<N> flagData);
```
NeroBurner: optional, but please explore if it is easily feasible:

this feels like it should be a private function :thinking: and even an implementation detail. Do you think you can completely move it into the `cpp` file? Maybe add a `WatchFacePrideFlag` reference parameter

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -0,0 +1,321 @@
...
+  bluetoothStatus = lv_label_create(lv_scr_act(), nullptr);
+  lv_label_set_text_static(bluetoothStatus, "");
+  lv_obj_align(bluetoothStatus, nullptr, LV_ALIGN_IN_TOP_RIGHT, -16, 0);
+
+  batteryValue = lv_label_create(lv_scr_act(), nullptr);
+  lv_label_set_recolor(batteryValue, true);
+  lv_label_set_align(batteryValue, LV_LABEL_ALIGN_CENTER);
+  lv_obj_set_auto_realign(batteryValue, true);
+
+  notificationIcon = lv_label_create(lv_scr_act(), nullptr);
```
mark9064: I haven't tested this on hardware and it's difficult to do with InfiniEmu, does the notification text look OK on all of the flags?

Also, maybe it'd be worth renaming this, as it's text now rather than an icon

src/displayapp/screens/WatchFacePrideFlag.h
Aperture32GLaDOS: I think the issue with moving it into the cpp file is that I'd have to either pass all the labels it changes individually as parameters, or make those labels public so I could access them from a WatchFacePrideFlag parameter.

I can definitely make it private, though (and I do agree, it should at least be private, if not a seperate function).

src/displayapp/screens/WatchFacePrideFlag.h
```diff
@@ -0,0 +1,132 @@
...
+        bool OnButtonPushed() override;
+
+        void Refresh() override;
+
+        void UpdateSelected(lv_obj_t* object, lv_event_t event);
+
+        template <std::size_t N>
+        void UseFlagData(PrideFlagData<N> flagData);
+
+        void UpdateScreen(Pinetime::Controllers::Settings::PrideFlag);
```
mark9064, line 44: Should this be private?

src/displayapp/screens/WatchFacePrideFlag.h
NeroBurner: then let's go for private for now and wait for @mark9064 to chime in on this one :)

src/displayapp/screens/WatchFacePrideFlag.cpp
Aperture32GLaDOS: Luckily enough, black has enough contrast on all the current flags to look good (or at least, they do on the emulator)
image
image
image
image
Yeah, it should definitely be renamed though.

src/displayapp/screens/WatchFacePrideFlag.h
mark9064: Private makes more sense to me as it mutates a lot of the watchface state :+1:

src/displayapp/screens/WatchFacePrideFlag.h
```diff
@@ -0,0 +1,132 @@
...
+          constexpr PrideFlagData(const std::array<lv_color_t, N>& sectionColours,
+                                  lv_color_t defaultTopLabelColour,
+                                  lv_color_t labelTimeColour,
+                                  lv_color_t defaultBottomLabelColour) {
+            this->sectionColours = sectionColours;
+            this->defaultTopLabelColour = defaultTopLabelColour;
+            this->labelTimeColour = labelTimeColour;
+            this->defaultBottomLabelColour = defaultBottomLabelColour;
+            // Space between adjacent text values calculated according to the following equation
+            spacing = (uint8_t) (1.5f * ((float) N) + 40.5);
```
mark9064: It would be better if this used C++ style casts rather than C style

Also, should it be 40.5f to avoid double promotion? Or maybe we use the cast all doubles to float compiler option, I don't remember

Aperture32GLaDOS: Yup, makes sense, I'll use 40.5f just to be sure

src/displayapp/screens/WatchFacePrideFlag.h
mark9064: Sorry I didn't make it clear, I was talking about the `UpdateScreen` method

Aperture32GLaDOS: No worries, it's an easy fix anyhow

Aperture32GLaDOS:
Yeah, no worries - I'm flashing it now

Aperture32GLaDOS:
wlw
bi
mlm
trans
The time is legible on all of them, but on the trans watch the day and date are invisible - if it's an issue we could change all text to be black

mark9064:
I don't think it's a big issue - I'll leave whether white or black works better for the trans face to you :)

Aperture32GLaDOS:
In that case, I'd prefer to leave it as-is, with black for the time and white for everything else

src/components/settings/Settings.h
```diff
@@ -319,6 +330,8 @@ namespace Pinetime {
 
         PineTimeStyle PTS;
 
+        PrideFlag prideFlag = PrideFlag::Gay;
```
NeroBurner, line 333: non-blocking, probably I should create a new issue:

@mark9064 should we move watchface specific settings to their own settings files? like alarm does?

mark9064: That's a good point, they probably should. The main settings file should be for system settings only

But I also agree that it is its own issue, other watchface settings are in the main file so this can go there for now until we refactor later

mark9064: Issue created

src/components/datetime/DateTimeController.cpp
```diff
@@ -144,6 +146,10 @@ const char* DateTime::DayOfWeekShortToString() const {
   return DaysStringShort[static_cast<uint8_t>(DayOfWeek())];
 }
 
+const char* DateTime::DayOfWeekToString() const {
+  return DaysString[static_cast<uint8_t>(DayOfWeek())];
+}
```
NeroBurner, line 151: optional, no biggie, honestly I'd like to keep the function: this function is unused. But I'm not sure if the compiler is smart enough to remove it itself from the resulting binary :thinking: 
could you try if the size numbers change with and without this unused function?

src/displayapp/screens/WatchFacePrideFlag.h
```diff
@@ -0,0 +1,132 @@
...
+        Utility::DirtyValue<bool> bleState;
+        Utility::DirtyValue<std::chrono::time_point<std::chrono::system_clock, std::chrono::seconds>> currentDateTime;
+        Utility::DirtyValue<uint32_t> stepCount;
+        Utility::DirtyValue<bool> notificationState;
+        Utility::DirtyValue<std::chrono::time_point<std::chrono::system_clock, std::chrono::days>> currentDate;
+        // Must be wrapped in a dirty value, since it is displayed in the day but is updated twice a day
+        Utility::DirtyValue<const char*> ampmChar {"AM"};
+
+        static Pinetime::Controllers::Settings::PrideFlag GetNext(Controllers::Settings::PrideFlag prideFlag);
+        static Pinetime::Controllers::Settings::PrideFlag GetPrevious(Controllers::Settings::PrideFlag prideFlag);
```
NeroBurner: those two functions could be moved completely into the `cpp` file into the private namespace

src/displayapp/screens/WatchFacePrideFlag.h
```diff
@@ -0,0 +1,132 @@
...
+          constexpr PrideFlagData(const std::array<lv_color_t, N>& sectionColours,
+                                  lv_color_t defaultTopLabelColour,
+                                  lv_color_t labelTimeColour,
+                                  lv_color_t defaultBottomLabelColour) {
+            this->sectionColours = sectionColours;
+            this->defaultTopLabelColour = defaultTopLabelColour;
+            this->labelTimeColour = labelTimeColour;
+            this->defaultBottomLabelColour = defaultBottomLabelColour;
+            // Space between adjacent text values calculated according to the following equation
+            spacing = static_cast<uint8_t>(1.5f * static_cast<float>(N) + 40.5f);
```
NeroBurner: even if it is just syntactic sugar, please use member initializers as we do for the `WatchFacePrideFlag` constructor (the member variables after the `:` symbol)

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -0,0 +1,321 @@
...
+  stepValue = lv_label_create(lv_scr_act(), nullptr);
+  lv_label_set_recolor(stepValue, true);
+  lv_obj_align(stepValue, lv_scr_act(), LV_ALIGN_CENTER, 0, 96);
+  lv_label_set_align(stepValue, LV_LABEL_ALIGN_CENTER);
+  lv_obj_set_auto_realign(stepValue, true);
+
+  UpdateScreen(settingsController.GetPrideFlag());
+
+  taskRefresh = lv_task_create(RefreshTaskCallback, LV_DISP_DEF_REFR_PERIOD, LV_TASK_PRIO_MID, this);
+  Refresh();
```
NeroBurner, line 167: CLion warning: Call to a virtual function inside a constructor is resolved at compile time

suggested fix: be explicit
```suggestion
  WatchFacePrideFlag::Refresh();
```

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -0,0 +1,321 @@
...
+    savedTick = xTaskGetTickCount();
+    return true;
+  }
+  if ((event == Pinetime::Applications::TouchEvents::DoubleTap) && !lv_obj_get_hidden(btnClose)) {
+    return true;
+  }
+  return false;
+}
+
+void WatchFacePrideFlag::CloseMenu() {
```
NeroBurner, line 189: optional: CLion suggests that `CloseMenu` member function can be made `const`
It doesn't feel like it should be `const` as we change the screen and hide some buttons

Thoughts?

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -0,0 +1,321 @@
...
+  }
+  if (themeChanged)
+    themeChanged = false;
+}
+
+void WatchFacePrideFlag::UpdateSelected(lv_obj_t* object, lv_event_t event) {
+  auto valueFlag = settingsController.GetPrideFlag();
+  bool flagChanged = false;
+
+  if (event == LV_EVENT_CLICKED) {
```
NeroBurner: variables can be moved into the if, then we call `GetPrideFlag()` only if we need it
```suggestion
  if (event == LV_EVENT_CLICKED) {
    auto valueFlag = settingsController.GetPrideFlag();
    bool flagChanged = false;

```

src/displayapp/screens/WatchFacePrideFlag.h
```diff
@@ -0,0 +1,132 @@
...
+        Utility::DirtyValue<const char*> ampmChar {"AM"};
+
+        static Pinetime::Controllers::Settings::PrideFlag GetNext(Controllers::Settings::PrideFlag prideFlag);
+        static Pinetime::Controllers::Settings::PrideFlag GetPrevious(Controllers::Settings::PrideFlag prideFlag);
+
+        TickType_t savedTick = 0;
+
+        std::vector<lv_obj_t*> backgroundSections;
+        bool themeChanged = false;
+        uint8_t numBackgrounds;
```
NeroBurner: unused, replaced by `PrideFlagData::N`
```suggestion
```

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -0,0 +1,321 @@
...
+    } else {
+      lv_label_set_text_static(notificationText, "");
+    }
+  }
+
+  currentDateTime = std::chrono::time_point_cast<std::chrono::seconds>(dateTimeController.CurrentDateTime());
+  if (currentDateTime.IsUpdated() || themeChanged) {
+    uint8_t hour = dateTimeController.Hours();
+    uint8_t minute = dateTimeController.Minutes();
+    uint8_t second = dateTimeController.Seconds();
```
NeroBurner: clang-tidy: can be declared const

```suggestion
    uint8_t hour = dateTimeController.Hours();
    const uint8_t minute = dateTimeController.Minutes();
    const uint8_t second = dateTimeController.Seconds();
```

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    lv_label_set_text_fmt(labelTime, "%02d:%02d:%02d", hour, minute, second);
+
+    currentDate = std::chrono::time_point_cast<std::chrono::days>(currentDateTime.Get());
+    if (currentDate.IsUpdated() || ampmChar.IsUpdated() || themeChanged) {
+      uint16_t year = dateTimeController.Year();
+      Controllers::DateTime::Months month = dateTimeController.Month();
+      uint8_t day = dateTimeController.Day();
+      Controllers::DateTime::Days dayOfWeek = dateTimeController.DayOfWeek();
```
NeroBurner:
```suggestion
      const uint16_t year = dateTimeController.Year();
      const Controllers::DateTime::Months month = dateTimeController.Month();
      const uint8_t day = dateTimeController.Day();
      const Controllers::DateTime::Days dayOfWeek = dateTimeController.DayOfWeek();
```

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -0,0 +1,321 @@
...
+      }
+    }
+  }
+
+  stepCount = motionController.NbSteps();
+  if (stepCount.IsUpdated() || themeChanged) {
+    lv_label_set_text_fmt(stepValue, "%lu steps", stepCount.Get());
+  }
+  if (themeChanged)
+    themeChanged = false;
```
NeroBurner: we should use braces here, too many oopsies because of no braces and later just added one command and forgot to add the braces later
```suggestion
  if (themeChanged) {
    themeChanged = false;
  }
```

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -0,0 +1,321 @@
...
+      UseFlagData<biFlagData.numSections>(biFlagData);
+      break;
+    case Pinetime::Controllers::Settings::PrideFlag::Lesbian:
+      UseFlagData<lesbianFlagData.numSections>(lesbianFlagData);
+      break;
+  }
+}
+
+Pinetime::Controllers::Settings::PrideFlag WatchFacePrideFlag::GetNext(Pinetime::Controllers::Settings::PrideFlag prideFlag) {
+  auto prideFlagAsInt = static_cast<uint8_t>(prideFlag);
```
NeroBurner: add const

```suggestion
  const auto prideFlagAsInt = static_cast<uint8_t>(prideFlag);
```

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -0,0 +1,321 @@
...
+  if (prideFlagAsInt < 3) {
+    nextFlag = static_cast<Controllers::Settings::PrideFlag>(prideFlagAsInt + 1);
+  } else {
+    nextFlag = static_cast<Controllers::Settings::PrideFlag>(0);
+  }
+  return nextFlag;
+}
+
+Pinetime::Controllers::Settings::PrideFlag WatchFacePrideFlag::GetPrevious(Pinetime::Controllers::Settings::PrideFlag prideFlag) {
+  auto prideFlagAsInt = static_cast<uint8_t>(prideFlag);
```
NeroBurner: more const

```suggestion
  const auto prideFlagAsInt = static_cast<uint8_t>(prideFlag);
```

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -0,0 +1,321 @@
...
+    CloseMenu();
+    return true;
+  }
+  return false;
+}
+
+template <std::size_t N>
+void WatchFacePrideFlag::UseFlagData(PrideFlagData<N> flagData) {
+  backgroundSections.reserve(N);
+  for (uint8_t i = 0; i < N; i++) {
```
NeroBurner: let's match the index variable with the type of the conditional `N`

```suggestion
  for (std::size_t i = 0; i < N; i++) {
```

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -0,0 +1,321 @@
...
+  lv_obj_set_style_local_text_color(labelDate, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, flagData.defaultTopLabelColour);
+  lv_obj_set_style_local_text_color(labelDay, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, flagData.defaultBottomLabelColour);
+  lv_obj_set_style_local_text_color(stepValue, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, flagData.defaultBottomLabelColour);
+  lv_obj_align(batteryValue, lv_scr_act(), LV_ALIGN_CENTER, 0, -2 * flagData.spacing);
+  lv_obj_align(labelDate, lv_scr_act(), LV_ALIGN_CENTER, 0, -1 * flagData.spacing);
+  lv_obj_align(labelDay, lv_scr_act(), LV_ALIGN_CENTER, 0, flagData.spacing);
+  lv_obj_align(stepValue, lv_scr_act(), LV_ALIGN_CENTER, 0, 2 * flagData.spacing);
+}
+
+void WatchFacePrideFlag::UpdateScreen(Pinetime::Controllers::Settings::PrideFlag prideFlag) {
```
NeroBurner:
```suggestion
void WatchFacePrideFlag::UpdateScreen(const Pinetime::Controllers::Settings::PrideFlag prideFlag) {
```

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -0,0 +1,321 @@
...
+bool WatchFacePrideFlag::OnButtonPushed() {
+  if (!lv_obj_get_hidden(btnClose)) {
+    CloseMenu();
+    return true;
+  }
+  return false;
+}
+
+template <std::size_t N>
+void WatchFacePrideFlag::UseFlagData(PrideFlagData<N> flagData) {
```
NeroBurner: The function can be moved into the `UpdateScreen` function as a lambda with the following declaration:

```cpp
  auto UseFlagData = [this]<size_t N>(PrideFlagData<N> flagData) {
```

with this the `PrideFlagData` class can be moved from the header to `cpp` file as well

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -0,0 +1,321 @@
...
+  constexpr lv_color_t cyan = LV_COLOR_MAKE(0x26, 0xce, 0xaa);
+  constexpr lv_color_t lightGreen = LV_COLOR_MAKE(0x98, 0xe8, 0xc1);
+  constexpr lv_color_t indigo = LV_COLOR_MAKE(0x50, 0x49, 0xcc);
+  constexpr lv_color_t steelBlue = LV_COLOR_MAKE(0x3d, 0x1a, 0x78);
+  constexpr std::array<lv_color_t, 7> gayColours {darkGreen, cyan, lightGreen, LV_COLOR_WHITE, lightBlue, indigo, steelBlue};
+  constexpr std::array<lv_color_t, 5> transColours {lightBlue, lightPink, LV_COLOR_WHITE, lightPink, lightBlue};
+  constexpr std::array<lv_color_t, 5> biColours {hotPink, hotPink, grayPurple, darkBlue, darkBlue};
+  constexpr std::array<lv_color_t, 7> lesbianColours {LV_COLOR_RED, orange, lightOrange, LV_COLOR_WHITE, lightPurple, darkPurple, magenta};
+  constexpr WatchFacePrideFlag::PrideFlagData<7> gayFlagData =
+    WatchFacePrideFlag::PrideFlagData(gayColours, LV_COLOR_BLACK, LV_COLOR_BLACK, LV_COLOR_WHITE);
```
NeroBurner: with the moved `PrideFlagData` class we can save a few typed redundant characters here. The `N` template parameter is taken from the `xxxColours` parameter

```suggestion
  constexpr PrideFlagData gayFlagData(gayColours, LV_COLOR_BLACK, LV_COLOR_BLACK, LV_COLOR_WHITE);
```

src/components/datetime/DateTimeController.cpp
mark9064: Linker stripping is on so this should be OK, all unreferenced functions aren't linked in to the binary

src/displayapp/screens/WatchFacePrideFlag.cpp
mark9064: Note: this is not a functionality change but just highlights that the `Refresh()` from the base type (here `WatchFacePrideFlag`) will always be executed instead of `Refresh()`  from an inheriting type

Personally I don't think this change is needed as it breaks convention with every other app and watchface, but equally it's worth knowing why it's been suggested (I learnt something new today!!)

src/displayapp/screens/WatchFacePrideFlag.cpp
mark9064: Sidenote: this is fixed by #2139

src/displayapp/screens/WatchFacePrideFlag.cpp
mark9064: I think this shouldn't be const: the method mutates members even if the compiler doesn't see this as LVGL is a C library

NeroBurner: I agree, you've put into words what I vaguely thought :)

src/components/datetime/DateTimeController.cpp
NeroBurner: TIL

src/displayapp/screens/WatchFacePrideFlag.cpp
Aperture32GLaDOS: Damn, that's a clever usage of lambda functions!

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -31,14 +75,10 @@ namespace {
...
   constexpr std::array<lv_color_t, 7> lesbianColours {LV_COLOR_RED, orange, lightOrange, LV_COLOR_WHITE, lightPurple, darkPurple, magenta};
-  constexpr WatchFacePrideFlag::PrideFlagData<7> gayFlagData =
-    WatchFacePrideFlag::PrideFlagData(gayColours, LV_COLOR_BLACK, LV_COLOR_BLACK, LV_COLOR_WHITE);
-  constexpr WatchFacePrideFlag::PrideFlagData<5> transFlagData =
-    WatchFacePrideFlag::PrideFlagData(transColours, LV_COLOR_WHITE, LV_COLOR_BLACK, LV_COLOR_WHITE);
-  constexpr WatchFacePrideFlag::PrideFlagData<5> biFlagData =
-    WatchFacePrideFlag::PrideFlagData(biColours, LV_COLOR_BLACK, LV_COLOR_WHITE, LV_COLOR_BLACK);
-  constexpr WatchFacePrideFlag::PrideFlagData<7> lesbianFlagData =
-    WatchFacePrideFlag::PrideFlagData(lesbianColours, LV_COLOR_WHITE, LV_COLOR_BLACK, LV_COLOR_WHITE);
+  constexpr PrideFlagData gayFlagData = PrideFlagData(gayColours, LV_COLOR_BLACK, LV_COLOR_BLACK, LV_COLOR_WHITE);
```
NeroBurner: you can use the constructor directly here to save some typing work

```suggestion
  constexpr PrideFlagData gayFlagData(gayColours, LV_COLOR_BLACK, LV_COLOR_BLACK, LV_COLOR_WHITE);
```

src/displayapp/screens/WatchFacePrideFlag.cpp
```diff
@@ -0,0 +1,337 @@
...
+    return prevFlag;
+  }
+
+  template <std::size_t N>
+  class PrideFlagData {
+  public:
+    constexpr PrideFlagData(const std::array<lv_color_t, N>& sectionColours,
+                            lv_color_t defaultTopLabelColour,
+                            lv_color_t labelTimeColour,
+                            lv_color_t defaultBottomLabelColour)
```
NeroBurner, line 43: more const I've overlooked until now

```suggestion
                            const lv_color_t defaultTopLabelColour,
                            const lv_color_t labelTimeColour,
                            const lv_color_t defaultBottomLabelColour)
```

NeroBurner (changes requested):
just one last `const`, and then I think we're good to go!

It's been lovely working with you on this MR @Aperture32GLaDOS

src/displayapp/screens/WatchFacePrideFlag.cpp
mark9064: I have to admit, I think you've gone a little overboard with consting :P my opinion aligns quite well with this blog https://quuxplusone.github.io/blog/2022/01/23/dont-const-all-the-things

Admittedly I much prefer the default immutability of rust

NeroBurner: he brings up very good reasoning. I just followed along with the CLion suggestions as I thought those are best practice. Marking as resolved

NeroBurner (approved):
all looking good. Thanks for the contribution and your great teamwork! I had fun reviewing and learning on this PR

Aperture32GLaDOS:
Brilliant! Thanks all for the advice and feedback, it's been very helpful

PR #2203, opened 2024-12-18
Adjust build instructions

zyphlar:
`cd build` before `cmake` seems to cause issues, as does a cmake command with insufficient details. Finally, many end users and developers will want to OTA flash their firmware, so handy DFU instructions in the correct folder might be appreciated.

### DIFF
--- a/doc/buildAndProgram.md
+++ b/doc/buildAndProgram.md
@@ -29,7 +29,6 @@ git clone https://github.com/InfiniTimeOrg/InfiniTime.git
 cd InfiniTime
 git submodule update --init
 mkdir build
-cd build
 ```
 
 ### Project generation using CMake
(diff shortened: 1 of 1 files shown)


doc/buildAndProgram.md
```diff
@@ -85,9 +86,19 @@ If you just want to build the project and run it on the Pinetime, using *pinetim
 Build:
 
 ```
+cd build
 make -j4 pinetime-app
 ```
 
+However if you want to flash the firmware OTA using an app like Gadgetbridge, using *pinetime-mcuboot-app* will create the `pinetime-mcuboot-app-dfu-$VERSION.zip` file you want in the `build/src` folder.
```
mark9064, line 93: This is a great addition. Maybe modify the above L84 to specify "PineTime devkit"? Currently these instructions kind of assume a devkit, whereas I think a lot of people work on sealed now as the bootloader is very stable

doc/buildAndProgram.md
```diff
@@ -65,8 +64,10 @@ DFU files are the files you'll need to install your build of InfiniTime using OT
 
 #### CMake command 
 
+ARM_NONE_EABI_TOOLCHAIN_PATH and NRF5_SDK_PATH must be specified (see above):
+
 ```
-cmake -DARM_NONE_EABI_TOOLCHAIN_PATH=... -DNRF5_SDK_PATH=...
+cmake -S "." -B "build" -DCMAKE_BUILD_TYPE=Release -DARM_NONE_EABI_TOOLCHAIN_PATH=... -DNRF5_SDK_PATH=.. -DBUILD_DFU=1 -DBUILD_RESOURCES=1 -DNRF_TARGET=nrf52
```
mark9064, line 70: `-DNRF_TARGET=nrf52` isn't needed, it's already specified in the cmake file. Release is also the default type, so I don't think that needs to be specified either (it's listed above, and I think people who need debug will know when they need it, you have to be pretty involved at that point)

As for DFU/resources, how about giving two example commands, one which builds neither (as current) and one which builds both, I think that way users can easily get the idea of what the options do

mark9064 (commented):
Thanks for the contribution!

So you might already know but there is #2157 open which suggests some similar changes cmake wise. Any advantage to doing cmake with `-B`?

PR #2207, opened 2024-12-23
DirtyValue: add possibility to mark DirtyValue updated

ljahn:
Adds the ability to manually mark a DirtyValue as updated.

This simplifies situations where a the user (or any other condition) selects if a value should be updated.

Example: In my Star Trek watch face I give the possibility to show weather or not. Using this function, I do not have to duplicate the weather updating code to the button press handler or introduce another variable `weatherNeedsUpdate`. I can just make the corresponding element visible, mark the weather data as updated (although it really isn't), and the next  normal refresh loop will put the correct weather information.

### DIFF
--- a/src/utility/DirtyValue.h
+++ b/src/utility/DirtyValue.h
@@ -18,6 +18,10 @@ namespace Pinetime {
         return false;
       }
 
+      void MarkUpdated() {
+        this->isUpdated = true;
+      }
+
       T const& Get() {
         this->isUpdated = false;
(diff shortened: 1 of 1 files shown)


mark9064:
I don't love this conceptually as it kind of breaks the idea that the DirtyValue is a state container that informs of updates only on state changes. However, I'm not dead set on not including this

For your use-case, could you update the weather irrespective of whether it's visible or not?

Issue #2082, opened 2024-06-21
Use separate screens for switching off ringing alarm and for alarm configuration


kyeastmood:
### Verification

- [X] I searched for similar bug reports (including closed issues) and found none was relevant.

### What happened?

Its too easy to disable the alarm completely instead of just switching it off when the alarm is ringing

### What should happen instead?

There should be separate screens for alarm configuration and for swithining off a ringing alarm

### Reproduction steps

Currently, when the alarm is ringing you can push the hardware button or the red button to switch the alarm off (so the button will turn green). But then it's easy to push the button again by accident (especially when being in bed and not looking at the watch) which will disable the alarm completely and it won't ring the next day.

### More details?

Solution:  use two separate screens for dealing with the alarm:
1. For configuring the alarm (enable/disable/set time/set recurrence)
2. For switching off or snoozing a ringing alarm

### Version

1.14

kyeastmood:
Hi @ljahn, thank you for working on this problem! It is the most annoying issue in InfiniTime for me.

tituscmd:
If I understand correctly, this should be fixed by #2211

ljahn:
Yes, it is.  
This issue wasn't auto closed when my PR was merged, because the fix for this was the result of dealing with another issue. Now there are two separate screens: One for configuration, and one when alerting, where one can only press Stop, nothing more.  

The only remaining this that is not covered is the mention of a snooze function. That is not implemented yet, but could well be its own issue/feature request.

mark9064:
Agree, closing this as solved as the headline issue is resolved. For snooze discussion I think a new issue / discussion post would be best

PR #2211, opened 2024-12-28
Alarm: Simplify alarm alerting screen (and fix bug with alerting on time value change)

### LINKED ISSUE #2206: Setting time while alarm is going off reboots watch upon new alarm going off
### Verification

### What happened?

I choose a later time while the alarm is going off to act as a sooze function.  However, it now reboots upon the 'snooze' alarm triggering.

In previous versions, the 'snooze' worked with no reboots. I would get the viberation, select 10 to 15 mins later, and dismiss the alarm. The alarm would then wake me up at that later date. 

### Reproduction steps

1-Set Alarm
2-As alarm going off, change to a later time.
3- Dismiss alarm
4- When new alarm is to go off, watch reboots and alarm seems to retain that time but will go off the follow day.

### More details?

It's worth noting that I had initially thought this was an intended feature as I was awake and the alarm must be for the next day as it would the following day. 

Also worth noting, that this works as expected if I 'silence' the alarm then set a new time.  No reboot occurs. 

I was told this shouldn't matter, but just in case it's relevant, I haven't flashed a new resource file since for a few versions. 

### Version

V1.15.0 

### Companion app

Gadgebridge 0.82.1 (f-droid)

ljahn:
Fixes #2206

When Alerting, the handlers for swiping and hard- and software buttons correctly deal with this.
The handler for time value changes did not.

~~Now a value change will also stop the alert.
For the scenario presented in #2206 (makeshift snooze), this means that after setting a new time, the new alarm also needs to be enabled again. However, I think this is not a problem, because the UI clearly presents that the alarm is disabled and it can be re-enabled exactly like setting an alarm always works.~~

Edit: Now fixing the problem by making the value change impossible in the first place by simplifying the alerting screen.

### DIFF
--- a/src/components/alarm/AlarmController.cpp
+++ b/src/components/alarm/AlarmController.cpp
@@ -111,7 +111,6 @@ uint32_t AlarmController::SecondsToAlarm() const {
 
 void AlarmController::DisableAlarm() {
   xTimerStop(alarmTimer, 0);
-  isAlerting = false;
   if (alarm.isEnabled) {
     alarm.isEnabled = false;
     alarmChanged = true;
--- a/src/displayapp/screens/Alarm.cpp
+++ b/src/displayapp/screens/Alarm.cpp
@@ -77,7 +77,7 @@ Alarm::Alarm(Controllers::AlarmController& alarmController,
   btnStop = lv_btn_create(lv_scr_act(), nullptr);
   btnStop->user_data = this;
   lv_obj_set_event_cb(btnStop, btnEventHandler);
-  lv_obj_set_size(btnStop, 115, 50);
+  lv_obj_set_size(btnStop, 240, 70);
   lv_obj_align(btnStop, lv_scr_act(), LV_ALIGN_IN_BOTTOM_LEFT, 0, 0);
   lv_obj_set_style_local_bg_color(btnStop, LV_BTN_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_RED);
   txtStop = lv_label_create(btnStop, nullptr);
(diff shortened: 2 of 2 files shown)


mark9064:
Though I hate to break the makeshift snooze, I think it's the right way to go here as the UI should be consistent with disabling the alarm anytime the parameters are changed

I'd like to understand why it's currently broken though - did you manage to get to the bottom of this?

ljahn:
I think the reason is this: 
`SetAlerting` creates an `lv_task` to automatically call `StopAlerting` after one minute. This task will call an invalid function reference and lead to a crash under the following condition:

All exit paths but the time value change (so not considering this fix) call `StopAlerting` themselves, which also terminates the `lv_task`. However, the value change callback only calls `DisableAlarm`, because its normal use case is for setting up an alarm, where you have to re-confirm enabling the alarm after every change you make. `DisableAlarm` still sets `isAlerting` in the `alarmController` to `false`, probably because someone thought a currently alerting but also disabled alarm makes no sense, this was introduced in a0cd439. That causes the destructor of `Alarm` to think there is nothing to do regarding the alerting when the alarm screen is dismissed. Therefore it does not call `StopAlerting` and the `lv_task` is left with an invalid function pointer, because `Alarm` does not exist anymore once the `lv_task` finally goes to call the callback function.

ljahn:
I also think we should do what you said and fully disable the time change buttons while the alarm is ringing. While we are at it, lets go one step further and disable all other buttons as well. The info button does not do anything useful in that situation anyway, it shows some meaningless value, and why should you be able to change the recurrence when you are not allowed to change the time. So in line with #2082, I would make the alarm disable screen have only one button, the `Stop` button.

I updated the PR to reflect this. This is what it looks like:
InfiniSim_2024-12-29_001905

Pressing the big red or the physical button brings you back to the normal alarm settings display and stops the alarm.

#2082 also calls for a snooze function, but this leads to additional questions that would warrant an own pull request and discussion.

mark9064:
Really nice analysis of the root cause, thanks for writing it up :)

The new UI looks good to me, agree that snooze can go in a separate PR (and the stop button could easily be half width with snooze for the other half, so that should work)

Code at a glance looks sensible, I should be able to review it properly soon

mark9064 (approved):

NeroBurner (approved):
Looking good. Thanks for the UI improvements and fixing the bug introduced by me 🙇

PR #2215, opened 2024-12-31
Add a fast mode to Counter where more than one is skipped on long press

ljahn:
Oftentimes I cannot be bothered to hold the + button of the timer down until 30s is reached, because it takes so long. Instead, I just set 30s less and wait around a bit after the timer rings.
When setting an alarm, I have the same problem with the minutes. It just takes too long. 

This adds a 'fast mode' to the `Counter` widget that is used in these cases.
An application designer can enable this for a widget and is then able to specify how many steps should be skipped on each trigger of the `LV_EVENT_LONG_PRESSED_REPEAT` event.

This PR enables this with value 5 for `Timer`, `Alarm` (minutes only) and `SettingDateTime` (days and minutes only).

Also includes a small fix for `Counter` where the `ValueChangedHandler` was not called if `SetValue` was used. I don't think there were any bugs caused by this yet, I checked all uses of the function.


ljahn:
The value of 5 I selected for the fast skip might be too high, maybe 3 would be better.
I will evaluate this in my daily usage.

PR #2218, opened 2025-01-04
Create a new `Color` method in the `Temperature` class in `SimpleWeatherService` for easier reusability.

JustScott:
It was brought up in #2204 to move the `TemperatureColor` function to the `Temperature` class in `SimpleWeatherService` as an easy to reuse method, however since then it has been decided to use a static temperature color in the terminal watch face  instead which means the `TemperatureColor` function is only used in one file (weather.cpp) making it currently "pointless" to move this function. However, I'm still creating this PR in case others think we should still move this function to the `Color` method for easy reuse in the future.


mark9064 (approved):
I think moving it still makes sense either way, thanks for splitting out this change :)

PR #2221, opened 2025-01-07
Star Trek Watchface (second attempt)

ljahn:
This is the same watch face as #1462. That PR has gone lengthy and opaque so I decided to start a fresh one.

I made a watch face inspired by Star Trek.

Here's a demonstration of its capabilities recorded with InfiniSim (the visible artifacts come from the GIF, not from the watch face)
InfiniSim_2025-01-07_183127

The features are:
Shows date and time (always) + selectable seconds display + two different fonts (Star Trek font only with resource pack)
Weather (can be turned off)
Heartbeat (only when active)
Steps (with nice progress bar for step goal)
AM/PM (not shown in GIF)
Animations: User selectable: None, Continuous color changes, Startup (visibility of objects scrolls down from the top), or both
Battery level with color for low battery
Alarm and timer status (timer takes precedence)
Bluetooth connection
Charging and notification status (charging takes precedence, normal vibrating mode is not shown)
Color scheme adopts to the "Always On Display" mode to increase visibility in reduced color space mode.

The watch face is usable without external resources, but then only the system font can be used for the time.
This watch face is quite demanding on the hardware but I did some optimizations, such that even when everything is visible (watch face with everything and menu overlay buttons), there is still more than 3000 heap space left.

There are three things that this PR changes in existing parts of InfiniTime, which I separated out in three small pull requests #2207, #2208 and #2209. I cherry picked those commits  into this branch so there should be no merge conflicts later.

Here's a compiled version for those who want to test/use this but can't be bothered to compile: 
[pinetime-mcuboot-app-dfu-1.15.0.zip](https://github.com/user-attachments/files/18336836/pinetime-mcuboot-app-dfu-1.15.0.zip)


romibi:
First:
> Star Trek font only with resource pack

Setup instructions would be nice.

Some more remarks:
- I'd prefer animations tuned back very much (like 1 change every 30s or so)
- I'd love some more color variants
- I had multiple crashes (sudden pine cone boot screen) while trying to get the font to load … but as i only have my watch since a bit more than a day i have no idea how to debug

I'm very tempted to use this watchface but currently I prefer the mint colored infineat face a tad more

ljahn:
@romibi 
> First:
> 
> > Star Trek font only with resource pack
> 
> Setup instructions would be nice.

Adding resources is a common procedure, it is described [here](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/doc/gettingStarted/updating-software.md) (scroll down a bit). I realize though, that I did not publish the necessary resource file, so If you did not self compile, you could not have it. I added it above.

> I'd prefer animations tuned back very much (like 1 change every 5 to 30s or so … not sure yet)

I thought about animation tempo and I might dial it back just a little bit, but I never look at the watch for too long, so 5s or more makes no sense in my opinion, then you could just turn the animation off (which you can by long pressing the watch face and then picking Static or Startup).

> I'd love some more color variants
I did some preparations and worked towards this, but the code needs some more changes for this feature. Currently I have no intent to work on this.

> I had multiple crashes (sudden pine cone boot screen) while trying to get the font to load … but as i only have my watch since a bit more than a day i have no idea how to debug
That is very interesting. I would like to know the exact circumstances to fix this.

romibi:
> Adding resources is a common procedure, it is described [here](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/doc/gettingStarted/updating-software.md) (scroll down a bit). I realize though, that I did not publish the necessary resource file, so If you did not self compile, you could not have it. I added it above.

I tried to just put the font .ttf into the fonts folder using https://infinitimeexplorer.netlify.app/ either with the name EdgeOfTheGalaxyRegular.ttf or EdgeOfTheGalaxyRegular.bin … which did not work …
But at least with that web app opening and uploading the resource file zip from above still does not make the font selection work …
And gadgetbridge does not recognize the resource file as an loadable file.
Shouldn't there be an EdgeOfTheGalaxyRegular file in that zip?

> > I'd prefer animations tuned back very much (like 1 change every 5 to 30s or so … not sure yet)
> 
> I thought about animation tempo and I might dial it back just a little bit, but I never look at the watch for too long, so 5s or more makes no sense in my opinion, then you could just turn the animation off (which you can by long pressing the watch face and then picking Static or Startup).

Yes 30s probably would only make sense with the always on display … 
But I'd like variation but no or rarely changes while actively looking at it …
… the way it is now I prefer Static. (Even Startup is too much for my taste)

> > I had multiple crashes (sudden pine cone boot screen) while trying to get the font to load … but as i only have my watch since a bit more than a day i have no idea how to debug
> 
> That is very interesting. I would like to know the exact circumstances to fix this.

Most of the times it happened when I was switching watchfaces. (to and from yours) 
Once it happened in another situation. I'm not sure anymore what i was doing. I think i was trying to open the weather app thingy.

ljahn:
> Shouldn't there be an EdgeOfTheGalaxyRegular file in that zip?

You are right, I uploaded the standard resource pack for the official watch faces. I corrected that mistake. Now it should work.
But make sure you remove everything you manually added to the file system.

> Most of the times it happened when I was switching watchfaces. (to and from yours)
> Once it happened in another situation. I'm not sure anymore what i was doing. 

Thanks, but that is very general, could be anything. Was this after of while you were copying random files to the watch to make the font work, or already before? I once had a bug with a double free on fonts, but I fixed that.

romibi:
> You are right, I uploaded the standard resource pack for the official watch faces. I corrected that mistake. Now it should work. But make sure you remove everything you manually added to the file system.

Thanks, now it works. And yes I've removed the wrong files again.

> > Most of the times it happened when I was switching watchfaces. (to and from yours)
> > Once it happened in another situation. I'm not sure anymore what i was doing.
> 
> Thanks, but that is very general, could be anything. Was this after of while you were copying random files to the watch to make the font work, or already before? I once had a bug with a double free on fonts, but I fixed that.

Now that you mention it: I think every time it happened I was connected to that File Explorer App … It can't be because of wrong files because it also happened after cleaning up.
But yes … It might be unrelated to your changes …

ljahn:
My second PineTime broke and I think I will not be buying a new one, so I'll close this and all related PRs, because I have no intention to maintain this in the future.
I really liked the project and tinkering with the watch, and wish you all the best of luck.

mark9064:
Shame to hear that it broke :( thanks for all of your contributions :heart:

PR #2222, opened 2025-01-09
alarm: Long press to stop alarm

vkareh:
This change prevents accidentally turning off alarm by ensuring that there is a deliberate long press, similar to resetting the Timer app.

InfiniSim_2025-03-27_213657

### DIFF
--- a/src/displayapp/screens/Alarm.cpp
+++ b/src/displayapp/screens/Alarm.cpp
@@ -74,6 +74,14 @@ Alarm::Alarm(Controllers::AlarmController& alarmController,
   lv_label_set_text_static(colonLabel, ":");
   lv_obj_align(colonLabel, lv_scr_act(), LV_ALIGN_CENTER, 0, -29);
 
+  progressStop = lv_bar_create(lv_scr_act(), nullptr);
+  lv_bar_set_range(progressStop, 0, progressBarSize);
+  lv_bar_set_value(progressStop, 0, LV_ANIM_OFF);
+  lv_obj_set_size(progressStop, progressBarSize, 10);
+  lv_obj_align(progressStop, nullptr, LV_ALIGN_IN_TOP_MID, 0, 0);
+  lv_obj_set_style_local_bg_color(progressStop, LV_BAR_PART_INDIC, LV_STATE_DEFAULT, LV_COLOR_ORANGE);
+  lv_obj_set_hidden(progressStop, true);
+
   btnStop = lv_btn_create(lv_scr_act(), nullptr);
   btnStop->user_data = this;
   lv_obj_set_event_cb(btnStop, btnEventHandler);
@@ -122,12 +130,28 @@ Alarm::Alarm(Controllers::AlarmController& alarmController,
   } else {
     SetSwitchState(LV_ANIM_OFF);
   }
+
+  taskRefresh = lv_task_create(RefreshTaskCallback, 50, LV_TASK_PRIO_MID, this);
+}
+
+void Alarm::Refresh() {
+  if (stopBtnPressTime.has_value()) {
+    TickType_t elapsed = xTaskGetTickCount() - stopBtnPressTime.value();
+    if (elapsed >= longPressTimeout) {
+      ResetStopProgress();
+      StopAlerting();
+    } else {
+      lv_coord_t stopPosition = (elapsed * progressBarSize) / longPressTimeout;
+      UpdateStopProgress(stopPosition);
+    }
+  }
 }
 
 Alarm::~Alarm() {
   if (alarmController.IsAlerting()) {
     StopAlerting();
   }
+  lv_task_del(taskRefresh);
   lv_obj_clean(lv_scr_act());
   alarmController.SaveAlarm();
 }
@@ -139,12 +163,32 @@ void Alarm::DisableAlarm() {
   }
 }
 
+void Alarm::StopButtonPressed() {
+  stopBtnPressTime = xTaskGetTickCount();
+  UpdateStopProgress(0);
+  lv_obj_set_hidden(progressStop, false);
+}
+
+void Alarm::ResetStopProgress() {
+  stopBtnPressTime = std::nullopt;
+  lv_obj_set_hidden(progressStop, true);
+  UpdateStopProgress(0);
+}
+
+void Alarm::UpdateStopProgress(lv_coord_t stopPosition) {
+  lv_bar_set_value(progressStop, progressBarSize - stopPosition, LV_ANIM_OFF);
+}
+
 void Alarm::OnButtonEvent(lv_obj_t* obj, lv_event_t event) {
-  if (event == LV_EVENT_CLICKED) {
-    if (obj == btnStop) {
-      StopAlerting();
-      return;
+  if (obj == btnStop) {
+    if (event == LV_EVENT_PRESSED) {
+      StopButtonPressed();
+    } else if (event == LV_EVENT_RELEASED || event == LV_EVENT_PRESS_LOST) {
+      ResetStopProgress();
     }
+    return;
+  }
+  if (event == LV_EVENT_CLICKED) {
     if (obj == btnInfo) {
       ShowInfo();
       return;
--- a/src/displayapp/screens/Alarm.h
+++ b/src/displayapp/screens/Alarm.h
@@ -24,6 +24,7 @@
 #include "displayapp/Controllers.h"
 #include "systemtask/WakeLock.h"
 #include "Symbols.h"
+#include <optional>
 
 namespace Pinetime {
   namespace Applications {
@@ -35,12 +36,15 @@ namespace Pinetime {
                        System::SystemTask& systemTask,
                        Controllers::MotorController& motorController);
         ~Alarm() override;
+        void Refresh() override;
         void SetAlerting();
         void OnButtonEvent(lv_obj_t* obj, lv_event_t event);
         bool OnButtonPushed() override;
         bool OnTouchEvent(TouchEvents event) override;
         void OnValueChanged();
         void StopAlerting();
+        void StopButtonPressed();
+        void ResetStopProgress();
 
       private:
         Controllers::AlarmController& alarmController;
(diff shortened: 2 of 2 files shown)


mark9064:
I haven't had a chance to test this with hardware since the ringing screen has been redesigned. Does the progress bar need to move at all? Maybe it could go between the time and the button in the blank space. Also if it looks a bit blocky (again I need to test), it will render more nicely if it is thinner since there will be fewer pixels to push.

mark9064 (commented):
Especially with the bigger button, I think this is a good addition. Thanks for the contribution! The only suggestion I have is calculating the bar progress more consistently and making sure it's super clear to the user that they have to hold the button

src/displayapp/screens/Alarm.cpp
```diff
@@ -122,12 +129,27 @@ Alarm::Alarm(Controllers::AlarmController& alarmController,
   } else {
     SetSwitchState(LV_ANIM_OFF);
   }
+
+  taskRefresh = lv_task_create(RefreshTaskCallback, 50, LV_TASK_PRIO_MID, this);
+}
+
+void Alarm::Refresh() {
+  if (stopBtnPressing && xTaskGetTickCount() > stopBtnPressTime + pdMS_TO_TICKS(150)) {
```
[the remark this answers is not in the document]
Saarsk: > `xTaskGetTickCount() - stopBtnPressTime > pdMS_TO_TICKS(150)` is tolerant to tick count wraparound Should we start showing the bar earlier than 150ms, as you need to hold the button to dismiss it. In fact, would it be better with no delay at all? A long press isn't super intuitive, so giving immediate feedback will make it clearer that holding it is needed IMO

I think immediately showing the bar makes perfect sense. It makes it responsive and informative about what's going on.

I also think it would make sense to show a different screen than the settings for the alarm (which is kind of confusing when the alarm goes off). Instead, my suggestion would be to show the time of the alarm only, as well as a larger dismiss button which will also acts as a long press indicator (the red background would gradually transition into a new background color).

A larger button will make the progress bar clearly visible even with a thumb on the button. And I think that using the button background as indicator is probably the most intuitive. 🙂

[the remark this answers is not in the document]
mark9064: It's worth noting the UI has already been changed since https://github.com/InfiniTimeOrg/InfiniTime/pull/2211 - so the larger button suggestion is already in :)

I like the idea of using the button as the bar, but I've got two concerns: 1) Performance will be poor: as the button is large, rendering will be choppy. 2) The button may be shrunk in the future to accommodate further functionality such a snooze button.

So maybe a thin indicator bar above the button could work? But I'm not sure I love that idea either. Perhaps a bar isn't needed at all?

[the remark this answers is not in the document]
vkareh: My first iteration used the button as the bar, but like @mark9064 said: performance was _very_ poor. The second issue was that I could not see the button changing colour (maybe I have fat fingers?) and so there was no feedback whatsoever.

Here's an updated capture:
InfiniSim_2025-03-27_213657

minacode:
Is the animation in the description still up to date? If not, can you please provide a new one? :)

I like the idea of a separate screen if the alarm goes off, but this PR is still very good for making sure that less alarms will be missed due to ghost touches.

minacode:
Thank you 🙂 

Can we make one more design iteration?

Maybe you can put the bar on the top of the screen, make it a little slimmer and also maybe not orange? Or feel free to suggest something else if you have another idea. Let's just see if we can make it fit in better.

Another more free idea would be to have a red background with the time in the middle and a transparent circle around it that fills clockwise while pressing the screen.

vkareh:
@minacode 
> Maybe you can put the bar on the top of the screen, make it a little slimmer and also maybe not orange? Or feel free to suggest something else if you have another idea. Let's just see if we can make it fit in better.

The orange bar is similar to the one in the timer reset button, so it's intended to keep it consistent/visible. However, I moved to bar to the top of the screen and made it a bit slimmer. What do you think?
InfiniSim_2025-05-15_085055

> Another more free idea would be to have a red background with the time in the middle and a transparent circle around it that fills clockwise while pressing the screen.

This seems like an unnecessary large change. This PR uses `lv_bar`, which is a native construct for this type of progress bar, and so has very little code/overhead.

minacode:
Thank you 🙂
I like the new version and think it looks cleaner.

tituscmd:
What if the stop button gradually loses saturation while it’s being pressed? The progress bar is helpful, but it might not be clear enough on its own that something’s actually happening when the button is being held down.

vkareh:
@tituscmd 
> What if the stop button gradually loses saturation while it’s being pressed? The progress bar is helpful, but it might not be clear enough on its own that something’s actually happening when the button is being held down.

My first iteration had the bar as part of the button (similar to the timer), but with the finger _on_ the button itself, I couldn't see anything happening.

minacode:
Thinking about it, we need a sufficient visual hint either from the start ("hold to stop") or at least after a short touch didn't work. Otherwise people might just press repeatedly and wonder why the button doesn't work.

What do you all think about using "Hold to stop" on the button without any other animation to get the alarm protection part of this PR merged soon and then we have more time to think about cool visual hints?

mark9064:
If you just tap it, the bar starts appearing immediately right? So there's feedback, and the eye sees motion so I guess it's reasonably clear it needs holding?

Changing the button colour might not look too good because it requires too much drawing to happen smoothly I think, but I guess we'd need to test

minacode:
Ok, then I am fine with merging the current version :)

mark9064 (changes requested):
Current visual design looks good to me, but I'd like to see code comments addressed (immediately rendering the bar on press and using total elapsed hold time rather than incrementing the position)

src/displayapp/screens/Alarm.cpp
```diff
@@ -122,12 +129,27 @@ Alarm::Alarm(Controllers::AlarmController& alarmController,
...
+    stopPosition += 15;
```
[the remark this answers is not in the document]
vkareh: Fixed.

src/displayapp/screens/Alarm.cpp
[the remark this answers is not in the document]
vkareh: Fixed to avoid delay.

vkareh:
> Current visual design looks good to me, but I'd like to see code comments addressed (immediately rendering the bar on press and using total elapsed hold time rather than incrementing the position)

Done!
InfiniSim_2025-11-14_120056

src/displayapp/screens/Alarm.cpp
```diff
@@ -139,12 +163,34 @@ void Alarm::DisableAlarm() {
...
+  stopBtnPressTime = xTaskGetTickCount();
+  stopBtnPressing = true;
+  stopPosition = 0;
+  lv_obj_set_hidden(progressStop, false);
+}
+
+void Alarm::ResetStopProgress() {
+  stopBtnPressing = false;
+  stopPosition = 0;
+  UpdateStopProgress();
```
mark9064: Very minor: I think it'd be more robust if the stop progress was set instead in `StopButtonPressed` just before unhiding it

src/displayapp/screens/Alarm.cpp
```diff
@@ -139,12 +163,34 @@ void Alarm::DisableAlarm() {
...
+  lv_obj_set_hidden(progressStop, true);
+}
+
+void Alarm::UpdateStopProgress() {
```
mark9064: This should take stop position as an argument, then we can avoid storing it as member

src/displayapp/screens/Alarm.h
```diff
@@ -62,8 +66,16 @@ namespace Pinetime {
         void HideInfo();
         void ToggleRecurrence();
         void UpdateAlarmTime();
+        void UpdateStopProgress();
         Widgets::Counter hourCounter = Widgets::Counter(0, 23, jetbrains_mono_76);
         Widgets::Counter minuteCounter = Widgets::Counter(0, 59, jetbrains_mono_76);
+
+        lv_obj_t* progressStop;
+        bool stopBtnPressing = false;
+        TickType_t stopBtnPressTime = 0;
```
mark9064: This could be `std::optional<TickType_t>`, where `nullopt` means not pressing and any integer value is the current stop pressing time

mark9064 (approved):
Looks great to me! Haven't tested on hardware yet

Couple refactoring suggestions that hopefully will only take a few mins

src/displayapp/screens/Alarm.h
vkareh: Clever. That removes the need for `stopBtnPressing`. I'm doing this.

src/displayapp/screens/Alarm.cpp
vkareh: Makes sense to me.

src/displayapp/screens/Alarm.cpp
vkareh: That's a good point. However, I'm tempted to keep it in this function as well to make sure it's in a clean state before hiding (feels weird to hide a progress in a halfway state).

src/displayapp/screens/Alarm.cpp
```diff
@@ -139,12 +163,32 @@ void Alarm::DisableAlarm() {
...
+void Alarm::StopButtonPressed() {
+  stopBtnPressTime = xTaskGetTickCount();
+  UpdateStopProgress(0);
+  lv_obj_set_hidden(progressStop, false);
+}
+
+void Alarm::ResetStopProgress() {
+  stopBtnPressTime = std::nullopt;
+  UpdateStopProgress(0);
+  lv_obj_set_hidden(progressStop, true);
```
mark9064: Super minor: I would hide the bar first then change the value to avoid a potential flash of the bar

I don't think it actually makes any difference in this specific case

vkareh: makes sense to me. fixed.

mark9064 (approved):
Looks great :)

mark9064:
I've stuck it on 1.17 but IMO it's all good to be included in 1.16 if someone else reviews it in time

PR #2227, opened 2025-01-15
Chime vibration patterns and 15 minute option

zischknall:
To combat my own time blindness I needed the option to have it vibrate every 15 minutes and additionally somehow indicate the time of the hour via the vibrations.

So I added an option for 15 minutes which was very straight-forward in the first commit and I think it has the highest chance of being merge-able. For the vibrations I changed the duration of motor activation as 35ms was very little and added a pattern of vibrations and pauses. So full hour is 1, first quarter hour is 2, half hour is 3 and 3 quarter hours is 4 vibrations.

It is still pretty archaically implemented. I think we should make it a bit more generic and make the vibration pattern (or if you don't want a pattern) configurable. But for me it's enough at the moment.

So if there is interest in something like this being added I will improve the implementation.


marigoldfish:
Oh, I really like this. I have time blindness problems myself, and the half hour chimes have been valuable. I have previously wished for 15min intervals and a distinction between the half hour and hour chimes.

I will give this a run for a while and see how it works out for me!

mark9064:
This PR is definitely an improvement, thanks! Sorry for the slow review- the main reason is that this PR highlights a wider problem with the way access to the vibration motor is handled. One such problem with this implementation is that doing vTaskDelay inside DisplayApp blocks the event loop stopping rendering and message processing

Reworking the motor controller is going to be challenging as there quite a few things that should probably be implemented:
- Non-blocking motor access (probably by usage of the timer task)
- Vibration patterns
- Motor intensity control through PWM (this allows running the vibration motor at half speed for example)

TLDR I think the functionality change here is very welcome here in InfiniTime, but unfortunately there's some tough maintenance work that needs to be done first

marigoldfish:
Chiming in (heh) to say that I've found the functionality of this PR very helpful over the last couple of weeks. I've definitely been zoning out less and the only occasions I've "lost track of time" for an extended period, my watch has been in sleep mode or not on my wrist. 

I did notice one minor bug, which may be in the scope of this PR or may be part of the larger reconfiguring @mark9064 brought up: when the chime goes off, the time does not update until it is finished. So, in the most extreme scenario, if I have not woken my watch since the previous chime (for example 3:45), then it will still display that time on the watch until the vibrations of the current chime is ended (at 4:00, the watch shows 3:45 while the four chimes go off, then updates to 4:00 when chiming is complete). 

But! This is a very small thing and is not in any way a deal breaker for me. This feature is very helpful for me and I hope that y'all also think it's worth doing the work to implement fully.

zischknall:
Thanks for your comments love to see people discussing this and also finding it useful. Fully agree with the motor controller needing to be reworked for this to work better. Did also not yet find the time to add a setting toggling between vibration behaviors.

zischknall:
For the first couple of weeks this worked great. But now it seems like sometimes it does not trigger. At first I thought I might just have gotten used to it but I can confirm that they have gotten unreliable after a while. Really strange...

Perdellian:
>     * Vibration patterns
> 
>     * Motor intensity control through PWM (this allows running the vibration motor at half speed for example)

Westminster ~~chimes~~ buzzes like an old grandfather clock?

PR #2230, opened 2025-01-18
Add sleep settings

JustScott:
Add a settings page for enabling/disabling actions to be done when sleep is turned on or off.

Added Options:
- [x] Allow Always on Display
- [x] Allow Notifications
- [x] Allow Chimes
- [x] Disable Ble (Ble is turned back on when sleep is turned off only if you had it on before enabling sleep)

I wanted to add an option for disabling steps, but failed to find a way in the code to do this. The step count would just update once sleep was turned off. Any help with this would be much appreciated... but I'm also fine with that option being left out.

InfiniSim_2025-01-18_110925
InfiniSim_2025-01-18_110929


mark9064:
I'm a bit confused as to what this is for. Sleep mode is for when you're sleeping - so I don't think that chimes, AOD or notifications should ever be allowed as you're not awake to see them. As for BLE, there are a couple of reasons I can think of, such as if you power off your phone at night and don't want anyone to be able to connect to your pinetime? But this seems like something that should be fixed with a security mode rather than worked around. I'd be interested to hear what you were aiming for here, I think there's definitely room to rework the modes but I'm not sure this is the best path forward

JustScott:
@mark9064 

> I'm a bit confused as to what this is for. Sleep mode is for when you're sleeping - so I don't think that chimes, AOD or notifications should ever be allowed as you're not awake to see them. As for BLE, there are a couple of reasons I can think of, such as if you power off your phone at night and don't want anyone to be able to connect to your pinetime? But this seems like something that should be fixed with a security mode rather than worked around. I'd be interested to hear what you were aiming for here, I think there's definitely room to rework the modes but I'm not sure this is the best path forward

I originally created this as I wanted AOD at night so that I could see the time whenever I wake up without having to click my watch, then I thought it would be nice to have the option to get notifications if you, for example, sleep with your phone outside of your room and recieve a call. The chime option being allowed could be helpful for those trying to get into lucid dreaming. The Ble setting is nice to turn on to conserve battery life if you don't need notifications at night anyways.

mark9064:
I see where you're coming from now. I guess the dilemma is that if InfiniTime supported every possible configuration someone could want, it'd just be too big and complex to fit on the watch and maintain. So there is the [InfiniTime vision](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/doc/InfiniTimeVision.md) where particularly there is "Prefer solid defaults over customisability". Of course this isn't a black and white rule, it's up to interpretation of whether a setting is a good idea or not (or there would be no settings at all right!).

So I'm not trying to hate on your idea! But more trying to weigh up whether this is a feature that is worth pulling in, or if it's something that would better live in a fork. A rule of thumb I try to use is "would the average user need to change this". When I go through the settings at the moment, I can say yes for everything: users need to choose watchface, imperial/metric units, switch 12h/24h, set the time etc. But I struggle to see this so much for this feature, which is why I'm more reserved towards it. It's hard to decide particularly for settings as they are part of the OS (as apps can be disabled/enabled by default easily)

JustScott:
@mark9064 I agree this probably goes beyond the infinitime vision, specifically for an official system setting, considering not everyone would find this useful. I think this feature could better serve the user as part of a sleep app, perhaps, if someone wants to adapt the code into there branch... otherwise I'm going to keep using this and keep it up to date with the main branch if anyone wants to add it to there fork. Thanks for the conversation mark!

JustScott:
Replaced `Allow AOD` with `Enable AOD`
  
Leaving it as `Allow AOD` assumes the user wants it the same during the day as during sleep. In my opinion the user would either want AOD on during sleep or they wouldn't.

new_enable_aod_sleep_option

JustScott:
Add an `Ignore Steps` option:
new_sleep_settings

When enabled and in sleep mode, motionController tracks the steps in the variable `ignoreSteps` and subtracts that from the total steps stored in `nbSteps` before returning the step count. This stop steps from counting up during sleep, and doesn't show them when sleep is disabled either.
show_sleep_disabling_steps

I believe this should also apply to the step count sent to companion apps, but I haven't tested this yet.

PR #2231, opened 2025-01-21
Adds vibration strength settings

Ghoelian:
This pr adds 2 new settings, one for notification and one for the hour/half hour chime vibration strengh.

Also previews the selected strength when you exit the menu. I'd prefer for this to happen when you select an option, but I couldn't figure out how to do that.

The only way I could think of how to get this to work was by changing the time the vibration motor is activated. With the values I gave them, there's a pretty pronounced difference between the lowest and highest setting, without it starting to feel like a long vibration.

Could use some suggestions on icons to use for the settings as I haven't added any yet.

PR #2235, opened 2025-01-26
Adds vibration strength settings

Ghoelian:
Redo of #2231, pulled in some things I didn't mean to pull in to that branch.

This pr adds 2 new settings, one for notification and one for the hour/half hour chime vibration strengh.

Also previews the selected strength when you exit the menu. I'd prefer for this to happen when you select an option, but I couldn't figure out how to do that.

The only way I could think of how to get this to work was by changing the time the vibration motor is activated. With the values I gave them, there's a pretty pronounced difference between the lowest and highest setting, without it starting to feel like a long vibration.

Could use some suggestions on icons to use for the settings as I haven't added any yet.


Ghoelian:
god damn it ok so it's the merge request solver thingy from github that keeps merging this way. at least i can just remove this one

mark9064:
Yeah all of the github buttons kinda suck, I reckon it's best to do things locally and push. Plus that way your local branch stays up to date

mark9064:
An idea that's been knocking around my head for a while: adaptive vibration strength.

When I'm still, the 35ms vibration is quite adequate and I notice it, but if I'm walking or otherwise moving it's easy to miss. So how about using recent motion activity to scale the vibration length? If the watch has been moving a lot the vibration can get strengthened. What do you think?

Ghoelian:
> An idea that's been knocking around my head for a while: adaptive vibration strength.
> 
> When I'm still, the 35ms vibration is quite adequate and I notice it, but if I'm walking or otherwise moving it's easy to miss. So how about using recent motion activity to scale the vibration length? If the watch has been moving a lot the vibration can get strengthened. What do you think?

Sounds like a great idea to me! I'll look in to implementing that, though I haven't done anything with motion in this project yet, so it might take a bit for me to figure stuff out.

PR #2237, opened 2025-01-27
Add background color to progress bar when flashing firmware

tituscmd:
This PR adds a background color to the progress bar when updating firmware.
It also removes the space between the value and the "%", since I feel like that's more commonly used and, in my opinion, looks better. Feel free to give your opinion on that, though!

Here's a pic of how it looks like on device:
(Hint: To see it in action, you’ll first need to flash the firmware. Then, flash any other firmware, as the firmware update screen is displayed by the firmware currently on the device.)
progress_bar


mark9064 (approved):
Haven't tested on HW, but looks good from the screenshot :)

Saarsk:
Just a note about the space between the value and percentage sign. A space there is the correct way of writing it. 🙂 I think it should be formatted the correct way.

I can understand that sometimes it happens that the correct way doesn't feel right. We tend to expect what we're conditioned to. I've had the same experience.

Though, it's just a matter of getting used to it, i.e. reconditioning. So rather than doing what subjectively feels right I think there's value to following objectively correct and commonly agreed conventions. 🙂 A lack there of will have negative effects regardless whether it's in a software project or in a book, as we adopt by reading and using language. So whenever something is incorrect, it spreads.

Other than that, I very much welcome the progress bar improvement. 😊

tituscmd:
"English style guides prescribe writing the percent sign following the number without any space between (e.g. 50%).[sources 1] However, the International System of Units and ISO 31-0 standard prescribe a space between the number and percent sign,[8][9][10] in line with the general practice of using a non-breaking space between a numerical value and its corresponding unit of measurement."
https://en.wikipedia.org/wiki/Percent_sign

It seems that both are correct. Let's do a poll!

tituscmd:
"However, the International System of Units and ISO 31-0 standard prescribe a space between the number and percent sign,[8][9][10] in line with the general practice of **using a non-breaking space between a numerical value and its corresponding unit of measurement.**"

Let's go for this then, shall we?

mark9064:
I think without the space is better. Wikipedia is quite clear: `English style guides prescribe writing the percent sign following the number without any space between (e.g. 50%)`

InfiniTime is not a technical document, but instead a user interface. Therefore I think following general English style and precedent makes sense. If it were a technical report then I would agree that recognising a standard such as ISO would be sensible (and if I were typsetting with LateX I would by using `siunitx`), and the same may apply in another language (it appears that some languages expect a space by default)

tituscmd:
Is everything fine from my side on this PR? Or is there any improvements wanted?

mark9064 (approved):

minacode (approved):

PR #2238, opened 2025-01-27
Notification -> Notifications

tituscmd:
This PR is a small change and it might very well be up to preference.

In the little notifications screen above the watchface, the "Notification" as well as "No notifications to display" is now changed to "Notification**s**" and "No notification**s** to display".

In my opinion using the plural makes more sense here, since the screen can show more than just one notification.
Feel free to tell me what you think!

Here's a pic of how it looks like in InfiniEmu:
Screenshot_2025-01-27_17-21-01

### DIFF
--- a/src/displayapp/screens/Notifications.cpp
+++ b/src/displayapp/screens/Notifications.cpp
@@ -246,8 +246,8 @@ namespace {
 
 Notifications::NotificationItem::NotificationItem(Pinetime::Controllers::AlertNotificationService& alertNotificationService,
                                                   Pinetime::Controllers::MotorController& motorController)
-  : NotificationItem("Notification",
-                     "No notification to display",
+  : NotificationItem("Notifications",
+                     "No notifications to display",
                      0,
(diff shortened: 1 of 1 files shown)


FintasticMan (approved):
Somehow, I haven't noticed this before!

SteveAmor:
I read the singular as:
Notification 0 of 0
Notification 1 of 1
Notification 1 of 2
.....
.....
Notification 4 of 5
etc, etc.

I also read the specific case for Notification 0/0 as the notification that you are currently on so the area which would normally display the notification singularly says (correctly in my opinion) "No notification to display".

It's only my opinion and would save 16B.

tituscmd:
I agree that either "Notification" or "Notifications" is in some way not perfectly descriptive. Something like "Notification Center" would be ideal, but it doesn't fit inside that space.

mark9064 (approved):
I also agree that this isn't optimal, but that it fits better with the s than without. May as well merge this for now, and then if someone suggests a better idea later on we can switch to it

PR #2239, opened 2025-01-27
Add mixed watchface

leroivi:
Here is the watchface I made and use so I open this PR to share it in case.
This watchface is a mix between the digital watchface and the analog watchface.

mixed_watchface

/!\ I added a new setting to adapt the date format to the user preferences.

date_format_setting


liamcharger:
The date format setting will need to be included in its own PR as it's not specific to this watch face. This will also make it easier for maintainers to review your code!

PR #2240, opened 2025-01-31
fonts: change from JetBrains Mono

liamcharger:
I don't know if this is the case for anyone else, but personally I don't like JetBrains Mono as a system font. In this PR, I've replaced it with Noto Sans. Arial or something similar might be better, but I haven't tested with anything other than Roboto, which did not look good at all.

I'm leaving this as a draft as I haven't changed any variable names, and I'd like to hear feedback from the community. The rest of the firmware also needs a couple of tweaks before this would be ready for merge. (as you can see from the screenshots)

Attached are some images of what Noto Sans looks like on the watch:
IMG_4749
IMG_4750
IMG_4751
IMG_4752
IMG_4753
IMG_4754


tituscmd:
> The monospace font definitely makes the watch feel more "for programmers" IMO, so I think I like this.

I agree with this 👆

liamcharger:
> The way Mon and Wed got wrapped in the weather app stands out to me, though.

The rest of the UI will definitely need a few tweaks. I didn't update anything because I wasn't sure if this would be the font we go with–I'll go ahead and start playing around with everything now that I know people are interested in this.

mark9064:
IMO kerning will be needed for this to look good (as it's not monospace), might be worth looking into how to enable it? It'll come with some size cost though

liamcharger:
I agree it would look better with kerning on. 

I looked at some docs, but didn't see a way to do it in LVGL 7. I might be missing something though.

src/displayapp/fonts/fonts.json
```diff
@@ -2,7 +2,7 @@
    "jetbrains_mono_bold_20": {
```
escoand, line 2: Doesn't this needs to be `noto_sans_bold_20` and so on?

liamcharger: Yeah, we'd want to change it I decided to continue work on this (which I don't think I will)

liamcharger:
Closing because I don't think I'll continue work on this (also the bpp has been increased so it matters less to me)

PR #2242, opened 2025-02-03
Add time since last weather update to weather app

### LINKED ISSUE #2140: Weather app needs to show time weather was acquired
### Verification

### Introduce the issue

I like the weather app but sometimes weather data will 'stall'.  I have noticed on many smartwatches from the Pebble up to WearOS that sometimes the weather is not updated for several hours.

### Preferred solution

On the weather page there should be the time the weather data was acquired, or a comment indicating this, eg. '2 hours old'.

### Version

a2356f2f

JustScott:
Provides the time since the last weather update, incrementing the duration from seconds to minutes at 60 seconds, and minutes to hours at 3600 seconds. I'm open to suggestions on changes to the texts position and/or color. This should resolve #2140.

Also, on the simulator the seconds old count is only updated when you open the app, while on hardware the seconds count up live while in the app... I'm not sure why this is.

weather_0s_old
weather_32s_old
weather_1m_old

### DIFF
--- a/src/displayapp/screens/Weather.cpp
+++ b/src/displayapp/screens/Weather.cpp
@@ -49,7 +49,7 @@ Weather::Weather(Controllers::Settings& settingsController,
 
   lastUpdated = lv_label_create(lv_scr_act(), nullptr);
   lv_obj_set_style_local_text_color(lastUpdated, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, Colors::bg);
-  lv_label_set_text_fmt(lastUpdated, "");
+  lv_label_set_text(lastUpdated, "");
 
   minTemperature = lv_label_create(lv_scr_act(), nullptr);
   lv_obj_set_style_local_text_color(minTemperature, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, Colors::bg);
@@ -150,20 +150,30 @@ void Weather::Refresh() {
       int8_t minutesSinceWeatherUpdate = secondsSinceWeatherUpdate / 60;
       int8_t hoursSinceWeatherUpdate = secondsSinceWeatherUpdate / 3600;
 
-      lv_obj_align(lastUpdated, nullptr, LV_ALIGN_CENTER, -31, -1);
-      if ((secondsSinceWeatherUpdate > 9 && secondsSinceWeatherUpdate < 60) ||
-          (minutesSinceWeatherUpdate > 9 && minutesSinceWeatherUpdate < 60) || hoursSinceWeatherUpdate > 9) {
-        lv_obj_align(lastUpdated, nullptr, LV_ALIGN_CENTER, -41, -1);
-      }
+      constexpr uint8_t Y_POSITION = 108;
+      constexpr uint8_t X_SINGLE_DIGIT_POSITION = 90;
+      constexpr uint8_t X_TWO_DIGIT_POSITION = 78;
+      constexpr uint8_t X_NOW_POSITION = 102;
+
+      lv_obj_set_pos(lastUpdated, X_SINGLE_DIGIT_POSITION, Y_POSITION);
 
       if (hoursSinceWeatherUpdate > 0) {
+        if (hoursSinceWeatherUpdate > 9) {
+          lv_obj_set_pos(lastUpdated, X_TWO_DIGIT_POSITION, Y_POSITION);
+        }
         lv_label_set_text_fmt(lastUpdated, "%dh ago", hoursSinceWeatherUpdate);
       } else if (minutesSinceWeatherUpdate > 0) {
+        if (minutesSinceWeatherUpdate > 9 && minutesSinceWeatherUpdate < 60) {
+          lv_obj_set_pos(lastUpdated, X_TWO_DIGIT_POSITION, Y_POSITION);
+        }
         lv_label_set_text_fmt(lastUpdated, "%dm ago", minutesSinceWeatherUpdate);
       } else if (secondsSinceWeatherUpdate > 30) {
+        if (secondsSinceWeatherUpdate > 9 && secondsSinceWeatherUpdate < 60) {
+          lv_obj_set_pos(lastUpdated, X_TWO_DIGIT_POSITION, Y_POSITION);
+        }
         lv_label_set_text_fmt(lastUpdated, "%ds ago", secondsSinceWeatherUpdate);
       } else if (secondsSinceWeatherUpdate < 31) {
-        lv_obj_align(lastUpdated, nullptr, LV_ALIGN_CENTER, -18, -1);
+        lv_obj_set_pos(lastUpdated, X_NOW_POSITION, Y_POSITION);
         lv_label_set_text_fmt(lastUpdated, "Now", secondsSinceWeatherUpdate);
       }
     } else {
--- a/src/displayapp/screens/Weather.cpp
+++ b/src/displayapp/screens/Weather.cpp
@@ -144,7 +144,7 @@ void Weather::Refresh() {
       lv_label_set_text_fmt(minTemperature, "%d°", minTemp);
       lv_label_set_text_fmt(maxTemperature, "%d°", maxTemp);
 
-      std::chrono::seconds secondsSinceEpoch =
+      auto secondsSinceEpoch =
         std::chrono::duration_cast<std::chrono::seconds>(dateTimeController.CurrentDateTime().time_since_epoch());
       int32_t secondsSinceWeatherUpdate = secondsSinceEpoch.count() - optCurrentWeather->timestamp;
       int8_t minutesSinceWeatherUpdate = secondsSinceWeatherUpdate / 60;
--- a/src/displayapp/screens/Weather.cpp
+++ b/src/displayapp/screens/Weather.cpp
@@ -163,14 +163,12 @@ void Weather::Refresh() {
         }
         lv_label_set_text_fmt(lastUpdated, "%dh ago", hoursSinceWeatherUpdate);
       } else if (minutesSinceWeatherUpdate > 0) {
-        if (minutesSinceWeatherUpdate > 9 && minutesSinceWeatherUpdate < 60) {
+        if (minutesSinceWeatherUpdate > 9) {
           lv_obj_set_pos(lastUpdated, X_TWO_DIGIT_POSITION, Y_POSITION);
         }
         lv_label_set_text_fmt(lastUpdated, "%dm ago", minutesSinceWeatherUpdate);
       } else if (secondsSinceWeatherUpdate > 30) {
-        if (secondsSinceWeatherUpdate > 9 && secondsSinceWeatherUpdate < 60) {
-          lv_obj_set_pos(lastUpdated, X_TWO_DIGIT_POSITION, Y_POSITION);
-        }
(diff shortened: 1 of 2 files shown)


liamcharger:
Personally, I would prefer to see the minutes instead of the seconds when I have recently updated my weather. For instance, instead of displaying “32 seconds old,” I would like to see “Now” and only show the number of minutes after one has passed.

Another aspect that feels more natural to me is using the phrase “X minutes ago” instead of “X minutes old.”

src/displayapp/screens/Weather.cpp
```diff
@@ -137,6 +144,18 @@ void Weather::Refresh() {
       lv_label_set_text_fmt(temperature, "%d°%c", temp, tempUnit);
       lv_label_set_text_fmt(minTemperature, "%d°", minTemp);
       lv_label_set_text_fmt(maxTemperature, "%d°", maxTemp);
+
+      int64_t secondsSinceEpoch = dateTimeController.CurrentDateTime().time_since_epoch().count() / 1000000000;
+      int64_t secondsSinceWeatherUpdate = secondsSinceEpoch - optCurrentWeather->timestamp;
+      if (secondsSinceWeatherUpdate < 0) {
```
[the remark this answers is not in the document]
JustScott: I did have it showing -1 a few times in the simulator, so figured I would add that just to cover it in the off chance it happened on the watch.

JustScott:
> Personally, I would prefer to see the minutes instead of the seconds when I have recently updated my weather. For instance, instead of displaying “32 seconds old,” I would like to see “Now” and only show the number of minutes after one has passed.
> 
> Another aspect that feels more natural to me is using the phrase “X minutes ago” instead of “X minutes old.”

I agree with you on using `ago` instead of `old`. As for using `Now` instead of displaying the seconds, I feel this looks a little akward and personally prefer the seconds count, I'd like to hear what others think about this.


now_instead_of_seconds
seconds_instead_of_now

Have to move the label up `-1px` to prevent the `g` from being cut off like in the picture above
move_update_text_up

And I move the temp up as well to prevent crowding:
move_all_up

tituscmd:
I like `Now` as well, but also agree that it looks a bit awkward.
I have some propositions as to how this could be perhaps made better:

`Updated now` / `Updated 12s ago`
`Fetched now` / `Fetched 12s ago`
`Refreshed now` / `Refreshed 12s ago`

Adding another word results in a less awkward look, as well as being more direct and intuitive to understand.

tituscmd:
> seconds_instead_of_now
>
> test

In these two images the label doesn't seem to be centered. Is that just me?

JustScott:
@tituscmd I fixed the centering issue by dynamically changing the position depending on weather there is on or two digits. I just tried out a bunch of examples of how this could look, everyone should let me know which they prefer!

### With a seconds counter (with dynamic alignment)
InfiniSim_2025-02-06_170016
InfiniSim_2025-02-06_170034

### Two version of up to date
InfiniSim_2025-02-06_170114
up_to_date

### variations of "updated now"
now_instead_of_seconds
InfiniSim_2025-02-06_170643
InfiniSim_2025-02-06_170704
InfiniSim_2025-02-06_170812

### "Updated:" with dynamic alignment
InfiniSim_2025-02-06_171040
InfiniSim_2025-02-06_171427

I can make more examples if anyones curious. Just let me know. I personally prefer to just include the seconds count, as it may be useful to some and keeps the label consistent... also the live second count on hardware makes the app feel more lively.

tituscmd:
> InfiniSim_2025-02-06_170034

In this image, if you look at the weather icon up top it still doesn't look centered. I don't know if you're doing that already, but I would suggest aligning the label with a pointer to the weather icon or the day in the middle of the table, those always seem to be centered.

I may be a bit of a perfectionist ;D

tituscmd:
@JustScott Thank you so much for making all these options!

I think after looking at everything, I like `Now` and `45s ago` the most. The others seem to make the app look to crowded.

JustScott:
> @JustScott Thank you so much for making all these options!
> 
> I think after looking at everything, I like `Now` and `58s ago` / `53m ago` / `15h ago` the most. The others seem to make the app look too crowded.

I agree with you on this, but am curious as to how long you think the app should display`Now` before switching to `Xs ago`?

tituscmd:
> > In this image, if you look at the weather icon up top it still doesn't look centered. I would suggest aligning the label with a pointer to the weather icon or the day in the middle of the table, those always seem to be centered.
> > I may be a bit of a perfectionist ;D
> 
> Ah I see what you mean about it not looking centered, I was just trying to keep the space between the `1m` and `ago` above the second letter of whatever the third day is, so it doesn't look like the label is moving when it moves from `9m` to `10m`.

I see, that makes sense.
I'd prefer it centered, but I like the idea of making it look like the label isn't moving! :)

tituscmd:
> I agree with you on this, but am curious as to how long you think the app should display`Now` before switching to `Xs ago`?

Oh, right. What about at `30s ago`?

JustScott:
> > I agree with you on this, but am curious as to how long you think the app should display`Now` before switching to `Xs ago`?
> 
> Oh, right. What about at `30s ago`? Or maybe `31s ago`, so that the first 30 seconds are displayed as "Now" and the next 30 seconds are displayed as `Xs ago`

Okay, just so anyone reading this is caught up, here's what it'll look like:

now
31s_ago
> I'd prefer it centered, but I like the idea of making it look like the label isn't moving! :)

I personally prefer it to be centered over the third words second letter, so I say we see what the others think. I'm open to changing it if people agree it doesn't look centered.
one_minute

liamcharger:
Just a summary of what I would expect to see—"Now" until a minute has passed, "Xmin ago" until an hour has passed, and so on. Support for "Xd ago" and "Xw ago" would also make sense so I'm not looking at "80h ago," or something similar in case my weather isn't updated for a while.

As far as positioning, I like the way the label looks in @JustScott's comment above.

Here's an idea for another PR: instead of squeezing everything onto one screen, what if it's separated into two that the user can scroll vertically—one for the current weather and one for the forecast? The forecast days could be laid out horizontally, and this would fix the text wrap issue in #2240 and would eliminate the need to pick an element to center on in this PR.

JustScott:
> Support for "Xd ago" and "Xw ago" would also make sense so I'm not looking at "80h ago," or something similar in case my weather isn't updated for a while.


I believe the weather service only keeps the weather data for 24 hours, so after that it wouldn't display anything until updated again.

liamcharger:
> I believe the weather service only keeps the weather data for 24 hours, so after that it wouldn't display anything until updated again.

👍 Just ignore that part then, haha

src/displayapp/screens/Weather.cpp
```diff
@@ -137,6 +144,18 @@ void Weather::Refresh() {
       lv_label_set_text_fmt(temperature, "%d°%c", temp, tempUnit);
       lv_label_set_text_fmt(minTemperature, "%d°", minTemp);
       lv_label_set_text_fmt(maxTemperature, "%d°", maxTemp);
+
+      int64_t secondsSinceEpoch = dateTimeController.CurrentDateTime().time_since_epoch().count() / 1000000000;
```
dariusarnold: Since C++ 14 you can use a `'` to split digits. I think it improves readability for large integer literals.
```suggestion
      int64_t secondsSinceEpoch = dateTimeController.CurrentDateTime().time_since_epoch().count() / 1'000'000'000;
```

dariusarnold: Sorry, I was not on the most recent commit, my bad.

src/displayapp/screens/Weather.cpp
```diff
@@ -137,10 +143,44 @@ void Weather::Refresh() {
       lv_label_set_text_fmt(temperature, "%d°%c", temp, tempUnit);
       lv_label_set_text_fmt(minTemperature, "%d°", minTemp);
       lv_label_set_text_fmt(maxTemperature, "%d°", maxTemp);
+
+      std::chrono::seconds secondsSinceEpoch =
```
dariusarnold: The coding-convention.md suggest to replace the type with auto when initializing with a cast because it avoids duplicating the type.
```suggestion
      const auto secondsSinceEpoch =
```

dariusarnold: The other three integer variables can also be made const.

src/displayapp/screens/Weather.cpp
```diff
@@ -137,10 +143,44 @@ void Weather::Refresh() {
...
+      int8_t hoursSinceWeatherUpdate = secondsSinceWeatherUpdate / 3600;
+
+      constexpr uint8_t Y_POSITION = 108;
+      constexpr uint8_t X_SINGLE_DIGIT_POSITION = 90;
+      constexpr uint8_t X_TWO_DIGIT_POSITION = 78;
+      constexpr uint8_t X_NOW_POSITION = 102;
+
+      lv_obj_set_pos(lastUpdated, X_SINGLE_DIGIT_POSITION, Y_POSITION);
+
+      if (hoursSinceWeatherUpdate > 0) {
```
dariusarnold, line 159: The `secondsSinceWeatherUpdate > 9` condition in line171  will always be true due to the if from the previous line and could be removed.

The `secondsSinceWeatherUpdate < 60` should also be always true because otherwise the flow would go through the `minutesSinceWeatherUpdate > 0` branch, but you can consider keeping it to be explicit (the same is true for `minutesSinceWeatherUpdate < 60` in line 166). The `secondsSinceWeatherUpdate < 31` condition in line 175 could then be transformed to just an `else`, probably with a comment to tell the reader that this is the < 31 s case.

The final code with all suggestions applied could look like this:
```cpp
      if (hoursSinceWeatherUpdate > 0) {
        if (hoursSinceWeatherUpdate > 9) {
          lv_obj_set_pos(lastUpdated, X_TWO_DIGIT_POSITION, Y_POSITION);
        }
        lv_label_set_text_fmt(lastUpdated, "%dh ago", hoursSinceWeatherUpdate);
      } else if (minutesSinceWeatherUpdate > 0) {
        if (minutesSinceWeatherUpdate > 9) {
          lv_obj_set_pos(lastUpdated, X_TWO_DIGIT_POSITION, Y_POSITION);
        }
        lv_label_set_text_fmt(lastUpdated, "%dm ago", minutesSinceWeatherUpdate);
      } else if (secondsSinceWeatherUpdate > 30) {
          lv_obj_set_pos(lastUpdated, X_TWO_DIGIT_POSITION, Y_POSITION);
        lv_label_set_text_fmt(lastUpdated, "%ds ago", secondsSinceWeatherUpdate);
      } else {
        // Update time less than 30 s
        lv_obj_set_pos(lastUpdated, X_NOW_POSITION, Y_POSITION);
        lv_label_set_text_fmt(lastUpdated, "Now", secondsSinceWeatherUpdate);
      }
```

dariusarnold:
Hello, I have been using this for some time now and really like it, thank you @JustScott. I did a small review to help this get merged. I have not experienced any issues or bugs.

JustScott:
@dariusarnold Thank you for the review, I've applied all of your suggestions outside of the final `else` statement instead of the explicit `else if` as I prefer this over a comment. Thank you for helping! Also, I don't really use weather much anymore on my pinetime... so if you find any bugs with these changes please let me know!

liamcharger:
Little bump on this to the maintainers :) would like to see this in 1.17

mark9064:
agree

PR #2248, opened 2025-02-15
Prevent blocking when Infineat watchface is charging

### LINKED ISSUE #2245: Return to Infineat watch face crashes when charging
### Verification

### What happened?

Swip off Infineat watch face, placing on charger, wait for screen to go blank and pressing button twice, screen remains blank and resets watchdog

Watch should not crash. It should display the watchface when you press the button (which it does for all watch faces except Infineat)

### Reproduction steps

Display timeout is 5 secs.
Only happens with the Infineat face
Go through the below quite quickly.
Reproduced 100% of the time.

Swipe watchface to access quick settings
Place on charger
Small charge icon is displayed at top of display
Display switches off after 5 seconds
Press button to switch on display (I wanted to be sure it was charging)
Display doesn't switch on*
Press button again.
Display doesn't switch on.
Remove from charger
Press button to switch on Display
Watchdog timeout occurs
See watchface for a fraction of a second and then white to green pine cone
Last reset = wtdg

### More details?

No shake to wake
Backlight low, med, high
24h
Infineat (works as expected on other faces)
Display timeout 5s
Bluetooth connected or off
Not using weather reporting
Watch is 5 years old from hardware point of view.
Battery level was <50% when I first noticed this and was reproduced multiple times at 100% charge.
Button presses were quick clicks. i.e. I didn't hold the button for 7 seconds to force a watchdog timer reset.

### Version

Standard V1.15 2105a7b6

### Companion app

_No response_

SteveAmor:
Fixes #2245
Blocking removed for the edge case where it reverts to the clock on wakeup if no app is loaded and the watchface is Infineat and the watch is charging.  In this particular case, there is a small UX watchface transition glitch on wakeup.  The alternative is blocking code which often results in the watch rebooting.

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -316,8 +316,11 @@ void DisplayApp::Refresh() {
             currentApp == Apps::Settings) {
           LoadScreen(Apps::Clock, DisplayApp::FullRefreshDirections::None);
           // Wait for the clock app to load before moving on.
-          while (!lv_task_handler()) {
-          };
+          // If Infineat is charging due to the constant charging animation, lv_task_handler will block so we will skip
+          if (settingsController.GetWatchFace() != Pinetime::Applications::WatchFace::Infineat || !batteryController.IsCharging()) {
+            while (!lv_task_handler()) {
+            }
+          }
         }
         // Clear any ongoing touch pressed events
         // Without this LVGL gets stuck in the pressed state and will keep refreshing the


mark9064:
Thanks for the patch! If we don't find a better solution before 1.16, I think this is a good temporary measure
It sounds like there are two core issues though:
- The Infineat charging animation is capable of using 100% drawing bandwidth (CPU/IO), and therefore causes LVGL to never finish. This sounds suspicious and should be reworked - it's a waste of resources. Drawing a battery indicator should not be so resource demanding
- The watchface load loop is overoptimistic and locks up DisplayApp when faced with 100% load

Both issues should be fixed. I've just looked on InfiniEmu and indeed the CPU is in the `Running` state the entire time when charging on Infineat (though my laptop CPU runs it very slowly, so this may be a factor), so this issue is pretty much confirmed. The watchface load loop should be fixed in LVGL. I have some testing code for this but it's super ugly. More broadly I need to really understand the LVGL render pipeline so I can point to the exact parts we need to change (i.e guaranteeing the watchface `Refresh()` runs exactly once, and then also that drawing this one refresh is fully finished)

mark9064:
Closing as #2256 is merged now :)

PR #2256, opened 2025-02-18
Prevents blocking when Infineat watchface is charging

### LINKED ISSUE #2245: Return to Infineat watch face crashes when charging

SteveAmor:
This PR prevents the blocking issue identified in #2245 
This is a better solution than the one proposed in #2248

### DIFF
--- a/src/displayapp/screens/WatchFaceInfineat.cpp
+++ b/src/displayapp/screens/WatchFaceInfineat.cpp
@@ -434,10 +434,10 @@ void WatchFaceInfineat::Refresh() {
 
   batteryPercentRemaining = batteryController.PercentRemaining();
   isCharging = batteryController.IsCharging();
-  if (batteryController.IsCharging() && (xTaskGetTickCount() - chargingAnimationTick > pdMS_TO_TICKS(150))) { // Charging battery animation
-    chargingBatteryPercent +=
-      100 /
-      lv_obj_get_height(logoPine); // Dividing 100 by the height gives the battery percentage required to shift the animation by 1 pixel
+  // Charging battery animation
+  if (batteryController.IsCharging() && (xTaskGetTickCount() - chargingAnimationTick > pdMS_TO_TICKS(150))) {
+    // Dividing 100 by the height gives the battery percentage required to shift the animation by 1 pixel
+    chargingBatteryPercent += 100 / lv_obj_get_height(logoPine);
     if (chargingBatteryPercent > 100) {
       chargingBatteryPercent = batteryPercentRemaining.Get();
     }
--- a/src/displayapp/screens/WatchFaceInfineat.cpp
+++ b/src/displayapp/screens/WatchFaceInfineat.cpp
@@ -434,13 +434,13 @@ void WatchFaceInfineat::Refresh() {
 
   batteryPercentRemaining = batteryController.PercentRemaining();
   isCharging = batteryController.IsCharging();
-  if (batteryController.IsCharging() && (uint32_t(lv_tick_get() - savedAnimationTick) > 150)) { // Charging battery animation
-    chargingBatteryPercent += 3; // 100% / pine_small height = 3
+  if (batteryController.IsCharging() && (xTaskGetTickCount() - chargingAnimationTick > pdMS_TO_TICKS(150))) { // Charging battery animation
+    chargingBatteryPercent += 100 / lv_obj_get_height(logoPine);
     if (chargingBatteryPercent > 100) {
       chargingBatteryPercent = batteryPercentRemaining.Get();
     }
     SetBatteryLevel(chargingBatteryPercent);
-    savedAnimationTick = lv_tick_get();
+    chargingAnimationTick = xTaskGetTickCount();
   } else if (isCharging.IsUpdated() || batteryPercentRemaining.IsUpdated()) {
     chargingBatteryPercent = batteryPercentRemaining.Get();
     SetBatteryLevel(chargingBatteryPercent);
--- a/src/displayapp/screens/WatchFaceInfineat.cpp
+++ b/src/displayapp/screens/WatchFaceInfineat.cpp
@@ -434,12 +434,13 @@ void WatchFaceInfineat::Refresh() {
 
   batteryPercentRemaining = batteryController.PercentRemaining();
   isCharging = batteryController.IsCharging();
-  if (batteryController.IsCharging()) { // Charging battery animation
-    chargingBatteryPercent += 1;
+  if (batteryController.IsCharging() && (uint32_t(lv_tick_get() - savedAnimationTick) > 150)) { // Charging battery animation
+    chargingBatteryPercent += 3; // 100% / pine_small height = 3
     if (chargingBatteryPercent > 100) {
       chargingBatteryPercent = batteryPercentRemaining.Get();
     }
     SetBatteryLevel(chargingBatteryPercent);
+    savedAnimationTick = lv_tick_get();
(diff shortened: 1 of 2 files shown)


SteveAmor:
I'm all for clean code and keeping things tidy but I don't like the way clang wants to right justify my comment, about the magic number 3, with the previous charging battery animation comment.
```(lv_tick_get() - savedAnimationTick > 150)) { // Charging battery animation```
```chargingBatteryPercent += 3;                  // 100% / pine_small height = 3```
But, if I have to make the change I will.

File: src/displayapp/screens/WatchFaceInfineat.cpp [hunk 1]
```diff
@@ -434,12 +434,13 @@ void WatchFaceInfineat::Refresh() {
 
   batteryPercentRemaining = batteryController.PercentRemaining();
   isCharging = batteryController.IsCharging();
-  if (batteryController.IsCharging()) { // Charging battery animation
-    chargingBatteryPercent += 1;
+  if (batteryController.IsCharging() && (lv_tick_get() - savedAnimationTick > 150)) { // Charging battery animation
```
NeroBurner, line 437: What happens when the tick counter wrapps around. Then the tick count is smaller than the saved tick count. What would be the resulting behaviour?

SteveAmor: Good point. I need to add a cast to fix this ```(uint32_t(lv_tick_get() - savedAnimationTick) > 150)```

SteveAmor: > Good point. I need to add a cast to fix this `(uint32_t(lv_tick_get() - savedAnimationTick) > 150)` Now, wrap around shouldn't be a problem.

I tested on the simulator with and without the ```uint32_t``` and the compiler applies the same modulo arithmetic for unsigned ints to either version with no change in the complied flash and ram sizes. However, it's probably safer to leave ```uint32_t``` in.

src/displayapp/screens/WatchFaceInfineat.cpp
```diff
@@ -434,12 +434,13 @@ void WatchFaceInfineat::Refresh() {
 
   batteryPercentRemaining = batteryController.PercentRemaining();
   isCharging = batteryController.IsCharging();
-  if (batteryController.IsCharging()) { // Charging battery animation
-    chargingBatteryPercent += 1;
+  if (batteryController.IsCharging() && (uint32_t(lv_tick_get() - savedAnimationTick) > 150)) { // Charging battery animation
+    chargingBatteryPercent += 3; // 100% / pine_small height = 3
```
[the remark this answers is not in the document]
SteveAmor: Thanks for the feedback @mark9064. I have made the changes and it works as expected in terms of the same timings as the original code and as per the first commit in this PR.
Let me know if any more changes are required.  If not, would you would like me to squash the commits?

src/displayapp/screens/WatchFaceInfineat.cpp
```diff
@@ -434,12 +434,13 @@ void WatchFaceInfineat::Refresh() {
 
   batteryPercentRemaining = batteryController.PercentRemaining();
   isCharging = batteryController.IsCharging();
-  if (batteryController.IsCharging()) { // Charging battery animation
-    chargingBatteryPercent += 1;
+  if (batteryController.IsCharging() && (uint32_t(lv_tick_get() - savedAnimationTick) > 150)) { // Charging battery animation
```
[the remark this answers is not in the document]
SteveAmor: Thanks for the feedback @mark9064. I have made the changes and it works as expected in terms of the same timings as the original code and as per the first commit in this PR.

src/displayapp/screens/WatchFaceInfineat.h
```diff
@@ -46,6 +46,7 @@ namespace Pinetime {
 
       private:
         uint32_t savedTick = 0;
+        uint32_t savedAnimationTick = 0;
```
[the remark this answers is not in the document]
SteveAmor: I think/hope the changes I made address this?

SteveAmor:
@mark9064 let me know if any more changes are required.  If not, would you would like me to squash the commits?

src/displayapp/screens/WatchFaceInfineat.cpp
```diff
@@ -434,12 +434,13 @@ void WatchFaceInfineat::Refresh() {
 
   batteryPercentRemaining = batteryController.PercentRemaining();
   isCharging = batteryController.IsCharging();
-  if (batteryController.IsCharging()) { // Charging battery animation
-    chargingBatteryPercent += 1;
+  if (batteryController.IsCharging() && (xTaskGetTickCount() - chargingAnimationTick > pdMS_TO_TICKS(150))) { // Charging battery animation
+    chargingBatteryPercent += 100 / lv_obj_get_height(logoPine);
```
mark9064, line 438: I think a code comment here to explain what is going on would be good (dividing 100 by the height means we can find the battery percentage change needed to shift the indicator by 1 pixel)

SteveAmor: Hopefully done 😀

mark9064: Maybe pop the comment the line before rather than inline? Might be easier to read that way

SteveAmor:
Thank you for your patience. I have been battling the clang formatting to get the test to pass. I always wanted to move the comment above the line of code.

src/displayapp/screens/WatchFaceInfineat.cpp
SteveAmor: Thank you for your patience. I have been battling the clang formatting to get the test to pass. I always wanted to move the comment above the line of code.

mark9064 (approved):
Thanks for the report and fix!

I think this is good to go, my only criticism would be is that when it's almost full the battery indicator might flicker in some way. But I think it did this before, so it's not a regression. How has it looked in your testing?

SteveAmor:
> I think this is good to go, my only criticism would be is that when it's almost full the battery indicator might flicker in some way. But I think it did this before, so it's not a regression. How has it looked in your testing?

On the PT hardware, it looks identical to the animation in V1.15 and the animation when the battery is almost full is also the same. It's quite tolerable, not flashing the last pixel or two too fast. You can see it's charging when you're at 90+ percent.

PR #2268, opened 2025-03-12
Show shake calibration being disabled when in sleep mode

dariusarnold:
I was confused by the issue with the shake calibration on my PineTime for which a partial fix exists in https://github.com/InfiniTimeOrg/InfiniTime/pull/2161 (thank you lmamane). I adopted those changes and made the demanded enum refactor as well as fix an issue with the text not disappearing when leaving the shake calibration screen.

Here is a screenshot of the new disabled message:
InfiniSim_2025-03-12_203033

All changes are:
- Switch from bool EnableForCal to always saving and restoring shake wake mode to avoid to many conditionals in the constructor and destructor
- Refactor calibrating into currentCalibrationStepEnum with descriptive names
- Initialize variables with default values in the class body
- Switch to forward declarations for references in the header and remove unused include
- Fix issue where text would stay on screen due to not calling lv_obj_clean() in the destructor

### DIFF
--- a/src/displayapp/screens/settings/SettingShakeThreshold.cpp
+++ b/src/displayapp/screens/settings/SettingShakeThreshold.cpp
@@ -2,8 +2,10 @@
 #include <lvgl/lvgl.h>
 #include "displayapp/DisplayApp.h"
 #include "displayapp/screens/Screen.h"
-#include "displayapp/screens/Symbols.h"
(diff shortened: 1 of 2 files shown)


mark9064:
After #2328 it now works all the time :) thanks for writing up this feature nonetheless

The refactoring you've done still looks good though, if you would like to open a new PR for that I will try and take a look

PR #2272, opened 2025-03-14
MotionService: Fix step notification on sleep

MB512:
When using BLE notifications for motion it seams that the code enabels the sending of each motion even in sleep mode.

Using BLE notifications for steps should work the same.

I don't know how much this would use the battery, but i think it could not be more than with motion notification.


mark9064:
Thanks for this contribution :) it all looked good but ended up being superseded by #2328 so I've closed it now

PR #2276, opened 2025-03-19
Refactor Colors in unique place

### LINKED ISSUE #559: Make color definitions more global in source code

leroivi:
Bring the PineTime colors, the Settings colors, the InfiniTimeTheme colors and the LVGL colors in one and unique place so it is easier for developper to be consistent between different screens and bring their own colors.

Fix #559

### DIFF
--- a/src/displayapp/InfiniTimeTheme.cpp
+++ b/src/displayapp/InfiniTimeTheme.cpp
@@ -1,5 +1,6 @@
 #include "displayapp/InfiniTimeTheme.h"
 #include <algorithm>
+#include "displayapp/Colors.h"
 
 // Replace LV_DPX with a constexpr version using a constant LV_DPI
 #undef LV_DPX
@@ -54,7 +55,7 @@ static void style_init_reset(lv_style_t* style) {
 static void basic_init() {
   style_init_reset(&style_bg);
   lv_style_set_bg_opa(&style_bg, LV_STATE_DEFAULT, LV_OPA_COVER);
-  lv_style_set_bg_color(&style_bg, LV_STATE_DEFAULT, LV_COLOR_BLACK);
+  lv_style_set_bg_color(&style_bg, LV_STATE_DEFAULT, Colors::Color(Colors::Named::Black));
   lv_style_set_text_font(&style_bg, LV_STATE_DEFAULT, theme.font_normal);
 
   style_init_reset(&style_box);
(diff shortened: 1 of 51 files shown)


src/displayapp/InfiniTimeTheme.cpp
```diff
@@ -54,7 +55,7 @@ static void style_init_reset(lv_style_t* style) {
 static void basic_init() {
   style_init_reset(&style_bg);
   lv_style_set_bg_opa(&style_bg, LV_STATE_DEFAULT, LV_OPA_COVER);
-  lv_style_set_bg_color(&style_bg, LV_STATE_DEFAULT, LV_COLOR_BLACK);
+  lv_style_set_bg_color(&style_bg, LV_STATE_DEFAULT, Colors::Color(Colors::Named::Black));
```
shaleh: Is there a way to achieve this otherwise great refactor without requiring developers to say the word Color three times in a row? Maybe having "Named" expose `Colors::Color` instances and having the raw numeric value be less exposed but available?

leroivi: Well, I was hoping that, by letting every constructors and cast operators be implicit, I could write `lv_style_set_bg_color(&style_bg, LV_STATE_DEFAULT, Colors::Named::Black);`
But it didn't work, I didn't figure out exactly why. This implies that the compiler realize all the following implicit cast : Colors::Named => uint32_t => Colors::Color => lv_color_t. Maybe one of the cast is ambiguous, or maybe it's simply too much, idk.

I was thinking that it's not a big problem, because it can be less verbose by using the `using namespace` instruction
So we can write `lv_style_set_bg_color(&style_bg, LV_STATE_DEFAULT, Color(Named::Black));` which could be acceptable.

shaleh: https://github.com/leroivi/InfiniTime/pull/1

Let me know what you think. If you like it, you can merge it on your side which will then update this PR. Happy to iterate with you.

minacode:
Sounds like a good refactoring, but why is the size difference so huge? Would this be possible with less impact?

leroivi:
> Sounds like a good refactoring, but why is the size difference so huge? Would this be possible with less impact?

Ok, I don't know, I will investigate. I suspect that the impact comes from numerous usage of the Color class that calls inline ctor and then inline cast. Maybe it'll be solve by not inlining them.
Which impact would be acceptable ?

minacode:
> Which impact would be acceptable ?

Please don't get me wrong. For me it's less about setting a specific number and more about understanding what is happening there. 
I can imagine a unified interface to be more efficient than three different ones. But even if it's not, I would like us to think about it.

minacode:
One thing that also immediately comes to my mind is to use the theme everywhere and to challenge how many colors we really need. But that would better be a separate PR, because it may restyle the UI.

PR #2278, opened 2025-03-22
Feature/custom font

mzrinsky:
Create a Fonts:: namespace for fonts.
Abstract the font definitions from the front names, so they can be changed without search & replace.


mzrinsky:
sorry wrong repo

PR #2279, opened 2025-03-23
Fix about screen vertical centering.

mzrinsky:
There was a newline at the end of the last line on the third screen, which was causing the label to not be centered vertically. Removing it fixes the centering.

Here are some before and after screenshots for clarity.

InfiniSim_2025-03-23_055820
InfiniSim_2025-03-23_055918

Thank You!


mark9064 (approved):

PR #2289, opened 2025-04-25
feat: add persistance to sotpwatch

hank-milliken:
> **DISCLAIMER**: I have no idea how C++ works. This is pure trial & error craziness.

This is basically copy-cat of the https://github.com/InfiniTimeOrg/InfiniTime/pull/2141 only applied on top of current `main`.

Todo:
- [ ] Display time on the Watch Face


hank-milliken:
I rebased this after 16, the persistence is already implemented in `main` so only the UI changes are relevant.

PR #2294, opened 2025-05-12
Dark App Launcher Tiles

tituscmd:
This is a simple PR to replace the aqua colored tiles in the app launcher with darker ones, going hand-in-hand with my #2292
This is most likely just up to taste, but I thought the cyan tiles were outdated since that color isn't used anywhere else, so I thought a darker color would match the current system theme better.

Let me know what you think!

image
image
image

### DIFF
--- a/src/components/ble/MusicService.h
+++ b/src/components/ble/MusicService.h
@@ -71,9 +71,9 @@ namespace Pinetime {
 
       uint16_t eventHandle {};
 
-      std::string artistName {"Waiting for"};
-      std::string albumName {};
-      std::string trackName {"track information.."};
+      std::string trackName;
+      std::string albumName;
+      std::string artistName {"Not Playing"};
 
       bool playing {false};
 
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -129,15 +129,6 @@ void DisplayApp::Start(System::BootErrors error) {
 
   bootError = error;
 
-  lvgl.Init();
-  motorController.Init();
-
-  if (error == System::BootErrors::TouchController) {
-    LoadNewScreen(Apps::Error, DisplayApp::FullRefreshDirections::None);
-  } else {
-    LoadNewScreen(Apps::Clock, DisplayApp::FullRefreshDirections::None);
-  }
-
   if (pdPASS != xTaskCreate(DisplayApp::Process, "displayapp", 800, this, 0, &taskHandle)) {
     APP_ERROR_HANDLER(NRF_ERROR_NO_MEM);
   }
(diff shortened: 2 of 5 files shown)


liamcharger:
The issue I see with this is it's a matter of preference: some people may prefer the aqua, and some may prefer this gray. 

I'm not sure if this would align with the vision, but a setting to change the theme would be nice, with options for the original color, this gray, maybe a red, and so on.

mark9064:
I think we'd probably want to stick to one theme, as supporting multiple adds some extra overhead from a code maintenance perspective.

I think it's nice that the colours are synchronised, though I don't feel super strongly

src/displayapp/screens/Tile.cpp
```diff
@@ -70,8 +70,8 @@ Tile::Tile(uint8_t screenID,
   lv_obj_align(btnm1, nullptr, LV_ALIGN_CENTER, 0, 10);
 
   lv_obj_set_style_local_radius(btnm1, LV_BTNMATRIX_PART_BTN, LV_STATE_DEFAULT, 20);
-  lv_obj_set_style_local_bg_opa(btnm1, LV_BTNMATRIX_PART_BTN, LV_STATE_DEFAULT, LV_OPA_50);
-  lv_obj_set_style_local_bg_color(btnm1, LV_BTNMATRIX_PART_BTN, LV_STATE_DEFAULT, LV_COLOR_AQUA);
+  // lv_obj_set_style_local_bg_opa(btnm1, LV_BTNMATRIX_PART_BTN, LV_STATE_DEFAULT, LV_OPA_50);
```
mark9064, line 73: This line can be erased (in general there's no need to push commented out code, what has been removed will be visible in the diff)

mmind:
As I found out in #2301 , the application launcher does track an enabled state for the app icons (for when the resources are missing for example).

So nothing against changing the color, just mentioning that is should probably still allow seeing a disabled app icon.

mark9064:
@JF002 Any chance you can chime in, I got interested in InfiniTime long after this colour was chosen so wondering if there's any history here (and therefore maybe reasons to change/not to change it)

JF002:
I'm not sure I remember how this specific color was chosen, to be honest. We used to have maintainers who took care of the UI (themes, colors, layout,..) but they are not around anymore. 
I think there's no reason not to change it, especially if it improves the user experience and consistency between all InfiniTime applications.

mark9064 (approved):
All LGTM to me then :)

minacode:
I agree that using the theme here is the correct way to go. Changing the theme would be another discussion.

minacode (approved):

minacode:
Thank you! 🙂

tituscmd:
> Thank you! 🙂

You're very welcome! :)

PR #2270, opened 2025-03-13
Fix nrf sdk download links

dariusarnold:
It appears that NRF has moved the download to a new location. The old URL is now a 404. 
Update the build script and README to use the same URL the download website currently provides.

Fixes #2267

PR #2296, opened 2025-05-22
fix nrf sdk path

yakov-olkhovskiy:


mark9064:
I believe the current name is correct for the download, that resolves to a file while the change suggested here is a dead link. What problems are you having that are solved by this PR?

You've probably seen it already, but #2270 only got merged recently so perhaps you need to rebase/merge those changes in to your local version

yakov-olkhovskiy:
@mark9064 I just cloned this repo and tried to build in dev container (windows VSCode) and it failed not finding nrf sdk, then I looked into /opt and found it under different name, made that change and it builds ok now.

NeroBurner:
The upstream NRF-SDK download url and zip archive filename changed, which was fixed with https://github.com/InfiniTimeOrg/InfiniTime/pull/2270

But the archive contents stayed the same, with the "old" folder name

After #2270 we have basically the same docker-container as before the MR, but the `GetNrfSdk` function of the `build.sh` script is called again during firmware build time as the expected foldername for the SDK isn't the same as the zip filename:
https://github.com/InfiniTimeOrg/InfiniTime/blob/9fb35cc073452d20a8c106693d86ffd332a85032/docker/build.sh#L33

```sh
```

Then during the build the `buils.sh` script tries to execute `GetNrfSdk` again, which fails as the files already exist:
https://github.com/InfiniTimeOrg/InfiniTime/actions/runs/15215211133/job/42804040210#step:5:41

```
replace /opt/nRF5_SDK_15.3.0_59ac345/components/802_15_4/api/HAL/hal_atomic.h? [y]es, [n]o, [A]ll, [N]one, [r]ename:  NULL
```

NeroBurner:
created https://github.com/InfiniTimeOrg/InfiniTime/pull/2299 with a (hopefully) full fix

PR #2299, opened 2025-05-24
docker: fix NRF_SDK download and subsequent build.sh

NeroBurner:
The upstream NRF-SDK download url and zip archive filename changed, which was fixed with https://github.com/InfiniTimeOrg/InfiniTime/pull/2270

But the archive contents stayed the same, with the "old" folder name.

After #2270 we have basically the same docker-container as before the PR, but the `GetNrfSdk` function of the `build.sh` script is called again during firmware build time as the expected foldername for the SDK isn't the same as the zip filename:

```sh
```

Then during the build the `buils.sh` script tries to execute `GetNrfSdk` again, which fails as the files already exist resulting in the following error:

```
replace /opt/nRF5_SDK_15.3.0_59ac345/components/802_15_4/api/HAL/hal_atomic.h? [y]es, [n]o, [A]ll, [N]one, [r]ename:  NULL
```

Fix this by reverting the `NRF_SDK_VER` to the folder name in the zip archive and by some character replacement generate the download URL from the above (the download is in lower-case without the `_` and `.` characters).

Furthermore add safeguards to check after the `GetNrfSdk` call if the expected folder is really created. Then we have an error early during container image creation if the contents of the zip-archive are unexpected.

### DIFF
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -32,20 +32,8 @@ main() {
   mkdir -p "$TOOLS_DIR"
 
   [ ! -d "$TOOLS_DIR/$GCC_ARM_PATH" ] && GetGcc
-  if [ ! -d "$TOOLS_DIR/$GCC_ARM_PATH" ]; then
-    echo "missing GCC path: $TOOLS_DIR/$GCC_ARM_PATH"
-    return 1
-  fi
   [ ! -d "$TOOLS_DIR/$NRF_SDK_VER" ] && GetNrfSdk
-  if [ ! -d "$TOOLS_DIR/$NRF_SDK_VER" ]; then
-    echo "missing NRF_SDK path: $TOOLS_DIR/$NRF_SDK_VER"
-    return 1
-  fi
   [ ! -d "$TOOLS_DIR/mcuboot" ] && GetMcuBoot
-  if [ ! -d "$TOOLS_DIR/mcuboot" ]; then
-    echo "missing mcuboot path: $TOOLS_DIR/mcuboot"
-    return 1
-  fi
 
   mkdir -p "$BUILD_DIR"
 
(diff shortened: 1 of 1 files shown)


JF002 (commented):
LGTM!
Did you have the opportunity to test that the docker image builds correctly and that it can build infinitime?

Also, any idea why 2 CI checks fail?

docker/build.sh
```diff
@@ -17,7 +17,9 @@ export npm_config_cache="${NPM_DIR}"
 
 export BUILD_TYPE=${BUILD_TYPE:=Release}
 export GCC_ARM_VER=${GCC_ARM_VER:="10.3-2021.10"}
-export NRF_SDK_VER=${NRF_SDK_VER:="nrf5sdk153059ac345"}
+export NRF_SDK_VER=${NRF_SDK_VER:="nRF5_SDK_15.3.0_59ac345"}
+NRF_SDK_VER_SLUG=${NRF_SDK_VER,,}
+export NRF_SDK_VER_SLUG=${NRF_SDK_VER_SLUG//[_.]/}
```
mark9064, line 22: Maybe my bash is just bad :sweat_smile: but I find these quite hard to read, could you add a comment explaining what these do

mark9064: Sorry to nitpick, could the message also clarify the reason why (the download name is different from the folder name). I find the best comments always explain why the code is doing what it is doing

NeroBurner: Sure!

```suggestion
# convert to lower case and remove _ and . character
# the download URL uses the SLUG, but the extracted folder is named like the original value
NRF_SDK_VER_SLUG=${NRF_SDK_VER,,}
export NRF_SDK_VER_SLUG=${NRF_SDK_VER_SLUG//[_.]/}
```

docker/build.sh
```diff
@@ -17,7 +17,12 @@ export npm_config_cache="${NPM_DIR}"
 
 export BUILD_TYPE=${BUILD_TYPE:=Release}
 export GCC_ARM_VER=${GCC_ARM_VER:="10.3-2021.10"}
-export NRF_SDK_VER=${NRF_SDK_VER:="nrf5sdk153059ac345"}
+export NRF_SDK_VER=${NRF_SDK_VER:="nRF5_SDK_15.3.0_59ac345"}
+# convert to lower case and remove _ and . character
+# convert to lower case and remove _ and . character
```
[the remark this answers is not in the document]
NeroBurner: Lol 😅 

```suggestion
```

mark9064:
https://github.com/InfiniTimeOrg/InfiniTime/actions/runs/15285543037/job/42994956633

NeroBurner:
> https://github.com/InfiniTimeOrg/InfiniTime/actions/runs/15285543037/job/42994956633

Failing size job fixed by using Nash AS shell for the job

https://github.com/InfiniTimeOrg/InfiniTime/commit/c3295d6d2a325f9a7418b15b943342635026926e

PR #2301, opened 2025-05-26
navigation: fix greying out the app icon if not enabled

mmind:
Commit 0aead42fdf51 ("navigation: Add is available (#1847)") added the ability to draw the app icon in grey and in a disabled state when some prerequisits were not met. Only the Navigation app was using this mechanism due to its icons being stored in the external memory and possibly missing.

Commit 63e0c4f4efb0 ("Application selection at build time") broke this by always setting the state as true:
      for (const auto& userApp : userApps) {
        apps[i++] = Screens::Tile::Applications {userApp.icon, userApp.app, true};
      }

Fix this by creating an isAvailable() strcuture in the app classes, similar
to how the Watchfaces handle the same problem of checking availability.

### DIFF
--- a/src/displayapp/screens/Navigation.h
+++ b/src/displayapp/screens/Navigation.h
@@ -67,6 +67,10 @@ namespace Pinetime {
       static Screens::Screen* Create(AppControllers& controllers) {
         return new Screens::Navigation(*controllers.navigationService);
       };
+
+      static bool IsAvailable(Pinetime::Controllers::FS& filesystem) {
+        return Screens::Navigation::IsAvailable(filesystem);
+      };
     };
   }
 }
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -517,8 +517,8 @@ void DisplayApp::LoadScreen(Apps app, DisplayApp::FullRefreshDirections directio
   switch (app) {
     case Apps::Launcher: {
       std::array<Screens::Tile::Applications, UserAppTypes::Count> apps;
-      std::ranges::transform(userApps, apps.begin(), [](const auto& userApp) {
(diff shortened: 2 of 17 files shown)


mmind:
@NeroBurner sure, I can do that :-) .

I stumbled upon that thing while doing the other navigation stuff and InfiniSim not showing the flags, but also did not have a grasp on how the whole thing works.

So you're definitly right, doing that only for Navigation does not scale :-) and also thanks for pointing out how things should look correctly.

mmind:
Hopefully that is what you had in mind :-) ... Follows now how the Wachfaces look

NeroBurner (approved):
I think you got it right. Good job!

src/displayapp/screens/Navigation.h
```diff
@@ -67,6 +67,10 @@ namespace Pinetime {
       static Screens::Screen* Create(AppControllers& controllers) {
         return new Screens::Navigation(*controllers.navigationService);
       };
+
+      static bool IsAvailable(Pinetime::Controllers::FS& filesystem) {
+        return Applications::Screens::Navigation::IsAvailable(filesystem);
```
mark9064:
```suggestion
        return Screens::Navigation::IsAvailable(filesystem);
```

mark9064 (approved):
Looks great, thanks for the PR :)

src/displayapp/screens/Navigation.h
mark9064: Hey, if you'd like I can apply this for you and merge?

mmind: Hi @mark9064 

sorry, was on vacation and didn't find the time to get into Infitime yet.
I've now amended the commit with the suggested change :-)

PR #2303, opened 2025-05-26
Navigation style updates

mmind:
Some updates for the navigation display coming from having a PineTime strapped to the handlebar off a motorbike and using it for navigation :-) .

* For one, dark green distance on black background is hard to read, so made it yellow
* Action-Icon and distance are the values needed at a glance and often, so having them on opposite ends of the screen makes it harder to just glance at them, so swapped distance and narrative element
* The current layout creates a lot of empty space around the elements, where the PineTime already has a black area in its frame around the display. So moved everything a bit more together, which in turn makes room for a 4th line of narrative, because sometimes the narrative OSMand provides can be a bit unwieldy :-)
* And lvgl actually provides nice layout alignment variants, so we don't need hardcode some pixel-values from the center, but can instead align elements to each other.

After the change, the display looks like:
navigation-after-reorder

Of course style is always perceived differently, so I'm of course open for suggestions.

### DIFF
--- a/src/displayapp/screens/Navigation.cpp
+++ b/src/displayapp/screens/Navigation.cpp
@@ -202,14 +202,6 @@ Navigation::Navigation(Pinetime::Controllers::NavigationService& nav) : navServi
   lv_obj_set_style_local_image_recolor(imgFlag, LV_IMG_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_CYAN);
   lv_obj_align(imgFlag, nullptr, LV_ALIGN_IN_TOP_MID, 0, 2);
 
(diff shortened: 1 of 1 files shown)


mark9064:
Rebase to fix build CI, sorry (same for other PRs)

NeroBurner:
one last rebase please. The GitHub Actions are now fixed. Sorry for the inconvenience

mmind:
> one last rebase please. The GitHub Actions are now fixed. Sorry for the inconvenience

No worries. I noticed the base-ref-size change in main, and tried my luck with #2301 first :-)

PR #2308, opened 2025-06-07
CMake Presets json file

sillydan1:
This PR adds a `CMakePresets.json` with the `Release` preset , which makes it much easier to get started on development, as code-editors such as VSCode, CLion and neovim support generating `compile_commands.json` databases using these presets. It does require a fairly recent cmake version though.

I only added one preset, `Release` for now and I didn't dare touch the CI stuff, but perhaps this could simplify the CI chain.

I also found that the `requirements.txt` file in `tools/mcuboot` didn't mention the `adafruit-nrfutil` dependency, so I added that to the build instructions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,10 @@ src/arm-none-eabi
 node_modules
 package.json
 package-lock.json
+
+# Toolchain and External dependencies
+sdk-toolchain/
+
+# Developer files
+CMakeUserPresets.json
+compile_commands.json
--- a/doc/buildAndProgram.md
+++ b/doc/buildAndProgram.md
@@ -12,9 +12,9 @@ To build this project, you'll need:
     python -m venv .venv
     source .venv/bin/activate
     python -m pip install wheel
-    python -m pip install -r tools/mcuboot/requirements.txt
+    python -m pip install -r tools/mcuboot/requirements.txt adafruit-nrfutil
     ```
-- A reasonably recent version of CMake (I use 3.16.5)
+- A reasonably recent version of CMake (minimum 3.10)
 - lv_font_conv, to generate the font .c files
   - see [lv_font_conv](https://github.com/lvgl/lv_font_conv#install-the-script)
   - install npm (commonly done via the package manager, ensure node's version is at least 12)
@@ -24,7 +24,7 @@ To build this project, you'll need:
 
 ### Clone the repo
 
-```
+```sh
 git clone https://github.com/InfiniTimeOrg/InfiniTime.git
 cd InfiniTime
 git submodule update --init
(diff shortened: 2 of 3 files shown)


doc/buildAndProgram.md
```diff
@@ -34,6 +34,40 @@ cd build
 
 ### Project generation using CMake
 
+#### CMakePresets
+
+Note that this requires CMake 3.19 or later to be installed.
+This project provides a `CMakePresets.json` file, enabling easy and fast development with modern editors such as VSCode, neovim or CLion.
+The `CMakePresets.json` file looks for the arm-gcc toolchaing and NRF52 SDK in the `sdk-toolchain` directory.
```
mark9064:
```suggestion
The `CMakePresets.json` file looks for the arm-gcc toolchain and NRF52 SDK in the `sdk-toolchain` directory.
```

doc/buildAndProgram.md
```diff
@@ -34,6 +34,40 @@ cd build
...
+So to download and exctract these in the correct directory, do the following:
```
mark9064:
```suggestion
So to download and extract these in the correct directory, do the following:
```

doc/buildAndProgram.md
```diff
@@ -34,6 +34,40 @@ cd build
...
+
+```sh
+mkdir -p sdk-toolchain
+
+# Get the nRF5 SDK
+wget https://developer.nordicsemi.com/nRF5_SDK/nRF5_SDK_v15.x.x/nRF5_SDK_15.3.0_59ac345.zip
```
mark9064: This link was previously dead (#2270), but it seems to be alive again now
I think it'd better if the "new" link is here used also so it's consistent

.gitignore
```diff
@@ -50,3 +50,13 @@ src/arm-none-eabi
 
 # clangd
 .cache/
+
+# Toolchain and External dependencies
+sdk-toolchain/
+
+# Developer files
+CMakeUserPresets.json
+compile_commands.json
```
mark9064, line 64: Is this needed? From what I understand compile_commands always lives in the `build` folder which is already gitignore'd

mark9064 (commented):
Looks like a nice improvement overall :) Thanks for sending it in

I'm not sure about storing the toolchain inside the tree. Maybe environment variables like with the [VSCode instructions](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/doc/buildWithVScode.md#environment-setup) would be more flexible? What are your thoughts, I do agree that having to configure env vars can be a bit annoying

.gitignore
sillydan1: Some tools put it in the project root... At least [the one I use](https://github.com/Civitasv/cmake-tools.nvim) does

sillydan1:
> Looks like a nice improvement overall :) Thanks for sending it in
> 
> I'm not sure about storing the toolchain inside the tree. Maybe environment variables like with the [VSCode instructions](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/doc/buildWithVScode.md#environment-setup) would be more flexible? What are your thoughts, I do agree that having to configure env vars can be a bit annoying

Thanks! :grin: Hmm... I see your point that using raw environment variables is very flexible, I just found it a bit annoying to use though - which is why I added the cmake presets file... These changes will technically not prevent users from manually setting the environment variables and calling `cmake` without the `--preset` option, but then that means we need to support multiple workflows. Sadly, there's no way to "use the environment variable if defined, otherwise, use the `${fileDir}/sdk-toolchain` value(s)"... At least I haven't been able to figure it out without needing to compromise on flexibility...

To me, ideally, the buildsteps should just be:
 - Clone the project (with submodules)
 - `cmake --preset Release`
 - `cmake --build build/Release`

Where the CMakeLists.txt would download the SDK and toolchain _for you_ at configure-time. I would love to try to get that done but it would require touching the CI and it would be a fairly big change to the now normal work flow, so I'm not sure if it would be too disruptive. Perhaps adding the presets just adds confusion and we should just close this PR?

sillydan1:
Closing as this became stale.

PR #2309, opened 2025-06-07
Remove leading zero for HRM value

tituscmd:
Hello everybody!

This is another simple PR of mine.
This PR removes the leading before the BPM value in the Heart Rate app. Previously it would say "083", but now it looks like this:
bpm
I think this is cleaner than _with_ the leading zero, but it might be up to preference.

Let me know what you think!


minacode (approved):

mark9064:
Should we change the default `---` since it's no longer always 3 characters? Or maybe it's a complete non-issue and looks fine in practice, would be interested to hear :)

tituscmd:
> Should we change the default `---` since it's no longer always 3 characters? Or maybe it's a complete non-issue and looks fine in practice, would be interested to hear :)

imageimage
In my custom version of this UI, I also use two dashes instead of the three. The reason being that, most of the time, your heart rate will likely be double digit.

minacode:
Or just one? I think, I am fine with any of the variants.

tituscmd:
> Or just one? I think, I am fine with any of the variants.

I think two digits "--" would be the most logical approach here, because that's the amount of digits your heart rate is going to be most of the time.
If that makes sense to y'all, I could definitely open another PR to replace every 0 (in the context of heart rate) with "--"

minacode:
Okay, let's try two and get this merged! 🙂

PR #2236, opened 2025-01-26
BatteryInfo App Redesign

tituscmd:
This PR is a redesign of the BatteryInfo app.

The colors I used are as follows:
Discharging (default): green
Charging: lime
Battery full: blue
Battery low (<10%): red

Here's a pic of it in InfiniEmu, as well as how each state looks like on-device.
emu
discharging
charging
fully_charged
image
(Can't manage to rescale this image on iOS, I'll get to it later)

PR #2310, opened 2025-06-10
BatteryInfo App Redesign: Fixup

tituscmd:
This is a simple little fixup for my PR #2236

It seems I have overlooked some things in my original PR, so here's to fixing those mistakes! ;D

- I removed the leading zero for the battery percentage, as that is not commonly used at all
- I also changed the coloring of the arc to match the color thresholds for the battery icon, so that they match

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,9 +39,12 @@ jobs:
           apt-get -y install --no-install-recommends python3-pil
       - name: Build
         shell: bash
-        run: /opt/build.sh all
+        run: |
+          git config --global --add safe.directory /__w/InfiniTime/InfiniTime
+          /opt/build.sh all
       - name: Output build size
         id: output-sizes
+        shell: bash
         run: |
           . /opt/build.sh
           .github/workflows/getSize.sh "$BUILD_DIR"/src/pinetime-app-*.out >> $GITHUB_OUTPUT
@@ -159,6 +162,7 @@ jobs:
 
     - name: Output build size
       id: output-sizes
+      shell: bash
       run: |
         . /opt/build.sh
         .github/workflows/getSize.sh "$BUILD_DIR"/src/pinetime-app-*.out >> $GITHUB_OUTPUT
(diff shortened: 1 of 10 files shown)


src/displayapp/screens/BatteryInfo.cpp
```diff
@@ -60,15 +60,18 @@ void BatteryInfo::Refresh() {
...
-    lv_label_set_text_static(status, "Battery low");
-  } else {
+  } else if (batteryPercent > 15) {
     lv_obj_set_style_local_line_color(chargingArc, LV_ARC_PART_INDIC, LV_STATE_DEFAULT, LV_COLOR_GREEN);
     lv_label_set_text_static(status, "Discharging");
+  } else if (batteryPercent > 5) {
+    lv_obj_set_style_local_line_color(chargingArc, LV_ARC_PART_INDIC, LV_STATE_DEFAULT, LV_COLOR_ORANGE);
+    lv_label_set_text_static(status, "Battery low");
+  } else {
+    lv_obj_set_style_local_line_color(chargingArc, LV_ARC_PART_INDIC, LV_STATE_DEFAULT, Colors::deepOrange);
```
NeroBurner, line 70: Critical was red before. Any reasoning why it is deep-orange now?

tituscmd: I'm using the same color scheme that the battery icon uses:

```
if (colorOnLowBattery) {
    static constexpr int lowBatteryThreshold = 15;
    static constexpr int criticalBatteryThreshold = 5;
    if (percentage > lowBatteryThreshold) {
      SetColor(LV_COLOR_WHITE);
    } else if (percentage > criticalBatteryThreshold) {
      SetColor(LV_COLOR_ORANGE);
    } else {
      SetColor(Colors::deepOrange);
    }
  }
```

I wouldn't mind going back to red, but this way it always matches the battery icon.

mark9064: The "deep orange" colour is basically red anyway, if you want a demo you can set the battery voltage in infiniemu with the slider at the bottom, reset the simulation after changing it to force a battery measurement. I think it looks good

tituscmd: > if you want a demo you can set the battery voltage in infiniemu with the slider at the bottom, reset the simulation after changing it to force a battery measurement.

Here are some screenshots:
Screenshot 2025-06-12 at 00-03-12 InfiniEmu
grafik

NeroBurner: Thanks! All good

PR #2312, opened 2025-06-11
set static text as static text

lmamane:
and save a few bytes of heap memory

### DIFF
--- a/src/displayapp/screens/settings/SettingWakeUp.cpp
+++ b/src/displayapp/screens/settings/SettingWakeUp.cpp
@@ -8,8 +8,6 @@
 
 using namespace Pinetime::Applications::Screens;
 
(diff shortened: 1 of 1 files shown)


src/displayapp/screens/settings/SettingWakeUp.cpp
```diff
@@ -8,8 +8,6 @@
 
 using namespace Pinetime::Applications::Screens;
 
-constexpr std::array<SettingWakeUp::Option, 5> SettingWakeUp::options;
```
NeroBurner, line 11: Why can this be removed?

mark9064: I think as it's an unnecessary forward declaration. It's already provided by SettingWakeUp.h

mark9064 (approved):

PR #2313, opened 2025-06-13
Ignoring npm files

owenfromcanada:
Added files created from npm to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,8 @@ src/arm-none-eabi
 
 # clangd
 .cache/
+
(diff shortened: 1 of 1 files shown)


mark9064 (approved):
Perhaps we should ship a `package-lock` to make installs more reproducible but this is certainly an improvement compared to the git tree disaster `node_modules` creates at the moment

owenfromcanada:
Fair enough, I don't know that much about npm or package-lock, but figured we'd want to avoid committing them.

PR #2318, opened 2025-06-20
Garden watch face

owenfromcanada:
This is a new watch face that creates a "garden" that grows throughout the day.  The flowers are different every day (chosen using a PRNG, so they're still deterministic).  It's based on the Digital watch face, and includes all the same information.

There are four commits here:
1. Adding a return type to the `StatusIcons::Update` method.  This is because I wanted to put the notification icon next to the status icons, and wanted to know when we need to call `lv_obj_realign` on it (i.e. when the status icons have changed).  This seems to fit with the overall approach, and shouldn't impact any existing code.
2. Adding the Azeret Mono font
3. Implementation (and integration) of the watch face
4. Addition of the new watch face to the documentation


owenfromcanada:
Corresponding discussion #2317

owenfromcanada:
Added a couple extra commits for clang-format results.

Side note: is there a way to run those same checks locally?  It looks like the `tests/test-format.sh` is only meant to be run through the GitHub workflow.  I have a clang extension in Code OSS, but it didn't flag the same issues.  Is there something others are using locally to test before pushing commits?

mark9064:
Yeah you should be able to run `clang-format` (optionally with `-i` for inplace) from the repository root and give it the path to any file IIRC

You will need to set the driver path for the clangd extension to work properly, but if you want to run clang-tidy on a specific file you can use the invocation in the comments of #2139. It's pretty slow though - the caching clangd engine is much faster

mark9064:
Oh, and you might also want to enable the pre-commit hook that automatically runs clang-format when you commit

PR #2320, opened 2025-06-22
New App: Tally

owenfromcanada:
As requested by my wife--for anyone who forgets where they were when counting things.

Basic functionality is tap-to-count, but shake-to-count can be enabled (using the same calibration as shake-to-wake).  Includes a delay to avoid accidental double counting, and helpful "toast" messages when options are changed.

Note that the added Adwaita Mono font is for the Tally 5 character (𝍸) which I used as the app icon.  This symbol is only implemented in a bare handful of existing fonts, but Adwaita is released under the OFL so should be fine to use.


src/displayapp/screens/Tally.cpp
```diff
@@ -0,0 +1,186 @@
...
+#include "displayapp/screens/Screen.h"
+#include "displayapp/screens/Symbols.h"
+#include "components/settings/Settings.h"
+#include "components/motion/MotionController.h"
+#include "components/motor/MotorController.h"
+
+using namespace Pinetime::Applications::Screens;
+
+namespace {
+  void countButtonEventHandler(lv_obj_t* obj, lv_event_t event) {
```
mark9064: Functions should be PascalCase

src/displayapp/screens/Tally.h
```diff
@@ -0,0 +1,75 @@
...
+        void Increment();
+        void Reset();
+        void ToggleShakeToCount();
+        void ToggleKeepAwake();
+
+      private:
+        static const int PERIOD = 100;               // milliseconds
+        static const int INCREMENT_DELAY_TIME = 100; // milliseconds
+        static const int SHAKE_DELAY_TIME = 200;     // milliseconds
+        static const int MESSAGE_TIME = 3000;        // milliseconds
```
mark9064: These can be `static constexpr`. Also, by InfiniTime convention constants are still camelCase (I would rather they be upper case too)

src/displayapp/screens/Tally.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+  UpdateCount();
+
+  refreshTask = lv_task_create(RefreshTaskCallback, PERIOD, LV_TASK_PRIO_MID, this);
+}
+
+Tally::~Tally() {
+  if (keepAwakeEnabled) {
+    wakeLock.Release();
+  }
```
mark9064: Wakelock destructor handles this for you

src/displayapp/screens/Tally.h
```diff
@@ -0,0 +1,75 @@
...
+        lv_obj_t* shakeLabel;
+        lv_obj_t* awakeButton;
+        lv_obj_t* awakeLabel;
+        int count = 0;
+        bool shakeToCountEnabled = false;
+        bool shakeToWakeTempEnable = false;
+        bool keepAwakeEnabled = false;
+        int incrementDelay = 0;
+        int shakeToCountDelay = 0;
+        int messageTimer = 0;
```
mark9064: Consider using fixed width unsigned integers instead of plain `int`. It looks like you don't need negative numbers here, and fixed size integers can be smaller than an int which saves memory. Sizing integers to their minimum is less important for an app as you know nothing else will be on the screen at the same time, but it's still good practice to be minimal. It looks like most of these could be `uint16_t`, though if you're patient maybe count should stay as `uint32_t` :)

If you retype these, it will probably make sense to retype the above constants as well

src/displayapp/screens/Tally.cpp
```diff
@@ -0,0 +1,186 @@
...
+        Increment();
+      }
+    } else if (shakeToCountDelay > 0) {
+      shakeToCountDelay--;
+    }
+  }
+
+  if (incrementDelay > 0) {
+    incrementDelay--;
+  }
```
mark9064, line 112: These delay loops are nice and simple - no need to change anything

As extra information which might be helpful another time: if the system is running slowly `Refresh()` will get called slower than expected, so you may prefer to store the actual time using `xTaskGetTickCount()` and then calculate the time since in order to decide when timers expire. Certainly when resolution or accuracy is more important I would do this rather than use delay loops. (You could also replace them entirely with the `Timer` class and use callbacks, but this gets very complicated and is only really useful when you want to deliver events when the app is closed)

src/displayapp/screens/Tally.cpp
```diff
@@ -0,0 +1,186 @@
...
+  lv_obj_align(countButton, nullptr, LV_ALIGN_IN_TOP_MID, 0, 0);
+  countLabel = lv_label_create(lv_scr_act(), nullptr);
+  lv_obj_set_style_local_text_font(countLabel, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, &open_sans_light);
+  lv_obj_align(countLabel, nullptr, LV_ALIGN_IN_TOP_MID, 0, 30);
+
+  messageLabel = lv_label_create(lv_scr_act(), nullptr);
+  lv_obj_align(messageLabel, nullptr, LV_ALIGN_IN_BOTTOM_MID, 0, -65);
+  SetMessage("Tap to count");
+
+  resetButton = lv_btn_create(lv_scr_act(), nullptr);
```
mark9064, line 61: I haven't tried this on hardware, but looking at InfiniEmu these buttons seem quite short. Maybe they could be a bit taller as they'd be easier to press

mark9064 (commented):
Thanks for sending this in :)

Code looks really good overall, just a few comments. Haven't run clang-tidy yet

src/displayapp/screens/Tally.h
owenfromcanada: Thanks for all the feedback!  For "plain" constants like this, is there a reason to use `constexpr` instead of `const`?  Is it just a best practice, or is there a functional difference?

src/displayapp/screens/Tally.h
owenfromcanada: Yeah, these could be a smaller type, but by my understanding, there's sometimes a performance difference when using smaller types than the default `int` size (depending on how things are packed, you can end up with an additional shift operation).  But it might be negligible, not sure about this hardware.

In any case, looking through the other apps and code, it seems like we generally use explicit sizing, so I'll change those over to match.

src/displayapp/screens/Tally.cpp
owenfromcanada: Thanks for this!  I considered using `xTaskGetTickCount` but figured the simpler method (which I shamelessly copied from the Dice app) would be fine.  And in this case, it's delaying an additional detection in the `Refresh` function, so the simple method might actually work better for what I'm attempting to do.  But good to keep in mind!

src/displayapp/screens/Tally.h
mark9064: Yeah generally InfiniTime is not too bothered about performance compared to memory savings, so the overhead from having to implement the wraparound semantics of smaller types tends not to be a concern. If performance is critical it may be worth considering but at that point you'd probably rather set that file/subsection to O3 rather than Os and you'll probably get a lot more. In this case neither the performance or memory matter directly really so it's all a moot point 😅 but generally I'd go in the direction of memory rather than performance

I'm sure you've noticed that InfiniTime overall is really memory bound, so saving a few bytes here and a few there generally is worth it for static memory / long-lifetime heap allocations

src/displayapp/screens/Tally.h
mark9064: I believe in this case (with a modern compiler) it's the same, but the strong guarantees of `constexpr` are much more predictable which means it's definitely the go-to choice to express a compile time constant IMO

mark9064:
One other thing: I'm guessing this is designed to go to 999 but it seems that the numbers are slightly cut off when showing 3 digits. Maybe they could be a bit smaller so they can keep a bit of margin even at 3 digits? I think the next font size down is a lot smaller though - what do you think?

owenfromcanada:
Good call--I mostly anticipated usage for lower values, but I pushed another commit that changes the font when it reaches 100.  This way it's still large for (what I'm guessing is) the majority of use cases, but will still work well above that.

src/displayapp/screens/Tally.h
```diff
@@ -0,0 +1,75 @@
+#pragma once
+
+#include "displayapp/apps/Apps.h"
+#include "displayapp/screens/Screen.h"
+#include "displayapp/Controllers.h"
+#include "systemtask/SystemTask.h"
+#include "systemtask/WakeLock.h"
+#include "Symbols.h"
+
+#include <array>
```
mark9064: Unused

owenfromcanada: Fixed

Saarsk:
> Good call--I mostly anticipated usage for lower values, but I pushed another commit that changes the font when it reaches 100. This way it's still large for (what I'm guessing is) the majority of use cases, but will still work well above that.

Hehe, I counted to 863 on my Pebble smartwatch just two days ago, with a tally counter app. 😅 So it's great that it'll not stop at 100. Four digis support would be nice too but I understand if you don't wanna do that. 😊

This is a great addition to the app suite! 🙂

PR #1718, opened 2023-03-31
Heartrate measurements in background

patricgruber:
This implements heart rate measurements when the screen is turned off.
The ticket I found for this is: https://github.com/InfiniTimeOrg/InfiniTime/issues/183

When starting the heart rate measurements through the HeartRate screen and turning off the screen, the heart rate task doesn't stop but keeps running in the background until the heart rate measurement is stopped through the screen again.
The task wait delay is set to ~10 seconds (10k ticks) so the task doesn't run all the time and drains the battery too much.

Already tested it on my PineTime and works great.
Right now it was more a proof of concept, therefore the interval between background measurements is hard-coded to ~5 minutes. But I'd be happy to implement a settings screen to configure the interval or other features that are wanted.

PR #2322, opened 2025-06-26
Background HR

mark9064:
Largely based on #1718

Changes:
- Simplified to 4 states
- Reworked display of heart rate to avoid stale values being shown for a long time in some scenarios
  - Note: this does not cover all scenarios, but I think it's good enough and as close to the current behaviour as possible while still making sense for background measurement
- Constant frequency measurements that take into account successful screen on measurements (and also long periods of unsuccessful screen on measurements)
- Refactored away a few annoying bits of the PPG code like `lastBpm` and the bodge to get correct not enough data messages
- Removed the 15s mode. The sensor already takes ~6s to get a measurement in ideal conditions so realistically it makes sense to use continuous - the sensor on time will be very similar and you benefit from proper history averaging with continuous so the measurement accuracy is much better
- Misc code bits such as using std::optional more where sensible. I haven't used it in the settings struct as the layout is compiler dependent AFAIK

Additionally, I've refactored the sampling to be constant frequency instead of relying on delays which may sleep for longer than requested. This should improve sampling accuracy. As part of this I've also raised the priority of the HR task to 1 so that it does not get out-competed by DisplayApp when it is busy

I'll admit this is still more complex than I'd like it to be. I promise it's a tricky problem to solve 😅


mark9064:
Sim fix should be trivial, just a method rename it seems

tituscmd:
Merged this onto my main device and will try it out compared to the "old"(er) system. Anything to look out for in particular?

mark9064:
I'm most interested in hearing whether it feels natural in terms of how it behaves when you have screen on measurements interacting with background measurements. Also any bugs ofc :)
Thanks for testing!

tituscmd:
Two things I can definitely say already:

1. I'm a big fan of the feature that, after about 30 seconds with not enough data, the measurement gets set to _failed_. That way the amount of wrong measurements has been reduced a good bit
2. I don't know if this is actually a causation, but I haven't had a reboot in 19hrs on a busy day like today where I usually get a reboot every few hours. This might have been fixed by a bug that was in the old version of background HR, or maybe I messed up the merge when I first merged background HR onto my device (long ago, more towards the beginning of my PineTime journey so I was definitely less experienced and the chances of me having messed up there aren't small)

I'll keep an eye out on 2. and on bugs in general of course.


PS: Now that measurements can fail, the value will be set to 0 and look awkward on the watch face. I solved it like this and I think this might be a good idea going forward (of course, again, the amount of dashes (-) is debatable)
image

tituscmd:
Thinking about it, it would actually make a lot of sense if this updated background HR fixes my reboots:

Most of the reboots I have encountered have been when I am moving. I usually get a lot on the bike, and not so many just sitting at home. Sometimes when I'm sick and don't go out at all, I noticed uptimes of up to multiple days. All of this points to the idea that the reboots have something to do with me moving.
I originally thought that something was up with the accelerometer - of course I did, because the accelerometer is specifically for movement. But now that I thought about the updated background HR more today, maybe that's not the case.

The HR sensor on PineTime is notoriously not very good, if it works at all, when moving. That leads me to believe, that maybe the background HR was responsible for my reboots.
The new version of background HR (this PR) has the fail save function to mark measurements as _failed_ when there hasn't been enough data for 30 seconds.
On the old version, I could imagine me being on the bike, a background measurement gets called and, since I'm moving so much, it can't gather enough data and gets stuck in that measurement loop and eventually crashes.

So maybe the fail save function in this PR has fixed that. Has anybody noticed similar behavior?

mark9064:
> PS: Now that measurements can fail, the value will be set to 0 and look awkward on the watch face. I solved it like this and I think this might be a good idea going forward (of course, again, the amount of dashes (-) is debatable)

I'd very much welcome a PR which replaces zero with - everywhere. And TBH we should also refactor it all to expect `std::optional<uint8_t>` rather than `uint8_t` so we can discard the hacky zero value altogether

WhyNotHugo:
> > PS: Now that measurements can fail, the value will be set to 0 and look awkward on the watch face. I solved it like this and I think this might be a good idea going forward (of course, again, the amount of dashes (-) is debatable)
> 
> I'd very much welcome a PR which replaces zero with - everywhere. And TBH we should also refactor it all to expect `std::optional<uint8_t>` rather than `uint8_t` so we can discard the hacky zero value altogether

See: https://github.com/InfiniTimeOrg/InfiniTime/pull/2342

src/heartratetask/HeartRateTask.cpp
```diff
@@ -1,19 +1,104 @@
...
+  return xTaskGetTickCount() - lastMeasurementTime >= backgroundPeriod.value();
+};
+
+TickType_t HeartRateTask::CurrentTaskDelay() {
+  auto backgroundPeriod = BackgroundMeasurementInterval();
+  TickType_t currentTime = xTaskGetTickCount();
+  auto CalculateSleepTicks = [&]() {
+    TickType_t elapsed = currentTime - measurementStartTime;
+
+    // Target system tick is the the elapsed sensor ticks multiplied by the sensor tick duration (i.e. the elapsed time)
```
JF002: Type : `the` is written twice.

mark9064: Good catch, thanks

mark9064:
OK should be good to go now. Will try get InfiniSim PR up soon

WhyNotHugo:
@mark9064 https://github.com/InfiniTimeOrg/InfiniSim/pull/180 might serve as a base :)

Koman10:
@JF002 Hoi, ik zou graag even privé met je praten. Heb je een momentje?

mark9064:
Sim PR is up https://github.com/InfiniTimeOrg/InfiniSim/pull/184

Compiles locally
Merge this first, then bump the sim InfiniTime version? (@NeroBurner let me know what works best for you, I don't use the sim too often :sweat_smile: )

NeroBurner:
either make the sim work for both code versions or I'll merge the sim PR after this one is merged

arjan-s:
Due to this PR being merged, I built InfiniTime myself from current master (7128fc0) to test it. I enabled background HR from the settings, but I never got results in Gadgetbridge, whatever interval (even continuous) I selected.
Looking at the bottom of the watch, the green led doesn't even light up like it does during manual measurements. Do I need more PRs merged or settings changed to get this to work?

mark9064:
HR needs to be turned enabled for background measurement to happen - do you have it turned on in the heart rate app?

arjan-s:
Oh, that's it! Thanks!

It wasn't clear to me that it isn't enabled automatically when the background hr setting is enabled. Perhaps that would be a good improvement? Or at least a warning that it needs to be enabled from the hr app as well?

mark9064:
Glad it's working :)

Maybe a warning would be good? Supporting background measurements with no screen on feels a bit weird - how should we communicate that a heart rate value on screen is not being currently measured?

PR #2323, opened 2025-06-27
Tooling fixes

mark9064:
First commit fixes build for GCC 14

Second one adds nano specs to the compiler flags (needed as it sets compiler options) and uses nosys specs rather than linking manually. This made the group entries redundant also. Nice explanation of the spec files: https://metebalci.com/blog/demystifying-arm-gnu-toolchain-specs-nano-and-nosys/

### DIFF
--- a/gcc_nrf52-mcuboot.ld
+++ b/gcc_nrf52-mcuboot.ld
@@ -18,7 +18,6 @@
  */
 
 SEARCH_DIR(.)
-GROUP(-lgcc -lc -lnosys)
 
 MCUBOOT_SIZE = 0x8000;
 SCRATCH_SIZE = 0x1000;
--- a/gcc_nrf52.ld
+++ b/gcc_nrf52.ld
@@ -24,7 +24,6 @@
  */
 
 SEARCH_DIR(.)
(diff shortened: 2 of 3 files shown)


shaleh:
Why does GCC14 matter when we are using a much older ARM cross compiler chain? Or is this an attempt to move the cross compile chain to a more recent version?

mark9064:
Yeah exactly, this enables us to upgrade the compiler later. Also it's more semantically correct anyway, so it's a win-win

JF002 (approved):
Great, thank you for those changes! Are there any other change needed before we can upgrade the compiler?

mark9064:
No other changes I think. Been using GCC 14.2 locally without problems (14.3 out now but should be fine as well)

mark9064:
One exception: newer compilers generate the linker warnings about the function stubs. These stubs are later removed by linker stripping as they're unused, but there's no way to disable the warning without either:
- Shipping our own stubs without warnings
- Removing the warnings directly from the .a being linked against (one liner command that needs to be run once for each compiler install)
- Doing some shell scripting on the linker output to cut the warnings out

or we could just put up with them

Issue #2189, opened 2024-12-08
Where the issues about bugs on the weather service, a double tap feature and the pong game and went?


tokiwee:
Where the issues about bugs on the weather service, a double tap feature and the pong game and disappeared??

mark9064:
I can assure you that no issues have been deleted or otherwise removed by the InfiniTime team. If you have not deleted them yourself, you will need to get in touch with GitHub support

vkareh:
yeah that's weird... I definitely have those threads in my email:
![Screenshot at 2024-12-09 16-58-51](https://github.com/user-attachments/assets/6f4c1b75-28d9-4dd7-a39a-3b94a7b360a3)

but trying to get there I get a 404:
![Screenshot at 2024-12-09 16-57-53](https://github.com/user-attachments/assets/46c6f06c-8a30-4770-bdb0-ab8cbb321f0d)

JF002:
As @mark9064 said, we haven't removed any of those issues. Github might have an anti-spam feature that removed them for some reasons?
Anyway, I suggest you contact Github support about this.
Feel also free to create them (maybe just one to begin with) to see what happens.

PR #2325, opened 2025-06-27
Touchscreen timeout fix

mark9064:
Fixes part of #2189

Still have no clue why github deleted the original issue??


src/drivers/Cst816s.cpp
```diff
@@ -56,28 +56,33 @@ bool Cst816S::Init() {
...
+  // The reset kills the current touch point, so paint/pong etc breaks
+  twiMaster.Write(twiAddress, 0xFB, 0, 1);
+
   return true;
 }
 
 Cst816S::TouchInfos Cst816S::GetTouchInfo() {
   Cst816S::TouchInfos info;
-  uint8_t touchData[7];
+  uint8_t touchData[6];
```
NeroBurner: while editing this, we could port to `std::array<unit8_t, 6>` :thinking:

mark9064: Good idea

src/drivers/Cst816s.cpp
```diff
@@ -56,28 +56,33 @@ bool Cst816S::Init() {
...
 
-  auto ret = twiMaster.Read(twiAddress, 0, touchData, sizeof(touchData));
+  constexpr uint8_t addressOffset = 1;
+  auto ret = twiMaster.Read(twiAddress, addressOffset, touchData, sizeof(touchData));
```
NeroBurner: so we're effectively ignoring the first byte?

mark9064: According to the datasheet there is nothing at register 0. So I figured we probably shouldn't be trying to read it. It's a machine translation though so maybe something is missing? https://wiki.pine64.org/images/2/2f/CST816S.zip

src/drivers/Cst816s.cpp
```diff
@@ -56,28 +57,33 @@ bool Cst816S::Init() {
...
 }
 
 Cst816S::TouchInfos Cst816S::GetTouchInfo() {
   Cst816S::TouchInfos info;
-  uint8_t touchData[7];
+  std::array<uint8_t, 6> touchData {};
 
-  auto ret = twiMaster.Read(twiAddress, 0, touchData, sizeof(touchData));
+  constexpr uint8_t addressOffset = 1;
+  auto ret = twiMaster.Read(twiAddress, addressOffset, touchData.data(), sizeof(touchData));
```
NeroBurner: now we can use .size()

```suggestion
  auto ret = twiMaster.Read(twiAddress, addressOffset, touchData.data(), touchData.size());
```

src/drivers/Cst816s.cpp
```diff
@@ -56,28 +57,33 @@ bool Cst816S::Init() {
...
   return true;
 }
 
 Cst816S::TouchInfos Cst816S::GetTouchInfo() {
   Cst816S::TouchInfos info;
-  uint8_t touchData[7];
+  std::array<uint8_t, 6> touchData {};
 
-  auto ret = twiMaster.Read(twiAddress, 0, touchData, sizeof(touchData));
+  constexpr uint8_t addressOffset = 1;
```
NeroBurner: maybe add a comment that informs that we skip the first byte as it always is 0 anyways

src/drivers/Cst816s.cpp
mark9064: Oops missed that, thanks!

mark9064: Actually, sizeof is probably more semantically correct as we care about the number of bytes rather than the number of elements

NeroBurner:
can you separate the only read what we need part into its own PR we can approve and merge first. Then each PR does (or fix) exactly one thing (and ping me in the new PR so I can approve :)

mark9064:
if it's useful it's already two commits, but i can definitely split out if youd prefer

mark9064:
Opened #2349 for the touchpanel reading refactor

mark9064:
Yep, has been working great locally. A friend has confirmed it fixed jumps in paddle for them too, so I think it's good to go

PR #2326, opened 2025-06-27
Disable legacy FreeRTOS functions

mark9064:
We don't use any of them, so may as well turn them off


NeroBurner (approved):
seems to compile so LGTM

PR #2327, opened 2025-06-27
Analog face constexpr fix

mark9064:


src/displayapp/screens/WatchFaceAnalog.cpp
```diff
@@ -15,8 +15,10 @@ namespace {
   constexpr int16_t MinuteLength = 90;
   constexpr int16_t SecondLength = 110;
 
+  // LVGL sine isn't constexpr (thought it could be) so fix size here
```
NeroBurner: typo?
```suggestion
  // LVGL sin isn't constexpr (thought it could be) so fix size here
```

NeroBurner (approved):

src/displayapp/screens/WatchFaceAnalog.cpp
mark9064: It's the full spelling https://en.wikipedia.org/wiki/Sine_and_cosine
I think it's probably clearer your way

mark9064:
Just realised I wasn't using `<limits>` which is best practice now, I think it's a bit clearer

NeroBurner (approved):
maybe my non-native-english and programmer view, but yes `sin` and `cos` are more familiar for me :grin:

mark9064:
Spotted a typo and clarified why it can't be constexpr

PR #2329, opened 2025-06-27
Constant frequency SystemTask state updates

### LINKED ISSUE #1425: Display turns on every 30 min when Shake Wake is active
### Verification

### What happened?

While Shake Wake is active, the display turns on automatically every full and half hour.

I expect that the display stays off unless there was a shake.

### Reproduction steps

* Turn on Shake Wake
* Observe the display every 30 minutes to the full and half hour
  (you may place the watch on the table to prevent accidental shake activation)

### More details?

* Chimes are off.
* Other Wake Up options seam not to influence this behaviour.

### Version

v1.10.0 and v1.11.0

### Companion app

Gadgetbridge

mark9064:
Hopefully fixes #1425 (I haven't tested) and other jank caused by motion measurements happening more often than usual

The period isn't exactly 100ms due to rounding, but I think it's close enough since there is nothing here that depends on super high precision for accuracy (this is a problem for PPG for example)

### DIFF
--- a/src/systemtask/SystemTask.cpp
+++ b/src/systemtask/SystemTask.cpp
@@ -180,13 +180,28 @@ void SystemTask::Work() {
   measureBatteryTimer = xTimerCreate("measureBattery", batteryMeasurementPeriod, pdTRUE, this, MeasureBatteryTimerCallback);
   xTimerStart(measureBatteryTimer, portMAX_DELAY);
 
+  constexpr TickType_t stateUpdatePeriod = pdMS_TO_TICKS(100);
+  // Stores when the state (motion, watchdog, time persistence etc) was last updated
+  // If there are many events being received by the message queue, this prevents
+  // having to update motion etc after every single event, which is bad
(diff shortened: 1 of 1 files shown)


tituscmd:
I've been using this today, and it feels like the display is polling slower than before. When I swipe between screens, there's a slight delay before I can swipe again, it seems less responsive than it used to be. But only slightly

mark9064:
It should be more responsive as the systemtask spends less time doing things like motion updates when there are many events leaving it free to process touch events etc

tituscmd:
> It should be more responsive as the systemtask spends less time doing things like motion updates when there are many events leaving it free to process touch events etc

It seems the watch needs some "getting used to" this, as it definitely feels more responsive after a day or two of use.

mark9064:
(Rebased onto the motion update changes)

JF002 (approved):
I compared with my wife's pinetime running InfiniTime 1.14, and I couldn't spot any difference in reactivity or polling speed.
But those changes will definitely make motion processing more reliable!

PR #2330, opened 2025-06-27
LittleFS upgrade + fixes

mark9064:
The optimised parameters result in slightly less efficient flash usage with many small files but much better access performance. External fonts should load slightly faster after this.

After this PR, the on disk format will be updated to a newer version so loading an older version of InfiniTime will cause the flash to be reformatted

The optimised parameters require formatting the filesystem fully. I haven't included this in the PR, but we probably should? I haven't thought about how to do this in a nice way yet

This might close some issues relating to FS thread safety, but I'm not convinced it's fully fixed yet. It's definitely an improvement though - currently there is no locking at all


mark9064:
Bumped to latest LittleFS

PR #2331, opened 2025-06-27
Initialise DisplayApp components in DisplayApp context

mark9064:
Loading the watchface in the SystemTask context creates a lot of stack pressure as the SystemTask stack is much smaller

This might also fix other bad behaviour that occurs with watchfaces on first load e.g as reported by feksa in the pinetime-dev channel. Not sure though, haven't looked at the exact cause

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -129,15 +129,6 @@ void DisplayApp::Start(System::BootErrors error) {
 
   bootError = error;
 
-  lvgl.Init();
-  motorController.Init();
-
-  if (error == System::BootErrors::TouchController) {
-    LoadNewScreen(Apps::Error, DisplayApp::FullRefreshDirections::None);
-  } else {
-    LoadNewScreen(Apps::Clock, DisplayApp::FullRefreshDirections::None);
-  }
-
   if (pdPASS != xTaskCreate(DisplayApp::Process, "displayapp", 800, this, 0, &taskHandle)) {
     APP_ERROR_HANDLER(NRF_ERROR_NO_MEM);
   }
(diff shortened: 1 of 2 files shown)


src/displayapp/DisplayApp.cpp
```diff
@@ -146,17 +137,25 @@ void DisplayApp::Start(System::BootErrors error) {
...
+  }
 
   while (true) {
     app->Refresh();
   }
 }
 
-void DisplayApp::InitHw() {
+void DisplayApp::Init() {
+  lcd.Init();
```
NeroBurner, line 154: any reasoning on why to change the order of the init calls? Or doesn't the order matter?

src/displayapp/DisplayApp.cpp
```diff
@@ -129,15 +129,6 @@ void DisplayApp::Start(System::BootErrors error) {
 
   bootError = error;
 
-  lvgl.Init();
-  motorController.Init();
```
NeroBurner, line 133: just to be sure, SystemTask doesn't need lvgl and motorcontroller initialized directly after the `Start()` call, does it?

NeroBurner: I assume SystemTask is calling `DisplayApp.Start()` haven't actually checked

src/displayapp/DisplayApp.cpp
mark9064: I thought it made sense to initialise all the hardware first (LCD etc) and then initialise bits relying on it like LVGL afterwards. Before LVGL was loading a screen before lcd.Init() which didn't make any sense to me

NeroBurner: your reasoning makes sense to me, classical case of "How did that ever work?" :grin:

src/displayapp/DisplayApp.cpp
mark9064: > I assume SystemTask is calling DisplayApp.Start() haven't actually checked

That's right

Motor and LVGL are both "owned" by DisplayApp (no other task ever runs any functions from them) so it should be OK to defer them

PR #2332, opened 2025-06-28
QR Code Watchface

Feksaaargh:
Add a watchface encoded in a QR code.

### Features

- Technically a watchface
- Need a phone or something to scan it, so why not use that as a watch
- Can display other pieces of text in the QR code
- Comes preloaded with a rick roll

### Known Issues

When rebooting the system with the QR code watchface selected, the code will be replaced with a gray square. Refresh the screen by going to a different menu and back to see the correct code. #2331 fixes this on hardware.

### Images

QR code of main watchface
QR code claiming to be a cat
Cappy blappy wearing a watch and saying queer rights matter
<sub>Note: Only the text in the first image is available in the default watchface. You would need to add the text in the second and third images yourself.</sub>

### Usage

The main watchface displays the time in JSON.
The keys are as follows:
- time: String. Contains time either in format "HH:MM" ("13:57") or "HH:MM XM" ("1:57 PM"), depending on user preference.
- date: String. Contains date in format "DD Mmm YYY" ("13 May 2025").
- bt: Boolean. Represents if bluetooth is connected.
- batt: String, contains an integer 0-100. Is prepended with a plus if charging.

Long tapping cycles through alternate texts. When you switch to a different text (including back to the main watchface screen), it buzzes out its index in binary - strong buzz for 1, weak buzz for 0. Index 0 is the main watchface, and 1 and on is the user-defined ones. By default, the texts are:
- https://www.youtube.com/watch?v=dQw4w9WgXcQ
- ++++[>++++<-]>+[>+++++>++++>++<<<-]>-.>++++.---.>--.<++.------.<-------.>++++.
- While you were reading this message, I snuck a blueberry into your pocket.

Exercise caution when showing this feature to others. Trust is hard to regain among peers.
Note: If you are on an alternate text and press the side button, the watchface will eat the press and go back to the main watchface. If you leave the watch on an alternate text for more than 1 minute, it will switch back to the main watchface automatically (or the next time the screen turns on).
To add more alternate texts, simply add your strings to `altTexts` in the WatchFaceQRCode class in WatchFaceQRCode.h. Your string cannot be longer than 78 bytes, else an error will be displayed instead of your string.

### Limitations

- The QR code is hardcoded to be a 4L code in byte mode, which allows 78 bytes of ISO-8859-1 encoded data.
- While the generated codes are always valid, the generation is technically not spec-compliant. The generator does not test each possible XOR mask to find the most readable one, and instead hardcodes it at 1 (which is y%2). This is why the codes sometimes look blotchy and not nice and even like a correctly generated code.

### Why?

To squander the gift of creativity is a terrible sin.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -428,6 +428,7 @@ list(APPEND SOURCE_FILES
         displayapp/screens/WatchFacePineTimeStyle.cpp
         displayapp/screens/WatchFaceCasioStyleG7710.cpp
         displayapp/screens/WatchFacePrideFlag.cpp
+        displayapp/screens/WatchFaceQRCode.cpp
 
         ##
 
--- a/src/displayapp/UserApps.h
+++ b/src/displayapp/UserApps.h
@@ -15,6 +15,7 @@
 #include "displayapp/screens/WatchFacePineTimeStyle.h"
 #include "displayapp/screens/WatchFaceTerminal.h"
 #include "displayapp/screens/WatchFacePrideFlag.h"
+#include "displayapp/screens/WatchFaceQRCode.h"
 
 namespace Pinetime {
   namespace Applications {
--- a/src/displayapp/apps/Apps.h.in
+++ b/src/displayapp/apps/Apps.h.in
@@ -54,6 +54,7 @@ namespace Pinetime {
       Infineat,
       CasioStyleG7710,
       PrideFlag,
+      QRCode,
     };
 
     template <Apps>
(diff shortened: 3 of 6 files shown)


Feksaaargh:
Boot screen fix has been merged

minacode:
I like your meme-level and dedication to implement this highly unusual watchface!

On a more serious note I am not quite sure how likely we would be to merge this, because of how unusable this is as a watch face, but feel free to disagree and discuss with me. I am open, just not convinced yet 🙂 

If you would implement a QR-code app that works, that will be a totally different story though. I think a lot of people would like to have that. There were some attempts in the past. If you are interested, you can take a look at them.

Feksaaargh:
That's okay, I was putting it here more as a proof of concept and for archival purposes than with actual intention for it to be merged (I heard this is where to put apps and watchfaces until some formal app store is made).

I did see the other qr code app, I can try making something similar! I'd first need to make my qr generator more robust though, which will take a good bit of work. My current implementation has too many things hardcoded for general use.

PR #2333, opened 2025-06-30
Alarm: Add button to set alarm time to current time

tituscmd:
Hey all! :)

This PR is a QoL addition to the Alarm screen. I've added a button to the screen that, when pressed, sets the alarm time to the current time.
image

Tell me what you think! :smiley:


mark9064:
I'm interested, what's your use case?

tituscmd:
Say it's 14:37 (time of writing this) and I want to set an alarm for roughly 2hrs from now. Then I go into the alarm, press my new button and just up the hour slider by 2. And then if I want it to be a "clean" number like 16:30, I can easily do that as well.

mark9064:
Ahhh gotcha. I think that functionality should be provided by the timer app really. It should be improved to allow hours, and #1971 solves the problem of the timer end being too subtle

mark9064:
Would this work for you? Or is the timer app not suitable for this?

tituscmd:
Yeah, I think that works! I agree that the timer app with the added hours makes more sense for this use case

tituscmd:
I'll just keep this PR open tho, in case anyone wants to use it :)

liamcharger:
It definitely makes more sense to use the timer app, but I use the alarm for timers because timers don't ring, which makes them pretty useless. 

I wouldn't necessarily merge this, but until timer ringing is merged, it's a nice addition

mark9064:
Timer ringing is queued for 1.16 so hopefully that should be resolved soon

PR #2335, opened 2025-07-23
Minor documentation tweaks

WhyNotHugo:
See commit messages for finer details.

### DIFF
--- a/doc/buildAndProgram.md
+++ b/doc/buildAndProgram.md
@@ -4,7 +4,7 @@
 
 To build this project, you'll need:
 
-- A cross-compiler : [ARM-GCC (10.3-2021.10)](https://developer.arm.com/downloads/-/gnu-rm)
+- A cross-compiler : [ARM-GCC (10.3-2021.10)](https://developer.arm.com/Tools%20and%20Software/GNU%20Toolchain)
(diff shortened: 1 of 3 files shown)


WhyNotHugo:
Kind ping

mark9064:
Sorry for the delay! I'm away for a couple weeks but I'll try look over this when I can :)

doc/buildAndProgram.md
```diff
@@ -4,7 +4,7 @@
 
 To build this project, you'll need:
 
-- A cross-compiler : [ARM-GCC (10.3-2021.10)](https://developer.arm.com/downloads/-/gnu-rm)
+- A cross-compiler : [ARM-GCC (10.3-2021.10)](https://developer.arm.com/Tools%20and%20Software/GNU%20Toolchain)
```
mark9064: I think the previous link is correct - since we're still on 10.3 we should point to the old versions page rather than the new one. Once we update to a newer compiler you're totally right that we should change it though

WhyNotHugo: True. Updated.

mark9064 (approved):
Looks great! Thanks :)

Reviewer note: will merge in a week or so if no other reviews as this is a doc change only

Issue #1592, opened 2023-02-19
The current time cannot be seen in apps


Riksu9000:
### Verification

- [X] I searched for similar issues and found none was relevant.

### Introduce the issue

The time is only displayed in quicksettings, applist and the watchface. The time cannot be seen when using an app. We also don't have the screen space to spare for displaying the time in every app.

### Preferred solution

Instead of putting the time in apps, the time could be temporarily overlaid on the screen, in a small font at the top edge of the screen for example, every time the watch is woken up, and perhaps with a strong shake as well.

minacode:
I would like to have a lock-screen that appears first on wake-up before the watch goes to the last open screen. It would make sense that it would be (similar to) the watch face. 

All we need to show the time with an "overlay" would be a method to reach the lock-screen from the running system (without the lock of course). I am thinking of a swipe gesture or long touch or something like that.

My phone works like that.

Riksu9000:
This would be a problem if you have a timer for example that you keep open so you can quickly check it. It would be the same for stopwatch, if it didn't disable sleep.

minacode:
That's true. Keeping them running in the background and showing them on the lock screen if needed is the common solution for this, I guess.

JF002:
> Instead of putting the time in apps, the time could be temporarily overlaid on the screen, in a small font at the top edge of the screen for example

Maybe we can use the top and sys layers from LVGL for this?

https://docs.lvgl.io/7.11/overview/layer.html#top-and-sys-layers

lorhof1:
It is already possible to go to notifications by double-pressing the button and then go back to the previous app with a single press or upwards swipe. The easiest solution would probably be to add the time to the notifications screen as it can be opened from apps already.

PR #2337, opened 2025-07-26
Music: Redesign Part 2

tituscmd:
Hey everybody,

Since creating PR #2292 and introducing a small little redesign of the music app, I've had some more time on my hands and finally came around to creating a better version of it.

I've moved the artist and track name up a bit to add a new progress bar that shows the progress of the song you're listening to, with the time that has passed shown on the left and the time remaining shown on the right.
Sadly, I've had to remove (comment out) the disc animation in the top right because of space issues. If anyone is especially fond of the disc I'm sure space can be made to fit it back in. It is definitely not a final decision by me, hence why it's only commented out and not removed altogether.
But now that I've temporarily removed the disc animation, the top of the screen seemed quite empty to me. And after some thinking and digging through the code, I stumbled upon these lines of code in the Music.cpp:
```
294 - // Let's assume it starts playing instantly
295 - // TODO: In the future should check for BT connection for better UX
```
And so I thought "why not do that?" and I added an indicator text at the top telling you if you are connected via bluetooth. Please share your opinion on this.

Finally, here are some pictures of it all:
WhatsApp Image 2025-12-28 at 01 30 44
Screenshot 2025-12-28 at 01-34-08 InfiniEmu


liamcharger:
Definitely like this better than the current UI!

But what if you replaced the "connected" with the status bar/icons instead? That makes more sense imo

P.S I always thought the disc was a waste of space :)

mark9064:
Aw I quite like the little disc

Looks like a good change overall

I interpret the comment about Bluetooth as clearing the presented data when Bluetooth is disconnected, rather than showing an indicator. Since if Bluetooth isn't connected, we don't know anything about the music state and none of the controls are going to do anything. So I think it might be better UX to disable the controls and show "No connection" when Bluetooth is down rather than having a status bar, that way you get that space back. What do you think?

tituscmd:
I think that's a good idea.

The initial reason for that status bar/"Connected" text was the fact that the little disc physically didn't fit with the new design, but without it the top seemed so empty.

I'll share some screenshots of the different ideas talked about in this PR later when I have the time again. Then we can see and decide which one we like best.
How does that sound? 🙂

mark9064:
Sounds good to me :)

tituscmd:
I've played around with the design a bit and read some more older discussions on the music app design, what do you think of this?

image
image

It shows everything the regular old way, until you're disconnected. Then it says so where it would usually say "Not Playing" and the buttons below are a bit darker.
I've also added the time up top, since some people in those discussions proposed that idea and I like it. How does this look?

mark9064:
Looks nice! What do you think about having the full status bar rather than just the time? I'm thinking if we're already using that top space we may as well fill it with the full bar - I get the feeling that putting anything else in that top space is going to look weird

Or maybe left align the time and bring back the disc top right :P

I think it's good as is though, you don't need to like either idea :)

tituscmd:
Im not necessarily against any of those ideas, it just seems to me that, from a UI perspective, the status bar is only ever present in watchfaces or "transition screens" i.e. quick settings, app launcher. 
I feel like an app should feel like its own separate entity from those, so including the status bar breaks that.

I like the idea of moving the time to the left and adding the disc back, but it's ever so slightly too large for the new UI. Would it be easy/doable to make that disc a bit smaller?

liamcharger:
I agree with @mark9064 here about the status bar. It shouldn't be exclusive to the launcher and quick settings (usually there just isn't enough space). I can think of too many times I would've liked to check the time while I was inside an app I didn't want to close

mark9064:
Yeah more broadly there is #1592

The disc could be resized, it's a small PNG that's been converted to LVGL format. The original PNG is in the tree (icons/music/disc.png)

tituscmd:
> I agree with @mark9064 here about the status bar. It shouldn't be exclusive to the launcher and quick settings (usually there just isn't enough space). I can think of too many times I would've liked to check the time while I was inside an app I didn't want to close

I agree partly. In an ideal world, the status bar should be visible in any app. But since that is not the case right now, I think it’s better to stick to this pattern until we find a better solution for every app/globally, so everything remains predictable for now.

LinuxinaBit:
Could you integrate the volume buttons onto the main page instead of still requiring the swipe up? (A good PR demonstrating the current behavior: #1875 )

Something like #1901 would be very nice.

image

tituscmd:
> Could you integrate the volume buttons onto the main page instead of still requiring the swipe up? (A good PR demonstrating the current behavior: #1875 )
> 
> 
> 
> Something like #1901 would be very nice.
> 
> 
> 
> image
> 
> 

While I agree that the current system of swiping is not ideal, I am personally not a fan of how it looks like when the volume buttons are stacked on the right.
It's a lot of visual clutter and I don't think it's the way to go.
We could go with something like how the Paint app does it, i.e. tapping and holding the screen and that changes the buttons (with a little vibration)

LinuxinaBit:
That's still completely obtuse and gives the user no indication that the controls exist.
Far better to figure out *some* way to put it on one page imo.

tituscmd:
I mean yeah, I agree, but I really don't like how the UI looks that way. I don't believe that that would be the best solution to the problem.

liamcharger:
Regardless of if there's an indicator or not, swiping is a bad design. A button-based system is much simpler/more predictable, although I agree it does take up more space than I'd like

Like @LinuxinaBit mentioned (#1875), there should be a scroll indicator at the bare minimum

tituscmd:
image

I played around with the UI a bit. What do we think about a button like this? The button label switches between "Volume Controls" and "Track Controls".

We kinda have to sacrifice the time at the top for this, since otherwise the screen is gonna be too cramped to read, but I think that's fine

tituscmd:
image
Looks like this on-device, I quite like it

LinuxinaBit:
What about just this instead?

image

No need to make one tap into two, I figure...

dariusarnold:
@LinuxinaBit I like the design with the volume buttons always visible. I think that it might be possible to shrink them a little horizontally and use that to display the connected/disconnected state. Then there would still be two lines for the track + artist above the progress bar. Depending on how the connected state is displayed (maybe a blue and greyed out symbol), there might also be space for the time.
Then everything would be displayed. That might be worthwhile to consider, but maybe I am imagining the display larger than it really is.

liamcharger:
> image
> Looks like this on-device, I quite like it

While it informs the user of the volume controls, I don't think that's intuitive either. The hint itself looks like a button, and I don't like the idea of it being permanent. 

@LinuxinaBit 's solution is great, although the button colors should be uniform. We don't lose any current elements, and it fixes the entire issue. 

@dariusarnold Since we already have "disconnected", an icon isn't necessary. Shrinking them vertically also wouldn't leave enough room for the actual buttons (they'd be too small to hit reliably)

tituscmd:
> While it informs the user of the volume controls, I don't think that's intuitive either. The hint itself looks like a button, and I don't like the idea of it being permanent. 

It is indeed a button and achieves the same as the "old" swiping gesture, but as a button

> @LinuxinaBit 's solution is great, although the button colors should be uniform. We don't lose any current elements, and it fixes the entire issue. 

Very fair.
How about we place the buttons below the current three buttons though, so it looks more like it belongs to those control buttons?

tituscmd:
>How about we place the buttons below the current three buttons though, so it looks more like it belongs to those control buttons?

I like this a lot actually. I've changed the height of the buttons to match the InfiniTime Vision on UI and I think it looks really good.
I only added the "Some Track" string to visualize how it will look like with a song playing, it will be removed once I actually push all those commits
image

LinuxinaBit:
There are two reasons I didn’t group the buttons like that:
- So the touch targets could be made bigger (expanding down past the visual button)
- So there’s extra space providing some good visual separation

Both of these are sorely needed on a screen this small.

As for the colors, I themed it the same as your `Volume Controls` button to provide some visual balance and avoid making it too uniform (the volume icons should’ve actually been a bit more gray to match).
I think it looked a _lot_ better than when I tried making them the same shade.

Main thing is that I really wanted to avoid making it just another grid of icons because it just doesn’t look all that good, and is less usable to boot.

Edit: A unified grid also moves the song title/artist text a bit too high for my preference...

tituscmd:
> As for the colors, I themed it the same as your `Volume Controls` button to provide some visual balance and avoid making it too uniform (the volume icons should’ve actually been a bit more gray to match, but I forgot to do that). I think it looked a lot better than when I tried making them the same shade.

The darker color is actually already used when the watch is disconnected, showing that they're disabled. As far as I can tell, that's the visual cue for the user to understand that the button currently won't work in general in InfiniTime (see stopwatch app -> lap button)


> Main thing is that I really wanted to avoid making it just another grid of icons because it just doesn’t look all that good, and is less usable to boot.

Not sure what you mean by less usable to boot. Otherwise, the grid of icons is down to preference I think. I think having the volume buttons up at the top separates them too much from the controls and makes them look out of place.

liamcharger:
The colors should be the same as it doesn’t make sense to have them different. And as @tituscmd mentioned, the darker gray is used for the disabled state. 

As far as positioning, I believe keeping them on the top of the screen is more practical, as the user is less likely to hit another button accidentally. The volume controls are also not "main” actions, so keeping them separated is more natural imo

tituscmd:
I'm really not a fan of buttons on the top edge of the screen :sweat_smile:

Here are the two variants next to each other, as well as a third one I just made which I think is even better than No. 1
image ..... image ..... image

If we were to go off the InfiniTime UI guidelines, it says buttons should generally be on the bottom edge. Now of course that can be disregarded if we don't have enough space, I'm not going to say we should 100% stick to those guidelines no matter what, but I get where they are coming from because imo it looks the best (picture 3)

LinuxinaBit:
I think part of the problem is that the spacing isn’t consistent with the bottom ones, and they aren’t tall enough. I think fixing both of those issues will help.

LinuxinaBit:
It might also be worth just making them a *different* shade (not the disabled button or the standard button shade) just for visual variety.
I think it could end up looking really good if you find the right combination.

tituscmd:
> I think part of the problem is that the spacing isn’t consistent with the bottom ones, and they aren’t tall enough. I think fixing both of those issues will help.

The spacing I chose is just the usual spacing of two buttons next to each other in InfiniTime, i.e. Firmware Validation screen

>It might also be worth just making them a different shade (not the disabled button or the standard button shade) just for visual variety.

We have to keep in mind what @liamcharger said "The colors should be the same as it doesn’t make sense to have them different."
I'll try to play around with the spacing a bit to make the screen less cramped maybe

tituscmd:
With double the spacing between the two sets of buttons, it looks a lot better imo.
Now the only downside is that the rest is really smushed into the top edge of the screen

image
image
image

tituscmd:
Another pic from on my wrist. I think it looks good!
image

liamcharger:
> image
> 
> 
> 
> No need to make one tap into two, I figure...

While I'm not necessarily against this, I think the screenshot above (once the colors are updated) is the best. Its buttons are a little bigger and they're separated from the main controls (both contributing to better usability, which I believe is more important than a opinion of which side looks better that will vary from user to user)

liamcharger:
> image
> 
> 
> 
> No need to make one tap into two, I figure...

While I'm not necessarily against what's pictured in the images above, I think the screenshot I quoted (once the colors are updated) is the best. Its buttons are a little bigger and they're separated from the main controls (both contributing to better usability, which I believe is more important than a opinion of which side looks better that will vary from user to user)

LinuxinaBit:
Here's what I meant by the spacing being too big btw:

image

Fixing that could help make it look a lot better, on-device especially.
If you can, making the top buttons _taller_ would likely help a lot as well.

<hr>

Here's a recolored version of my concept from earlier:

image

tituscmd:
example

These are the UI guidelines I was going off of. That's why the gap is bigger. The pixels on the images I uploaded aren't exactly the same because, for whatever reason, they were rendered as 288x288 instead of 240x240.

LinuxinaBit:
Those are so old they don't even show the taller height buttons used on the stopwatch etc.
I think that, especially when combining buttons in the same UI, consistency within the UI and general usability are far more important.

tituscmd:
> Those are so old they don't even show the taller height buttons used on the stopwatch etc. I think when combining buttons in the same UI, consistency within the UI is far more important anyway.

Very fair.

tituscmd:
I'll send a picture of the two "final" versions (volume buttons on top vs. bottom) and then maybe we can get some more opinions before we settle on one?

tituscmd:
Screenshot 2025-08-23 at 18-40-06 InfiniEmu
(spacing)Screenshot 2025-08-23 at 18-33-14 InfiniEmu

.

If you like the left version more, vote the thumbs up. If you like the version on the right more, vote the thumbs down

src/displayapp/screens/Music.cpp
```diff
@@ -24,6 +24,7 @@
 #include "displayapp/icons/music/disc_f_1.c"
```
dariusarnold: These disc files can probably removed when fully unused.

dariusarnold:
@tituscmd Can I ask about the Track/Artist info (currently placeholder "Some Track")? Is the "Disconnected" replaced with the artist when connected? I liked the two line design from the first post, since it would reduce text scrolling.

---

The current commit ed30e5 from https://github.com/InfiniBros/InfiniTime/tree/music_redesign_2 makes InfiniSim crash on my end when entering the music app. It happens reliably with the following callstack:
```
draw_indic(lv_obj_t * bar, const lv_area_t * clip_area) (/workspaces/InfiniTime/src/libs/lvgl/src/lv_widgets/lv_bar.c:498)
lv_bar_design(lv_obj_t * bar, const lv_area_t * clip_area, lv_design_mode_t mode) (/workspaces/InfiniTime/src/libs/lvgl/src/lv_widgets/lv_bar.c:378)
lv_refr_obj(lv_obj_t * obj, const lv_area_t * mask_ori_p) (/workspaces/InfiniTime/src/libs/lvgl/src/lv_core/lv_refr.c:895)
lv_refr_obj(lv_obj_t * obj, const lv_area_t * mask_ori_p) (/workspaces/InfiniTime/src/libs/lvgl/src/lv_core/lv_refr.c:935)
lv_refr_obj_and_children(lv_obj_t * top_p, const lv_area_t * mask_p) (/workspaces/InfiniTime/src/libs/lvgl/src/lv_core/lv_refr.c:837)
lv_refr_area_part(const lv_area_t * area_p) (/workspaces/InfiniTime/src/libs/lvgl/src/lv_core/lv_refr.c:768)
lv_refr_area(const lv_area_t * area_p) (/workspaces/InfiniTime/src/libs/lvgl/src/lv_core/lv_refr.c:510)
lv_refr_areas() (/workspaces/InfiniTime/src/libs/lvgl/src/lv_core/lv_refr.c:400)
_lv_disp_refr_task(lv_task_t * task) (/workspaces/InfiniTime/src/libs/lvgl/src/lv_core/lv_refr.c:199)
lv_task_exec(lv_task_t * task) (/workspaces/InfiniTime/src/libs/lvgl/src/lv_misc/lv_task.c:386)
lv_task_handler() (/workspaces/InfiniTime/src/libs/lvgl/src/lv_misc/lv_task.c:134)
Pinetime::Applications::DisplayApp::Refresh(class Pinetime::Applications::DisplayApp * const this) (/workspaces/InfiniTime/src/displayapp/DisplayApp.cpp:260)
Pinetime::Applications::DisplayApp::Process(void * instance) (/workspaces/InfiniTime/src/displayapp/DisplayApp.cpp:149)
sdl_function_wrapper(void * instance) (/workspaces/InfiniSim/sim/task.cpp:61)
libSDL2-2.0.so.0![Unknown/Just-In-Time compiled code] (Unknown Source:0)
libc.so.6!start_thread(void * arg) (pthread_create.c:447)
libc.so.6!clone() (clone.S:100)
```

The debugger (gdb) prints
```
Thread 37 "displayapp" received signal SIGFPE, Arithmetic exception.
[Switching to Thread 0x7fb1d7fff6c0 (LWP 5533)]
0x00005625e98d981c in draw_indic (bar=0x7fb1cc001d58, clip_area=0x7fb1d7ffe8f8) at /workspaces/InfiniTime/src/libs/lvgl/src/lv_widgets/lv_bar.c:498
498	        anim_start_value_x = (int32_t)((int32_t)anim_length * (ext->start_value - ext->min_value)) / range;
```

I can see that the issue is caused by `range` being zero when attached in gdb. I am not familiar enough with InfiniSim + LVGL to diagnose this quickly though. It might be a simulator-only issue as well, I have been using an earlier commit of this on the watch for some time now where it worked well.

tituscmd:
> @tituscmd Can I ask about the Track/Artist info (currently placeholder "Some Track")? Is the "Disconnected" replaced with the artist when connected? I liked the two line design from the first post, since it would reduce text scrolling.

Yes, sure!
The "Some Track" text is just for testing right now, so you can get an idea of how the UI would look like with a song playing (since I can't simulate that in InfiniEmu)

In the actual version. When your watch is not connected to any bluetooth device, it would just show "Disconnected" and the upper text would be blank.
When you are connected to a bluetooth device, the field that is now showing "Some Track" would show the song's name and the field saying "Disconnected" would be replaced by the artist.

On another note, I can tell that we're going with the first option where the music buttons are up top, so I'll be pushing that in a bit. Once that's done, I'll also show a picture of both states on device and remove all the commented out code

dariusarnold:
@tituscmd Great, looking forward to try it on the watch.

tituscmd:
> The current commit ed30e5 from https://github.com/InfiniBros/InfiniTime/tree/music_redesign_2 makes InfiniSim crash on my end when entering the music app. It happens reliably with the following callstack:
> 
> ```
> draw_indic(lv_obj_t * bar, const lv_area_t * clip_area) (/workspaces/InfiniTime/src/libs/lvgl/src/lv_widgets/lv_bar.c:498)
> lv_bar_design(lv_obj_t * bar, const lv_area_t * clip_area, lv_design_mode_t mode) (/workspaces/InfiniTime/src/libs/lvgl/src/lv_widgets/lv_bar.c:378)
> lv_refr_obj(lv_obj_t * obj, const lv_area_t * mask_ori_p) (/workspaces/InfiniTime/src/libs/lvgl/src/lv_core/lv_refr.c:895)
> lv_refr_obj(lv_obj_t * obj, const lv_area_t * mask_ori_p) (/workspaces/InfiniTime/src/libs/lvgl/src/lv_core/lv_refr.c:935)
> lv_refr_obj_and_children(lv_obj_t * top_p, const lv_area_t * mask_p) (/workspaces/InfiniTime/src/libs/lvgl/src/lv_core/lv_refr.c:837)
> lv_refr_area_part(const lv_area_t * area_p) (/workspaces/InfiniTime/src/libs/lvgl/src/lv_core/lv_refr.c:768)
> lv_refr_area(const lv_area_t * area_p) (/workspaces/InfiniTime/src/libs/lvgl/src/lv_core/lv_refr.c:510)
> lv_refr_areas() (/workspaces/InfiniTime/src/libs/lvgl/src/lv_core/lv_refr.c:400)
> _lv_disp_refr_task(lv_task_t * task) (/workspaces/InfiniTime/src/libs/lvgl/src/lv_core/lv_refr.c:199)
> lv_task_exec(lv_task_t * task) (/workspaces/InfiniTime/src/libs/lvgl/src/lv_misc/lv_task.c:386)
> lv_task_handler() (/workspaces/InfiniTime/src/libs/lvgl/src/lv_misc/lv_task.c:134)
> Pinetime::Applications::DisplayApp::Refresh(class Pinetime::Applications::DisplayApp * const this) (/workspaces/InfiniTime/src/displayapp/DisplayApp.cpp:260)
> Pinetime::Applications::DisplayApp::Process(void * instance) (/workspaces/InfiniTime/src/displayapp/DisplayApp.cpp:149)
> sdl_function_wrapper(void * instance) (/workspaces/InfiniSim/sim/task.cpp:61)
> libSDL2-2.0.so.0![Unknown/Just-In-Time compiled code] (Unknown Source:0)
> libc.so.6!start_thread(void * arg) (pthread_create.c:447)
> libc.so.6!clone() (clone.S:100)
> ```
> 
> The debugger (gdb) prints
> 
> ```
> Thread 37 "displayapp" received signal SIGFPE, Arithmetic exception.
> [Switching to Thread 0x7fb1d7fff6c0 (LWP 5533)]
> 0x00005625e98d981c in draw_indic (bar=0x7fb1cc001d58, clip_area=0x7fb1d7ffe8f8) at /workspaces/InfiniTime/src/libs/lvgl/src/lv_widgets/lv_bar.c:498
> 498	        anim_start_value_x = (int32_t)((int32_t)anim_length * (ext->start_value - ext->min_value)) / range;
> ```
> 
> I can see that the issue is caused by `range` being zero when attached in gdb. I am not familiar enough with InfiniSim + LVGL to diagnose this quickly though. It might be a simulator-only issue as well, I have been using an earlier commit of this on the watch for some time now where it worked well.

This seemed to have been caused by some 0 division on the progress bar. Should be fixed by latest commit - works for me in InfiniEmu now

src/displayapp/screens/Music.cpp
tituscmd: Yep! Good catch

tituscmd:
Latest commit makes the UI switch on reconnection much smoother, since we now force fetch the data and update the UI if we just reconnected - otherwise we only fetch new data if it changed.
While adding that I also optimized the code a bit, putting stuff into separate functions so the code is cleaner and less cluttered. I'll send photos of the 3 UI states and a video showcasing the new reconnection shortly

tituscmd:
image
image
image

tituscmd:
WhatsAppVideo2025-08-25at23 43 48-ezgif com-video-to-gif-converter(1)

liamcharger:
Looking good, @tituscmd!

Thanks for your work on this!

tituscmd:
This PR should be ready for review from my side :)

LinuxinaBit:
Ooh, what if the progress bar was blue?

image

tituscmd:
> Ooh, what if the progress bar was blue?
> image

I personally like the gray and don't see any big improvement in terms of readability with a blue one.

LinuxinaBit:
Yeah, not a big deal. Though it isn't really for readability; it's just about adding some visual flair to an otherwise colorless interface :)

0x0000ff:
I'm of the belief that track info isn't needed on a watch or should be at least be minimized to a progress bar. The buttons could them be laid out in a way that would make them easier to press.

Just my 2 cents.

LinuxinaBit:
@0x0000ff 
Something like https://github.com/InfiniTimeOrg/InfiniTime/pull/558 is always an option, to achieve a similar effect to what you're talking about without removing UI elements that other people (like me!) want.

That being said, I think it's more important at this point to get the UI merged so people can actually start using it when the next version of InfiniTime is released... More changes can always be made in the future.

tituscmd:
> I'm of the belief that track info isn't needed on a watch or should be at least be minimized to a progress bar. The buttons could them be laid out in a way that would make them easier to press.
> 
> 
> 
> Just my 2 cents.

Not quite sure what you mean by this. Care to elaborate?

tituscmd:
> @0x0000ff 
> 
> Something like https://github.com/InfiniTimeOrg/InfiniTime/pull/558 is always an option, to achieve a similar effect to what you're talking about without removing UI elements that other people (like me!) want.
> 
> 
> 
> That being said, I think it's more important at this point to get the UI merged so people can actually start using it when the next version of InfiniTime is released... More changes can always be made in the future.

Agreed. Should be all good to merge from my side I think.

.vscode/settings.json
```diff
@@ -65,6 +65,7 @@
         "stdexcept": "cpp",
         "streambuf": "cpp",
         "cinttypes": "cpp",
-        "typeinfo": "cpp"
+        "typeinfo": "cpp",
```
mark9064: Could this change be moved to another PR

src/displayapp/screens/Music.cpp
```diff
@@ -150,74 +158,94 @@ Music::~Music() {
 }
 
 void Music::Refresh() {
-  if (artist != musicService.getArtist()) {
+  bleState = bleController.IsConnected();
+  if (bleState.Get() == false) {
```
mark9064:
```suggestion
  if (!bleState.Get()) {
```

src/displayapp/screens/Music.cpp
```diff
@@ -150,74 +158,94 @@ Music::~Music() {
...
-      lastIncrement = xTaskGetTickCount();
-    }
-  } else {
-    lv_label_set_text_static(txtPlayPause, Symbols::play);
-  }
 }
 
 void Music::UpdateLength() {
+  int remaining = totalLength - currentPosition;
+  if (remaining < 0)
```
mark9064: Could you add braces here

src/displayapp/screens/Music.cpp
```diff
@@ -150,74 +158,94 @@ Music::~Music() {
 }
 
 void Music::Refresh() {
-  if (artist != musicService.getArtist()) {
+  bleState = bleController.IsConnected();
+  if (bleState.Get() == false) {
+    SetDisconnectedUI();
+    lastConnected = false;
```
mark9064: Could `lastConnected` be avoided by relying on `bleState.IsUpdated()`?

src/displayapp/screens/Music.cpp
```diff
@@ -150,74 +158,94 @@ Music::~Music() {
...
+      SetConnectedUI();
+      RefreshTrackInfo(true);
+    } else {
+      RefreshTrackInfo(false);
+    }
+    lastConnected = true;
+  }
+}
+
+void Music::SetDisconnectedUI() {
```
mark9064, line 179: If the state variables (playing, artist, track, album etc) were updated here, could we get rid of force? I think the UI will be more robust if the state variables are always kept up to date, that way connection/disconnection don't need special handling

Would also need handling during in the constructor

src/displayapp/screens/Symbols.h
```diff
@@ -42,6 +42,7 @@ namespace Pinetime {
         static constexpr const char* sleep = "\xEE\xBD\x84";
         static constexpr const char* calculator = "\xEF\x87\xAC";
         static constexpr const char* backspace = "\xEF\x95\x9A";
+        static constexpr const char* swap = "\xef\x81\xb9";
```
mark9064: Is this used anywhere?

src/displayapp/screens/Music.h
```diff
@@ -59,13 +67,14 @@ namespace Pinetime {
...
+        lv_obj_t* barTrackDuration;
 
         lv_style_t btn_style;
 
-        /** For the spinning disc animation */
-        bool frameB;
-
         Pinetime::Controllers::MusicService& musicService;
+        const Controllers::Ble& bleController;
+        Pinetime::Controllers::DateTime& dateTimeController;
```
mark9064: Time controller unused?

src/displayapp/screens/Music.h
```diff
@@ -59,13 +67,14 @@ namespace Pinetime {
...
 
         std::string artist;
```
mark9064: Looks like these could all use DirtyValue

mark9064 (commented):
Looking good :)

Haven't tested locally or ran clang-tidy yet

.vscode/settings.json
tituscmd: Totally accidental - not even sure how it happened but VSC seems to do that sometimes. I'll remove it ASAP

src/displayapp/screens/Music.cpp
tituscmd: Do you mean

`int remaining = (totalLength - currentPosition);`

?

src/displayapp/screens/Symbols.h
tituscmd: Nope, unused symbol from the experimental "Swap Controls"-button. I'll remove it ASAP

src/displayapp/screens/Music.cpp
mark9064: On the if statements:
```cpp
if (remaining < 0) {
...
}
```

tituscmd: Right, makes sense. I'll do that as well as soon as I get the time to

src/displayapp/screens/Music.h
tituscmd: Yep, that was from when I wanted to display the time in the Music app as well. I'll remove it

src/displayapp/screens/Music.cpp
tituscmd: Yep! Good catch.

src/displayapp/screens/Music.cpp
tituscmd: Fixed. Check out the new commit and tell me if I've understood correctly what you said :D

src/displayapp/screens/Music.cpp
```diff
@@ -20,10 +20,8 @@
 #include <cstdint>
 #include "displayapp/DisplayApp.h"
 #include "components/ble/MusicService.h"
-#include "displayapp/icons/music/disc.c"
-#include "displayapp/icons/music/disc_f_1.c"
-#include "displayapp/icons/music/disc_f_2.c"
```
mark9064, line 25: The disc icon files should be deleted as well. Though I'm a little (quite) sad to see it go... it does allow more functionality to fit in

src/displayapp/screens/Music.cpp
```diff
@@ -152,22 +156,71 @@ Music::~Music() {
...
+    lv_label_set_text(txtArtist, artist.Get().data());
+  }
+
+  track = musicService.getTrack();
+  if (track.IsUpdated()) {
+    lv_label_set_text(txtTrack, track.Get().data());
+  }
+
+  album = musicService.getAlbum();
+  if (album.IsUpdated()) {
```
mark9064: Album unused? If so we can remove it as a member

src/displayapp/screens/Music.h
```diff
@@ -60,30 +68,29 @@ namespace Pinetime {
...
-        /** Last time an animation update or timer was incremented */
-        TickType_t lastIncrement = 0;
 
         bool playing;
 
         lv_task_t* taskRefresh;
 
-        Widgets::PageIndicator pageIndicator = Widgets::PageIndicator(0, 2);
+        Utility::DirtyValue<bool> bleState {};
+        Utility::DirtyValue<bool> bleRadioEnabled {};
```
mark9064: Unused?

src/displayapp/screens/Music.cpp
mark9064: Yeah pretty much spot on, looks great. I'd ideally like to have artist/track be the actual values displayed (so they'd be set to "Disconnected"/"--:--" here) because then only one rendering path is needed (i.e. only one place where `txtArtist` etc. are upated), but I think this is a bit of a pointless refactor (mainly as the connected/disconnected states are reasonably different)

mark9064 (commented):
Pretty much there I think :)

src/displayapp/screens/Music.cpp
tituscmd: Yeah I was wondering about that, should we try to include it in the UI somehow? Although I don't think ever seen it used.

src/displayapp/screens/Music.cpp
tituscmd: I'm also not particularly fond of getting rid of it, but it makes too much sense for usability 😔

src/displayapp/screens/Music.cpp
mark9064: I don't think it's too important compared to the track title and artist so it probably doesn't justify the space. Especially as a lot of the time the field is probably blank (since podcasts/video content etc. don't really have an album)

tituscmd: Yep, that's fair. I'll go ahead and remove it next time I'm available. Might take a few days because of the upcoming holidays.

mark9064: No rush, drop a message whenever :) happy holidays!

tituscmd: Thanks, will do :) 
Happy holidays to you as well!

src/displayapp/screens/Music.h
tituscmd: Yep, not anymore

src/displayapp/screens/Music.cpp
tituscmd: Maybe one day we will find a way to implement them, it feels wrong to get rid of them entirely...

mark9064: I think our code policy is generally if it isn't used, it should be removed. The Git history will remember, we can always bring it back. Mainly as it's confusing to keep dead code around: someone browsing the source after this change might see these files and then try to figure out where they're from.

If there's ever a chance to use them again, being completely realistic I doubt they're going to be remembered any more than they would be by continuing to exist in tree. And by then they'll also probably be incompatible with a new LVGL version or whatever. In short keeping unused code around is confusing and causes problems later as it becomes incompatible with the evolving codebase

mark9064:
One other thing, could you edit the first message to show screenshots of the latest version? Just makes it a bit easier for anyone looking at the PR for the first time

src/displayapp/screens/Music.cpp
tituscmd: Sounds reasonable. I've removed the files in the latest commit and believe this PR should be done now.

tituscmd:
I made the button labels a bit more gray when the buttons are disabled, so it’s easier to tell at a glance what’s inactive.
I also cleaned up SetConnectedUI and SetDisconnectedUI by using arrays for the buttons and labels, since repeating the same line over and over was getting annoying.

I updated the first message of this PR as well so it matches the current UI, making it clearer for anyone checking it out for the first time - thanks for pointing that out @mark9064 :D

mark9064:
[out.patch](https://github.com/user-attachments/files/24367919/out.patch)

Had a poke around, few suggestions here (mainly to avoid storing copies of all the buttons/labels again). I think the only substantial code change required is to fix the the progress bar narrowing issue, `lv_bar_set_range` takes int16s and currently the code assumes it can handle int32s

You can apply the changes locally with `git apply <path/to/patchfile>`

tituscmd:
> I think the only substantial code change required is to fix the the progress bar narrowing issue, lv_bar_set_range takes int16s and currently the code assumes it can handle int32s

Totally didn't think of that, good catch.
I've clamped both the range and value of the bar to the max value of int16_t. With the latest commit, any overflow issues should be solved :+1:

CodeGuyEli:
This looks super cool!

tituscmd:
> This looks super cool!

Thank you so much! 😊

liamcharger:
Could we get this added to 1.17? It's a nice qol update

mark9064:
Added to milestone

@tituscmd can you sync with main (merge/rebase)?

tituscmd:
> @tituscmd can you sync with main (merge/rebase)?

Should be all good to go!

PR #2338, opened 2025-07-28
Firmware validation: UI redesign

tituscmd:
Hey everyone!

This is a redesign of the UI for the settings page for validate the firmware.

This is the screen for when the firmware is validated
Screenshot 2025-07-29 at 00-09-20 InfiniEmu

And this is the screen for when the firmware is not yet validated. I changed the text to explain what happens when the firmware isn't validated, rather than just telling you to validate it, so that new users understand what they’re doing :)
Screenshot 2025-07-29 at 00-09-17 InfiniEmu

Tell me what you think!


skarn958:
I really like it. It's clearer and more straight forward.

mark9064:
This is way better :)

The only bit I'm not sure about is the rounded button corners. I feel like the rest of the corners in the OS are a bit more square?

tituscmd:
> The only bit I'm not sure about is the rounded button corners. I feel like the rest of the corners in the OS are a bit more square?

I originally based that decision on the buttons like in the Steps and Timer screen, but after a second look it seems that most other screens in InfiniTime where there are two buttons next to each other do in fact use more square-looking corner radiuses (or radii if you will ;))

Here's how that would look like:
image
I like that as well and, as far as I'm concerned, we can for sure go with that design rather than the circle corner radius

mark9064:
Also, just thinking about the text would "Rollback" describe what the button does better? Not sure if the longer label would fit nicely though...

marigoldfish:
How about "Revert"?

tituscmd:
image

I actually like "Rollback" a lot. Not only does it match with the text above, but the longer label also matches the longer label on the "Validate" button.

If everyone is happy with this I'll push the two commits (the squircle buttons and the label change) and then the PR should be all set from my side :)

tituscmd:
I’ve now pushed those two commits (squircle buttons and label change)

> image
> 
> I actually like "Rollback" a lot. Not only does it match with the text above, but the longer label also matches the longer label on the "Validate" button.
> 
> If everyone is happy with this I'll push the two commits (the squircle buttons and the label change) and then the PR should be all set from my side :)

I’ve now pushed those two commits

src/displayapp/screens/FirmwareValidation.cpp
```diff
@@ -15,34 +16,50 @@ namespace {
...
 
   labelIsValidated = lv_label_create(lv_scr_act(), nullptr);
-  lv_obj_align(labelIsValidated, labelVersion, LV_ALIGN_OUT_BOTTOM_LEFT, 0, 0);
+  lv_obj_align(labelIsValidated, nullptr, LV_ALIGN_CENTER, 0, 10);
+  lv_label_set_align(labelIsValidated, LV_LABEL_ALIGN_CENTER);
+  lv_obj_set_auto_realign(labelIsValidated, true);
   lv_label_set_recolor(labelIsValidated, true);
-  lv_label_set_long_mode(labelIsValidated, LV_LABEL_LONG_BREAK);
-  lv_obj_set_width(labelIsValidated, 240);
+  // lv_label_set_long_mode(labelIsValidated, LV_LABEL_LONG_BREAK);
```
dariusarnold, line 47: Sorry if this is still in progress then please ignore for now, but before the merge commented out code should likely be removed.

tituscmd:
Just a little friendly bump that I've now pushed the last changes and this PR should be ready to merge, if desired :)

src/displayapp/screens/FirmwareValidation.cpp
```diff
@@ -15,34 +16,50 @@ namespace {
...
+  // lv_obj_set_width(labelIsValidated, 240);
```
mark9064: Can be removed

src/displayapp/screens/FirmwareValidation.cpp
```diff
@@ -15,34 +16,50 @@ namespace {
...
-                             "Please #00ff00 Validate# this version or\n#ff0000 Reset# to rollback to the previous version.");
+    lv_label_set_text_static(labelIsValidated, "Any reboot will\nrollback to last\nvalidated firmware");
 
     buttonValidate = lv_btn_create(lv_scr_act(), nullptr);
     buttonValidate->user_data = this;
     lv_obj_set_size(buttonValidate, 115, 50);
     lv_obj_align(buttonValidate, nullptr, LV_ALIGN_IN_BOTTOM_LEFT, 0, 0);
     lv_obj_set_event_cb(buttonValidate, ButtonEventHandler);
     lv_obj_set_style_local_bg_color(buttonValidate, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, Colors::highlight);
+    // lv_obj_set_style_local_radius(buttonValidate, LV_BTN_PART_MAIN, LV_STATE_DEFAULT, LV_RADIUS_CIRCLE);
```
mark9064: Can be removed

src/displayapp/screens/FirmwareValidation.cpp
```diff
@@ -53,9 +70,10 @@ FirmwareValidation::FirmwareValidation(Pinetime::Controllers::FirmwareValidator&
     lv_obj_align(buttonReset, nullptr, LV_ALIGN_IN_BOTTOM_RIGHT, 0, 0);
     lv_obj_set_style_local_bg_color(buttonReset, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_RED);
     lv_obj_set_event_cb(buttonReset, ButtonEventHandler);
+    // lv_obj_set_style_local_radius(buttonReset, LV_BTN_PART_MAIN, LV_STATE_DEFAULT, LV_RADIUS_CIRCLE);
```
mark9064: Can be removed

tituscmd:
Reviewed changes have been pushed :+1:

tituscmd:
Should be ready to merge from my understanding. No rush, though. I’m just going through my PRs to check which ones need some attention :)

PR #2341, opened 2025-08-21
docs: list exact dependencies on Alpine

WhyNotHugo:
Add an copy-pasteable snippet for quickly installing all build dependencies on Alpine.

### DIFF
--- a/doc/buildAndProgram.md
+++ b/doc/buildAndProgram.md
@@ -20,6 +20,19 @@ To build this project, you'll need:
   - install npm (commonly done via the package manager, ensure node's version is at least 12)
   - install lv_font_conv: `npm install lv_font_conv`
 
+On Alpine Linux (edge), these can be installed via:
+
+```sh
+apk add -t .infinitime-deps \
+	cmake \
+	gcc-arm-none-eabi \
+	lv_font_conv \
+	nrf5-sdk \
+	py3-cbor2 \
+	py3-cryptography \
+	py3-intelhex
+```
+
 ## Build steps
 
 ### Clone the repo


mark9064:

I'm not familiar with APK, does this pin the dependencies to the correct versions? e.g ARM GCC 10.3, nRF SDK 15.3 etc
If so, where are the version pins stored?

WhyNotHugo:
It does not pin the exact versions. Alpine mirrors only keep the latest version of each package, so pinning older versions is not possible.

At this time, this installs the following versions:

```
cmake-4.1.0
gcc-arm-none-eabi-15.2.0
lv_font_conv-1.5.3
nrf5-sdk-17.1.0
py3-cbor2-5.6.5
py3-cryptography-44.0.3
py3-intelhex-2.3.0
```

I have confirmed that builds work fine with these.

mark9064:
Ah OK. I think without a version pin it's not possible to include this in the official instructions, as different compiler and SDK versions will change the firmware image produced and cause incompatibilities.

For example, the current main does not compile with ARM GCC 14.2, though it compiles fine with 10.3. Also the inverse might happen - someone may unknowingly write code that relies on GCC 15 or a new SDK feature and then be surprised when they try and upstream (and waste their time and reviewers time trying to figure out what is happening :sweat_smile:  )

Hopefully you understand where I'm coming from :)

WhyNotHugo:
I understand your concern.

Realistically, omitting this from the documentation merely results in people taking extra steps to figure out the exact names of the packages to install. They'll eventually end with the same result and install the (only) versions available anyway.

> current main does not compile with ARM GCC 14.2, though it compiles fine with 10.3

Tangential, but I'm somewhat surprised that 14.2 doesn't work, but 15.2 works fine.

mark9064:
Are the normal install instructions of downloading the toolchain / NRF SDK impossible to follow on Alpine? libc issue or something?

WhyNotHugo:
Following the the link for ARM-GCC in [buildAndProgram.md](https://github.com/InfiniTimeOrg/InfiniTime/blob/9afc23cba9bcf938d8c49d6e15e7662ee8e6385d/doc/buildAndProgram.md#dependencies) leads to [this page](https://developer.arm.com/downloads/-/arm-gnu-toolchain-downloads), where the oldest available version is 11.2-2022.02. The version in the documentation (10.3-2021.10) seems to no longer be ~~available~~ published.

Edit: I guess the direct link to the tarball works, but it's not linked on their website any more.

WhyNotHugo:
Regardless of the documented versions no longer being available, the binary builds on that side are for GNU/Linux. Alpine uses musl, so those builds won't work without additional patching.

mark9064:
It's still linked, but you're right it's behind an extra page now

image

Unfortunate that there are libc problems :( I think in that case we can't "officially" support Alpine as the dependency versions need to be right - sorry for the disappointing conclusion

WhyNotHugo:
I don't think there's a strict dependency on GNU libc. The binaries produced on a host running musl seem to work fine.

mark9064:
What I meant is that the GCC 10.3 precompiled toolchain is incompatible with musl libc. And since there seems to be no other way to get 10.3 on Alpine, we can't support it officially in the install guide.

The builds *should* be OK at the moment at least compiler wise, I have no clue what the newer NRF SDK changes and that may cause some issues, not sure. Good that they appear to work though :)

PR #2342, opened 2025-08-30
Use std::optional for heart rate

WhyNotHugo:
If no value is available, use null. Use this distinction to differentiate in the UI between missing data and zero value.

Still sends value 0 over BLE. I'm unsure whether this should change or not


WhyNotHugo:
Checks have not completed because they require maintainer approval.

This isn't something I can do anything about, I'm not sure what's the purpose of this warning.

WhyNotHugo:
Building the simulator fails.

`docker run --rm -it -v ${PWD}:/sources infinitime-build` doesn't build the simulator too. I tried tweaking it to build it, but apparently multiple other dependencies are missing.

Is there an existing setup for this which I can use?

WhyNotHugo:
See: https://github.com/InfiniTimeOrg/InfiniSim/pull/180

PR #2349, opened 2025-10-07
Remove redundant touchpanel read

mark9064:
Split from #2325

> According to the datasheet there is nothing at register 0. So I figured we probably shouldn't be trying to read it. It's a machine translation though so maybe something is missing? https://wiki.pine64.org/images/2/2f/CST816S.zip


NeroBurner (approved):
LGTM

PR #2359, opened 2025-11-03
alarm: Add support for multiple alarms

### LINKED ISSUE #950: Multiple Alarms
### Verification

### Pitch us your idea!

Have multiple alarms running, for example 7am and 15pm daily

### Description

It would be nice to have the possibility to have multiple alarms running at the same time. For example one recurring alarm from Mo-Fr at 7am and then a second alarm daily at 1pm. Now I have to change the alarm time each time. That means, if I want to set a different alarm besides my recurring alarm for work Mo-Fr 7am I have to change this alarm and hopefully not forget to set it again. I think 2 different alarms are enough, more are welcome.

vkareh:
Replace single alarm with launcher screen showing 4 different alarms. Each alarm displays time, recurrence, and enable/disable toggle in a vertical list. Tapping alarm time opens configuration screen for that specific alarm.

InfiniSim_2025-10-31_093240
InfiniSim_2025-10-31_093246

Fixes #950

### DIFF
--- a/src/components/alarm/AlarmController.h
+++ b/src/components/alarm/AlarmController.h
@@ -19,6 +19,7 @@
 
 #include <FreeRTOS.h>
 #include <timers.h>
+#include <array>
 #include <cstdint>
 #include "components/datetime/DateTimeController.h"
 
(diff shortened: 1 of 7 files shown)


minacode:
This is a very beautiful UI! 🙂

tituscmd:
> This is a very beautiful UI! 🙂

Absolutely agree!

src/components/alarm/AlarmController.h
```diff
@@ -79,16 +108,19 @@ namespace Pinetime {
...
       bool alarmChanged = false;
+      uint8_t alertingAlarmIndex = 0;
+      uint8_t nextAlarmIndex = 0;
 
       Controllers::DateTime& dateTimeController;
       Controllers::FS& fs;
       System::SystemTask* systemTask = nullptr;
       TimerHandle_t alarmTimer;
-      AlarmSettings alarm;
+      AlarmSettings alarms[MaxAlarms];
```
minacode: Would it be better to use a `std::array` here? This seems to be the standard in other parts of the code base.

vkareh: Yes, good catch. I've updated the branch.

Itai-Nelken:
I've been using this for the past few days and it's great so far, thanks for implementing this!

One bug I noticed is that pressing the red "stop ringing" button doesn't make the ringing screen go away. Pressing the button does bring you back to the alarms screen.

Something else is that the alarms screen always shows time in 24H format instead of following the setting.

vkareh:
@Itai-Nelken  
> One bug I noticed is that pressing the red "stop ringing" button doesn't make the ringing screen go away. Pressing the button does bring you back to the alarms screen.
> 
> Something else is that the alarms screen always shows time in 24H format instead of following the setting.

Fixed both issues. Thanks for calling those out.

Itai-Nelken:
> Fixed both issues. Thanks for calling those out.

Thanks, works fine so far!

Something else I noticed is that the time to alarm always shows the next alarm regardless of the selected alarm.
But otherwise this has been working perfectly well for me for the past week!

PR #2363, opened 2025-11-07
timers: add hour to countdown timer

daniebker:
A small change to add an hour selector to the timer countdown. I found only being able to set minutes too limiting and sometimes need timers that go longer than 1 hour. 

## Testing

I’ve tested this in InfiniSim and on the pinetime.

1. Build and open up infinisim with this branch checked out
2. Navigate to the timer app
3. Expected you see H : M : S selectors, all on 0
4. Set H to 1  M to 0 and S to 1
5. Start the timer
6. After ticking down the timer should tick down to 0:59:59
7. Pause the timer. 
8. Set minutes to 50 and start the timer
9. Expected previous functionality works as expected and there are no regresions. 

I also have another change, unrelated to the timer, because I was having issues building the chrono on Mac.

```cpp
  currentDateTime += std::chrono::seconds(correctedDelta);
  uptime += std::chrono::seconds(correctedDelta);

  std::time_t currentTime = std::chrono::system_clock::to_time_t(
    std::chrono::time_point_cast<std::chrono::system_clock::duration>(currentDateTime));
  localTime = *std::localtime(&currentTime);

  auto minute = Minutes();
```

Using the `time_point_cast` resolved the issue for me, but I haven’t tested it on linux so can’t guarantee it still works there, which is why I didn’t include it as part of this MR. Has this issue been raised before?

### DIFF
--- a/src/components/ble/MotionService.cpp
+++ b/src/components/ble/MotionService.cpp
@@ -64,7 +64,7 @@ int MotionService::OnStepCountRequested(uint16_t attributeHandle, ble_gatt_acces
     NRF_LOG_INFO("Motion-stepcount : handle = %d", stepCountHandle);
     uint32_t buffer = motionController.NbSteps();
 
-    int res = os_mbuf_append(context->om, &buffer, 4);
+    int res = os_mbuf_append(context->om, &buffer, sizeof(buffer));
     return (res == 0) ? 0 : BLE_ATT_ERR_INSUFFICIENT_RES;
   }
   if (attributeHandle == motionValuesHandle) {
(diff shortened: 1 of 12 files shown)


minacode:
If you build the code using the docker container you should be able to verify your additional fix under Linux.

Do you have an explanation for why you encountered the problem under macOS? What is the problem there?

daniebker:
> If you build the code using the docker container you should be able to verify your additional fix under Linux.
> 
> Do you have an explanation for why you encountered the problem under macOS? What is the problem there?

Yes, I’ve built the code under the docker container and it was working fine, in fact that’s the one I sent to the watch. I just tried removing the block I have locally for the mac compilation, but now it’s building fine, even after wiping the build, so I can’t reproduce the error I was seeing. Anyway, I didn’t include it as part of the MR because I wasn't sure it wasn’t just a ‘me’ issue.

minacode:
Please fix the formatting error. Overall the code looks good to me, but I haven't tested it.

PR #2365, opened 2025-11-07
Refactor Timer component to provide expiry information

mark9064:
Also resolve the overflow in GetTimeRemaining for long timers

Prerequisite for #1971 (@vkareh any feedback please do chip in!)

It's a bit annoying to store a duplicate of the expiry time but I think it's the best way around xTimerGetExpiryTime being undefined after a timer expires

Should the API for timer remaining/since expired be two methods? I went with this as it allows getting the complete state with one function, but maybe this isn't a sensible API for a generic timer component

Untested!

### DIFF
--- a/src/components/timer/Timer.cpp
+++ b/src/components/timer/Timer.cpp
@@ -9,18 +9,33 @@ Timer::Timer(void* const timerData, TimerCallbackFunction_t timerCallbackFunctio
 void Timer::StartTimer(std::chrono::milliseconds duration) {
   xTimerChangePeriod(timer, pdMS_TO_TICKS(duration.count()), 0);
   xTimerStart(timer, 0);
+  expiry = xTimerGetExpiryTime(timer);
+  triggered = true;
 }
 
-std::chrono::milliseconds Timer::GetTimeRemaining() {
+// nullopt if timer stopped (StopTimer called / StartTimer not yet called)
+// otherwise TimerStatus with the ticks until/since expiry (depending on state of expired flag)
+std::optional<Timer::TimerStatus> Timer::GetTimerState() {
   if (IsRunning()) {
-    TickType_t remainingTime = xTimerGetExpiryTime(timer) - xTaskGetTickCount();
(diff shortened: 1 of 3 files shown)


src/components/timer/Timer.cpp
```diff
@@ -9,18 +9,33 @@ Timer::Timer(void* const timerData, TimerCallbackFunction_t timerCallbackFunctio
...
   xTimerStart(timer, 0);
+  expiry = xTimerGetExpiryTime(timer);
+  triggered = true;
 }
 
-std::chrono::milliseconds Timer::GetTimeRemaining() {
+// nullopt if timer stopped (StopTimer called / StartTimer not yet called)
+// otherwise TimerStatus with the ticks until/since expiry (depending on state of expired flag)
+std::optional<Timer::TimerStatus> Timer::GetTimerState() {
   if (IsRunning()) {
```
minacode, line 19: Is it really necessary to check if the timer is running?
I would like to argue that we only check whether the timer was triggered. And if it was, we can then check whether `expiry >= xTaskGetTickCount()` or not, to calculate the distance between the two and set `expired` to either true or false.

The function could start with
```
if (!triggered) {
    return std::nullopt;
}
// ...
```

minacode (commented):
I think the API design is good and general enough, because you basically just implemented a signed timedelta w.r.t the expiry time. I see no problem with this being to specific for other use cases where timers are needed. Whenever the expiry is not needed, it can easily be ignored.

src/components/timer/Timer.cpp
mark9064: The tick count wraps around so that gets tricky - you'd have to calculate which is longer out of the duration to the expiry or the duration from the expiry to the current time, and then assume the state based off that. Half of the timer's range is lost though:

Suppose the tick count wraps around at 1000
If the current time is 500, and we start a timer expiring in 700 ticks the expiry will be 200. But if we take the closest direction, we'd assume the timer has already expired and did so 300 ticks ago (500-200).

So I think the most robust way is just to check whether it's actually expired or not

minacode: I see, thank you :)

PR #2371, opened 2025-11-14
PPGv2

mark9064:
- New heart rate algorithm
   - Adaptive filtering to remove noise caused by motion
   - Harmonic analysis to match measured waveform with a real PPG waveform to reject noise
   - Frequency tracking and refinement over time
- Heart rate detection can now function in more challenging environments e.g. while running
- Full automatic gain and drive implementation for HRS3300 resulting in power saving in most scenarios while being able to adapt to full sunlight conditions on light skin, also adapts to darker skin tones better
  - Including no touch detection
- Much better noise tolerance: this algorithm reports failure instead of an incorrect heart rate value when the signal is poor. The heart rate values reported can be trusted a lot more
   - Algorithm accuracy validated against an ECG chest strap (Polar H10)
   - Also good performance on PPG-DaLiA (public PPG dataset): Mean absolute error (MAE) ~8BPM

Just to set expectations, while this is a solid improvement the HRS3300 sensor is really not that great so this implementation is nowhere the performance of flagship smartwatches today in terms of accuracy and recall

Big thanks to everyone over at wasp-os who [investigated the sensor performance](https://github.com/wasp-os/wasp-os/pull/363#issuecomment-1279733038), this work wouldn't have been possible without

I'm not sure it will be possible to merge this due to the memory usage of the adaptive filter. It might conflict with the G7710 watchface (I haven't tested as I use a custom G7710 version with seconds, so the fonts are smaller and use less memory)

This changes the way heart rate values are presented. With this implementation, you will only see a number when the algorithm is actively tracking the heart rate, or if the last background measurement was successful and the algorithm does not yet have enough data to begin searching for a heart rate (8 seconds of data)

### DIFF
--- a/src/displayapp/screens/HeartRate.cpp
+++ b/src/displayapp/screens/HeartRate.cpp
@@ -1,6 +1,6 @@
 #include "displayapp/screens/HeartRate.h"
-#include <lvgl/lvgl.h>
 #include <components/heartrate/HeartRateController.h>
+#include <lvgl/lvgl.h>
 
 #include "displayapp/DisplayApp.h"
 #include "displayapp/InfiniTimeTheme.h"
@@ -12,12 +12,16 @@ namespace {
     switch (s) {
       case Pinetime::Controllers::HeartRateController::States::NotEnoughData:
         return "Not enough data,\nplease wait...";
+      case Pinetime::Controllers::HeartRateController::States::Searching:
+        return "Searching...";
       case Pinetime::Controllers::HeartRateController::States::NoTouch:
         return "No touch detected";
-      case Pinetime::Controllers::HeartRateController::States::Running:
+      case Pinetime::Controllers::HeartRateController::States::Ready:
         return "Measuring...";
       case Pinetime::Controllers::HeartRateController::States::Stopped:
         return "Stopped";
+      case Pinetime::Controllers::HeartRateController::States::Disabled:
+        return "Disabled";
     }
     return "";
   }
@@ -30,7 +34,7 @@ namespace {
 
 HeartRate::HeartRate(Controllers::HeartRateController& heartRateController, System::SystemTask& systemTask)
   : heartRateController {heartRateController}, wakeLock(systemTask) {
-  bool isHrRunning = heartRateController.State() != Controllers::HeartRateController::States::Stopped;
+  bool isHrRunning = heartRateController.State() != Controllers::HeartRateController::States::Disabled;
   label_hr = lv_label_create(lv_scr_act(), nullptr);
 
   lv_obj_set_style_local_text_font(label_hr, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, &jetbrains_mono_76);
@@ -50,7 +54,7 @@ HeartRate::HeartRate(Controllers::HeartRateController& heartRateController, Syst
 
   label_status = lv_label_create(lv_scr_act(), nullptr);
   lv_obj_set_style_local_text_color(label_status, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, LV_COLOR_GRAY);
-  lv_label_set_text_static(label_status, ToString(Pinetime::Controllers::HeartRateController::States::NotEnoughData));
+  lv_label_set_text_static(label_status, ToString(Pinetime::Controllers::HeartRateController::States::Disabled));
 
   lv_obj_align(label_status, label_hr, LV_ALIGN_OUT_BOTTOM_MID, 0, 10);
 
@@ -77,18 +81,10 @@ HeartRate::~HeartRate() {
 void HeartRate::Refresh() {
 
   auto state = heartRateController.State();
-  switch (state) {
-    case Controllers::HeartRateController::States::NoTouch:
-    case Controllers::HeartRateController::States::NotEnoughData:
-      // case Controllers::HeartRateController::States::Stopped:
-      lv_label_set_text_static(label_hr, "---");
-      break;
-    default:
-      if (heartRateController.HeartRate() == 0) {
-        lv_label_set_text_static(label_hr, "---");
-      } else {
-        lv_label_set_text_fmt(label_hr, "%03d", heartRateController.HeartRate());
-      }
+  if (heartRateController.HeartRate() == 0) {
+    lv_label_set_text_static(label_hr, "---");
+  } else {
+    lv_label_set_text_fmt(label_hr, "%03d", heartRateController.HeartRate());
   }
 
   lv_label_set_text_static(label_status, ToString(state));
@@ -97,14 +93,14 @@ void HeartRate::Refresh() {
 
 void HeartRate::OnStartStopEvent(lv_event_t event) {
   if (event == LV_EVENT_CLICKED) {
-    if (heartRateController.State() == Controllers::HeartRateController::States::Stopped) {
+    if (heartRateController.State() == Controllers::HeartRateController::States::Disabled) {
       heartRateController.Enable();
-      UpdateStartStopButton(heartRateController.State() != Controllers::HeartRateController::States::Stopped);
+      UpdateStartStopButton(heartRateController.State() != Controllers::HeartRateController::States::Disabled);
       wakeLock.Lock();
       lv_obj_set_style_local_text_color(label_hr, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, Colors::highlight);
     } else {
       heartRateController.Disable();
-      UpdateStartStopButton(heartRateController.State() != Controllers::HeartRateController::States::Stopped);
+      UpdateStartStopButton(heartRateController.State() != Controllers::HeartRateController::States::Disabled);
       wakeLock.Release();
       lv_obj_set_style_local_text_color(label_hr, LV_LABEL_PART_MAIN, LV_STATE_DEFAULT, Colors::lightGray);
     }
--- a/src/displayapp/screens/HeartRate.cpp
+++ b/src/displayapp/screens/HeartRate.cpp
@@ -16,7 +16,7 @@ namespace {
         return "Searching...";
       case Pinetime::Controllers::HeartRateController::States::NoTouch:
         return "No touch detected";
-      case Pinetime::Controllers::HeartRateController::States::Ready:
+      case Pinetime::Controllers::HeartRateController::States::Measuring:
         return "Measuring...";
       case Pinetime::Controllers::HeartRateController::States::Stopped:
         return "Stopped";
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,6 +4,3 @@
 [submodule "src/libs/littlefs"]
 	path = src/libs/littlefs
 	url = https://github.com/littlefs-project/littlefs.git
-[submodule "src/libs/arduinoFFT"]
-	path = src/libs/arduinoFFT
-	url = https://github.com/kosme/arduinoFFT.git
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -675,9 +675,9 @@ set(INCLUDE_FILES
         heartratetask/HeartRateTask.h
         components/heartrate/Ppg.h
         components/heartrate/HeartRateController.h
-        libs/arduinoFFT/src/arduinoFFT.h
-        libs/arduinoFFT/src/defs.h
-        libs/arduinoFFT/src/types.h
+        components/heartrate/FFT.h
+        components/heartrate/IIR.h
+        components/heartrate/RLS.h
         components/motor/MotorController.h
         buttonhandler/ButtonHandler.h
         touchhandler/TouchHandler.h
--- a/src/components/heartrate/FFT.h
+++ b/src/components/heartrate/FFT.h
@@ -1,0 +1,125 @@
+#include <complex>
+#include <array>
+#include <bit>
+#include <numbers>
+
+namespace Pinetime {
+  namespace Utility {
+    // Fast Fourier transform
+    // Implements in-place N to N point complex-to-complex FFT
+    // Implements in-place 2N to N point real-to-complex FFT
+    // Performing these transforms requires some "twiddling" constants to be known
+    // These constants depend only on the size of the transform
+    // Since they are expensive to compute, they can only be computed at compile time (consteval)
+    class FFT {
+    public:
+      static consteval std::size_t IntegerLog2(std::size_t n) {
+        return std::bit_width(n) - 1;
+      }
+
+      template <std::size_t N>
+      static consteval std::array<std::complex<float>, IntegerLog2(N)> GenComplexTwiddle() {
+        using namespace std::complex_literals;
+
+        std::array<std::complex<float>, IntegerLog2(N)> result;
+        for (std::size_t i = 0; i < IntegerLog2(N); i++) {
+          result[i] = exp_consteval(-2.i * std::numbers::pi / static_cast<double>(1 << (i + 1)));
+        }
+        return result;
+      }
+
+      template <std::size_t N>
+      static consteval std::array<std::complex<float>, (N / 4) - 1> GenRealTwiddle() {
+        using namespace std::complex_literals;
+
+        std::array<std::complex<float>, (N / 4) - 1> result;
+        for (std::size_t i = 0; i < (N / 4) - 1; i++) {
+          result[i] = exp_consteval(-2.i * std::numbers::pi * static_cast<double>(i + 1) / static_cast<double>(N));
+        }
+        return result;
+      }
+
+      template <std::size_t N>
+      static void ComplexFFT(std::array<std::complex<float>, N>& array, const std::array<std::complex<float>, IntegerLog2(N)>& twiddle) {
+        // In-place Cooley-Tukey
+        InplaceBitReverse(array);
+        for (std::size_t s = 1; s < IntegerLog2(N) + 1; s++) {
+          std::size_t m = 1 << s;
+          for (std::size_t k = 0; k < N; k += m) {
+            std::complex<float> omega = 1.f;
+            for (std::size_t j = 0; j < m / 2; j++) {
+              std::complex<float> t = omega * array[k + j + (m / 2)];
+              std::complex<float> u = array[k + j];
+              array[k + j] = u + t;
+              array[k + j + (m / 2)] = u - t;
+              omega *= twiddle[s - 1];
+            }
+          }
+        }
+      }
+
+      template <std::size_t N>
+      static void RealFFT(std::array<std::complex<float>, N>& array,
+                          const std::array<std::complex<float>, (N / 2) - 1>& realTwiddle,
+                          const std::array<std::complex<float>, IntegerLog2(N)>& complexTwiddle) {
+        using namespace std::complex_literals;
+
+        // See https://www.robinscheibler.org/2013/02/13/real-fft.html for how this works
+        FFT::ComplexFFT(array, complexTwiddle);
+        // Compute DC bin directly (xe/xo simplify)
+        // Nyquist bin ignored as unneeded
+        array[0] = array[0].real() + array[0].imag();
+        // Since computations depend on the inverse of the index (mirrored)
+        // compute two at once, outside to inside
+        for (std::size_t index = 1; index < N / 2; index++) {
+          std::size_t indexInv = N - index;
+          std::complex<float> xeLo = (array[index] + std::conj(array[indexInv])) / 2.f;
+          std::complex<float> xoLo = -1.if * ((array[index] - std::conj(array[indexInv])) / 2.f);
+          std::complex<float> xeHi = (array[indexInv] + std::conj(array[index])) / 2.f;
+          std::complex<float> xoHi = -1.if * ((array[indexInv] - std::conj(array[index])) / 2.f);
+          array[index] = xeLo + (xoLo * realTwiddle[index - 1]);
+          array[indexInv] = xeHi + (xoHi * -std::conj(realTwiddle[index - 1]));
+        }
+        // Middle element not computed by above loop
+        // Since index == indexInv
+        // the middle simplifies to the conjugate as the twiddle is always -i
+        std::size_t middle = N / 2;
+        array[middle] = std::conj(array[middle]);
+      }
+
+    private:
+      // consteval wrappers of builtins
+      template <typename _Tp>
+      static consteval std::complex<_Tp> exp_consteval(const std::complex<_Tp>& __z) {
+        return polar_consteval<_Tp>(__builtin_exp(__z.real()), __z.imag());
+      }
+
+      template <typename _Tp>
+      static consteval std::complex<_Tp> polar_consteval(const _Tp& __rho, const _Tp& __theta) {
+        return std::complex<_Tp>(__rho * __builtin_cos(__theta), __rho * __builtin_sin(__theta));
+      }
+
+      template <class T, std::size_t N>
+      static void InplaceBitReverse(std::array<T, N>& array) {
+        // Gold-Rader algorithm
+        // Faster algorithms exist, but this is sufficient
+        std::size_t swapTarget = 0;
+        for (std::size_t index = 0; index < N - 1; index++) {
+          // Only swap in one direction
+          // Otherwise entire array gets swapped twice
+          if (index < swapTarget) {
+            T temp = array[index];
+            array[index] = array[swapTarget];
+            array[swapTarget] = temp;
+          }
+          std::size_t kFactor = N / 2;
+          while (kFactor <= swapTarget) {
+            swapTarget -= kFactor;
+            kFactor /= 2;
+          }
+          swapTarget += kFactor;
+        }
+      }
+    };
+  }
+}
--- a/src/components/heartrate/HeartRateController.cpp
+++ b/src/components/heartrate/HeartRateController.cpp
@@ -1,11 +1,15 @@
 #include "components/heartrate/HeartRateController.h"
-#include <heartratetask/HeartRateTask.h>
-#include <systemtask/SystemTask.h>
+
+#include <cstdint>
+#include "heartratetask/HeartRateTask.h"
 
 using namespace Pinetime::Controllers;
 
-void HeartRateController::Update(HeartRateController::States newState, uint8_t heartRate) {
+void HeartRateController::UpdateState(HeartRateController::States newState) {
   this->state = newState;
+}
+
+void HeartRateController::UpdateHeartRate(uint8_t heartRate) {
   if (this->heartRate != heartRate) {
     this->heartRate = heartRate;
     service->OnNewHeartRateValue(heartRate);
@@ -14,14 +18,14 @@ void HeartRateController::Update(HeartRateController::States newState, uint8_t h
 
 void HeartRateController::Enable() {
   if (task != nullptr) {
-    state = States::NotEnoughData;
+    state = States::Stopped;
     task->PushMessage(Pinetime::Applications::HeartRateTask::Messages::Enable);
   }
 }
 
 void HeartRateController::Disable() {
   if (task != nullptr) {
-    state = States::Stopped;
+    state = States::Disabled;
     task->PushMessage(Pinetime::Applications::HeartRateTask::Messages::Disable);
   }
 }
--- a/src/components/heartrate/HeartRateController.h
+++ b/src/components/heartrate/HeartRateController.h
@@ -15,12 +15,13 @@ namespace Pinetime {
   namespace Controllers {
     class HeartRateController {
     public:
-      enum class States : uint8_t { Stopped, NotEnoughData, NoTouch, Running };
+      enum class States : uint8_t { Disabled, Stopped, NotEnoughData, Searching, Ready, NoTouch };
 
       HeartRateController() = default;
       void Enable();
       void Disable();
-      void Update(States newState, uint8_t heartRate);
+      void UpdateState(States newState);
+      void UpdateHeartRate(uint8_t heartRate);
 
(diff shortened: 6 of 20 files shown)


mark9064:
Oh, this needs GCC 14 as well

mark9064:
GCC 14 here #2372, not going to rebase on to it as I don't think that will fix CI

tituscmd:
Ooooh, I like this a lot.

If this is ready to test on-device, I would be more than willing to daily drive it for a while and report its results!

mark9064:
Yeah this is ready, I've been working on this for >2 years so I'm pretty happy with everything here (of course there's gonna be review feedback etc. but the design of the detector is done in my eyes)

tituscmd:
I'm gonna merge both GCC 14 and this PR onto my main device and run it for a while. I'll report back here when something noteworthy happens!

tituscmd:
I’m particularly excited about my gym session and the bike ride there tomorrow. Interested to see how it performs there!

tituscmd:
> Frequency tracking and refinement over time

What does this mean? I've had a few measurements spaced apart more than my background measurement interval setting which confused me.

minacode:
Wow, good work!

> Including no touch detection

Can we now reliably detect if the watch is not worn?

mark9064:
> > Frequency tracking and refinement over time
> 
> What does this mean? I've had a few measurements spaced apart more than my background measurement interval setting which confused me.

The heart rate is tracked internally by the algorithm rather than smoothing over a buffer of the last values so accuracy is a bit better. There's no outward behaviour change

If measurements are spaced farther than the background interval, it's probably because a measurement failed

mark9064:
> Wow, good work!
> 
> > Including no touch detection
> 
> Can we now reliably detect if the watch is not worn?

I'm not sure I'd call it super reliable, any surface reflects light in the same way as skin. But it does catch the sensor not being near any surface

Andersama:
I should've known to check the PR's for interesting changes. I'd love to test this.

minacode:
I would like to review this from a more theoretical perspective. Can you do a little write-up of the logic behind this? I tried to get it from the code, but got a little overwhelmed 😄

mark9064:
Yeah it's quite a lot, and understanding each step exactly requires some DSP knowledge that can be quite tricky to learn (at least it was for me!)

The pipeline has two overall stages, the first of which runs for every ingested sample and the second of which is evaluated on the first stage output every half of a second.

Stage 1:
- Pass the incoming tuples of `(reflectance, accx, accy, accz)` through individual high-pass filters (high-pass filters transmit higher frequencies and attenuate lower ones; the cut off frequency and steepness of the cut off is a choice made when designing the filter)
  - Why? Removes DC baseline from the signal and other long term trends (low frequency components)
- Use the filtered acceleration channels `acc` to estimate noise in the reflectance signal and remove it (this is performed by the adaptive filter)
  - Why? Removes (some) noise caused by motion artefacts, which is the biggest contributor to the reflectance signal

We now have the filtered reflectance signal which we stream into an 8s long buffer. Twice per second: 
(Stage 2)
- Energy normalise a copy of the buffer data
  - Why? Provides input scale invariance, i.e. tolerance against different skin reflectivities, different sensor drive/gain modes, different ambient lighting conditions
- Transform it to the frequency domain using the Fourier transform
  - Why? Heart rate BVP signal is periodic so easier to identify in the frequency domain
- Detect peaks in the signal with a prominence filter (bandpass the spectrum)
  - Why? Suppresses wideband spectral noise caused by discontinuties in the signal (usually transient motion artefacts)
- Filter the peaks based on whether the signal matches the well known shape of a heart rate signal at that frequency
  - Why? Attenuates peaks caused by things other than the BVP
- Threshold resulting filtered spectrum to find strong peaks
  - Why? Want to avoid tracking peaks with little energy (weak peaks) as these are more likely to be noise
- Accept strong peaks and track peaks over time
  - Why? Strength of the true HR peak will vary over time as signal quality changes, so once it has been detected tracking it over time provides much better recall (as opposed to only reporting a value when a strong enough peak is present)

Aside: The behaviour around the stopped state is a bit broken currently, will fix when I have some time

mark9064:
(Will squash fixups down later, I know a few people have merged this in locally so trying to avoid rebasing for now)

Andersama:
@mark9064 attempted a build merging my commit as well, it looks like there's a lot of compiler errors. I'll patch them up and give this a shot.

Ok from an initial quick overview: I'd replace the lines:
```c++
std::array<std::complex<float>, IntegerLog2(N)> result;
```
with
```c++
std::array<std::complex<float>, IntegerLog2(N)> result{};
```
I don't know about GCC, but I've had plenty of experience with constexpr functions failing because variables aren't initialized*. It's a bit silly but you should `{}` even though the values will be overwritten in the for loop.

and then I've never seen this as an expression, might be a gcc extension? I can't compile this with GCC on godbolt.
```c++
-2.i * std::numbers::pi
```
my best guess is
```c++
-2. * i * std::numbers::pi
```
I haven't touched FFTs in ages so I'd have to work out what's going on here.

I'd drop bringing in `<numbers>` just for the pi constant. That seems to resolve the issues in the FFT section.

Update: Assuming the guess above was ok, I made edits that got it to compile and build. I'm not sure that it was a correct assumption because the samples I'm getting back are anywhere between 150 to 200 bpm, which is a bit high considering I was maybe at 70-90 bpm.

To clarify this is buildable with a few small edits. The constexpr errors can be resolved by doing the calculations needed in a constructor for the `std::complex<float>`s.

Andersama:
@mark9064 I'm going to guess I messed up the FFT in some way trying to patch up the errors around `std::complex`, it doesn't appear to me that the code would need a compiler version bump, for reference here were my edits off of this PR: https://github.com/Andersama/InfiniTime/tree/ppg_upgrade

mark9064:
Does it compile for you with GCC14? It does on my machine

Upgrading the compiler is not a big deal, so I wouldn't worry too much about trying to workaround unimplemented functionality. We try to keep the compiler up to date anyway (as well as other dependencies) but haven't recently due to some linker warnings (which we now understand)

I appreciate you looking into this though :)

mark9064:
Also anyone got any feedback? Criticism is welcome too, this approach is no good if it only works well for me :)

Andersama:
@mark9064 I'm building off of the included docker image, so whichever GCC's included there, it wouldn't compile, but also I tested the same expression on godbolt (with GCC14 and trunk) https://godbolt.org/z/5hnEo5ej6 and it also wouldn't compile so I'm not clear on what `-2.i * std::numbers::pi` was supposed to do, because to me, that looks like an error, and it appears to be treated as an error. If it were me I'd just write whatever mathematically works out to calculate the twiddle values into the `std::complex<float>` constructor for the real and imaginary components, because float operations have been constexpr for quite a while and the constructor is simple enough it didn't appear GCC complained about that approach when I tried it. Although I probably messed up the math.

I did with some fiddling and edits and managed to get it to compile with whichever GCC's in use now...overall I'm just not 100% certain I did quite the right thing, because the heart rate values I was getting back were wildly off.

Also: It's the same errors that are appearing in the build checks for the PR here: https://github.com/InfiniTimeOrg/InfiniTime/actions/runs/19586678888/job/56096912190?pr=2371
I wanted to test this out so, I at least got past the compiler errors, but I'm not sure if I'm dealing with a semantic error.

mark9064:
Use #2372 and rebuild the docker container from there 

The compiler error in Godbolt tells you exactly what you need to add :) and you'll see that `using namespace std::complex_literals;` is in the original code too (this provides `i` so inline imaginary constants are easy to write)

Andersama:
Ah, well serves me right for not paying attention to the godbolt logs, weird that it's not the same error. Ah, you know what I bet it's buried behind the constexpr failure. The simplified expression in godbolt I dropped constexpr in the surrounding function so that's probably the actual issue, no contexpr `i` operator.

Oh, and that probably explains my bad results, because I thought `i` was the loop variable...so my twiddle values in my code are probably garbage.

Andersama:
@mark9064 slight update, fixed up the apparent math error, compiles and appears to work as intended. I'll have to grab my other heart rate sensor to verify, but the numbers make much more sense now, currently recording ~80 bpm. I rewrote the offending blocks like so:
```c++
      template <std::size_t N>
      static consteval std::array<std::complex<float>, IntegerLog2(N)> GenComplexTwiddle() {
        using namespace std::complex_literals;

        std::array<std::complex<float>, IntegerLog2(N)> result {};
        for (std::size_t i = 0; i < IntegerLog2(N); i++) {
          std::complex<double> tmp = std::complex<double> {0.0, -2.0} * std::numbers::pi / static_cast<double>(1 << (i + 1));
          std::complex<float> value = exp_consteval(std::complex<float> {(float) tmp.real(), (float) tmp.imag()});
          result[i] = value;
        }
        return result;
      }

      template <std::size_t N>
      static consteval std::array<std::complex<float>, (N / 4) - 1> GenRealTwiddle() {
        using namespace std::complex_literals;

        std::array<std::complex<float>, (N / 4) - 1> result {};
        for (std::size_t i = 0; i < ((N / 4) - 1); i++) {
          std::complex<double> tmp =
            std::complex<double> {0.0, -2.0} * std::numbers::pi * static_cast<double>(i + 1) / static_cast<double>(N);
          std::complex<float> value = exp_consteval(std::complex<float> {(float) tmp.real(), (float) tmp.imag()});
          result[i] = value;
        }
        return result;
      }
```
      
There were more errors that cropped up since the compiler didn't appreciate the conversion from `std::complex<double>` to `std::complex<float>` so I had to break the expression apart into steps. Apparently there's an option to use `if` as a literal operator to generate `std::complex<float>` from a literal instead of a double which probably would simplify this mess...

I suppose the only thing I would wonder is if a windowing function was used, it seemed like the old arduino fft based code just performed an fft with no windowing function at all, and I'm wondering if that would've helped.

Update, after sitting with a finger ppg sensor, this seems very accurate, I'd say it almost stays in complete sync with the finger sensor, except for the random blips where it drops to 0, will be testing it on a workout later.

Further update: Two different workouts later, did very poorly on an outdoor bike ride most of the readings were 0's virtually unusable for feedback for how I was doing. Second one indoors did much much better, only a few samples of 0s across a whole hour, and was responsive enough I actually managed to complete a workout with set target heartrates, so very nice...in good conditions. My best guess, lighting and potentially an amount of motion is a significant issue.

Andersama:
This conceptually is the same result, compliable with the current version of GCC.
```c++
      template <std::size_t N>
      static consteval std::array<std::complex<float>, IntegerLog2(N)> GenComplexTwiddle() {
        using namespace std::complex_literals;

        std::array<std::complex<float>, IntegerLog2(N)> result {};
        for (std::size_t i = 0; i < IntegerLog2(N); i++) {
          result[i] = exp_consteval(std::complex<float> {0.0f, (float) (-2.0 * (std::numbers::pi / static_cast<double>(1 << (i + 1))))});
        }
        return result;
      }

      template <std::size_t N>
      static consteval std::array<std::complex<float>, (N / 4) - 1> GenRealTwiddle() {
        using namespace std::complex_literals;

        std::array<std::complex<float>, (N / 4) - 1> result {};
        for (std::size_t i = 0; i < ((N / 4) - 1); i++) {
          result[i] = exp_consteval(
            std::complex<float> {0.0f, (float) (-2.0 * (std::numbers::pi * static_cast<double>(i + 1) / static_cast<double>(N)))});
        }
        return result;
      }
```

Andersama:
So in testing this for a bit @mark9064 I've found some weird quirks, namely it seems like the orientation of the watch drastically impacts how well it records, I'm wondering if maybe the motion compensation code might be a bit off.

mark9064:
When you say orientation, what exactly do you mean?

mark9064:
PPG Debug
Also when I say the sensor is just too bad in some scenarios, here's an example from right now where my hands are very cold. No algorithm can get a heart rate from this data - there's just no signal (look at the raw HRS channel) despite me being perfectly still

Andersama:
@mark9064 I'm finding that when using a stationary (somewhat reclined) bike if my hands are resting on the handlebars it eventually fails to estimate my heart rate, whereas if I lift my arm in the air, rest it on my chest or head it gets data again? I'm wondering if maybe it's picking up my cadence on the bike as a signal via motion where maybe it doesn't as much on the bar.

But since it so quickly fails when I rest my hand on the bar I was wondering if maybe the motion compensation code was accidentally taking the orientation of the watch and using that in some calculation.

Where'd you get the graph from? I'd be curious to see what the watch is working with as well.

mark9064:
The motion filtering takes all acceleration channels into account and treats them equally, so it's not sensitive to the orientation. However, the watch itself is sensitive to orientation as the contact pressure between the sensor and the skin changes with orientation, and contact pressure changes the signal strength a lot.

What you're probably seeing is when your hands are on the handlebars there's a combination of:
1. The motion becoming much weaker in intensity according to the watch accelerometer (as the handlebars don't move much) so it's harder to filter motion noise
2. The wrist position changing the contact pressure in a way that degrades the signal

The graph is from some debug tooling, I can try and publish it but honestly it is very brittle and it will be difficult for me to package it in a way that other people can run. It also requires an extension to InfiniTime to stream PPG values over bluetooth

mark9064:
Re your specific scenario: you could try wearing it a bit further up your arm when you cycle, see if that works better

Andersama:
@mark9064 I wouldn't mind a tweak to the firmware, given I'm now running a completely separate and custom branch of my own.

Well I feel a bit silly given how much feedback from issues on how the sensors aren't exactly reliable and then when I went to use it I found exactly that. The thing I noticed is that sweat buildup negatively impacts the sensor readings. Which prompted me to compare my old fitbit and a garmin watch, where I found both of those an interesting design difference. They have two leds with diffusers surrounding the photoresistor and then the IR led off to one side. That suggests to me that part of the struggle with getting good readings is in the packaging / design of the leds of the sensor, and maybe not so much the rest of the components.

tituscmd:
Can anyone report anything similar to me?
I keep getting the occasional super high reading of 150BPM+, sometimes isolated values, sometimes just a lot of them:

image 
image

mark9064:
Interesting, I'm guessing from the timestamps that's while sleeping?

I'd say it's mostly likely a problem with the algorithm rather than something you're doing wrong. How tight do you wear the watch?

If you have access to a computer with working bluetooth, I can send a build that allows you to do data collection and I can review it after to see what's gone wrong

tituscmd:
> Interesting, I'm guessing from the timestamps that's while sleeping?

Yes, these screenshots are. I've had it occur while just sitting at my desk as well, tho.


> I'd say it's mostly likely a problem with the algorithm rather than something you're doing wrong. How tight do you wear the watch?

Decently tight. It's not so tight that I constantly feel it, but tight enough that it doesn't move when I shake my wrist


> If you have access to a computer with working bluetooth, I can send a build that allows you to do data collection and I can review it after to see what's gone wrong

For sure, that would work. Thanks to InfiniLink tho I can also export my HRM data to a csv file if that's any good for you?

mark9064:
Alright

How this works is it saves the raw PPG and accelerometer data to a file on the watch, so the file can be analysed later.

You'll need this commit (or the ppg-debug branch)

https://github.com/mark9064/InfiniTime/commit/b30c8eba4f7d6ab3760209b03c306225f0068824

Since it uses extra memory to store the data, you might have to change to the digital watchface to avoid crashes.

Every time the heart rate measurement starts, the file on the watch gets erased to make way for the new data. In other words, the file on disk represents a contiguous recording. To make a recording, use the continuous mode and start heart rate measurement. When you've reproduced the issue, stop heart rate measurement. After this you must not start measurement again (until you've transferred the file off the watch) or you will lose your recording!

Note: you must create a file called `hrs-acc-bin` in the root of the watch's filesystem. The contents don't matter, it will be overwritten. If no `hrs-acc-bin` file exists, recording is disabled.

To get the data from the watch these two scripts are required. Place them in the same directory. You will need to install `bleak`, a Python library used for Bluetooth LE communication. If you're new to Python, I would recommend setting up a virtual environment (aka venv), activating that and then installing bleak (guides for setting up virtual environments are easy to find online).


[file.py](https://github.com/user-attachments/files/24438746/file.py)
[run_full_collect.py](https://github.com/user-attachments/files/24438816/run_full_collect.py)


Then run the second script: `python run_full_collect.py`. This will stream the file to your watch twice in a row (as a basic integrity check measure), and then save the file in the same directory. The watch can store a bit more than 4 hours of continuous recording (capped at 3MiB), and streaming that entire amount will take a few minutes. Smaller recordings transfer faster as you'd expect.

I know that's a lot, let me know how it goes. If successful drop the `.bin` file here and I will try and see what's going on. Feel free to send multiple recordings or longer ones where the issue happens multiple times, more data is better :)

tituscmd:
Totally fine! I'm sure I'll be able to handle it.

I can't guarantee any timeframe of when I'll get to that tho. School is about to start in a few days so I'll be quite busy.

mark9064:
Oh yeah no rush, whenever you have the time

tituscmd:
For the time being, this is today's chart of heart rate data and the time span where you can see all those outlier points is exactly where I was sleeping (don't judge my sleep schedule 😅)
Might have something to do with the fact that I use a 2min interval while sleeping and a 5min interval while anytime else.image

mark9064:
Resolved possible hang after no touch reset (probably not noticeable in use but would cause spinning of the heart rate task)

tituscmd:
After some more time of usage I've also noticed that it almost always seems to continuously turn on when not on my wrist and laying on some surface.

mark9064:
Yeah no touch detection is very hard, a lot of surfaces are about as reflective as skin. It works sometimes and I think that's about the best we can do. I think in most cases you can use the relative strength of the different light wavelengths to judge skin vs not skin, but single the HRS3300 is single wavelength we can't do this.

I also think there is some room for tuning the no touch thresholds though, I feel they might be a bit too high.

tituscmd:
What's weird to me tho is that my interval is set to 10mins, but its seemingly running continuously. I imagine it's looping trying to find a skin-like surface?

mark9064:
Oh right. It shouldn't do that and doesn't for me. Can you try the latest build? (and any more specific repro steps would be great)

tituscmd:
Sure, I'll try latest.
Where I usually notice it is when I go to take a shower and place my watch on the windowsill. I don't modify the heart rate settings in any way, just take it off and put it down. I put it down on it's side but not facing towards the window, moreso facing into the room. When I go to take my shower, after a few minutes I notice that the green light is running endlessly.

mark9064:
Thanks, will see if I can observe this. Mine is set to 5 mins and I treat it the same when showering, but have spotted the light coming on for 30s as expected and then giving up

tituscmd:
Anything else that's new/different I should be on the lookout for?

mark9064:
Nothing other than the exact issue you described (sorry missed the notification for your reply)

Andersama:
@mark9064 popping back by, I haven't been using the watch for a while. Ended up needing a more reliable way to monitor my heart rate. Good news is, doing better, and now have a watch to compare results against. Bad news, don't exactly need two watches...but I do enjoy a side project, if you'd like comparative data. I can definitely do that.

mark9064:
Hi Alex! Yeah very reasonable choice to pursue other hardware, the pinetime sensor isn't competitive with modern ones.

I'm unsure how good the comparison data will be. The accuracy of smartwatch HR is quite variable so it'll be difficult to know which watch is wrong. For my own testing and validation I used an ECG based chest strap as the ground truth which is generally regarded as reliable. It still might be an interesting test though, if you try it I'd be interested to see results :)

mark9064:
Hmm, I'm becoming more convinced that infinite running is sometimes possible. But I think I've only (possibly) seen it in the no touch state. Do you have any ideas what conditions cause it to reproduce?

tituscmd:
> Hmm, I'm becoming more convinced that infinite running is sometimes possible. But I think I've only (possibly) seen it in the no touch state. Do you have any ideas what conditions cause it to reproduce?

I was thinking it has something to do with reflections. Because when I put my fingers over the continuously flashing light, it stops (until it starts again)

mark9064:
When you say it's continuously flashing, is that the every ~2s no touch flashing or the constant (every 50ms) flashing

tituscmd:
Constant flashing

tituscmd:
Here's a video of what I'm talking about!
https://github.com/user-attachments/assets/d7a14f75-699c-4efa-a183-83dfe93b4aa4

tituscmd:
The heart app says "Stopped" while it's doing its flashing

mark9064:
Well that doesn't sound good. That implies there's something busted with the state machine, or somehow some other change is interfering. Anything else on your branch that could be relevant?

I've found the issue I was talking about which only applies to the no touch state. If the sensor is reading exactly 0, the values get discarded as sensor startup values so the exit condition is never set

tituscmd:
> Anything else on your branch that could be relevant?

Well I use the background measurements for HR branch, and a lot of other things, but none of those touch the heart rate stuff. It's stuff like a new watch face, ANCS, new apps etc.

mark9064:
Hmm, maybe they are the same issue, but through a different path

mark9064:
I've fixed this, but I've decided I don't like the state machine at all. Going to rework

mark9064:
Don't really see a way to avoid the complexity of the current machine so ended up reworking it a little

The problematic case was that on a PPG value of zero it would never hit the gain control or the stop condition, so it would hang

Does this fix it for you? @tituscmd

tituscmd:
> Does this fix it for you? @tituscmd

Will check it out and report back ASAP!

src/displayapp/screens/HeartRate.cpp
```diff
@@ -30,7 +34,7 @@ namespace {
 
 HeartRate::HeartRate(Controllers::HeartRateController& heartRateController, System::SystemTask& systemTask)
   : heartRateController {heartRateController}, wakeLock(systemTask) {
-  bool isHrRunning = heartRateController.State() != Controllers::HeartRateController::States::Stopped;
+  bool isHrRunning = heartRateController.State() != Controllers::HeartRateController::States::Disabled;
```
tituscmd, line 37: I'm wondering, why is `isHrRunning` not considered true if the State is _either_ `Disabled` or `Stopped`?

tituscmd:
> > Does this fix it for you? @tituscmd
> 
> Will check it out and report back ASAP!

I have updated my local branch to include this fix and will test drive it for the current week. I'll report back at the end of the week, or whenever I notice something is off. @mark9064 

In the meantime, I can confirm @Andersama's rewrite of GenRealTwiddle fixes the same constexpr build error I was hitting on current GCC. Rebuilt clean after applying it. Might be worth noting for the future of this PR

mark9064:
That will be solved soon as the 10.3 compiler isn't even available anymore :D

tituscmd:
> That will be solved soon as the 10.3 compiler isn't even available anymore :D

Hahaha alright, fine by me! :D

src/displayapp/screens/HeartRate.cpp
mark9064: In the stopped state, HR is enabled, but it's not running as it's waiting for a background measurement

The stopped state is never shown to the user as the screen is always off i.e. right now it makes no difference. But it's a good point, sounds like this needs reworking

tituscmd:
> I'll report back at the end of the week, or whenever I notice something is off. @mark9064

It's now the beginning of another week, actually, so I'm here to happily report the bug has been completely fixed! Great job!!

mark9064:
Glad to hear :)

So remaining there's still the harmonic reading problem you have (where it reads 2x the real HR) and also there's the states which look a bit dubious. Anything else?

PS the PPG debug branch is also updated if you have a chance to collect data

tituscmd:
> So remaining there's still the harmonic reading problem you have (where it reads 2x the real HR) and also there's the states which look a bit dubious. Anything else?

I also tend to have the problem where I get half the real HR. I don't really know if it is precisely half the HR or just generally something in the 30s - 40s.

> PS the PPG debug branch is also updated if you have a chance to collect data

What exactly does this mean?

mark9064:
https://github.com/InfiniTimeOrg/InfiniTime/pull/2371#issuecomment-3711642423

As per this: just saying the branch is up to date if you want to collect debug data

Yeah low HRs are a big problem, it's really difficult to discern them from noise. I think the current solution is as close to a fair tradeoff as possible (without a better sensor)

Reading the subharmonic (i.e. half the true heart rate) is possible as well, but HRs <50 are more likely to be random motion noise than a multiple of the true HR

Andersama:
@mark9064 honestly I don't know that it's so much the sensor at issue as the package design. The instinct I bought has two sets of leds and four surrounding sensors. I think they've run into data reliability issues and compensate with multiple sensors as a result. That at least became my suspicion when the orientation of my arm was causing issues, but I haven't comes back around to this in a while, so if I test these patches I'll let you know how they do.

Andersama:
@mark9064 there's a relatively recent research paper "A Real-Time PPG Peak Detection Method for Accurate Determination of Heart Rate during Sinus Rhythm and Cardiac Arrhythmia" which describes how they go about analyzing raw PPG data. It looks like they're using a samsung band in testing and a 30 second buffer of data.

mark9064:
I think we're on the same page - when I say sensor I mean the hardware. The sensor is poor - only single channel, low quality LDC, limited LED drive control and gain options etc.

But yes you're right, having multiple ideally distributed sensors is even better - their outputs can be correlated for noise rejection

I had a look, the paper you mention is unfortunately not super useful since it deals with motion artefacts by just giving up - I think that's probably a good approach for medical settings, but not so suitable for daily life heart rate measurements. 

If you're interested, I would recommend the following papers. These are some of the ones I read/built on while developing this approach

```
Alessio Burrello et al. “Q-PPG: Energy-Efficient PPG-Based Heart Rate
Monitoring on Wearable Devices”. In: IEEE Transactions on Biomedical
Circuits and Systems 15.6 (Dec. 2021), pp. 1196–1209. issn: 1940-9990.
doi: 10.1109/tbcas.2021.3122017. url: http://dx.doi.org/10.
1109/TBCAS.2021.3122017.

Nishan Chettri et al. “Advances in PPG Sensors Data Acquisition With
Light-to-Digital Converters: A Review”. In: IEEE Sensors Journal 24.16
(Aug. 2024), pp. 25261–25274. issn: 2379-9153. doi: 10.1109/jsen.2024.
3420170. url: http://dx.doi.org/10.1109/JSEN.2024.3420170.
3420171. 

Nicholas Huang and Nandakumar Selvaraj. “Robust PPG-based Ambu-
latory Heart Rate Tracking Algorithm”. In: 2020 42nd Annual Interna-
tional Conference of the IEEE Engineering in Medicine and Biology Soci-
ety (EMBC). IEEE, July 2020, pp. 5929–5934. doi: 10.1109/embc44109.
2020.9175346. url: http://dx.doi.org/10.1109/EMBC44109.2020.
9175346.

Manh Pham Hung et al. Pushing the Limit of PPG Sensing in Seden-
tary Conditions by Addressing Poor Skin-sensor Contact. 2025. doi: 10.
48550/ARXIV.2504.02735. url: https://arxiv.org/abs/2504.02735.

Christodoulos Kechris et al. “KID-PPG: Knowledge Informed Deep Learn-
ing for Extracting Heart Rate From a Smartwatch”. In: IEEE Trans-
actions on Biomedical Engineering 72.3 (Mar. 2025), pp. 870–877. issn:
1558-2531. doi: 10.1109/tbme.2024.3477275. url: http://dx.doi.
org/10.1109/tbme.2024.3477275.

Hooseok Lee et al. “Adaptive scheduling of acceleration and gyroscope
for motion artifact cancelation in photoplethysmography”. In: Computer
Methods and Programs in Biomedicine 226 (Nov. 2022), p. 107126. issn:
0169-2607. doi: 10.1016/j.cmpb.2022.107126. url: http://dx.doi.
org/10.1016/j.cmpb.2022.107126.

Attila Reiss et al. “Deep PPG: Large-Scale Heart Rate Estimation with
Convolutional Neural Networks”. In: Sensors 19.14 (July 2019), p. 3079.
issn: 1424-8220. doi: 10.3390/s19143079. url: http://dx.doi.org/
10.3390/s19143079.

Seyed Salehizadeh et al. “A Novel Time-Varying Spectral Filtering Algo-
rithm for Reconstruction of Motion Artifact Corrupted Heart Rate Sig-
nals During Intense Physical Activities Using a Wearable Photoplethys-
mogram Sensor”. In: Sensors 16.1 (Dec. 2015), p. 10. issn: 1424-8220. doi:
10.3390/s16010010. url: http://dx.doi.org/10.3390/s16010010.

Tim Schack, Michael Muma, and Abdelhak M. Zoubir. “Computationally
efficient heart rate estimation during physical exercise using photoplethys-
mographic signals”. In: 2017 25th European Signal Processing Conference
(EUSIPCO). IEEE, Aug. 2017, pp. 2478–2481. doi: 10.23919/eusipco.
2017.8081656. url: http://dx.doi.org/10.23919/EUSIPCO.2017.
8081656.

Philip Schmidt et al. “Introducing WESAD, a Multimodal Dataset for
Wearable Stress and Affect Detection”. In: Proceedings of the 20th ACM
International Conference on Multimodal Interaction. ICMI ’18. ACM, Oct.
2018, pp. 400–408. doi: 10.1145/3242969.3242985. url: http://dx.
doi.org/10.1145/3242969.3242985.

Menglian Zhou and Nandakumar Selvaraj. “Heart Rate Monitoring us-
ing Sparse Spectral Curve Tracing”. In: 2020 42nd Annual International
Conference of the IEEE Engineering in Medicine and Biology Society
(EMBC). IEEE, July 2020, pp. 5347–5352. doi: 10 . 1109 / embc44109 .
2020.9175349. url: http://dx.doi.org/10.1109/EMBC44109.2020.
9175349.
```

mark9064:
CI is green :)

No workarounds are needed to build this branch anymore

PR #2372, opened 2025-11-14
Update to GCC 15

### LINKED ISSUE #2465: GCC version in docs is no longer available
The link to the gcc download from the docs, https://developer.arm.com/downloads/-/gnu-rm, now points to https://gitlab.arm.com/tooling/gnu-toolchains-for-arm, which doesn't have any versions below 11.2-2022.02 and the legacy downloads page only contains releases up to 2019 (see https://developer.arm.com/tools-and-software/gnu-toolchain#Downloads ). 10.3-2021.10 doesn't seem to be available anywhere. I also think the download URLs in the Docker build scripts may no longer work, but I may also be checking wrong

mark9064:
Rebase of #1512

~~The linker warnings still exist~~

### DIFF
--- a/doc/buildAndProgram.md
+++ b/doc/buildAndProgram.md
@@ -4,7 +4,7 @@
 
 To build this project, you'll need:
 
-- A cross-compiler : [ARM-GCC (10.3-2021.10)](https://developer.arm.com/downloads/-/gnu-rm)
+- A cross-compiler : [ARM-GCC 15.3-Rel1 (arm-none-eabi)](https://gitlab.arm.com/tooling/gnu-toolchains-for-arm/-/tree/releases/15.3.rel1)
 - The NRF52 SDK 15.3.0 : [nRF-SDK v15.3.0](https://nsscprodmedia.blob.core.windows.net/prod/software-and-other-downloads/sdks/nrf5/binaries/nrf5sdk153059ac345.zip)
 - The Python 3 modules `cbor`, `intelhex`, `click` and `cryptography` modules for the `mcuboot` tool (see [requirements.txt](../tools/mcuboot/requirements.txt))
   - To keep the system clean, you can install python modules into a python virtual environment (`venv`)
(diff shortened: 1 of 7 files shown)


mark9064:
The size difference definitely isn't zero though, I guess the CI job isn't built to handle toolchain changes

mark9064:
14.2 (I haven't downloaded 14.3 yet but should be very similar)
```
           FLASH:      379616 B     474632 B     79.98%
     SPARE_SPACE:           0 B        12 KB      0.00%
             RAM:       23868 B        64 KB     36.42%
```
10.3
```
           FLASH:      382428 B     474632 B     80.57%
             RAM:       23576 B        64 KB     35.97%
```

mark9064:
Linker warnings now resolved by implementing stubs
For this branch:
```
10.3

           FLASH:      385104 B     474632 B     81.14%
             RAM:       23584 B        64 KB     35.99%
Build time: 13 seconds

14.2

           FLASH:      382248 B     474632 B     80.54%
     SPARE_SPACE:           0 B        12 KB      0.00%
             RAM:       23876 B        64 KB     36.43%
Build time: 18 seconds

15.2

           FLASH:      379940 B     474632 B     80.05%
     SPARE_SPACE:           0 B        12 KB      0.00%
             RAM:       23876 B        64 KB     36.43%
             
Build time: 17 seconds
```

So the firmware is continuing to get smaller :)
Unfortunately there's a slight bump in memory usage somewhere in the last 5 versions, but I think we have to accept it.

src/stdlib.c
```diff
@@ -49,3 +51,58 @@ void* realloc(void* ptr, size_t newSize) {
 void* __wrap_realloc(void* ptr, size_t newSize) {
   return realloc(ptr, newSize);
 }
+
+// Implement functions required by libc as stubs
+// These functions aren't linked into the final binary
+
+__attribute__((error("stub"))) void _close(int fp) {
```
mark9064: The attributes might be a no-op as these are technically definitions rather than declarations. But they can't hurt I guess... the docs aren't super clear

src/CMakeLists.txt
```diff
@@ -1118,7 +1118,7 @@ add_dependencies(${EXECUTABLE_MCUBOOT_RECOVERYLOADER_NAME} ${EXECUTABLE_RECOVERY
 
 set_target_properties(${EXECUTABLE_MCUBOOT_RECOVERYLOADER_NAME} PROPERTIES
         SUFFIX ".out"
-        LINK_FLAGS "-mthumb -mabi=aapcs -std=gnu++98 -std=c99 -L ${NRF5_SDK_PATH}/modules/nrfx/mdk -T${NRF5_LINKER_SCRIPT_MCUBOOT} --specs=nosys.specs -mcpu=cortex-m4 -mfloat-abi=hard -mfpu=fpv4-sp-d16 -Wl,--gc-sections -Wl,--print-memory-usage --specs=nano.specs -Wl,-Map=${EXECUTABLE_MCUBOOT_RECOVERYLOADER_FILE_NAME}.map"
```
NeroBurner, line 1121: just so I understand correctly. To get down size we exluded the stlib with the nosys.specs linker flag.

Why can we remove it now? I'd like to learn more

mark9064: libnosys provides stub implementations for system calls like read, write etc.

Since GCC11 these stubs emit warnings at the end of compilation which are noisy and annoying

From my research it seems like the best solution is to reimplement those calls as stubs ourselves, as we're sure that they're not being used. So since we're implementing those calls ourselves, we don't need libnosys anymore

mark9064:
I think we need to prioritise this based on #2465

doc/buildAndProgram.md
```diff
@@ -4,7 +4,7 @@
 
 To build this project, you'll need:
 
-- A cross-compiler : [ARM-GCC (10.3-2021.10)](https://developer.arm.com/downloads/-/gnu-rm)
+- A cross-compiler : [ARM-GCC 15.3-Rel1 (arm-none-eabi)](hhttps://gitlab.arm.com/tooling/gnu-toolchains-for-arm/-/tree/releases/15.3.rel1)
```
NeroBurner: hhttps. little typo in link

NeroBurner (changes requested):
little typo in link. otherwise looks sane.

didn't test locally but it runs in CI and you tested it, so I believe it works!

doc/buildAndProgram.md
mark9064: Thanks, fixed :)

PR #2375, opened 2025-11-18
Show blue icon when connected, grey when disconnected and hide when disabled

vkareh:
This is a rebased version of #1974 since there are longstanding conflicts in the branch. Attribution in the commit remains to the original author.

> This PR changes behaviour of bluetooth icon display.
> 
> Previous behaviour was that the icon showed coloured white when enabled and connected.
> 
> New behaviou is that the icon shows blue when connected, grey when disconnected and is hidden when disabled. This provides more info in the same space and is consistent with many common implementations which may offer a more familiar workflow.


minacode (commented):
I think it would be cleaner to use the color from src/displayapp/InfiniTimeTheme.h.

Issue #2369, opened 2025-11-10
Disable heart rate tracking while charging


Andersama:
### Verification

- [x] I searched for similar issues (including closed issues) and found none was relevant.

### Introduce the issue

The watch faces appear to be able to indicate when the watch is charging, use that information to disable background and foreground heart rate tracking.

Andersama:
I'm not 100% clear on the state machine as of yet, but I'm imagining something like:
```c++
    case States::BackgroundMeasuring:
    case States::ForegroundMeasuring:
      if (batteryController.IsCharging()) {
        return portMAX_DELAY;
      }
     return CalculateSleepTicks();
```
here:
https://github.com/InfiniTimeOrg/InfiniTime/blob/99ae2f368bd3f9ca5521d2d04a426addbb00b04b/src/heartratetask/HeartRateTask.cpp#L68

Might be what we're after.

Andersama:
Alternatively modifying `HeartRateTask::Work()` would probably make more sense.

Andersama:
May also make sense to modify the condition inside of `HandleSensorData()`:

```c++
if (ambient > 0 || batteryController.IsCharging()) {
```

Andersama:
On second look, probably should use `IsPowerPresent()`

Andersama:
I've made the changes above, but I'm clearly missing something, the sensor remains active.

Saarsk:
Can this please be optional? 🥺 It would render the watch temporarily inoperable when charging it on the arm (without removing it from the wrist) and create gaps in health statistics. Then there's notifications, alarms, timers etc.

marigoldfish:
How on earth do you charge the watch on your wrist? Wouldn't you be blocking the heart rate sensor with the charger anyway? 🤔

Saarsk:
That's only if you use the original unmodified charging puck. 😊 It's just some pogo-pins (or anything conductive, really) that need to make contact.

The watch will be fitted with USB-C when I get the time (I'm between houses for the time being). I'm USB-C-ing everything. No more dongle life.

Everyone has a battery bank in their pocket at all times that that can charge other devices. A battery bank will also sit surprisingly comfortable in my elastic watch strap enabling a 100 % untethered arm operation. 😁 There are also these tiny USB-C batttery banks that could attach right into a USB-C on the side of the watch when that time comes.

marigoldfish:
[...]

PR #2376, opened 2025-11-18
[WIP] Disable heart rate tracking while power is connected (charging)

Andersama:
This does not appear to actually disable heart rate monitoring. I think it may be worth adding new message states to communicate to the heart rate task via `SystemTask.cpp`.

Conceptually, the idea here was to force the task into a waiting state if the battery controller indicated power was present.

A similar idea would apply to the step tracker. I did not add a boolean flag to control whether this feature was on or off, but conceptually if someone built an external wearable battery or charger then this would be incomplete.

See #2369

Edit: Do not use as is in [0e67cff](https://github.com/InfiniTimeOrg/InfiniTime/commit/0e67cff623d0ed43d7b89b0261a96341fc0ecc8a), this introduces a crash when the heart rate setting is turned on in the heart rate app.

### DIFF
--- a/src/heartratetask/HeartRateTask.cpp
+++ b/src/heartratetask/HeartRateTask.cpp
@@ -1,6 +1,7 @@
 #include "heartratetask/HeartRateTask.h"
 #include <drivers/Hrs3300.h>
 #include <components/heartrate/HeartRateController.h>
+#include <components/battery/BatteryController.h>
 #include <limits>
 
 using namespace Pinetime::Applications;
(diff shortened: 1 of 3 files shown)


Andersama:
The best guess I can make for the crash, which I'll test I guess tomorrow is that the slight change to the state machine now allows going from the foreground measuring state (enabled) directly to the waiting state. There is a comment about an assumption being made in the switch statement for handling events. I'll test a modification to the state machine to see if this works smoothly if background measuring. Hopefully there's another approach because otherwise this will mean the heart rate checks will continue while the watch face is on.

Update: My change to keep the modifications to the new state to the background tracking states did fix the crash. However now the watch appears to rewake itself, staying in the foreground state, never allowing the tracker to disable.

Issue #2364, opened 2025-11-07
Keep heart tracker enabled after reboot


Andersama:
### Verification

- [x] I searched for similar issues (including closed issues) and found none was relevant.

### Introduce the issue

I'm assuming there's some form of persistence that can be enabled after reboot, since a number of settings for the watch stick around. Consider adding a toggle to the heart tracker app to keep the setting after reboot, make it enabled by default.

PR #2377, opened 2025-11-22
Keep heart rate tracking setting after reboot

Andersama:
When the watch boots check the user's last known setting.

See: #2364


Andersama:
I've tested this, it appears to work, no issues so far.

Update: A small issue, the settings aren't saved when the button is enabled / disabled (which is vaguely what I hoped would happen), they're saved whenever the system decides to write the settings. The hacky approach I've found is to toggle any of the options in the settings area.

PR #2379, opened 2025-11-25
flashlight: Add multiple display modes

vkareh:
Adds various modes to the flashlight app:
- Off
- Red (night mode)
- White (normal)
- Strobe (safety/emergency signaling)

Tapping the screen cycles through these modes. This makes it behave more like standard headlamps or safety hiking/cycling lights, with the first mode always red to not affect your pupils in the dark.

InfiniSim_2025-11-25_170421

Note: The strobe looks strange in the animation because of the InfiniSim refresh rate. It looks better on the real watch, since it turns the backlight on and off directly.

Also the strobe is 2Hz at 20% duty cycle (100ms on, 400ms off), which provides good visibility from a distance and is low enough that it won't trigger people with photosensitive epilepsy.

### DIFF
--- a/src/displayapp/screens/FlashLight.cpp
+++ b/src/displayapp/screens/FlashLight.cpp
@@ -18,6 +18,7 @@ FlashLight::FlashLight(System::SystemTask& systemTask, Controllers::BrightnessCo
   : wakeLock(systemTask), brightnessController {brightnessController} {
 
   previousBrightnessLevel = brightnessController.Level();
+  currentMode = Mode::Off;
   brightnessController.Set(Controllers::BrightnessController::Levels::Low);
 
   flashLight = lv_label_create(lv_scr_act(), nullptr);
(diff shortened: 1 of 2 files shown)


Andersama:
What would you think about adding an option for an SOS indicator in addition to the strobe? Shouldn't be much different from the strobe option, just an additional amount of memory for a small state machine.

vkareh:
@Andersama 
> What would you think about adding an option for an SOS indicator in addition to the strobe? Shouldn't be much different from the strobe option, just an additional amount of memory for a small state machine.

To be honest, I don't think the display refresh is fast enough for this. At best you'd get something that looks like a random pattern blinker... I played with different frequencies, duty cycles, and patterns, and none of them look like what you'd expect from a simple hardware-controlled LED...

vkareh:
:facepalm: doh! Of course! As soon as I said "simple hardware-controlled LED", it dawned on me that redrawing the screen between black and white is kind of pointless if I can just set the screen as white and just turn it on-and-off at the hardware level. So I did that and the strobe is significantly nicer now.

Also I was able to get a somewhat reasonable SOS signal:

https://github.com/user-attachments/assets/d274dcb1-e65c-4697-b4d6-4cf684317a1d

(but I haven't pushed that change yet - maybe if more folks think it's a good idea I can add it?)

TheScienceotter:
A red light would be a really nice addition to the torch, but I think it would be better if the light type was toggled through a separate button in the UI. Firstly so we can keep the convenience of a single tap to turn the torch on or off, but also because if I'm trying to keep my night vision, I don't want to have to go through the main light and strobe light to turn the torch off (especially if there are more types of light/flash sequences also to be added). 

Maybe a button in the lower left to cycle through light colour and then a button in the bottom right to cycle through strobe patterns?

vkareh:
@TheScienceotter to turn the torch off without going through the white light I just press the hardware button...

Andersama:
@vkareh sorry for the delay, that's cool!

PR #2389, opened 2025-12-14
Dynamic Themes

ScanuNicco:
## Overview
This pull request adds basic support for loading a theme stored on persistent storage. On boot, InfiniTime will look for a file located at `/themes/theme.cfg` and load the colors from the file. A new setting has been added to allow reloading the theme file without requiring a reboot.

## Theme file format
The theme file itself is a custom key-value style configuration file. The goal is to be easily editable, but simple to parse to avoid taking a dependency on a potentially large JSON/YAML parsing library. The following theme was used to create the attached screenshots:
```
# InfiniTime Red by ScanuRag
accent_light=880000
accent=440000
accent_dark=281818
highlight=00B000
text_primary=BBBBBB
text_header=FF0000
page_bg=111111
icon=FF0000
```

## Process for applying a theme
In order to apply a theme, the user needs to write or find a theme file and move it to the PineTime's persistent storage using a compainion app. Not all apps support moving arbitary files to persisent storage, but it is currently possible using ITD with a physical device, or on InfiniSim using the bundled `littlefs-do` tool. One the theme has been sent to the device, the user needs to navigate to Settings>Theme>Reload Theme to apply it.

## Notes
- The page background is applied in DisplayApp. I am not sure if that is an acceptable approach, or if I should modify all apps to apply the background. Previously, I found that some (but not all) apps manually apply the background leading to inconsistency.
- Themes in general do not apply to watch faces, as they have their own theming
- This has been tested on a PineTime, as well as in the Simulator
- The simulator provides its own version of LittleVgl.cpp, which is modified by this PR. To test with the simulator, it's version of that file needs the same modifications as the one in this PR.

## Pictures
InfiniSim_2025-12-14_173615
InfiniSim_2025-12-14_173606
InfiniSim_2025-12-14_173600
InfiniSim_2025-12-14_173543
InfiniSim_2025-12-14_173541
InfiniSim_2025-12-14_173537
InfiniSim_2025-12-14_173729
InfiniSim_2025-12-14_173846


ScanuNicco:
I have been running this for more than a month now and have not run into any instability. I'm going to mark it as ready for review.

PR #2390, opened 2025-12-15
Fix negative temperature rounding

mark9064:
Suggested implementation from #2194 

@JF002

(to whoever merging: please rebase merge to keep authorship)

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -53,10 +53,13 @@
 #include "displayapp/screens/settings/SettingBluetooth.h"
 #include "displayapp/screens/settings/SettingOTA.h"
 
+#include "utility/Math.h"
+
 #include "libs/lv_conf.h"
 #include "UserApps.h"
 
 #include <algorithm>
+#include <limits>
 
 using namespace Pinetime::Applications;
 using namespace Pinetime::Applications::Display;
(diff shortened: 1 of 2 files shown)


FintasticMan:
This looks good! Sorry for not being very involved recently. I do suggest using an implementation that rounds away from zero as opposed to one that rounds towards zero, as that is more commonly used and is more intuitive.

Also, are the static casts necessary for the integer literals? Can't the compiler figure the type out without it? If they are necessary it might be good to have the function take 2 type parameters to avoid that (that might make the overflow cases harder to reason about though, haven't thought it through much).

mark9064:
No worries on being busy :smile: really appreciate the review

Good point on the rounding mode, I've switched it to round half away from zero :+1: 

The static casts are unfortunately necessary. If we allow two different types we get the really annoying integer promotion rules we ran into last time, so I think it's best to stick with something stricter where we can be confident the behaviour is correct. I agree that the required casts are annoying though :/ just not sure there's a good alternative

FintasticMan:
> If we allow two different types we get the really annoying integer promotion rules we ran into last time

In theory we should only run into those issues if either of the types is unsigned, which can be documented? Not sure if that's a good solution though haha.

mark9064:
Yeah I'd prefer to avoid any scenario where it's easy to get bad behaviour without realising (i.e. relying on documentation). I think a couple of extra casts are preferable to an incorrect output

FintasticMan:
We could add something like this: https://stackoverflow.com/questions/36406333/fixed-width-integer-literals-in-c.

Then we don't have to use static casts directly, and we can just do `100_i16` for example.

mark9064:
A rust style literal suffix would be really nice actually from a usability standpoint. Why isn't this part of the language itself yet...

I'm a bit split if it's a good idea for the project though. Contributors probably won't expect a language extension like that and I don't like having two (well three if you count C style casts) ways to do the same thing. Maybe I'm overthinking this?

src/utility/Math.h
```diff
@@ -1,10 +1,33 @@
...
+    constexpr T RoundedDiv(T dividend, T divisor) {
+      bool neg = divisor < 0;
+      if (neg) {
+        // overflows if divisor is minimum value for T
+        divisor = -divisor;
+      }
+
+      T m = dividend % divisor;
+      T h = divisor / 2 + divisor % 2;
+      T res = (dividend / divisor) + (!(dividend < 0) & (m >= h)) - ((dividend < 0) & ((m + h) <= 0));
```
NeroBurner, line 25: I may be missing something, but m and h are purely positive or 0. so the m+h should only ve true if both are 0?

so maybe I dont understand the negative divident case 😅

mark9064: `m` can be negative. For example `RoundedDiv(-23, 17)` would make `m = -23 % 17 = -6`

Issue #2132, opened 2024-10-02
Swiping to watchface is slower on Casio watchface


tituscmd:
### Verification

- [X] I searched for similar issues (including closed issues) and found none was relevant.

### Introduce the issue

Swiping from the app launcher, the quick settings or the notification screen to the watchface is slower than swiping e.g. through the pages of the app launcher.

### Preferred solution

The swiping duration should be consistent.

### Version

d6e9361a

mark9064:
Casio watchface? It takes time to load the fonts. Try Digital, see if speed is consistent

tituscmd:
Yeah, works normally on Digital. Seems you're right.

tituscmd:
Updated the title accordingly.

mark9064:
In progress, but being discussed over in #2197 ^^

tituscmd:
> In progress, but being discussed over in [#2197](https://github.com/InfiniTimeOrg/InfiniTime/issues/2197) ^^

Whoops, my bad!

mark9064:
Should be slightly improved by #2330. You might need to wipe your flash with lfs_format. Would be interested to see if you notice any changes

PR #2392, opened 2025-12-16
Fast fonts

### LINKED ISSUE #2197: Copy in-use resources from SPI flash to built-in flash
Since reading files from the SPI flash is very slow we could copy the files that the current watchface needs into the internal flash, then use them from there instead of the flash. This could be implemented as an [LVGL file system](https://docs.lvgl.io/7.11/overview/file-system.html), although I don't think we need to use LittleFS and could instead use a trivial flat file system structure.

For example, when the casio watchface is selected the three font files it uses will be copied into the internal flash, and when the watchface is shown it will load the files from there much quicker.

Alternatively, we could implement a sort of caching LVGL file system that sits in between the LittleFS file system and the LVGL adapter that caches some files in the internal flash, but that might be trickier due to fragmentation.

mark9064:
Requires #2372, #2391

- Fonts load much faster. With #2330, the G7710 watchface should load in about 30ms versus the current 250-400ms (10x speedup)
  - Basically it is no longer noticeable whether fonts are on the external flash or not
- Fonts use less memory when loaded as they are in one large allocation
- Saves ~2K flash size (LVGL font loader eliminated)
- Fonts changed to use smart pointers so they deallocate themselves

Closes #2197, #2132

@tituscmd you might be interested in trying this :)

This works by compiling the fonts separately at compile time. The data sections (i.e. the structs) are pulled out and serialised to a file, and all the internal pointers are set to their offsets within the file. This makes loading the font extremely easy: allocate a block of memory the size of the file, copy the entire file to that memory, and then fix the pointers inside by adding the address of the memory to all of them. This allows fonts to be loaded with minimal overhead.

### DIFF
--- a/doc/ExternalResources.md
+++ b/doc/ExternalResources.md
@@ -53,18 +53,17 @@ lv_obj_t* logo = lv_img_create(lv_scr_act(), nullptr);
 lv_img_set_src(logo, "F:/images/logo.bin");
 ```
 
-Load a font from the external resources: you first need to check that the file actually exists. LVGL will crash when trying to open a font that doesn't exist.
+Load a font from the external resources. 
+
+You don't need to check if it exists in the constructor as LVGL can handle the font not existing gracefully (nothing will render).
+However, do check that it exists in the `IsAvailable()` method so users can see when resources are missing (the watchface will not be selectable in the list if `IsAvailable()` returns `false`).
+Remember to free any loaded fonts (with `free()`) in the screen destructor.
 
 ```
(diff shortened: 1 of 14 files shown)


tituscmd:
Very nice! I am indeed interested in this :) 

I'll put this PR on the list of stuff I wanna tackle once I've got the free time again.
When that time comes, I'll report whatever I can!

mark9064:
Looks like CI won't work as it depends on the updated container. I promise it builds :P

(The build requires pyelftools now, but the current build container doesn't have it)

PR #2395, opened 2025-12-27
Add pawn scripting language

pipe01:
This PR adds support for using the [Pawn](https://www.compuphase.com/pawn/pawn.htm) language to write apps and watchfaces and load them at runtime. It's only a proof of concept right now, currently two apps use it:

  * The digital watchface has been [rewritten in Pawn](https://github.com/pipe01/InfiniTime/blob/pawn/src/resources/apps/watchface_digital.p), its code is loaded from program ROM
  * The app with an "L" icon displays a list of all apps in the external flash and lets the user run them

Simple hello world example:

```pawn
#include "infinitime"

@start() {
	var lv_obj: label = lv_label_create(screen)
	lv_obj_set_pos(label, 20, 20)
	lv_label_set_text(label, "Hello world")
}
```

Note: there's currently some bug in the web InfiniEmu emulator and the watchface fails to load initially, just swipe up and down to reload it

### DIFF
--- a/src/pawn/amx.c
+++ b/src/pawn/amx.c
@@ -17,6 +17,10 @@
  *  Version: $Id: amx.c 7487 2025-08-21 18:37:33Z thiadmer $
  */
 
+#undef __LINUX__
+#undef __linux
+#undef __linux__
+
 #define WIN32_LEAN_AND_MEAN
 #if defined _UNICODE || defined __UNICODE__ || defined UNICODE
 # if !defined UNICODE   /* for Windows API */
--- a/src/pawn/amx.h
+++ b/src/pawn/amx.h
@@ -23,6 +23,10 @@
 #include <stdlib.h>   /* for size_t */
 #include <limits.h>
 
+#undef __LINUX__
+#undef __linux
+#undef __linux__
+
 #if (defined __linux || defined __linux__) && !defined __LINUX__
   #define __LINUX__
 #endif


mark9064:
This looks super cool. Question: could the example program here live in ROM rather than static memory (I assume that's where the +600 is coming from)? Otherwise if all programs live in RAM all the time I think it'll be really tough memory wise

pipe01:
Yeah, we can even dynamically load the program from the external flash in small chunks to the RAM

mark9064:
Cool, so if we made that char array `static constexpr` it'd still work fine? In that case the static memory usage added by this is only a few bytes which sounds very promising (and the program size looks reasonable). As I investigated in #2392, loading a few Ks from flash can be done in \<40ms or so having apps/watchfaces loaded this way sounds very plausible. I guess my one concern is larger apps might have enough bytecode that we can't load it all into memory at once. But it sounds like from what you say we can load in chunks, so that might not be an issue

@JF002 would be very interesting to hear your thoughts on this

pipe01:
I've implemented the chunk loading I mentioned, now the code will only use up to 512 bytes (configurable, should probably be closer to a couple KBs). Obviously right now it's useless since it's loading from program ROM, but when loading from external flash it will also act as an LRU cache

mark9064:
Awesome :)

How does calling native functions e.g. LVGL functions work? Can we autogenerate bindings or do we need to implement our own micro-API for apps/watchfaces. I guess I'm not opposed to a micro-API anyway as we can make it more user friendly

pipe01:
I was thinking of a mix of autogenerating bindings for LVGL and some auxiliary functions on top, e.g. the [handful](https://github.com/pipe01/InfiniTime/blob/8cdc948488f8c643ae34cf450ddf18baa6b5197d/src/displayapp/screens/Pawn.cpp#L393) currently provided. Some sort of high level API would definitely make the programs smaller although I don't know what such an API would look like; both options are definitely possible

pipe01:
I've reimplemented the digital watchface in Pawn [here](https://github.com/pipe01/InfiniTime/blob/56fa587634f7ab90ff40516c43757e8869686820/src/pawn/programs/watchface_digital.p) and removed the original one. I'm not sure if we should actually do this or not but it's a good exercise, although if we do we could consider moving it to the external flash to free up internal flash space

pipe01:
Loading programs from flash works, and with it comes hot reloading:

[pawn reload.webm](https://github.com/user-attachments/assets/5ee940b2-41ee-4774-858f-7331f73f22ea)

PR #2396, opened 2025-12-29
Fix DateTime::SetTime logging before initialized

pipe01:
The `SetTime` method is indirectly called by the global variable constructor in [`main.cpp`](https://github.com/InfiniTimeOrg/InfiniTime/blob/f2814dd5ed26b1b6478d781048c0df27f03ca1e3/src/main.cpp#L104), which runs before the [logger is initialized](https://github.com/InfiniTimeOrg/InfiniTime/blob/f2814dd5ed26b1b6478d781048c0df27f03ca1e3/src/main.cpp#L318) and thus crashes when trying to log a message. This doesn't affect release builds since logging is disabled there, but it does affect debug builds

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -357,7 +357,7 @@ int main() {
     dateTimeController.SetCurrentTime(NoInit_BackUpTime);
   } else {
     // __DATE__ is a string of the format "MMM DD YYYY", so an offset of 7 gives the start of the year
-    dateTimeController.SetTime(Pinetime::Utility::compileTimeAtoi(&__DATE__[7]), 1, 1, 0, 0, 0);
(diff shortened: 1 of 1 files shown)


mark9064:
Hmm, I don't really like adding an extra argument to solve this. Could we defer the set time call to after logging is initialised?

pipe01:
You're right, I think it's better this way

src/utility/Math.h
```diff
@@ -29,5 +29,14 @@ namespace Pinetime {
       }
       return res;
     }
+
+    constexpr int compileTimeAtoi(const char* str) {
```
mark9064:
```suggestion
    constexpr int CompileTimeAtoi(const char* str) {
```
May as well fix the name while here :)

PR #2399, opened 2026-01-06
Antialias smaller system fonts

mark9064:
This adds 2 bits per pixel (BPP) antialiasing for smaller system fonts, following the discussion in #2315 

All of the icons look a lot nicer IMO, and text generally looks better too


batje:
Wow, what a difference. Really nice.

kieranc:
LGTM!

RedBigz:
Hi! I don't mean to cause any thread hijacking here, but would the implementation of a cleaner font family be desired for InfiniTime? I found that using IBM Plex Mono instead of JetBrains Mono made text easier to read on the PT's small display.

(Left: JetBrains Mono 1bpp, Right: IBM Plex Mono 4bpp, however it still looks great at 2bpp)
image

kieranc:
> Hi! I don't mean to cause any thread hijacking here, but would the implementation of a cleaner font family be desired for InfiniTime? I found that using IBM Plex Mono instead of JetBrains Mono made text easier to read on the PT's small display.

Maybe? You could open a discussion or PR, get some input? I think a fairer comparison for your images would be with both at the same BPP, FWIW.

PR #2400, opened 2026-01-06
Persist time and steps across reboots

### LINKED ISSUE #2293: Rebooting a watch resets the number of steps to 0
### Verification

### What happened?

Having some steps counted, then after a reboot of the watches the number resets

Number of steps shall be persistent through reboot

### Reproduction steps

- Start device,
- count some steps
- reboot device by long pressing a button
- steps are 0

### More details?

_No response_

### Version

1.15.0

### Companion app

_No response_

mark9064:
Persist current steps, trip steps and time across reboots with an easily extendable struct for persisting more data in the future.

I've spent a lot of time reading, and I'm pretty sure now that the volatile semantics here are now correct. This means that loading values after reboot should be much more reliable since the compiler is aware that the memory used has special semantics.

Fixes #2293

### DIFF
--- a/src/components/motion/MotionController.cpp
+++ b/src/components/motion/MotionController.cpp
@@ -38,12 +38,15 @@ namespace {
 void MotionController::AdvanceDay() {
   --nbSteps; // Higher index = further in the past
   SetSteps(Days::Today, 0);
+  carrySteps = 0;
   if (service != nullptr) {
     service->OnNewStepCountValue(NbSteps(Days::Today));
   }
 }
 
 void MotionController::Update(int16_t x, int16_t y, int16_t z, uint32_t nbSteps) {
+  // Offset the sensor value by whatever we are carrying forward
+  nbSteps += carrySteps;
   uint32_t oldSteps = NbSteps(Days::Today);
   if (oldSteps != nbSteps && service != nullptr) {
     service->OnNewStepCountValue(nbSteps);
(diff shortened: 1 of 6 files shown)


Andersama:
I'm assuming this has been built and does its thing. The question I suppose I have is this: why not use the filesystem? Is that at risk of not persisting as well?

I was literally in the process of making my own edit yesterday for this reason...but what I've done is taken a struct and spat it out into a file.

Edit: I'll modify my rng pr in relation to this for seeding purposes.

mark9064:
Mainly as these values are updated 10 times per second, all the time, and writing to the flash 10 times per second is best avoided

Andersama:
Makes sense, I may have made the delay slightly more than 10x per second. Suppose a filesystem write is a backup? EG: Update this struct and on the off chance that may fail as it seems it might in the documentation then lookup a file?

mark9064:
I don't think a file backup is needed, it's extra complexity for a case that realistically only happens after a firmware update, which is once per year currently

SteveAmor:
Could a similar approach be taken to maintain the state of the 2's game if you are playing the game and an alarm, notification or chime take you out of the app?

mark9064:
Yes, though in that case the flash probably makes more sense rather than RAM (since memory is limited, and we'd only expect to be writing out the game state on app exit)

#2171 may be of interest

PR #2404, opened 2026-01-10
Remove placeholder LVGL values

mark9064:
Lots of the screens currently use placeholder values for labels etc.

Almost all of them can be eliminated by calling `Refresh()` in the constructor and then correcting any alignment problems.

The main advantage of this is that the screen content is up to date when the screen scrolls into view, rather than scrolling in with placeholder values and then loading the real content.

It also eliminates redundant data which is nice

One other pattern here is that the refresh task creation is moved to the end of the constructor, since it makes sense to start the refresh timer after a successful refresh rather than before it IMO


mark9064:
Relevant discussion and testing in #2257 also

mark9064:
Fixed placeholder in StopWatch. Honestly could do with refactoring more broadly to always render everything on load

PR #2420, opened 2026-02-22
Add quiet hours (scheduled notification silencing)

sofar:
Automatically switches notification status to Sleep at a configurable start hour and back at a configurable end hour. Includes a new settings screen with enable toggle and start/end hour controls. Disabled by default (21:00-09:00). Also adds moon glyph to the default UI font so it displays in the settings menu.

This feature draws on extensive community discussion across several related PRs. The following feedback was reviewed and incorporated:

Adopted:

- "Quiet hours" naming (kieranc, PR #1461): renamed from "sleep setting" / "auto sleep" to avoid confusion with the existing sleep mode.

- Preserve notification state across transitions (Itai-Nelken, PR #1461): quiet hours now saves the user's notification status (On/Off/Sleep) before entering and restores it when exiting, instead of unconditionally forcing On. Uses transient (non-persisted) state following the existing bleRadioEnabled/dfuAndFsEnabledTillReboot pattern.

- Previous state was Sleep edge case (FintasticMan, PR #1461): if the previous state was already Sleep when quiet hours began, it is restored faithfully. FintasticMan suggested restoring to Off instead, but preserving the actual state is more predictable and consistent.

- Alarm overrides quiet hours (FintasticMan, PR #1461): when an alarm fires, quiet hours are exited so the alarm can wake the user. This ensures alarms are never silenced by scheduled quiet hours.

- Disable wrist-lower-to-sleep during sleep mode (kieranc, PR #2415, approved by NeroBurner): wrist-raise wake was already suppressed during sleep mode but wrist-lower-to-sleep was not, which is inconsistent. Moved the lower-wrist check inside the existing != Sleep guard per mark9064's code review suggestion to avoid a duplicate condition check.

- Respect explicit user choices (chmeeedalf, PR #2002): if a user manually changes notification status via QuickSettings during quiet hours, that works normally; the original pre-quiet-hours state is still restored when quiet hours end.

- Chimes suppressed during quiet hours: the existing chime handlers already gate on notificationStatus != Sleep, so setting Sleep during quiet hours suppresses chimes automatically with no additional code.

Not adopted:

- Separate auto-start/auto-stop toggles (Boteium, PR #1461): would let a user manually enter sleep early but still auto-wake. Adds UI complexity for a niche use case; a single toggle is simpler and aligns with the InfiniTime vision of "prefer solid defaults over customisability" (mark9064, PR #2230).

- Sleep Bluetooth checkbox (escoand, PR #1461): BLE control during sleep is a separate security concern that deserves its own feature, not a quiet hours sub-option (mark9064, PR #2230).

- Configurable sleep mode behaviors -- AOD, chimes, notifications, step tracking (JustScott, PR #2230): maintainer mark9064 noted that sleep mode means the user is sleeping, so allowing notifications/chimes/AOD contradicts its purpose. The author agreed this belongs in forks, not mainline.

- Red/dim screen during sleep (minacode/lman0, PR #1261): a larger UX change outside the scope of notification scheduling.

- Vibration priority system (minacode, PR #1328): a proper priority queue (phone > timer > alarm > notification) would be ideal for centralized DND management, but requires a motor controller rework that is a much larger effort.

- 30-minute or 15-minute granularity for quiet hours times (LinuxinaBit, PR #1461; zischknall, PR #2227): hour granularity is sufficient for scheduling sleep/wake times and keeps the UI simple.


sofar:
I've swapped the chime vs. the start/end of the quiet hours around and updated the commit. Previously, on quiet hour start, the chime would fire at the start hour. And at the quiet hour end, the chime would not. This changes that behavior around: At quiet hours start, the chime will not fire, and, at quiet hours end, it will fire. This should be much more intuitive - e.g. "don't disturb me between 10 and 7" means to me and probaby most people "at 9:59 you may still disturb me, and at 7:01 you may as well".

kieranc:
I've been testing this for the last week and I've noticed a couple of times that sleep mode hasn't been disabled in the morning. I can't offer much more debugging information but I do think there's a problem. I don't have an alarm or chimes enabled.

sofar:
I've found my watch twice on "notifications disabled" but attribute it to thick-fingering it. Will pay some more attention to this as I am really hoping to continue using this long-term.

sofar:
I also see this needs a rebase. I'll try and get to it soon.

sofar:
Rebased. I've looked into the "stuck in sleep mode" bug and it seems there's several paths that could lead into it. A reboot, time sync, possibly others.

For this reason I've modified the patch to also save "We're in quiet hours" as a state bit, so that when the watch reboots, it can properly undo or continue it, and doesn't stay "stuck" in that mode.

Will continue testing this :)

src/components/settings/Settings.h
```diff
@@ -351,10 +351,64 @@ namespace Pinetime {
...
+
+      bool IsInQuietHours() const {
+        return settings.inQuietHours;
+      }
+
     private:
       Pinetime::Controllers::FS& fs;
 
-      static constexpr uint32_t settingsVersion = 0x000a;
+      static constexpr uint32_t settingsVersion = 0x000c;
```
FintasticMan: Why skip `0x000b`?

sofar:
I'm now using this patch on 2 devices, and, it's well over 20+ days update on one of them. The other has some additional patches and is at 10+ days of uptime. No issues anymore with lost state or stuck in sleep mode.

Theweasels:
I have been using it for about 4 days and so far so good. One minor issue I noticed is with the Always On display mode. Normally sleep mode will turn off the always on display. However the quiet hours don't turn it off until I wake the watch at least once.

Probably beyond the scope of this feature,  but I do wish it would adjust the brightness as well. I feel like I have to choose between leaving it on high brightness so I can see it easier during the day, or leaving it on low brightness so I don't blind myself if I check the time in the middle of the night. Having it adjust the brightness with my quiet hours would be a godsend.

PR #2421, opened 2026-02-22
Add Railway watch face

sofar:
Swiss/Dutch railway station clock-inspired analog watch face with 12 bold hour notches, thick hour and minute hands (no second hand), and a tap-to-toggle overlay that shows the full Digital watch face. The overlay auto-dismisses after 5 seconds.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -430,6 +430,7 @@ list(APPEND SOURCE_FILES
         displayapp/screens/WatchFacePineTimeStyle.cpp
         displayapp/screens/WatchFaceCasioStyleG7710.cpp
         displayapp/screens/WatchFacePrideFlag.cpp
+        displayapp/screens/WatchFaceRailway.cpp
 
         ##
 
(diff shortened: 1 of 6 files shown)


vkareh:
@sofar can you upload a screenshot?

sofar:
Here's a screencap from infiniemu (as linked by the bot):
image

GaryM0101:
Nice! Sword hands would look give it a better look. Add minute markers to give it a Bauhaus look of the era. Also have an option to invert, most station clocks have a white face.
burk

sofar:
I went for a minimalist style for performance reasons, and aesthetic. The minute hands are extra clock cycles that all need to be computed and drawn. As is, the code is relatively short, and the straight edges avoid a particular complication I encountered when trying to draw a non-flat end.

I don't think inverted is a good fit myself. But if there's any consideration of merging something like this watch face and it would help adoption (and not hurt it due to bloat risk) then I'll entertain it.

sofar:
image

sofar:
Rebased, and did one more change to the face itself:

I found myself at nights unable to see whether it was midnight or something else. For this I've made the 12/3/6/9 hour ticks a little longer towards the middle (exactly like the image posted earlier) which should hopefully help with this.

There's a screenshot of the simulator above this comment that shows how it looks now.

PR #2426, opened 2026-03-15
Configurable Notification Timeouts

Thulium-Drake:
Hi there!

After setting my GF up with a PineTime she complained to me that the timeout for notifications was too short :-)

So I've attempted to make a screen that will allow the user to set the timeout for notifications between 7 seconds (the default), 15 and 30.

I started by copying the existing menu for setting the TimeFormat, took some inspiration on how the ScreenTimeout is handled and wrapped it all together in this PR. I've tested it in InfiniSim (there's also a PR coming to add this setting to `littlefs-do settings`) and it seems stable, but I'm not a programmer by trade, so I'd really love some feedback if this code is sane and sensible.

Here's some screenshots from InfiniSim:

image
image

And the setting is stored in the same way as the ScreenTimeout:

```
$ build/littlefs-do settings
ClockFace: 0 Digital
Chimes: 0 None
PTSColorTime: Teal
PTSColorBar: Teal
PTSColorBG: Black
AppMenu: 0
SettingsMenu: 0
ClockType: H24
NotificationStatus: 0 On
NotificationTimeout: 30000 ms  #<<<<
ScreenTimeOut: 15000 ms
ShakeThreshold: 150
WakeUpModes: 
- SingleTap:  OFF
- DoubleTap:  OFF
- RaiseWrist: OFF
- Shake:      OFF
Brightness: 3 Medium
StepsGoal: 10000
BleRadioEnabled: true
```

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -409,6 +409,7 @@ list(APPEND SOURCE_FILES
         displayapp/screens/settings/Settings.cpp
         displayapp/screens/settings/SettingWatchFace.cpp
         displayapp/screens/settings/SettingTimeFormat.cpp
+        displayapp/screens/settings/SettingNotificationTimeout.cpp
         displayapp/screens/settings/SettingWeatherFormat.cpp
         displayapp/screens/settings/SettingWakeUp.cpp
         displayapp/screens/settings/SettingDisplay.cpp
--- a/src/components/settings/Settings.h
+++ b/src/components/settings/Settings.h
@@ -196,6 +196,17 @@ namespace Pinetime {
         return settings.clockType;
       };
 
+      void SetNotificationTimeout(uint32_t notificationTimeout) {
+        if (notificationTimeout != settings.notificationTimeout) {
+          settingsChanged = true;
+        }
+        settings.notificationTimeout = notificationTimeout;
+      };
+
+      uint32_t GetNotificationTimeout() const {
+        return settings.notificationTimeout;
+      };
+
       void SetWeatherFormat(WeatherFormat weatherFormat) {
         if (weatherFormat != settings.weatherFormat) {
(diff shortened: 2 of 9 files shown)


muesli:
Love this. We might want to think about copying the layout and options of the "Display timeout" settings (with the exception of "Always On"):
settings_display

Thulium-Drake:
I tried, but I couldn't figure out how that menu worked from the code :sweat_smile:

I am running into an interesting issue on my unsealed dev watch. When uploading the firmware, I do see the menu, and I can save my preference. But instead of it popping up for X seconds, I only see a green `i` in the left upper corner.

I already tried deleting `settings.dat` with `itctl fs rm settings.dat` and rebooting the watch, no dice...

But when I flash `1.16.0` again with WatchMate or ITD, it is working again (be it non-configurable :-) )

EDIT: forgot to add, meanwhile this change _does_ work in InfiniSim... 

EDIT2: I was able to fix this by bumping `settingsVersion` and `project(pinetime VERSION ..`
Is the project version also related to it? I was pretty sure I tried it earlier with only the bumped `settingsVersion` :-)

muesli:
Looking at the code, I believe bumping `settingsVersion` should be enough. That being said, I recently had similar issues after I added a new struct member in the middle of the `SettingsData` struct.

mark9064:
Hmm, we really try to avoid adding additional options and I'm not sold that this is something most users might want to change. You can always tap the screen to stop the timer, and I feel that 7 seconds is enough time to get to the watch. What's her use case?

Thulium-Drake:
Hi @mark9064 her usecase is the opposite, 7 seconds isn't enough for her. If she's busy (she works in a shop) and has her hands full, she needs more time to check the notification and decide whether or not she needs to grab her phone (or let it slide).

PR #2430, opened 2026-03-19
notifications: Add phone alarm support with dismiss/snooze

vkareh:
Map HighPrioritizedAlert category to a new PhoneAlarm notification type that shows dismiss and snooze buttons and rings until the user interacts or times out.

Watch sends 2-byte BLE responses back to the phone to dismiss or snooze the alarm.

This allows the user to set alarms on the normal phone clock, and have it ring on the watch:
InfiniSim_2026-03-19_161623

Companion: https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/5913


Eve1374:
Hi, I've been using it for more than a week and it has been working good and is very useful (until gadget bridge auto updated back to its main version ;) )
However I had in mind that the main version of Infinitime was displaying the alarm's text on the notification, but now I can't find it back even when switching back to main; did I dream ? I would find it quite useful to have the alarm's text displayed instead of "phone alarm", but don't have the time to dig into it now.

PR #2431, opened 2026-03-28
Add Home Assistant service and app

pipe01:
This PR adds an app and its corresponding BLE service class. A companion app can send the watch a screen layout with components arranged in a grid, including labels and buttons that will notify the companion app when pressed. When used together with [PinePartner](https://github.com/pipe01/PinePartner) and [HomePlugin.js](https://gist.githubusercontent.com/pipe01/a0f0b148186d135e548a86d488b18317/raw/HomePlugin.js), the app can talk to Home Assistant and call services (e.g. toggle a light), as well as display values from sensors.

TODO:
  - [ ] Sanitize received layout data to not crash if invalid
  - [ ] Add more component types, e.g. switches, sliders


liamcharger:
I'd be happy to add support for this into InfiniLink if this is merged🙂

Seeing the plugin you linked, it looks very simple, but it'd be nice to have a doc or two as well :)

PR #2440, opened 2026-05-13
Adjust display init sequence to fix a rotation issue on the new ST7789P3 display controller

### LINKED ISSUE #2439: PineStore announcement : the current display is EoL, a new one will be used for future PineTime batches
A new display (ST7789P3) will be installed in future PineTime production batches since the current one (ST7789V2) is EoL (End Of Life).

This new chip should be compatible with the previous one, I couldn't spot any meaningful change in the datasheet.

However, PineStore noticed a small issue when the new display is installed : the display is rotated 90° when displaying the pinecone in the bootloader. According to their analysis, it's probably caused by a small issue in the initialization sequence : the command SLEEPOUT is issued too soon. It should be called AFTER all the other registers are initialized, with an additional 120ms delay.

ST7789P3 datasheet : [ST7789P3_SPEC_V0.0_Preliminary-1.pdf](https://github.com/user-attachments/files/27796477/ST7789P3_SPEC_V0.0_Preliminary-1.pdf)

https://github.com/user-attachments/assets/ecef7794-ee61-4ffb-a97c-915ac524ea32

# Analysis

The change looks quite simple : Move `SleepOut();` at the end of the initialization sequence, just before `DisplayOn()`, in [St7789::Init()](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/src/drivers/St7789.cpp#L13).

# Tests and potential impacts

We need to make sure that this change does not produce any unwanted side effect on the current hardware equipped with the ST7789V2 display. The suggested changes make sense, but we need to test them carefully.

And we also obviously need to make sure those changes actually work on the new hardware. Those tests will be done by PineStore engineering team at first and we'll make sure the Core  developer team also receive new units to be able to maintain the new HW revision.

JF002:
Those changes were suggested by PineStore engineering team to ensure that the new ST7789P3 will be supported by InfiniTime.

They basically move the call to the SLEEPOUT command after all the registers have been set. This should fix the rotation issue they detected while testing the new display.

Fixes #2439

### DIFF
--- a/src/drivers/St7789.cpp
+++ b/src/drivers/St7789.cpp
@@ -17,7 +17,6 @@ void St7789::Init() {
   HardwareReset();
   SoftwareReset();
   Command2Enable();
-  SleepOut();
   PixelFormat();
   MemoryDataAccessControl();
   SetAddrWindow(0, 0, Width, Height);
@@ -32,6 +31,7 @@ void St7789::Init() {
   SetVdv();
   PowerControl();
   GateControl();
+  SleepOut();
   DisplayOn();
 }
 


mfalkvidd:
In https://github.com/InfiniTimeOrg/InfiniTime/issues/2439 you mentioned a 120ms delay. I don’t see that in this change. Is it required?

mark9064 (approved):
Approval conditional on datasheet allowing us to set this while sleeping :)

JF002:
> In #2439 you mentioned a 120ms delay. I don’t see that in this change. Is it required?

Yes, you're right. The 120ms delay is suggested by PineStore. However, according to the datasheet, 5ms should be enough except if we want to send a SLEEPIN command again. In this case, we have to wait 120ms. This is already handled by [St7789::EnsureSleepOutPostDelay()](https://github.com/InfiniTimeOrg/InfiniTime/blob/main/src/drivers/St7789.cpp#L96C14-L96C37).

> Approval conditional on datasheet allowing us to set this while sleeping :)

According to my reading of the datasheet, all the commands that are now sent before SLEEPOUT are available in sleep mode, so we should be OK.

PR #2453, opened 2026-07-19
Keep external flash awake during AOD

### LINKED ISSUE #2451: Navigation app flags stop working with always on mode
### Verification

### What happened?

The navigation app flags disappear after the screen dims in always on mode, when the icons file changes

They should not disappear

### Reproduction steps

1. Enable always on mode
2. Open navigation app and start navigating
3. Wait for screen to dim
4. Trigger a navigation scenario which causes the icons file [1] to change
5. Observe there is no flag

[1]: https://github.com/InfiniTimeOrg/InfiniTime/blob/main/src/displayapp/screens/Navigation.cpp#L198

### More details?

If you keep the screen active and cause a situation where the icon file changes again, the flag will re-appear.

Reverting 44d1798 "fixes" this. 

### Version

main

### Companion app

Gadgetbridge v0.92.0

mark9064:
Fixes #2451 

According to https://pine64.org/documentation/PineTime/_full/#reducing-power-consumption this costs 14uA.
Given that AOD already uses ~3mA, which is ~200x more, I expect no battery life changes

### DIFF
--- a/src/systemtask/SystemTask.cpp
+++ b/src/systemtask/SystemTask.cpp
@@ -310,14 +310,14 @@ void SystemTask::Work() {
           if (state != SystemTaskState::GoingToSleep) {
             break;
           }
-          if (BootloaderVersion::IsValid()) {
-            // First versions of the bootloader do not expose their version and cannot initialize the SPI NOR FLASH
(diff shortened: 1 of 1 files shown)


LMBishop:
Can confirm this has fixed it for me :)

PR #2454, opened 2026-07-19
Improve MotorController flexibility

mark9064:
Refactors motor controller to:
- Support ringing any number of times, with any period between rings
- Adaptively change the vibration strength based on how much the watch is moving
- Use predefined strength values (light/medium/strong) rather than arbitrary numbers across the codebase

Other changes:
- Chime is now implemented as 2 rings 200ms apart. This sets it nicely apart from normal notifications

Internal changes:
- Only one timer is used for ringing
- State machine (Idle, Buzz, RingOn (ringing with motor on), RingOff (ringing with motor off))

Would be interested to hear what people make of the adaptive vibration strength. I think it's definitely led to me missing fewer notifications when on the move

### DIFF
--- a/src/components/motor/MotorController.cpp
+++ b/src/components/motor/MotorController.cpp
@@ -1,43 +1,146 @@
 #include "components/motor/MotorController.h"
+#include <cstdint>
 #include <hal/nrf_gpio.h>
-#include "systemtask/SystemTask.h"
+#include "components/motion/MotionController.h"
 #include "drivers/PinMap.h"
 
 using namespace Pinetime::Controllers;
+
+MotorController::MotorController(Pinetime::Controllers::MotionController& motionController) : motionController {motionController} {
+}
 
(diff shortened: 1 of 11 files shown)


mark9064:
10.3 is too stupid to build this apparently...

mark9064:
Oh, this compiles with LTO, but not without

mark9064:
Should be possible to try now :)

tituscmd:
> Adaptively change the vibration strength based on how much the watch is moving

Could this same idea be applied to the frequency of heart rate background measurements, as a sort of "Dynamic Mode" option?

mark9064:
In theory, yes. But I'm not sure this will have the intended effect since during motion is when heart rate measurement is least reliable

I think an exercise mode which is triggered by motion / walking might work better? Though only on a watch where the sensor can actually work well under motion - maybe one for the PT Pro?

PR #2457, opened 2026-07-27
FSService: Initialise WriteResponse::status on all paths

JoeStrout:
WriteResponse has no default member initialisers and both write handlers assigned status on only one branch, so the other returned uninitialised stack memory to the client over BLE:

- commands::WRITE assigned status only on success, so a failed FileOpen reported garbage. (The ternary was also dead code, as res is known to be 0 inside that branch.)
- commands::WRITE_DATA assigned status only on failure, so every successful chunk reported garbage, which readily looks like a negative LittleFS error code.

This change value-initialises both, matching DelResponse, MKDirResponse and ListDirResponse in the same file, and assign status on both paths.

The visible symptom is a companion app (like InfiniTimeExplorer) reporting a file transfer error ("Error creating file [0]") even when writing the file correctly.

### DIFF
--- a/src/components/ble/FSService.cpp
+++ b/src/components/ble/FSService.cpp
@@ -164,7 +164,7 @@ int FSService::FSCommandHandler(uint16_t connectionHandle, os_mbuf* om) {
       memcpy(filepath, header->pathstr, plen);
       filepath[plen] = 0; // Copy and null terminate string
       fileSize = header->totalSize;
-      WriteResponse resp;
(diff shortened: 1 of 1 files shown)


JoeStrout:
Note: `ReadResponse` and the other response structs are also declared without `{}`, so their `padding` fields go out over the air as uninitialized stack bytes. Harmless functionally, but it is a small information leak.  It might be worth fixing just on general principles — happy to add this to my PR if you like.

PR #2460, opened 2026-07-31
Resets the appStack when the display is automatically moved back to the watch face

### LINKED ISSUE #2461: Unexpected jump back to watch face when button pressed
### Verification

### What happened?

Irrespective of where you are in the menu system, pressing the button takes you back to the watch face

Irrespective of where you are in the menu system, pressing the button takes you back to the screen you were previously on

### Reproduction steps

Set display timeout to 5 seconds or the below will take a while to run through.

Now navigate to a screen deep in the menu (eg About).  Press the button and the screen jumps back to the previous, list of settings you can change.

Loop 10 times: 
  Turn the watch on by pressing the button
  Watch face appears
  Swipe to the Settings screen
  Wait for the watch to turn the display off
/ Loop

Now navigate to a screen deep in the menu (eg About).  Press the button and the screen jumps back to your watch face.

Note that I also got to this state by Swiping to the Notifications screen and letting the watch turn the display off.
It looks like on the 8th loop, you jump back two screens and the 9th loop you jump back to the watch face.  10 loops it to make sure you get to the "bad" state.

### More details?

the appStack is getting filled up because the previous screen is not popped off the appStack if the display goes to sleep when you are on the settings or navigation screens.

### Version

V1.16.1

### Companion app

Gadegetbridge 0.89.1

SteveAmor:
Button presses and gestures to move to another screen pop the previous screen off the App Stack. 

When the display times out to go to sleep, if you are on the Notifications or Settings screen, the display automatically goes back to the watch face.  This change in screen isn't popped off the App Stack which eventually fills the appStack and causes a button press to jump straight to the watch face rather than the previous screen. 

Given the watch face can be considered as the top level screen, this PR prevents the stack overflow/stale-entry behaviour.  This is in fact what happens when you do a button:longPress

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -323,6 +323,8 @@ void DisplayApp::Refresh() {
         if (currentApp == Apps::Launcher || currentApp == Apps::Notifications || currentApp == Apps::QuickSettings ||
             currentApp == Apps::Settings) {
           LoadScreen(Apps::Clock, DisplayApp::FullRefreshDirections::None);
+          appStackDirections.Reset();
+          returnAppStack.Reset();
           // Wait for the clock app to load before moving on.
           while (!lv_task_handler()) {
           };


SteveAmor:
Tested on PineTime hardware and it works as expected and I can't force it into a bad state as per the process in #2461

mark9064 (approved):
Great catch

PR #2464, opened 2026-08-02
Release wakelock when timer Reset is pressed so screen will go to sleep

SteveAmor:
Fixes #2419 
Tested on watch hardware

### DIFF
--- a/src/displayapp/screens/Timer.cpp
+++ b/src/displayapp/screens/Timer.cpp
@@ -182,6 +182,7 @@ void Timer::ToggleRunning() {
   auto timerStatus = timer.GetTimerState();
   if (timerStatus && timerStatus->expired) {
     motorController.StopRinging();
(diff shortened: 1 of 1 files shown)


mark9064 (approved):
I'm wondering if this should be in the `Reset()` function instead. But I think this might be more correct, the wakelock should release on the transition from ringing <10s to >10s, and also from ringing <10s to stopped / reset.

SteveAmor:
I'm still not really sure if it should go in ```Reset()``` or not.  There were a few reasons I put it where it is. First was so that it was very clear what is happening, second was to save unnecessarily releasing the wakelock when other functions call ```Reset()``` but the release is not strictly required, and the third was that I didn't spot any refactoring that would result in a reduction in program size to move it to ```Reset()```.
Happy to move to ```Reset()``` if my thought process is flawed/not a good design etc.

Issue #2118, opened 2024-09-04
The SPI flash memory ID is displayed as “ff-ff-ff” once the PineTime has gone into sleep mode (on new chips only).


JF002:
### Verification

- [X] I searched for similar bug reports (including closed issues) and found none was relevant.

### What happened?

The SPI flash ID are set to "ff-ff-ff" in System Info after deep sleep

### What should happen instead?

They should be displayed as "68-40-16"

### Reproduction steps

Using a new PineTime equipped with the [new flash chip - BY25Q32ESWIG](https://github.com/InfiniTimeOrg/InfiniTime/issues/2096), boot the device and before it has the chance to go to sleep, go to the SystemInfo app (Quick Settings -> Settings -> About -> 3rd page). The SPI Flash field displays "68-40-16".

Let the watch go to sleep, and repeat the procedure again : the Spi Flash field now displays "ff-ff-ff"

### More details?

I guess something prevents InfiniTime from fetching those IDs after deep sleep. I'll need to debug this further and check the data sheet to check if there's something specific related to the deep sleep mode.

Note that the SPI flash seems to work fine even if InfiniTime couldn't get the IDs.

### Version

1.14.1

### Companion app

N/A

Issue #2096, opened 2024-07-21
PineStore announcement : the current flash chip is EoL, a new one will be used for future PineTime batches


JF002:
### Verification

- [X] I searched for similar issues (including closed issues) and found none was relevant.

### Introduce the issue

I've just received this information from TL Lim from PineStore : the current SPI flash memory chip (XTX XT25F32B) is EoL (end of life) and will be replaced by another one : BY25Q32ESWIG. 

Here is the datasheet : [BY25Q32ESWIG-03.A.02.00372.pdf](https://github.com/user-attachments/files/16323586/BY25Q32ESWIG-03.A.02.00372.pdf)

The only change expected on the software point of view are the JEDEC IDs :

| Chip | Manufacturer | Memory type | Density |  
|---------|---------------------|---------------------|-------------|
| XTX XT25F32B    | 0x0B | 0x40 | 0x16 |
| BY25Q32ESWIG | **0x68** | 0x40 | 0x16 |

PineStore asked me to provide them with a new firmeware that supports this new chip.

**Feel free to forward this announcement to all firmware developers that might be interested to support this new flash chip!**

### Preferred solution

At first sight, no change should be necessary in InfiniTime, since it does not check the IDs (it only logs the value, but does not fail if they are not the ones that were expected).

However, the SPI flash driver in the [bootloader](https://github.com/InfiniTimeOrg/pinetime-mcuboot-bootloader) will most probably fail since those IDs are specified and checked at boot time. 

Since the bootloader is a very critical part of the project, please feel free to cross-check my analysis and also test and provide feedback when a new version of the bootloader will be available :)

### Version

1.14.0

SteveAmor:
Do you think there are going to be any issues with people who receive one of these new PTs and flash the bootloader to use WASPOS and then use the process/file to flash back to Infinitime?

JF002:
@SteveAmor This is a very good question! According to TL Lim and my observations, the new chip behaves exactly the same way than the old one, so no code change should be needed except if the bootloader/reloader checks the chip IDs of the memory chip. 
Do you know if this announcement has reached the WaspOS development team?

PR #2466, opened 2026-08-06
Fix OTA silently transferring a zero-length image

fluffyspace:
## The bug

`DfuService::OnServiceData` arms the 10 s one-shot inactivity timer before dispatching each access:

```cpp
if (bleController.IsFirmwareUpdating()) {
  xTimerStart(timeoutTimer, 0);
}
```

The image-size write then blocks the BLE host task twice before it can respond: an unbounded wait for SystemTask to disable sleeping, then `DfuImage::Erase()`, which erases the whole image slot sector by sector. Measured from the host (the gap between writing the image sizes and the `10 01 01` response, which is only sent after the erase returns), the erase finishes in under ~8 s on my watch (BY25Q32ESWIG) — but the datasheet worst case is ~35 s (116 sectors × 300 ms max), and the sleep-disable wait ahead of it is unbounded, so the blocked stretch can outlast the 10 s window. The timer runs on the FreeRTOS timer task, so it fires while the BLE task is blocked and calls `Reset()`, which zeroes `applicationSize`.

The erase then completes and the handler carries on regardless — it sets `state = Init` and reports success. `ReceiveFirmwareImage` later calls `dfuImage.Init(20, 0, crc)`, so `IsComplete()` (`totalWriteIndex == totalSize`, now `0 == 0`) is **already true when the first data packet arrives**. The watch tells the host the entire image has been received, emits no packet-receipt notifications, writes nothing to flash, and validation CRCs zero bytes.

## What it looks like from the host

- `10 03 01` ("whole image received") arrives a fraction of a second after Begin DFU — no real transfer can be that fast
- **zero** `11 …` packet-receipt notifications for the entire "transfer"
- validation fails, and nothing was written

There is no indication that anything timed out, which makes this very hard to attribute. It cost me a couple of evenings before I instrumented the watch.

Recovery mode is unaffected: its blocked stretch finishes inside the 10 s window. That is why OTA can fail consistently from the running firmware while succeeding from recovery — which reads like a host or radio problem rather than a firmware one.

## The fix

Stop the timer for the whole blocked stretch — the sleep-disable wait and the erase — and restart it once the handler can react to packets again. While the BLE task is blocked there it cannot process packets anyway, so there is no inactivity for the timer to detect in that window; it still guards the actual transfer, and a host that disappears mid-erase is caught ~10 s after the response goes out.

## Testing

Tested on a PineTime (bootloader 1.0.1, InfiniTime 1.16.0, BY25Q32ESWIG flash). Before this change, every OTA from the running firmware failed exactly as described. After it, updates complete and validate normally, and recovery mode is no longer needed to flash — verified across several successful ~390 kB updates.

For scale: host-side logs of seven successful runs put the `10 01 01` response at 8.4–9.3 s after script start *including* scan, connect and service discovery, so the erase itself is under ~8 s on this watch — the failures need the extra contribution of the sleep-disable wait, or a slower/worn part. I don't have an instrumented failing run (those predate the logging), so the exact split on the failing attempts is unknown.

### DIFF
--- a/src/components/ble/DfuService.cpp
+++ b/src/components/ble/DfuService.cpp
@@ -145,7 +145,13 @@ int DfuService::WritePacketHandler(uint16_t connectionHandle, os_mbuf* om) {
         vTaskDelay(pdMS_TO_TICKS(5));
       }
 
+      // Erasing the image slot takes far longer than the 10s inactivity timeout, and that
+      // timer runs on the FreeRTOS timer task, so it fires while this call blocks. Its
+      // Reset() zeroes applicationSize, the erase then completes and the transfer carries
+      // on with a zero-length image. Hold the timeout off for the duration of the erase.
+      xTimerStop(timeoutTimer, 0);
       dfuImage.Erase();
+      xTimerStart(timeoutTimer, 0);
 
       uint8_t data[] {16, 1, 1};
       notificationManager.Send(connectionHandle, controlPointCharacteristicHandle, data, 3);
--- a/src/components/ble/DfuService.cpp
+++ b/src/components/ble/DfuService.cpp
@@ -138,6 +138,15 @@ int DfuService::WritePacketHandler(uint16_t connectionHandle, os_mbuf* om) {
                    bootloaderSize,
                    applicationSize);
 
+      // From here until the response below, this task blocks and cannot service BLE:
+      // first the unbounded wait for SystemTask to disable sleeping, then the image-slot
+      // erase, which can approach or exceed the 10s inactivity timeout (116 sectors at a
+      // datasheet-max 300ms each). The timer runs on the FreeRTOS timer task, so it fires
+      // while this task is blocked; its Reset() zeroes applicationSize, the erase then
+      // completes and the transfer carries on with a zero-length image. Hold the timeout
+      // off until this handler can react to packets again.
+      xTimerStop(timeoutTimer, 0);
+
       // Wait until SystemTask has disabled sleeping
       // This isn't quite correct, as we don't actually know
       // if BleFirmwareUpdateStarted has been received yet
@@ -145,11 +154,6 @@ int DfuService::WritePacketHandler(uint16_t connectionHandle, os_mbuf* om) {
         vTaskDelay(pdMS_TO_TICKS(5));
       }
 
-      // Erasing the image slot takes far longer than the 10s inactivity timeout, and that
-      // timer runs on the FreeRTOS timer task, so it fires while this call blocks. Its
-      // Reset() zeroes applicationSize, the erase then completes and the transfer carries
-      // on with a zero-length image. Hold the timeout off for the duration of the erase.
-      xTimerStop(timeoutTimer, 0);
       dfuImage.Erase();
       xTimerStart(timeoutTimer, 0);
 


SteveAmor:
@fluffyspace are you absolutely sure that ```DfuImage::Erase()``` takes tens of seconds?
If so, there could be a difference between the old spiNOR and new spiNOR chip when it comes to the time taken to erase sectors.

I just checked the datasheets and the new NOR is faster for a sector erase @ typically 35ms vs the old @ typically 50ms.

I belive there are 116 (decimal) erases in the loop so I'd expect all sectors to be erased in under 10 seconds.

mfalkvidd:
Could be a damaged sector (or multiple) that cause the erase to take longer than specified?

mfalkvidd:
Btw, the datasheet linked from https://pine64.org/documentation/PineTime/Further_information/Datasheets_schematics_and_certifications/ says 70ms typical which gives a bit more than 8 seconds for 116 sectors. Maybe that leaves a too thin margin?

SteveAmor:
Good spot. For the new NOR, I used the datasheet from here https://github.com/InfiniTimeOrg/InfiniTime/issues/2096
Also noted that these values are typical and not max.

mfalkvidd:
Is there an easy way to check which nor chip is in a watch? Accelerometer variant is shown in Settings->About but the nor manufacturer is not (yet) surfaced there.

SteveAmor:
> Is there an easy way to check which nor chip is in a watch? Accelerometer variant is shown in Settings->About but the nor manufacturer is not (yet) surfaced there.

It's on the 3rd page of About. SPI Flash. The old NOR is 0b-40-16. I believe the new NOR will show ff-ff-ff if the watch has gone into deep sleep #2118

mfalkvidd:
Mine shows all ff so that seems to be as expected then. @fluffyspace maybe you could check your version?

Maybe the timer should be kept, but increased to give better margin for the older chips, especially as they get slower over time because of natural wear?

Edit: The datasheet of the new nor mentions 300ms max erase time which would result in 35s for 116 sectors.

SteveAmor:
> Edit: The datasheet of the new nor mentions 300ms max erase time which would result in 35s for 116 sectors.

Gosh, yeah, just saw that.
Perhaps the timer should be excessive, just in case the flash is borked and then act accordingly to avoid a complete lock up. May be even create a specific erase timeout.

fluffyspace:
@mfalkvidd Checked mine. Read straight after a reboot, before the watch had a chance to sleep, About page 3 shows `68-40-16` — the new BY25Q32ESWIG. Once it has slept it reads `ff-ff-ff`, exactly as described in #2118. So this reproduced on the *newer* chip, not an old worn one.

Which I think is the useful data point here: on this watch the erase demonstrably overran the 10 s timeout on *every* attempt, consistently enough that OTA from the running firmware never once succeeded. So the 300 ms max is not a theoretical worst case that only shows up on degraded parts — the typical figures just don't describe what these chips actually do. And `maxSize` is 475136, exactly 116 sectors, so 116 × 300 ms ≈ 35 s is the number to design against.

@SteveAmor On the erase-specific timeout — I don't think a timer at this layer can do the job you want, for two reasons.

It can't detect anything. While `Erase()` runs, the BLE host task is blocked inside it and cannot process an incoming packet even if one arrives. There is no inactivity to detect in that window, so the timer can only produce false positives. It earns its keep during the actual transfer, which is why this patch restarts it immediately afterwards — a host that disappears mid-erase is still caught ~10 s later.

It can't prevent the lock-up either. `OnTimeout()` calls `Reset()`, which doesn't abort anything; it zeroes `applicationSize` on the timer task while the erase keeps running on the BLE task. That race *is* the bug. Making the timeout larger only makes it rarer — at 35 s worst case you'd need a very generous margin to be safe, and you'd still have a silent-corruption path rather than a clean failure.

And if the flash really is borked, the hang isn't in DFU at all:

```cpp
// src/drivers/SpiNorFlash.cpp:87-102
  WriteEnable();
  while (!WriteEnabled())
    vTaskDelay(1);
  ...
  while (WriteInProgress())
    vTaskDelay(1);
```

Two unbounded spins. A timer callback can't break a blocked task out of those, so an erase timeout would fire, call `Reset()`, and leave the BLE task still wedged in the driver — inconsistent state *and* the hang, which is worse than where we are now.

Your underlying concern is fair, I just think it belongs a layer down: bound those two waits against the datasheet max, have `SectorErase` report failure, and let `DfuImage::Erase()` propagate it so DFU aborts cleanly instead of hanging. I'm happy to open that as a separate PR — it's the same shape as the unbounded-loop fixes in #2467 and #2468. Keeping it out of this one means the zero-length-image bug can be fixed now without waiting on the larger change.

mfalkvidd:
Great info. I wonder why your watch fails every time while mine has not had any failures. We both have the new chip.

Maybe I do OTA differently and don’t hit this path?
I use ITD on a Linux laptop to update firmware. Have not looked into which commands it sends.

SteveAmor:
I have the old chip and regularly OTA while developing.  I assume the firmware update bar percentage stays at zero when the erase is in progress.  Mine is roughly paused for 5 seconds before the OTA starts to progress.  I use Gadgetbridge.

fluffyspace:
@mfalkvidd To answer the client question first: I don't use ITD — I drive siglo's `InfiniTimeDFU` class headlessly from a Python script on Linux. But I don't think the client is the difference: the erase is kicked off by the image-size write and runs to completion on the watch regardless of what the host does afterwards, so no client can shorten or skip it. Clients differ in how they tolerate the long silence, but that failure mode looks like the host giving up mid-erase, not like this bug.

Second — I owe the thread a correction. I went back through my flash logs to get a real number instead of an impression, and they don't support "tens of seconds" as written in the PR description. My script timestamps notifications from process start, and across seven successful runs the `10 01 01` response — which is only sent after `Erase()` returns — arrives at 8.4-9.3 s, *including* scanning, connecting, service resolution and the Start DFU handshake. So the erase itself is under ~8 s on my watch. @SteveAmor your skepticism was better founded than I credited; I'll amend the PR description.

What I don't have is a number from a failing run — the failures that led me to this bug predate that logging. What I can say is that OTA from the running firmware failed with this exact zero-length signature every time before the patch, and has succeeded every time since.

Which sent me back to the code, because a sub-8 s erase shouldn't trip a 10 s timer. And the timed window turns out to be more than the erase:

```cpp
      // Wait until SystemTask has disabled sleeping
      // This isn't quite correct, as we don't actually know
      // if BleFirmwareUpdateStarted has been received yet
      while (!systemTask.IsSleepDisabled()) {
        vTaskDelay(pdMS_TO_TICKS(5));
      }

      dfuImage.Erase();
```

That wait is unbounded, runs inside the armed 10 s window, and depends on how quickly SystemTask gets around to processing `BleFirmwareUpdateStarted` — which varies with what the watch is doing at that moment, in a way flash timing doesn't. An erase that reliably takes ~8 s leaves ~2 s of budget for that wait plus radio latency. I think that's the answer to "why my watch and not yours": two watches with the same chip can sit on opposite sides of a 10 s cliff without either flash being out of spec, and the side you land on can depend on the watch's state when the transfer starts, not just silicon.

Two consequences:

1. I've updated the PR to stop the timer *before* the sleep-disable wait rather than just before the erase, so the whole blocked stretch is covered — same size of change, strictly more correct — and amended the description to match the measured numbers.
2. The datasheet worst case matters independently of my logs: 116 sectors × 300 ms max ≈ 35 s is legal per spec, so a 10 s inactivity timeout spanning this window is wrong by construction even if most parts finish in 8.

@SteveAmor your ~5 s pause on the old chip is a useful third data point — that pause is dominated by the erase (Gadgetbridge's bar can't move until the watch answers the image-size write). So the picture so far is: old chip ~5 s, mine under ~8 s, both "working", and both a large fraction of a 10 s budget on healthy parts. That reads as a margin problem, not a broken-chip problem — notably the old chip is the *fastest* watch measured yet, which doesn't fit the wear theory.

@mfalkvidd if ITD logs its control-point traffic with timestamps, the gap between writing the image sizes and receiving `10 01 01` is your number for the same window. If yours also lands around 6-8 s, that confirms the thin-margin picture over the slow-chip lottery — either result would be informative.

mfalkvidd:
I’ll check how verbose ITD’s logs are. I guess I could use btmon or wireshark on the Linux host instead if the logs are insufficient?

Not sure when I can try it but I’ll report as soon as I can.

I don’t quite get why two watches with the same chip consistently would fall on opposite sides of the cliff though. Would you mind elaborating?

fluffyspace:
I don't know. I don't have logs from the failing runs so I can't say why mine failed every time and yours never does. Maybe some chips are just slower than others, or it depends on what the watch is doing when the transfer starts. All I know for sure is that before the patch every update from the running firmware failed like this, and after the patch every update works.

btmon sounds like a good idea. A real measurement will tell us more than my guessing.

SteveAmor:
I guess you could remove this patch and test to get some logs of the failure?

fluffyspace:
Yes, I can do that. I'll flash a build without the patch and capture timestamps of a failing attempt, btmon too. Might take me a day or two.

fluffyspace:
As promised: I put unpatched stock 1.16.0 on the watch (BY25Q32ESWIG) and tried OTA from the running firmware, with btmon capturing.

It did not reproduce. Two attempts, both transfers completed and validated normally.

The number the thread wanted, though, is better than any I had before. My script used to scan before connecting, so my timestamps bundled scan, connect and discovery into the "erase" figure. It now connects directly, which isolates the protocol:

- attempt 1: image-size write → `10 01 01` in **3.4 s**
- attempt 2: image-size write → `10 01 01` in **3.3 s**

Both attempts started with the watch asleep, which is when the sleep-disable wait has the most work to do — and the whole blocked stretch still fits in ~3.4 s.

So I owe the thread a second correction: the 8.4–9.3 s I posted earlier was mostly connection overhead. The real blocked stretch — sleep-disable wait plus full-slot erase — is ~3.4 s on this watch, about a third of the 10 s budget. More margin than I implied. @SteveAmor your skepticism keeps ageing well.

I'm sorry about the churn. This is the second time I've had to correct my own numbers on this thread, and you've spent review time on figures that didn't isolate what I claimed they did. I should have measured properly before arguing from impressions.

I also have to qualify "failed every time before the patch": those failing runs were OTAs from my development fork. Its DfuService was unmodified upstream code, but the rest of the firmware was not stock, and the length of the sleep-disable wait depends on what SystemTask is doing at that moment. I can't rule out that something in my fork inflated that wait past the window back then. I don't know. Whatever inflated it, the failure it produced was the silent zero-length transfer this PR describes, and stock can hit the same window with a slower part or a busier moment.

What doesn't depend on my logs either way: while the host task sits in the sleep-disable wait and the erase, it cannot process packets, so the inactivity timer can only false-positive in that window — and the erase alone is allowed ~35 s by datasheet (116 sectors × 300 ms max) against the 10 s timeout. Stopping the timer across the blocked stretch is correct by construction.

I'll keep the unpatched build around and try to catch a failing run with btmon attached; if anyone wants the btsnoop of a successful run in the meantime, say so.

PR #2469, opened 2026-08-07
Improve notification preview UX

SteveAmor:
Implements better notifications UX as discussed in #1844 as I have always been annoyed that you can't swipeDown to see earlier messages when in notification preview state but you can tap, dismiss or swipeUp.

```bool interacted``` no longer required because you now transition from Modes:Preview to Modes:Normal when you interact with a preview notification.

Maintains the same functionality as before but enables you to swipeDown to see earlier messages when you get a notification.

### DIFF
--- a/src/displayapp/DisplayApp.cpp
+++ b/src/displayapp/DisplayApp.cpp
@@ -320,8 +320,8 @@ void DisplayApp::Refresh() {
           brightnessController.Set(Controllers::BrightnessController::Levels::Off);
         }
         // Since the active screen is not really an app, go back to Clock.
-        if (currentApp == Apps::Launcher || currentApp == Apps::Notifications || currentApp == Apps::QuickSettings ||
-            currentApp == Apps::Settings) {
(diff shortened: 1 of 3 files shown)


SteveAmor:
Tested for a couple of days on the watch hardware and it is much less annoying :-D

mark9064:
This sounds great. The notification state machine is a beast so I will test this by daily driving for a bit :D
(Then I promise I'll try actually digging into it!)

mark9064:
Might also be worth adding `NotificationPreview` to the list of apps AOD can return to the clockface

mark9064 (approved):
Reviewed state machine changes: makes a lot more sense to properly transition out of the preview mode rather than express this with the interacted variable

Tested on hardware, all good. Integrating above suggestion up to you, not merge blocking IMO

SteveAmor:
Happy to implement that. I need to get more familiar with AOD to add that to my physical tests.

SteveAmor:
Added ```NotificationPreview``` to the list of apps AOD can return to the clockface.
Much better AOD UX. Tested on hardware ok.

mark9064:
HW test all good from me as well :)

PR #2450, opened 2026-07-18
Add Legend watch face

DavisNT:
An elegant and very practical digital watch face.

Uses only fonts already in InfiniTime and no external resources. Has the standard functionality:
* shows time, date (without year) and day of week
* supports 24 and 12 hour time formats
* BLE, new notification and alarm clock icons
* battery percentage (green when charging)
* steps and heart rate
* temperature and weather icon

Additionally:
* visual indication (amber dot) for any notification(s)
* visual indication (red or blue line) when notifications are muted
* battery percentage turns red when 20% or less
* stopwatch and timer icons (when active)
* colorful sun, snow and thunderstorm weather icons
* generally uses brighter colors for more urgent things

InfiniSim_2026-07-18_223549 
InfiniSim_2026-07-18_215528 
InfiniSim_2026-07-18_215818

PR #2471, opened 2026-08-08
Implement DirtyValues on Tile, Quick Settings, Steps and Batt Info screens

SteveAmor:
Implemented as an alternative to #2257

### DIFF
--- a/src/displayapp/screens/BatteryInfo.cpp
+++ b/src/displayapp/screens/BatteryInfo.cpp
@@ -7,16 +7,12 @@ using namespace Pinetime::Applications::Screens;
 
 BatteryInfo::BatteryInfo(const Pinetime::Controllers::Battery& batteryController) : batteryController {batteryController} {
 
-  batteryPercent = batteryController.PercentRemaining();
-  batteryVoltage = batteryController.Voltage();
-
   chargingArc = lv_arc_create(lv_scr_act(), nullptr);
   lv_arc_set_rotation(chargingArc, 270);
   lv_arc_set_bg_angles(chargingArc, 0, 360);
   lv_arc_set_adjustable(chargingArc, false);
   lv_obj_set_size(chargingArc, 180, 180);
   lv_obj_align(chargingArc, nullptr, LV_ALIGN_CENTER, 0, -30);
-  lv_arc_set_value(chargingArc, batteryPercent);
   lv_obj_set_style_local_bg_opa(chargingArc, LV_ARC_PART_BG, LV_STATE_DEFAULT, LV_OPA_0);
   lv_obj_set_style_local_line_color(chargingArc, LV_ARC_PART_BG, LV_STATE_DEFAULT, Colors::bgAlt);
   lv_obj_set_style_local_border_width(chargingArc, LV_ARC_PART_BG, LV_STATE_DEFAULT, 2);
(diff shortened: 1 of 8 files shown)


SteveAmor:
@mark9064 I would very much appreciate your review of this.

I see that statusIcons.Update() already uses dirty values to only update if there is a change https://github.com/InfiniTimeOrg/InfiniTime/blob/71d1f5b45bd2bc9499acb16a5b6893f5aae0a15f/src/displayapp/widgets/StatusIcons.cpp#L34
So I only needed to update the time at the top of the screen to use dirty values. 

I thought I might have to use override for the UpdateScreen but the compiler didn't like that https://github.com/InfiniTimeOrg/InfiniTime/blob/71d1f5b45bd2bc9499acb16a5b6893f5aae0a15f/src/displayapp/screens/settings/QuickSettings.h#L33
If I got it right, I'll move on to a few other screens - like the apps screen which has a similar layout with status icons and the time. 
It works as expected on my watch.

mark9064 (approved):
This is a nicer solution!

The logic with dirtyvalue all looks good. When you said you thought you might have to use override, what did you mean by that?

There's a few other places that have `Refresh()` in the wrong place - maybe you will find this useful for your changes: https://github.com/mark9064/InfiniTime/commit/43bfddc1715143cd422d2bd1dc93657f4956b3f4

SteveAmor:
> This is a nicer solution!
> The logic with dirtyvalue all looks good. 

Thank you 

> When you said you thought you might have to use override, what did you mean by that?

I saw in other screens that use DirtyValues that the header file has `void Refresh() override;` and I'm getting out of my depth with that one. 

> There's a few other places that have `Refresh()` in the wrong place - maybe you will find this useful for your changes: [mark9064@43bfddc](https://github.com/mark9064/InfiniTime/commit/43bfddc1715143cd422d2bd1dc93657f4956b3f4)

Very useful, thank you.

I have updated Tile.cpp (not pushed yet) and then started on BatteryInfo and the lvgl was not lining up. Now I can see why! 
BatterInfo needs a lot of work to use DirtyValues. I think I need to track the charging/discharging status to see if that changes which is more of a refactor than I thought it would be. Rather than track changes in charging.

mark9064:
Ahhhhh right I see

The `Screen` class, which (almost?) all app screens inherit from provides a dummy `Refresh` method and a short helper which you can pass to LVGL (see `Screen.cpp`). `Refresh` is purposefully a virtual method: if a method is virtual, it means that classes inheriting from it are allowed to override it (i.e. replace the contents of the method with something else), and when you do override it you need to use the `override` keyword.

Since QuickSettings is using its own helper for LVGL and its own `UpdateScreen`, it's not overriding anything. But it would probably make sense to use `Refresh` and the normal helper here like other apps, so maybe worth changing?

SteveAmor:
Got it. Thank you. I had started to refactor using `Refresh()` and backed out for two reasons. First was the "override" or not question and not really understanding it (and not wanting to use the excuse "well it compiled so must be ok") and second because I wanted to go back and tidy up the few cases that use `UpdateScreen()` in a PR just for that change. And then I saw your lvgl PR and all I could think about was the merge conflicts 🤣

mark9064:
LOL don't worry that isn't a PR yet. I can rebase around any conflicts, I don't mind

SteveAmor:
@mark9064 

Forced push above was to refactor for `Refresh()`.

I now have tested changes to Tile.cpp, Steps.cpp and BatteryInfo.cpp that include dirty values for screen changes, lvgl refactoring and screenUpdate replaced with `Refresh()` where required.  I can add them to this PR, raise individual PRs for each screen or a new PR with those three screens.  What would you prefer me to do?

mark9064:
I think rolling them into this PR should be fine - sounds like the overall change isn't going to be huge

SteveAmor:
After the discussion about dirty values in https://github.com/InfiniTimeOrg/InfiniTime/pull/2450#issuecomment-5272703492 I have updated the steps goal to be a dirty value.

PR #2479, opened 2026-08-18
Remove redundant LFS logging information

mark9064:
The LFS log events are unique and don't need line/file information

Since some of the LFS logs use 4 parameters, this was causing too many parameters of the maximum 6 to be used

I'm not sure why this wasn't a problem with GCC 10.3

To be honest I'm not a macro expert so I can't really debug the actual expansion, but I don't think we actually lose debugability here by dropping these


mark9064:
Build is passing, but compare base branch is failing (since main is broken right now)

NeroBurner (approved):
I agree

PR #2480, opened 2026-08-21
Notification screen at max brightness when walking

SteveAmor:
This is possibly a proof of concept and I am more than open to any feedback on the code and/or enhancements/changes.

I usually have my Pinetime on Low brightness but when I'm walking outside and get notifications I put the brightness on High so I can read them in the light.  Given there's no ambient light detector, I realised I could check if I'm walking so I've created this PR.  

It would be great if it did make it into the main branch but it's possibly an unusual user case. ~~and there's no settings for it.~~ You can select Motion Auto Bright (MAB) in the Display settings screen to toggle it on and off.

Essentially, if you are walking and you get a notification (or check your notifications), the screen will be at max brightness.  There's a ~~5~~10 second decay for the last time your steps increased to keep this feature "active". After that, notifications will appear at normal brightness.

There may be scope for more types of activities to be triggered in the MotionController.
This PR may also give some ideas for the PTPro.

### DIFF
--- a/src/components/motion/MotionController.cpp
+++ b/src/components/motion/MotionController.cpp
@@ -76,6 +76,9 @@ void MotionController::Update(int16_t x, int16_t y, int16_t z, uint32_t nbSteps)
   int32_t deltaSteps = nbSteps - oldSteps;
   if (deltaSteps > 0) {
     currentTripSteps += deltaSteps;
+    locomotionDecay = 50; // 5 seconds x 10Hz
+  } else if (locomotionDecay > 0) {
+    locomotionDecay--;
   }
   SetSteps(Days::Today, nbSteps);
 }
--- a/src/components/motion/MotionController.cpp
+++ b/src/components/motion/MotionController.cpp
@@ -76,7 +76,7 @@ void MotionController::Update(int16_t x, int16_t y, int16_t z, uint32_t nbSteps)
   int32_t deltaSteps = nbSteps - oldSteps;
   if (deltaSteps > 0) {
     currentTripSteps += deltaSteps;
-    locomotionDecay = 50; // 5 seconds x 10Hz
+    locomotionDecay = 100; // 10 seconds x 10Hz
   } else if (locomotionDecay > 0) {
     locomotionDecay--;
   }
--- a/src/components/motion/MotionController.h
+++ b/src/components/motion/MotionController.h
@@ -74,6 +74,10 @@ namespace Pinetime {
         return service;
       }
 
+      bool Locomotion() const {
+        return (locomotionDecay != 0);
+      }
+
     private:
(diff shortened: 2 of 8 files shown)


SteveAmor:
Toggle for Motion Auto Bright (MAB)
 
InfiniSim_2026-08-22_174737

mark9064:
Could adding in a daytime check using the sunset/sunrise time be a good addition? Haven't actually tried this but seems like it might avoid being super bright when walking at night. Might be overcomplicating things though

Eve1374:
cool ! I just flashed it. I'd like to implement it for cycling, I'd have to check speed on a certain axis instead of steps increase I guess. Or maybe a shortcut like triple tap to activate notification max brightness (in case activity is not detected - or one is outside without cycling / walking...)

SteveAmor:
> cool ! I just flashed it. I'd like to implement it for cycling, I'd have to check speed on a certain axis instead of steps increase I guess. Or maybe a shortcut like triple tap to activate notification max brightness (in case activity is not detected - or one is outside without cycling / walking...)

Thanks for the feedback!

I think it would be great to add other "activities" to the MotionController which can then be used to trigger max brightness.  

Alternatively, if you don't use "shake to wake", you could use that to activate max brightness.

I'm also currently wondering if this feature should be available on all screens. Something I'm considering implementing globally to test. That would be useful for the alarm going off or just, simply, checking the time.

SteveAmor:
So, the watch does have a light sensor, it's just underneath the watch. I wonder if there is a way of calibrating it to capture the ambient light level 🤔

mark9064:
Difficult: depends massively on skin tone and strap tightness / contact pressure

SteveAmor:
> Could adding in a daytime check using the sunset/sunrise time be a good addition? 

Forgive me if I've got this wrong but I think that will rely on the user subscribing to a weather service (to get the sunset/sunrise times), if their companion app even has that ability to support weather.

The current weather info is not something I've explored yet.

mark9064:
Yeah it will - I meant it as something that could be active if weather information is present, but not a dealbreaker if it isn't

At least for anyone using GB, sunrise/sunset should work

SteveAmor:
Thanks @mark9064. I think it will make a great improvement and am going to start testing with some hard coded times before then moving to real value.

SteveAmor:
Got it all working using sunrise and sunset. I didn't realise there was a method `IsNight()`. The good thing is that if you are not using a weather service or this info isn't provided, it will return `false`.
I've also implemented it globally now - making the decision to go to max brightness in `DisplayApp`. This could be good for waking the display up for the time or screens like Navigation.
I have also highlighted the Quick Settings brightness icon in yellow if MAB is enabled.
I've been testing this weekend and all seems well.

PR #2307, opened 2025-05-30
Prime Watchface

tituscmd:
Hey all!

This PR introduces a new watch face I recently made. I went with the name "Prime" for no reason at all but it sounding _cool_—I’m not great at naming things, so I’m totally open to suggestions 😅. The design is based on the existing Digital watch face. Whether this should replace the current Digital face or live alongside it, I don’t have a strong preference either way.

This watch face uses a new font I added called **Karnivore**. It’s a free TTF font, so it should be fine licensing-wise—but feel free to flag anything I might’ve missed.

One neat addition is that it shows the current song playing directly on the face, below the line showing the day of the week and the date. Though it has to be said that its color might be a little bit dark on the black background, but I thought it looked best like that regardless.

The temperature above the time changes color depending on how warm it is:
```
-0°C or below → blue (freezing)
-1–5°C        → cyan (near freezing)
-6–15°C       → light blue (early spring)
-16–25°C      → orange (warm)
-over 25°C    → deep orange (hot)
```

I also styled the heart rate and step count icons/values to match the colors used in the InfiniLink app because I thought they looked nice(r) :)

The day of the week is shown in sentence case (Fri), instead of all caps like on the Digital face. The date is in DD.MM.YYYY format. I personally prefer using periods and leading zeroes for single-digit months.

The battery percentage is shown at the top. Right now it’s hard-coded, but this can of course be updated if PR #1959 is merged.

I think that covers everything! Hope you enjoy it 🙂

image
prime_on_device

PR #2481, opened 2026-08-30
Watchface click on weather and steps

SteveAmor:
Another proof of concept.
Shortcut from the digital watch face to jump to the weather app and steps app when you tap the watch face icons and then jump back to the watch face when you leave weather or steps.
~~It doesn't work so well on the digital watch face as the weather icon is in the middle of the screen. So swiping up and down can trigger a trip to the weather app.~~

I've implemented this on the Legend watch face (which is where I really wanted this feature), along with tapping the battery percent in the top right to take you to BatteryInfo. ~~It works really well with the battery percent, steps and weather in the corners of the Legend watch face.~~

My intention is to update this PR for the Legend watch face when it's merged into main. And after I've implemented any feedback 😀


SteveAmor:
Force push to fix formatting

tituscmd:
This is a great idea! I think I'll implement this as well for my watchface (#2307) and I'll make sure to @ you for this idea.

tituscmd:
After playing around with it for a while, you should use LV_EVENT_CLICKED instead of pressed, so that it doesnt fire when scrolling around

SteveAmor:
> After playing around with it for a while, you should use LV_EVENT_CLICKED instead of pressed, so that it doesnt fire when scrolling around

Cool. Thanks for the feedback. I was going to investigate that as the settings menu is quite happy with scrolling and not triggering a setting menu 😀  I'll update the PR.

tituscmd:
I noticed it because on my watch face, weather is in the top middle, which made scrolling down for notifications a pain in the ass 😅

SteveAmor:
Oh, wow, LV_EVENT_CLICKED is much, much, much better.  The Digital watch face also has the weather in the middle of the screen but swiping now works as expected.

SteveAmor:
@tituscmd there is one bug that I'm failing to get to the bottom of.  I'm pretty sure that when you flash it, clicking on the icons won't work unless you go to another screen first.  I can also recreate it by being on the watch face, long holding the button to take you to `about` and then short press of the button to take you back to the watch face.  Then you can't click an icon until you move off and back to the watch face.

I had the same issue before I passed the DisplayApp to the screen the icon takes you to.  Before that, when you returned to the watch face, you couldn't click on an icon until you left the watch face and returned. So it's something to do with that.

tituscmd:
> Oh, wow, LV_EVENT_CLICKED is much, much, much better. The Digital watch face also has the weather in the middle of the screen but swiping now works as expected.

Yes, I was surprised how big of an impact it had as well!

tituscmd:
> @tituscmd there is one bug that I'm failing to get to the bottom of. I'm pretty sure that when you flash it, clicking on the icons won't work unless you go to another screen first. I can also recreate it by being on the watch face, long holding the button to take you to `about` and then short press of the button to take you back to the watch face. Then you can't click an icon until you move off and back to the watch face.

Funnily enough, I just encountered this bug a few minutes ago too. I also don't know what's causing it yet...

> I had the same issue before I passed the DisplayApp to the screen the icon takes you to. Before that, when you returned to the watch face, you couldn't click on an icon until you left the watch face and returned. So it's something to do with that.

This seems to be a great lead tho. I'll try to look into it, but maybe someone with more knowledge about the deeper workings of the PT could chime in here :D

tituscmd:
What I find interesting, it seems to work on first press after waiting a few seconds just sitting on the watch face after wake

SteveAmor:
Also prevents clicking the icons if you long press the button to jump back to the watch face.  I think it may have something to do with https://github.com/InfiniTimeOrg/InfiniTime/blob/825056574f47a8187b410b860f326050566553e2/src/displayapp/DisplayApp.cpp#L464 which is probably the same state after a full reboot (caused by the watch dog timer or firmware update)

SteveAmor:
> What I find interesting, it seems to work on first press after waiting a few seconds just sitting on the watch face after wake

It's the same behaviour with swiping too. It's present on the current main branch, even compiled with GCC 10.3.
Good news, not introduced with this PR.  Not so good news, does need investigating / fixing.

Having said that, it could just be a delay in the touch screen software/hardware initialisation.

SteveAmor:
Refactor fixes the edge cases we have seen.
Delay when screen wakes up is assumed to be initialisation of the touch screen.

SteveAmor:
Running on PT hardware but also checked in InfiniSim to ensure the `lvgl memory usage` is stable.

tituscmd:
> > What I find interesting, it seems to work on first press after waiting a few seconds just sitting on the watch face after wake
> 
> 
> 
> It's the same behaviour with swiping too. It's present on the current main branch, even compiled with GCC 10.3.
> 
> Good news, not introduced with this PR.  Not so good news, does need investigating / fixing.
> 
> 
> 
> Having said that, it could just be a delay in the touch screen software/hardware initialisation.

Very interesting. I had noticed this before with swiping and had thought there was something wrong with my custom firmware. Good to know it's a different issue!

SteveAmor:
@tituscmd if you want instant gestures on the display when it wakes up, enable double tap to wake up as that doesn't put the touch controller to sleep. https://github.com/InfiniTimeOrg/InfiniTime/blob/825056574f47a8187b410b860f326050566553e2/src/systemtask/SystemTask.cpp#L324

https://github.com/InfiniTimeOrg/InfiniTime/blob/825056574f47a8187b410b860f326050566553e2/src/systemtask/SystemTask.cpp#L415

tituscmd:
What I find interesting, it seems to work on first press after waiting a few seconds just sitting on the watch face after wake

> @tituscmd if you want instant gestures on the display when it wakes up, enable double tap to wake up as that doesn't put the touch controller to sleep.
> 
> https://github.com/InfiniTimeOrg/InfiniTime/blob/825056574f47a8187b410b860f326050566553e2/src/systemtask/SystemTask.cpp#L324
> 
> https://github.com/InfiniTimeOrg/InfiniTime/blob/825056574f47a8187b410b860f326050566553e2/src/systemtask/SystemTask.cpp#L415

Thanks for the heads-up, but I rarely use any tap-to-wake at all. I'm a raise-to-wake kinda guy :sunglasses:

SteveAmor:
> Thanks for the heads-up, but I rarely use any tap-to-wake at all. I'm a raise-to-wake kinda guy

You don't need to make use of the double tap to wake. It just prevents the touch screen going to sleep (and having to wake up - slowely). You can still be a raise-to-wake guy 😉

tituscmd:
> > Thanks for the heads-up, but I rarely use any tap-to-wake at all. I'm a raise-to-wake kinda guy
> 
> You don't need to make use of the double tap to wake. It just prevents the touch screen going to sleep (and having to wake up - slowely). You can still be a raise-to-wake guy 😉

Oh! That's perfect then. I will definitely be making use of that!

SteveAmor:
> Oh! That's perfect then. I will definitely be making use of that!

Oof, typical current draw for leaving the touch sensor on is 2.5mA. I have noticed a rapid drop in battery level today so I'll be living with the slow initialisation in favor of significantly longer battery life.

tituscmd:
It seems to me that, not using double-tap to wake, trying to press one of the complications right after waking the watch crashes it.

